PR #10, opened 2021-01-13
Update package description

cvalarida:
Remove the mention of Formation in the package description.


bkjohnson:
:thinking: Aren't we still tightly coupled to formation due to the `className`s the components use?

cvalarida:
:man_shrugging: Maybe so, but we don't need it in our description if we're actively moving away from the name.

PR #16, opened 2021-01-21
Wrap Date in a fieldset

Mottie:
## Description

To improve accessibility, the `Date` component should be wrapped in a `<fieldset>` and include a `<legend>`.

Related: https://github.com/department-of-veterans-affairs/va.gov-team/issues/15364

## Testing done

Unit tests updated

## Screenshots

N/A - appearance is the same, but did require some classes to make it the same

## Acceptance criteria

- [x] Date component wrapped in a `fieldset`, and includes a `legend`

## Definition of done

- [ ] Events are logged appropriately
- [x] No sensitive information (i.e. PII/credentials/internal URLs/etc.) is captured in logging, hardcoded, or specs

### DIFF
--- a/src/components/Date/Date.jsx
+++ b/src/components/Date/Date.jsx
@@ -84,18 +84,18 @@ class Date extends React.Component {
     }
 
     return (
-      <div className={!isValid ? 'input-error-date' : undefined}>
-        <label>
-          {this.props.label ? this.props.label : 'Date of birth'}
+      <fieldset className={!isValid ? 'input-error-date' : undefined}>
+        <legend className="vads-u-font-size--base vads-u-font-weight--normal">
+          {this.props.label || 'Date of birth'}
           {this.props.required && (
             <span className="form-required-span">(*Required)</span>
           )}
-        </label>
+        </legend>
         {errorSpan}
         <div
           className={isValid ? undefined : 'usa-input-error form-error-date'}
         >
-          <div className="usa-date-of-birth">
+          <div className="usa-date-of-birth usa-datefields clearfix-text">
             <div className="form-datefield-month">
               <Select
                 errorMessage={isValid ? undefined : ''}
@@ -136,7 +136,7 @@ class Date extends React.Component {
             </div>
           </div>
         </div>
-      </div>
+      </fieldset>
     );
   }
 }
--- a/src/components/Date/Date.jsx
+++ b/src/components/Date/Date.jsx
@@ -95,7 +95,7 @@ class Date extends React.Component {
         <div
           className={isValid ? undefined : 'usa-input-error form-error-date'}
         >
-          <div className="usa-date-of-birth usa-datefields clearfix-text">
+          <div className="usa-date-of-birth usa-datefields clearfix">
             <div className="form-datefield-month">
               <Select
                 errorMessage={isValid ? undefined : ''}
--- a/src/components/MonthYear/MonthYear.jsx
+++ b/src/components/MonthYear/MonthYear.jsx
@@ -72,18 +72,18 @@ class MonthYear extends React.Component {
     }
 
     return (
-      <div className={!isValid ? 'input-error-date' : undefined}>
-        <label>
-          {this.props.label}
+      <fieldset className={!isValid ? 'input-error-date' : undefined}>
+        <legend className="vads-u-font-size--base vads-u-font-weight--normal">
+          {this.props.label || 'Date'}
           {this.props.required && (
             <span className="form-required-span">(*Required)</span>
           )}
-        </label>
+        </legend>
         {errorSpan}
         <div
           className={isValid ? undefined : 'usa-input-error form-error-date'}
         >
-          <div className="usa-date-of-birth">
+          <div className="usa-date-of-birth usa-datefields clearfix-text">
             <div className="form-datefield-month">
               <Select
                 errorMessage={isValid ? undefined : ''}
@@ -114,7 +114,7 @@ class MonthYear extends React.Component {
             </div>
           </div>
         </div>
-      </div>
+      </fieldset>
     );
   }
 }
--- a/src/components/MonthYear/MonthYear.jsx
+++ b/src/components/MonthYear/MonthYear.jsx
@@ -83,7 +83,7 @@ class MonthYear extends React.Component {
         <div
           className={isValid ? undefined : 'usa-input-error form-error-date'}
         >
-          <div className="usa-date-of-birth usa-datefields clearfix-text">
+          <div className="usa-date-of-birth usa-datefields clearfix">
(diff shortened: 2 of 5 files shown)


src/components/Date/Date.jsx
```diff
@@ -84,18 +84,18 @@ class Date extends React.Component {
     }
 
     return (
-      <div className={!isValid ? 'input-error-date' : undefined}>
-        <label>
-          {this.props.label ? this.props.label : 'Date of birth'}
+      <fieldset className={!isValid ? 'input-error-date' : undefined}>
+        <legend className="vads-u-font-size--base vads-u-font-weight--normal">
```
Mottie, line 88: Super large `legend` without these classes

src/components/Date/Date.jsx
```diff
@@ -84,18 +84,18 @@ class Date extends React.Component {
...
+          {this.props.label || 'Date of birth'}
```
Mottie, line 89: Not a fan of keeping "Date of birth" but oh well

ghe-emoji

src/components/Date/Date.jsx
```diff
@@ -84,18 +84,18 @@ class Date extends React.Component {
...
             <span className="form-required-span">(*Required)</span>
           )}
-        </label>
+        </legend>
         {errorSpan}
         <div
           className={isValid ? undefined : 'usa-input-error form-error-date'}
         >
-          <div className="usa-date-of-birth">
+          <div className="usa-date-of-birth usa-datefields clearfix-text">
```
Mottie, line 98: Not a fan of keeping `usa-date-of-birth` but there are some definitions that need it.

Also added the `clearfix-text` to override the floats on the date fields. I was going to add a flex box class here, but it's out of scope.

Mottie:
Also updated the `MonthYear` component - now wrapped in a fieldset.

src/components/Date/Date.jsx
bkjohnson: What exactly are these new classes doing for us?

Mottie: I added `usa-datefields` as a placeholder for the wrapper, so that eventually we can remove `usa-date-of-birth` and use this new class name. Right now, it's being targeted by [two](https://github.com/department-of-veterans-affairs/vets-website/blob/mater/src/applications/disability-benefits/all-claims/sass/disability-benefits.scss#L44) [different](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/applications/financial-status-report/sass/financial-status-report.scss#L115) forms to adjust margin & padding. It's also needed by [HLR's analytics](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/applications/disability-benefits/wizard/pages/bdd.jsx#L95) because a classname is needed for as a field type (for the wizard only).

The `clearfix-text` was added because the month & day selects and year input are all aligned using float. If you look at the screenshots in #15679 you'll see that the float isn't being cleared properly. I really wanted to change this to use flexbox, but I think that's out of scope for this PR.

bkjohnson: Oh, I see. So `usa-datefields` doesn't add styling, it is _just_ there to be a target?

bkjohnson (commented):
This seems fine to me! I just want to bring in @1Copenut and/or @noahgelman for a chance to look at it

src/components/Date/Date.jsx
1Copenut, line 98: I took a look and noticed the floats weren't clearing, and that's why the error red border wasn't as tall as the inputs. If I changed "clearfix-text" to "clearfix", the floats cleared as expected.

src/components/MonthYear/MonthYear.jsx
```diff
@@ -72,18 +72,18 @@ class MonthYear extends React.Component {
...
-        </label>
+        </legend>
         {errorSpan}
         <div
           className={isValid ? undefined : 'usa-input-error form-error-date'}
         >
-          <div className="usa-date-of-birth">
+          <div className="usa-date-of-birth usa-datefields clearfix-text">
             <div className="form-datefield-month">
               <Select
```
1Copenut, line 88: These select menus and input have an extra CSS rule that causes them to get wider when the error class is applied. Can we remove this extra error state width rule so the selects and input stay the same width?

```css
.user-input-error select {
  width: calc(100% + 1.5rem);
}
```

Screen Shot 2021-02-01 at 9 43 20 PM

1Copenut (changes requested):

src/components/Date/Date.jsx
Mottie: I'll take care of this first thing tomorrow!

src/components/MonthYear/MonthYear.jsx
Mottie: CSS changes will need to happen in the VFST repo. My plan was to eventually get rid of the float and use flex-box to keep these inputs inline. I didn't think that change would be in scope with this work.

Mottie: Also, would it be better to spell out the full month name? I thought I remembered the screenreader trying to read the abbreviations and it sounded weird.

1Copenut: If we can spell the names out, yes that would be great! Screen readers do a so so job of unpacking the abbreviations, and inconsistently at that.

1Copenut: I can go with the CSS adjustment being a separate ticket (out of scope). I'll write a separate ticket for the CSS change work now, and reference it here.

1Copenut: New issue for the CSS here: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/394

Mottie: Actually, I don't know if replacing the months with a full name would break stuff... I think it should be a separate PR.

1Copenut (approved):
👍 LGTM! Thank you @Mottie

joshkimux (approved):

Mottie:
@bkjohnson This is ready for review. Thanks!

src/components/Date/Date.jsx
ncksllvn: Would `'Date of birth'` make sense as a default prop?

(I sorry your message in vsp-platform-support and decided to surf in 🏄  😁 )

Mottie: It's used for so much more than just DOB. In 526, we use to for date of separation. There are also military service history, but that uses the `DateWidget` in the form library instead of this component. I was thinking just "Date" would be good enough.

bkjohnson (approved):
Sorry for the delay!

PR #17, opened 2021-01-21
Only find tests in src/

cvalarida:
This fixes the `chokidar` watcher error as well as the failing tests we were seeing.

Just don't ask me _why._ I find this `karma` setup positively baffling. :see_no_evil:


bkjohnson (approved):
Confirmed that it works locally

PR #22, opened 2021-01-26
ProgressBar a11y aria fixes

bkjohnson:
## Description
Part of https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/357

This makes changes to the `<SegmentedProgressBar>` and `<ProgressBar>` components. Once this is merged we will update the usage of these components in `vets-website` to finish off the ticket.

This also adds `propType` documentation for the existing `label` prop and the new `ariaValueText` prop and updates the story in Storybook.

## Acceptance criteria
- [x] Add a new prop: `ariaValueText` in `<SegmentedProgressBar>`

## Definition of done
- [ ] Events are logged appropriately
- [ ] No sensitive information (i.e. PII/credentials/internal URLs/etc.) is captured in logging, hardcoded, or specs

### DIFF
--- a/src/components/ProgressBar/ProgressBar.jsx
+++ b/src/components/ProgressBar/ProgressBar.jsx
@@ -19,6 +19,11 @@ export default function ProgressBar({ percent, label }) {
 
 ProgressBar.propTypes = {
   /**
+   * The `aria-label` to use for the component
+   */
+  label: PropTypes.isRequired,
+
+  /**
    * Percent of progress made. 0 to 100.
    */
   percent: PropTypes.number.isRequired,
--- a/src/components/ProgressBar/ProgressBar.jsx
+++ b/src/components/ProgressBar/ProgressBar.jsx
@@ -21,7 +21,7 @@ ProgressBar.propTypes = {
   /**
    * The `aria-label` to use for the component
    */
-  label: PropTypes.isRequired,
+  label: PropTypes.string.isRequired,
 
   /**
    * Percent of progress made. 0 to 100.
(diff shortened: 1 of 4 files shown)


src/components/ProgressBar/ProgressBar.jsx
```diff
@@ -18,6 +18,11 @@ export default function ProgressBar({ percent, label }) {
 }
 
 ProgressBar.propTypes = {
+  /**
+   * The `aria-label` to use for the component
+   */
+  label: PropTypes.isRequired,
```
bkjohnson, line 24: According to [an investigation](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/357#issuecomment-765665591), there are only two uses of `<ProgressBar>` in `vets-website`, neither of which use that `label` prop. So my questions are:

- _Should_ this be required?
- Should this be a prop at all?
    - If not, should the hard-coded value be something like `${percent} percent finished`?

src/components/SegmentedProgressBar/SegmentedProgressBar.jsx
```diff
@@ -32,10 +38,21 @@ export default function SegmentedProgressBar({ current, total, label }) {
 }
 
 SegmentedProgressBar.propTypes = {
+  /**
+   * The `aria-valuetext` to use for the component
+   */
+  ariaValueText: PropTypes.isRequired,
```
bkjohnson, line 44: According to the ticket, I'm only adding `ariaValueText` as a prop to the `<SegmentedProgressBar>`

PR #28, opened 2021-02-04
ProgressBar and LoadingIndicator axe fixes

ddzz:
## Description
This PR adds default `aria-label` values to the `ProgressBar`, `SegmentedProgressBar`, and `LoadingIndicator`. This fixes axe failures caused by an `aria-label` value not being set for these components.

This is the simplest solution to the axe failures, but I'm not sure if a default, generic name is sufficient for accessibility standards. @1Copenut

Closes two tickets: [`ProgressBar` axe failure](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/357) and [`LoadingIndicator` axe failure](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/366).

## Testing done
- No failures observed when running the axe browser plugin on pages that use these components.
- Ran the entire Cypress test suite locally using these changes and the latest versions of `axe-core` and `cypress-axe`. No axe failures observed for any `ProgressBar`s or `LoadingIndicator`s.

## Acceptance criteria
- [ ] No axe failures seen in the axe browser plugin for pages that use these components
- [ ] No axe failures in the `vets-website` Cypress test suite.

### DIFF
--- a/src/components/LoadingIndicator/LoadingIndicator.jsx
+++ b/src/components/LoadingIndicator/LoadingIndicator.jsx
@@ -42,5 +42,5 @@ LoadingIndicator.propTypes = {
 
 LoadingIndicator.defaultProps = {
   setFocus: false,
-  label: "loading-indicator",
+  label: "Loading",
 };
--- a/src/components/LoadingIndicator/LoadingIndicator.jsx
+++ b/src/components/LoadingIndicator/LoadingIndicator.jsx
@@ -42,4 +42,5 @@ LoadingIndicator.propTypes = {
 
 LoadingIndicator.defaultProps = {
   setFocus: false,
+  label: "loading-indicator",
 };
--- a/src/components/SegmentedProgressBar/SegmentedProgressBar.jsx
+++ b/src/components/SegmentedProgressBar/SegmentedProgressBar.jsx
@@ -12,7 +12,7 @@ export default function SegmentedProgressBar({ current, total, label }) {
   return (
     <div
       className="progress-bar-segmented"
-      role="progressbar"
+      // role="progressbar"
       aria-valuenow={current}
       aria-valuemin="0"
       aria-valuemax={total}
@@ -40,4 +40,8 @@ SegmentedProgressBar.propTypes = {
    * The total number of segments in the progress bar
    */
   total: PropTypes.number.isRequired,
+};
+
+SegmentedProgressBar.defaultProps = {
+  label: "progress-bar-segmented",
 };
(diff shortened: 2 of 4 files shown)


bkjohnson (commented):
I'll let Trevor weigh in on this one, since I'm not sure if those are the defaults that we want.

src/components/LoadingIndicator/LoadingIndicator.jsx
```diff
@@ -42,4 +42,5 @@ LoadingIndicator.propTypes = {
 
 LoadingIndicator.defaultProps = {
   setFocus: false,
+  label: "loading-indicator",
```
1Copenut, line 45: Let's change this to just read "Loading". I've been chatting with @jbalboni and @noahgelman about possible fast-follow improvements for this component using a role="status" and requiring a `{message}` prop. That's out of the scope for this ticket, and they're still reviewing it in VAOS. Having this label gets this PR to where we need it, with axe-core errors cleaned up and plain language for the spinner.

1Copenut (changes requested):
@ddzz Thank you for putting this together so quickly. I pulled your branch locally and used the `file:../` import in package.json to work with it briefly. I was getting good results and no axe-core errors. Just a few text changes and I'll retest and we should be ready to ship.

src/components/SegmentedProgressBar/SegmentedProgressBar.jsx
```diff
@@ -17,7 +17,7 @@ export default function SegmentedProgressBar({ current, total, label }) {
       aria-valuemin="0"
       aria-valuemax={total}
       tabIndex="0"
-      aria-label={label}
+      aria-label={`Step ${current} of ${total}`}
```
1Copenut, line 20: I looked more closely this morning for instances of `<SegmentedProgressBar />` to make sure we wouldn't introduce trouble upstream. I found only three, and no instances were passing in a `{label}` prop. The most pertinent one was [line 83 of FormNav.jsx](https://github.com/department-of-veterans-affairs/vets-website/blob/9c610f8981da2b3608af0e4bc10519072d78d158/src/platform/forms-system/src/js/components/FormNav.jsx#L83). I think this is a good solution, and definitely could be iterated on down the road.

1Copenut (approved):
👍 LGTM! I'll write two follow on tickets for out of scope improvements discovered during this PR review cycle. Thank you @ddzz !

bkjohnson (approved):
:+1:

PR #31, opened 2021-02-05
Add accordion web component to storybook

bkjohnson:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/19484

This is an example of how we might include Web Components from the `web-components` branch inside our deployed version of Storybook.

## Testing done

Local storybook

### Showing that a React component is no longer supported

image

### New Web Component stories

image

## Acceptance criteria
- [ ]

## Definition of done


.storybook/preview.js
```diff
@@ -2,6 +2,11 @@ import '@department-of-veterans-affairs/formation/dist/formation.min.css';
 import '@department-of-veterans-affairs/formation/dist/formation';
 import { withHTML } from '@whitespace/storybook-addon-html/react';
 
+import 'web-components/dist/component-library/component-library.css';
+import { defineCustomElements } from 'web-components/loader';
+
+defineCustomElements();
```
bkjohnson, line 8: I don't think we need to apply polyfills for our documentation.

package.json
```diff
@@ -19,7 +19,7 @@
     "build": "node ./scripts/build.js",
     "build-storybook": "build-storybook",
     "prepublishOnly": "pinst --disable && node ./scripts/prepublish.js",
-    "postinstall": "husky install",
+    "postinstall": "husky install && cd node_modules/web-components && npm install --only=prod && npm run-script build",
```
bkjohnson, line 22: This is because of how we are distributing our web component library right now (not publishing to npm). We can add the dependency as a git url, but we also need to do a build since the built files aren't tracked. In order to do a build, this will only install the dependencies, not the `devDependencies`.

package.json
```diff
@@ -91,6 +91,7 @@
     "react-html-parser": "^2.0.2",
     "react-router": "3",
     "rimraf": "^3.0.2",
-    "sinon": "^9.2.2"
+    "sinon": "^9.2.2",
+    "web-components": "https://github.com/department-of-veterans-affairs/component-library.git#wc-v0.2.0"
```
bkjohnson, line 95: This will add the release `wc-v0.2.0` which is based off of the `web-components` branch.

stories/wc-helpers.jsx
```diff
@@ -0,0 +1,67 @@
...
+  Title,
+  Subtitle,
+  Description,
+  Primary,
+  ArgsTable,
+  Stories,
+  PRIMARY_STORY,
+} from '@storybook/addon-docs/blocks';
+
+import componentDocs from 'web-components/component-docs.json';
```
bkjohnson, line 13: This json file is a result of our `web-components` build, and we are using it here to add "auto-generated" documentation into storybook using these helper functions.

package.json
cvalarida: :grimacing: Ugly, but necessary without expanding the scope of this PR, from what I can tell.

A cleaner way might be to use a GH action when the version bumps to
1. Build the library
2. Push the result to a branch
3. Create a new release from that branch which contains _only_ the built code

Let's loop back around to improve the process later, though.

cvalarida (commented):
This looks good, but I'm holding off on approval until I know configuring Storybook to look for WC stories isn't possible. (If we do the whole release-built-JS-files-from-another-branch dance, we'll have to figure out the stories there too. Maybe we "publish" them alongside the actual components for now. :man_shrugging: )

stories/va-accordion.stories.jsx
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
```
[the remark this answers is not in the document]
bkjohnson: I'll see what I can do.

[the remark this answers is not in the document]
bkjohnson: I don't think that it is _impossible_, but the work to get this done seems like it is out of scope. Our stories on the `web-components` branch are written in typescript, so we can't just import them as they are into a branch with no typescript support. 

One option would be to run the `build-storybook` command as part of our build so that we get the built stories as JS. At this point we would end up with something like `storybook-static/main.abc123.bundle.js`, where all of our story exports are. Ignoring the problem of finding the right file to import based on the hashed filename (and there [isn't an easy way](https://storybook.js.org/docs/react/api/cli-options#build-storybook) to have the command build the stories in their own file), I expect that there would be additional problems here around conflicting story names. For example, each component will likely have a `Default` story, so how do we import N `Default`s from the same file?

With that being said, I think that a custom webpack configuration would allow us to get individual JS files per component that could be imported, but this is just a guess. I don't know how well a built story from one project would work when it is imported into another project and used as a story.

Another option would be if we reworked the `web-components` branch so that our stories were written as they are here (using React in plain `.jsx` files), and we could treat those source files in the `web-components` dependency as part of our Storybook instance. This would involve updating our configuration to [include stories from `node-modules/web-components`](https://github.com/department-of-veterans-affairs/component-library/blob/533d4b80aac29cce8eda2d27de36371ac4c181d2/.storybook/main.js#L2). Out of all the options this one seems like the least amount of effort, but I'm not sure right now if it would be a step forward.

package.json
bkjohnson: I think that would be a good enhancement, because we are going to have to do this same thing in `vets-website`.

cvalarida (approved):

stories/wc-helpers.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Construct an object appropriate for Storybook subcomponent documentation based
+ * on the return object from `getWebComponentDocs`
+ */
+export const componentStructure = comp => {
+  const props = propStructure(comp);
+  return {
+    [comp.tag]: {
+      __docgenInfo: {
```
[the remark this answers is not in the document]
bkjohnson: Yup: https://github.com/department-of-veterans-affairs/component-library/pull/31/commits/c1fb0a7b05136603045a66a0d1fa53a2336b4813

PR #33, opened 2021-02-10
Update date display & storybook args for banner

bkjohnson:
## Description

Related to https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/198

This updates the `MaintenanceBanner` stories with some new content, and also puts a comma after the formatted day. It also replaces `moment` with `date-fns` _only in `MaintenanceBanner`_. This allows us to properly format the `a.m.`/`p.m.` according to [our own guidelines](https://design.va.gov/content-style-guide/dates-and-numbers#times-and-time-zones).

I also found a bug that _might_ be related to daylight savings time. The displayed time was an hour off from the time that was passed in as a prop. This has been fixed.

## Testing done

- Local Storybook
- Unit testing

### Before banner

image

### During banner

image

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,8 @@
   },
   "dependencies": {
     "classnames": "^2.2.6",
+    "date-fns": "^2.17.0",
+    "date-fns-tz": "^1.1.1",
     "lodash": "^4.17.15",
     "moment": "^2.23.0",
     "prop-types": "^15.6.2",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/component-library",
-  "version": "1.1.3",
+  "version": "2.0.0",
   "description": "VA.gov component library in React",
   "keywords": [
     "react",
--- a/src/components/MaintenanceBanner/MaintenanceBanner.jsx
+++ b/src/components/MaintenanceBanner/MaintenanceBanner.jsx
@@ -74,7 +74,7 @@ export class MaintenanceBanner extends Component {
       return (
         <>
           <p>
-            <strong>Date:</strong> {startsAtET.format('dddd MMMM D, YYYY')}
+            <strong>Date:</strong> {startsAtET.format('dddd, MMMM D, YYYY')}
           </p>
           <p>
             <strong>Start/End time:</strong> {startsAtET.format('h:mm a')} to{' '}
@@ -88,11 +88,11 @@ export class MaintenanceBanner extends Component {
       <>
         <p>
           <strong>Start:</strong>{' '}
-          {startsAtET.format('dddd MMMM D, YYYY, [at] h:mm a')} ET
+          {startsAtET.format('dddd, MMMM D, YYYY, [at] h:mm a')} ET
         </p>
         <p>
           <strong>End:</strong>{' '}
-          {expiresAtET.format('dddd MMMM D, YYYY, [at] h:mm a')} ET
+          {expiresAtET.format('dddd, MMMM D, YYYY, [at] h:mm a')} ET
         </p>
       </>
     );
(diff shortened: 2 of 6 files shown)


cvalarida (approved):
Two things:
1. Does this need any unit tests updated?
    - If not, why not?
2. I see in [your comment here](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/198#issuecomment-777028487) that you advise we move away from `moment` because we can't format the `a.m.` / `p.m.`.
    - This _is_ possible in `moment`—we're doing that on `vets-website` via some fancy `moment` magic.
    - That said, I'm all for adding `date-fns` or something else here to accomplish the job if that'll do the trick.
        - We'll just need to measure the net effect on the library's output.

rtwell:
lgtm!

package.json
```diff
@@ -34,6 +34,8 @@
   },
   "dependencies": {
     "classnames": "^2.2.6",
+    "date-fns": "^2.17.0",
+    "date-fns-tz": "^1.1.1",
```
bkjohnson, line 38: This is to get the formatting that we want, and also to move away from `moment`, which is in [maintenance mode only](https://momentjs.com/docs/#/-project-status/).

src/components/MaintenanceBanner/MaintenanceBanner.jsx
```diff
@@ -67,18 +80,17 @@ export class MaintenanceBanner extends Component {
...
-    const expiresAtET = expiresAt.clone().subtract(4, 'hours');
-
-    if (startsAt.isSame(expiresAt, 'day')) {
+    if (isSameDay(startsAt, expiresAt)) {
       return (
         <>
           <p>
-            <strong>Date:</strong> {startsAtET.format('dddd MMMM D, YYYY')}
+            <strong>Date:</strong>{' '}
+            {formatEastern(startsAt, 'EEEE, MMMM d, yyyy')}
```
bkjohnson, line 88: The `format` tokens used by `date-fns` are different than the custom ones that `moment` uses. [See this doc for more info](https://github.com/date-fns/date-fns/blob/master/docs/unicodeTokens.md).

src/components/MaintenanceBanner/MaintenanceBanner.stories.jsx
```diff
@@ -14,34 +14,29 @@ export default {
...
-  warnStartsAt: moment().subtract(30, 'minutes').valueOf(),
+  title: 'Site maintenance',
+  warnTitle: 'Upcoming site maintenance',
+  content:
+    'We’re working on VA.gov right now. If you have trouble signing in or using tools, check back after we’re finished. Thank you for your patience.',
+  warnContent:
+    'We’ll be doing some work on VA.gov. The maintenance will last 1 hour. During that time, you won’t be able to sign in or use tools.',
+  startsAt: new Date(Date.now()),
+  expiresAt: addHours(Date.now(), 1),
+  warnStartsAt: subMinutes(Date.now(), 30),
```
bkjohnson, line 39: The content changes here are for the original request in @rtwell's issue comment.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/component-library",
-  "version": "1.1.3",
+  "version": "2.0.0",
```
bkjohnson, line 3: Breaking change since the `<MaintenanceBanner>` props are expecting `Date` objects instead of `moments`

bkjohnson (commented):
I haven't tested this in `vets-website` yet, but based on the build output it looks like I'll have to add `date-fns` and `date-fns-tz` as dependencies there since the built files are still trying to `require` those dependencies.

package.json
cvalarida: Ooh, good call. Let's be sure to make an announcement about this when we release it.

PR #35, opened 2021-02-12
Add analytics events to Modal component

bsmartin-ep:
## Description
_Standardize data layer events and variables to track modals across va.gov_

re: https://github.com/department-of-veterans-affairs/va.gov-team/issues/14494

## Acceptance criteria
- [ ]

## Definition of done


src/components/Modal/Modal.jsx
```diff
@@ -41,6 +42,17 @@ class Modal extends React.Component {
...
     }
+    // Conditionally track the event.
+    if (this.props.enableAnalytics) {
+      dispatchAnalyticsEvent({
+        'type': 'Modal',
+        'action': 'show',
+        'status': this.props.status,
+        'title': this.props.title,
+        'primaryButtonText': this.props.primaryButton?.text,
+        'seondaryButtonText': this.props.secondaryButton?.text,
```
cvalarida, line 53: I'll have more opinions about the shape of this when we talk later, but a few things until then:
- You don't need to quote your object property names
  - If you merge the latest `master` branch in here, it'll set up commit hooks to auto-format and lint your code when you go to commit :sparkles: 
- Looks like you've got a typo for `secondaryButtonText`

src/components/Modal/Modal.jsx
```diff
@@ -41,6 +42,17 @@ class Modal extends React.Component {
     if (this.props.clickToClose) {
       document.addEventListener('click', this.handleDocumentClicked, true);
     }
+    // Conditionally track the event.
+    if (this.props.enableAnalytics) {
+      dispatchAnalyticsEvent({
+        type: 'Modal',
```
cvalarida, line 48:
```suggestion
        componentName: 'Modal',
```

src/components/Modal/Modal.jsx
```diff
@@ -41,6 +42,17 @@ class Modal extends React.Component {
...
+        action: 'show',
+        status: this.props.status,
+        title: this.props.title,
+        primaryButtonText: this.props.primaryButton?.text,
+        secondaryButtonText: this.props.secondaryButton?.text,
```
cvalarida, line 53: Like we talked about in the meeting (minutes ago from when I'm writing this :smile: ), can we group these together in a nested object. Maybe call it `details` or `context` or something. This will collect all the event-specific details, which are expected to be very different between components and events.

src/components/Modal/Modal.jsx
```diff
@@ -273,11 +285,18 @@ Modal.propTypes = {
    * is not the first focusable element in the document
    */
   initialFocusSelector: PropTypes.string,
+  /**
+   * Analytics tracking function(s) will be called
+   */
+  enableAnalytics: PropTypes.bool,
```
cvalarida, line 291: I'd recommend reversing this. Call it `disableAnalytics`. That way it tracks by default, and if we really don't want it to, we tell it to not with `<Modal disableAnalitycs />` instead of `<Modal enableAnalytics={false} />`.

src/helpers/analytics.js
```diff
@@ -0,0 +1,10 @@
+export default function dispatchAnalyticsEvent({type, action, ...data}) {
```
cvalarida, line 1: Can we get some JSDocs on this to describe the shape of the parameter here? Maybe something like
```js
/**
 * Dispatch a custom JavaScript event on document.body for tracking analytics.
 * A separate event listener will be required to record the events to GA.
 * 
 * event.componentName {string} - The name of the component
 * event.action {string} - The action that triggered the analytics event
 * event.details {object} - Contains any number of key-value pairs to further describe the analytics event
 */
```

src/helpers/analytics.js
```diff
@@ -0,0 +1,10 @@
...
+  const event = new CustomEvent("component.library.analytics", {
+    detail: {
```
cvalarida, line 3: Does this need to be nested like this?

src/helpers/analytics.js
```diff
@@ -0,0 +1,10 @@
+export default function dispatchAnalyticsEvent({type, action, ...data}) {
+  const event = new CustomEvent("component.library.analytics", {
```
cvalarida, line 2: Can we call this custom event `component-library-analytics`? Seems a bit more "JavaScripty." :man_shrugging:

cvalarida (commented):
Overall, it looks pretty good. I had a number of smallish details I wanted to point out, though.

src/components/Modal/Modal.jsx
bsmartin-ep: Do we need to add the `mocha` plugin and env to `.eslintrc.json`?  I'm seeing a lot of linting errors on the global vars mocha uses to run tests (_describe_, _it_, etc)

cvalarida: Yeah, good catch.

For posterity, @bsmartin-ep made this addition in [this separate PR](https://github.com/department-of-veterans-affairs/component-library/pull/38).

bsmartin-ep: Done.

src/components/Modal/Modal.jsx
bsmartin-ep: Done.  I like `details` since context has a lot of other meanings.

src/helpers/analytics.js
bsmartin-ep: Are you referring to the formatting or whether our details object needs to be inside an object?

src/components/Modal/Modal.jsx
```diff
@@ -224,6 +243,10 @@ Modal.propTypes = {
    * Click outside modal will call onClose prop
    */
   clickToClose: PropTypes.bool,
+  /**
+   * Child elements (content) of modal when displayed
+   */
+  children: PropTypes.node,
```
bsmartin-ep, line 249: I had to add this for linting purposes.  Got a less confusing description?

src/components/Modal/Modal.unit.spec.jsx
```diff
@@ -100,11 +102,82 @@ describe('<Modal/>', () => {
...
+        <Modal id="modal" visible={true} clickToClose onClose={() => {}}>
+          Modal contents
+        </Modal>,
+      );
+
+      global.document.body.removeEventListener(
+        'component-library-analytics',
+        handleAnalyticsEvent,
+      );
+    });
```
bsmartin-ep, line 143: Is this the best way to test these events?  Basically recreate the listener and look at `e.details`?

Or should we be using sinon.spy or something?

src/helpers/analytics.js
cvalarida: Yeah, does it need to be `{ detail: { componentName, action, details } }` or can it just be `{ componentName, action, details }`?

src/components/Modal/Modal.unit.spec.jsx
cvalarida: I think we can get away with not using `sinon.spy` here. Still, as it's written right now, this test will pass if `handleAnalyticsEvents` is never called.

The general approach of creating an event listener seems okay to me, but I have concerns that timing might get...weird. Like, if the test ends (because execution ends) before the event listener is called. My suspicion is the [execution phase](https://medium.com/@nazanindelam/all-you-need-to-learn-to-understand-javascript-execution-context-and-stack-3babbdd88868) may sometimes happen like so:
1. The event listener call will get queued up in the JS event loop
2. The `it()` test invocation will finish
3. The event loop will pop off the next event (calling the handler)...after our test has completed

(Plug for [this great talk](https://www.youtube.com/watch?v=8aGhZQkoFbQ) on how it all works together in JS.)

I don't know if that'll _actually_ happen, but I think we need to be careful here. I'll pull this branch down and run some tests on it myself to see if I can come up with a clever way to ensure that our event listener is called (if it's going to be called) before the end of the test while also ensuring that we don't wait around for some arbitrary amount of time.

Right now, I'm leaning toward grabbing the `done` callback in the test function definition, pulling up the `handleAnalyticsEvents` into its own function which returns a closure that runs that `expect` against an object we pass it, and then calls the `done` callback...which sounds way more complicated with English than it does with code. :dizzy_face:

src/components/Modal/Modal.unit.spec.jsx
```diff
@@ -100,11 +102,82 @@ describe('<Modal/>', () => {
...
+
+    it('should include title when present', () => {
+      const handleAnalyticsEvent = e => {
+        console.log(e.detail);
```
cvalarida, line 147:
```suggestion
```

src/components/Modal/Modal.jsx
cvalarida: Nah, this looks good to me. :+1:

src/components/Modal/Modal.jsx
```diff
@@ -58,7 +77,7 @@ class Modal extends React.Component {
     }
   }
 
-  handleDocumentKeyDown = event => {
+  handleDocumentKeyDown(event) {
```
cvalarida, line 80: :thinking: So these are different statements.

The former arrow function binds `this` to the current `this` context (creation phase). The "normal" class method syntax `handleDocumentKeyDown(event)` doesn't bind `this` to anything, so it'll change based on whatever calls it.

In general, I've found it less confusing to stick with the arrow functions so `this` doesn't change.

What was the reason for changing from an arrow function to a "normal" function?

src/components/Modal/Modal.unit.spec.jsx
cvalarida: Well good news! I can't get this test to fail your way. :smile: 

I tried it the way I was thinking and it also works, but I realized by calling `done()` in the event listener callback, I can't then remove the event listener from the DOM, so that's no good. I mean, it may be fine if each test runs with a new DOM, but I'm not super confident that that happens, so :man_shrugging:.

If we find this produces flaky tests somewhere down the line, we can reassess.

src/components/Modal/Modal.unit.spec.jsx
```diff
@@ -100,11 +102,82 @@ describe('<Modal/>', () => {
...
+          },
+        });
+      };
+
+      global.document.body.addEventListener(
+        'component-library-analytics',
+        handleAnalyticsEvent,
+      );
+
+      mount(
```
cvalarida, line 133: Don't forget to assign this to a variable and then `.unmount()` it.

src/components/Modal/Modal.jsx
bsmartin-ep:
```js
handleDocumentKeyDown = event => {
```
...gives me this eslint error:

```
Parsing error: Unexpected token = [79, 25]
```

src/helpers/analytics.js
bsmartin-ep: From what I can tell, any additional info has to be in the `detail` property...

https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent

For fun, I tried moving the object and it looks it was just disregarded...

image

src/components/Modal/Modal.unit.spec.jsx
bsmartin-ep: OK cool - I'll be writing a ton of these tests so I will keep an eye on things.

src/components/Modal/Modal.jsx
bsmartin-ep: @cvalarida -

This error (predictably) goes away when I add this to my eslintconfig:

`"parser": "babel-eslint",`

...and I do get the error when I pop open another component using an arrow function (like DropDownPanel).  

image

cvalarida: Ahh, okay. So this is a linting issue then. I'll make a PR for that today (hopefully). These declarations probably shouldn't be changed, though.

bsmartin-ep: Sounds good to me @cvalarida.  Let me know when I should update my branches.

cvalarida (approved):
With the addition of the `.bind()` statements in the constructor (which it looks like I just missed the first go-round? :confused:), this looks good. I'll make the ESLint updates and change these back into arrow functions in a separate PR.

Thanks for your patience! :sweat_smile:

cvalarida:
If you wanna get these analytics out there sooner rather than later, don't forget to bump the package number so we can publish and use a new version.

PR #38, opened 2021-02-15
Add mocha to eslint

bsmartin-ep:
From:
https://dsva.slack.com/archives/C01DBGX4P45/p1613414971056900

We need to add mocha to eslint to allow the global vars it uses in tests.

image

I'm not super familiar with this so please have at it...

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -17,6 +17,6 @@
     "prettier/prettier": ["error"]
   },
   "globals": {
-    "global": "writeable"
+    "global": "writable"
   }
 }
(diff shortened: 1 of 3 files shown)


cvalarida (commented):
I think you'll need to add the `eslint-plugin-mocha` addition to the `package.json` and `yarn.lock` files for this to work.

bsmartin-ep:
`package.json` and `yarn.lock` have been updated @cvalarida.

cvalarida (approved):
Thanks for this!

PR #39, opened 2021-02-16
Remove moment

bkjohnson:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/19954

This removes `moment` as a dependency from `component-library`. It was only used in `validations.js`, but some of those helpers are used in the date components.

I tried to measure a difference in bundle size, but it doesn't look there is much of one unless I'm doing something wrong. I did a `yarn pack` on this branch and the size of `package.tgz` was 46K. I switched to master, did a `yarn install` followed by `yarn pack` and the package was the same size.

## Testing done

Local unit testing

### moment only used in one file

image

## Acceptance criteria
- [x] `moment` is no longer a dependency

## Definition of done


src/components/Date/Date.jsx
```diff
@@ -2,7 +2,6 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import _ from 'lodash';
-import moment from 'moment';
```
bkjohnson, line 5: I don't understand why our linter didn't complain about this being here and being unused, but when I removed all of the `moment` things from `validations.js` it was complaining. Maybe it is due to the different file extension.

src/helpers/validations.js
```diff
@@ -61,21 +53,7 @@ function isValidRequiredField(validator, field) {
...
-  );
-}
-
-function isValidYearOrBlank(value) {
-  return isValidYear(value) || value === '';
-}
-
-function isValidCurrentOrPastYear(value) {
-  return Number(value) >= 1900 && Number(value) < moment().year() + 1;
+  return Number(value) >= 1900 && Number(value) <= maxYear;
```
bkjohnson, line 56: Here is a place where `moment` was replaced.

src/helpers/validations.js
```diff
@@ -1,16 +1,8 @@
...
-const maxYear = moment().add(100, 'year').year();
-
-function dateToMoment(dateField) {
-  return moment({
-    year: dateField.year.value,
-    month: dateField.month.value ? parseInt(dateField.month.value, 10) - 1 : '',
-    day: dateField.day ? dateField.day.value : null,
-  });
-}
+const maxYear = addYears(Date.now(), 100).getFullYear();
```
bkjohnson, line 5: Here is the other place where `moment` was replaced.

src/components/Date/Date.jsx
cvalarida: 1. Our linter isn't running in CI, just in the commit hooks (should probably fix this sooner rather than later)
2. The linting commit hook only lints staged files, and this file hasn't been staged since it was added

Mind you, I don't know how it got here in the first place (from the old repo) without linting catching it. :man_shrugging:

cvalarida (commented):
Like we talked about in Slack, I'll hold off on approving for now since we may want to use the `MonthYear` component in the forms library definition. So until we make a deliberate decision one way or the other, let's keep `MonthYear` around a little while longer.

bkjohnson:
I've added `<MonthYear>` back and removed `moment` from it.

src/helpers/validations.js
```diff
@@ -61,17 +53,7 @@ function isValidRequiredField(validator, field) {
  * Date Validations *
  */
 function isValidYear(value) {
-  return (
-    Number(value) >= 1900 && Number(value) <= moment().add(100, 'year').year()
-  );
-}
-
-function isValidYearOrBlank(value) {
```
cvalarida, line 69: Were these originally removed because you removed the `MonthYear` component?

bkjohnson: Yup - that was the only place where they were used.

cvalarida: There are a lot of removed functions in here. Should they all still be removed now that you've added `MonthYear` back in?

bkjohnson: Yeah. All of these functions were removed before I removed `MonthYear`:

image

This isn't super pressing to merge, so we can talk about it next week when we have more time.

cvalarida (approved):
Thanks for merging master into here. I was about to ask that to make sure we get the tests run in CI. :raised_hands:

packages/react-components/src/helpers/validations.js
```diff
@@ -1,16 +1,7 @@
...
-const maxYear = moment().add(100, 'year').year();
-
-function dateToMoment(dateField) {
-  return moment({
-    year: dateField.year.value,
-    month: dateField.month.value ? parseInt(dateField.month.value, 10) - 1 : '',
-    day: dateField.day ? dateField.day.value : null,
-  });
-}
+const maxYear = new Date().getFullYear() + 100;
```
bkjohnson, line 4: Here I'm replacing `moment` with some native Date code.

packages/react-components/src/helpers/validations.js
```diff
@@ -61,17 +52,7 @@ function isValidRequiredField(validator, field) {
...
-  );
-}
-
-function isValidYearOrBlank(value) {
-  return isValidYear(value) || value === '';
-}
-
-function isValidCurrentOrPastYear(value) {
-  return Number(value) >= 1900 && Number(value) < moment().year() + 1;
+  return Number(value) >= minYear && Number(value) <= maxYear;
```
bkjohnson, line 55: Here I replaced `moment` with the date const from above.

bkjohnson (commented):
Everything else that's removed was unused. I've avoided doing additional refactoring, but we may want to come back in and clean additional unused code up.

I originally tried adding a linting rule that would throw errors on unused exports, but I ran into problems. We also may want to take a look at our linting config and get that working like we want.

packages/react-components/src/helpers/validations.js
```diff
@@ -61,17 +52,7 @@ function isValidRequiredField(validator, field) {
  * Date Validations *
  */
 function isValidYear(value) {
-  return (
-    Number(value) >= 1900 && Number(value) <= moment().add(100, 'year').year()
-  );
-}
-
-function isValidYearOrBlank(value) {
```
bkjohnson, line 69: Uses `isValidYear`

packages/react-components/src/helpers/validations.js
```diff
@@ -90,28 +71,6 @@ function isFullDate(date) {
   return /\d{4}-\d{1,2}-\d{1,2}/.test(date);
 }
 
-function isValidDate(day, month, year) {
```
bkjohnson, line 93: Uses `isValidYear`

packages/react-components/src/helpers/validations.js
```diff
@@ -132,18 +91,6 @@ function validateIfDirtyDate(dayField, monthField, yearField, validator) {
   return true;
 }
 
-function isValidAnyDate(day, month, year) {
```
bkjohnson, line 135: Uses `isValidYear` _and_ `moment` directly.

packages/react-components/src/helpers/validations.js
```diff
@@ -152,28 +99,6 @@ function isValidPartialDate(day, month, year) {
   return true;
 }
 
-function isValidDateField(field) {
```
bkjohnson, line 155: Uses `isValidYear` under the hood

packages/react-components/src/helpers/validations.js
```diff
@@ -152,28 +99,6 @@ function isValidPartialDate(day, month, year) {
...
-  return isValidDate(field.day.value, field.month.value, field.year.value);
-}
-
-function isValidPartialDateField(field) {
```
bkjohnson, line 159: Uses `isValidYear` under the hood, and is also unused.

packages/react-components/src/helpers/validations.js
```diff
@@ -185,64 +110,10 @@ function isValidPartialMonthYear(month, year) {
...
-function isValidCurrentOrFutureMonthYear(month, year) {
-  const momentDate = moment({ month: parseInt(month, 10) - 1, year });
-  return momentDate.isSameOrAfter(moment(), 'month');
-}
-
 function isBlankMonthYear(field) {
   return isBlank(field.month.value) && isBlank(field.year.value);
 }
 
-function isValidDateOver17(day, month, year) {
```
bkjohnson, line 233: Uses `isValidYear` _and_ `moment` directly.

bkjohnson (commented):
Besides the `maxYear` const, `isValidYear` was the only function that used `moment` directly and was also used. My approach was to try to remove any unused helper that used `isValidYear` so that I could see if it also could be removed. I ended up leaving it and replacing the `moment` call, but I went ahead and identified which unused helpers were using `isValidYear`.

packages/react-components/src/helpers/validations.js
k80bowman: nice use of predefined variables here!

k80bowman (approved):

PR #40, opened 2021-02-16
Add analytics events to Additional Info component

bsmartin-ep:
Standardize data layer events and variables to track additional info component across www.va.gov

For:
https://github.com/department-of-veterans-affairs/va.gov-team/issues/14488


bkjohnson:
Waiting on #35 to be merged and for master to be merged in.

src/components/AdditionalInfo/AdditionalInfo.jsx
```diff
@@ -3,21 +3,36 @@ import PropTypes from 'prop-types';
...
   constructor(props) {
     super(props);
     this.expandedContentId = _.uniqueId('tooltip-');
     this.state = { open: false };
+
+    this.toggle = this.toggle.bind(this);
   }
 
-  toggle = () => {
+  toggle() {
```
cvalarida, line 17: I think this change was before I made the linting updates, yeah? Are you still running into problems with the arrow function syntax?

src/components/AdditionalInfo/AdditionalInfo.jsx
```diff
@@ -3,21 +3,36 @@ import PropTypes from 'prop-types';
...
     this.setState({ open: !this.state.open });
+
+    // Conditionally track the event.
```
cvalarida, line 20: So this will work fine since `this.setState()` is asynchronous, but for a little better readability, I'd recommend putting the tracking code before the `setState` call. That way it's clear that the `action: !this.state.open...` refers to the state _before_ it was switched.

src/components/AdditionalInfo/AdditionalInfo.jsx
```diff
@@ -69,4 +82,18 @@ AdditionalInfo.propTypes = {
...
   onClick: PropTypes.func,
+  /**
+   * Child elements (content) of modal when displayed
+   */
+  children: PropTypes.node,
+  /**
+   * Function to pass in events and parameters to the dataLayer
+   * and (eventually) into Google Analytics
+   */
+  trackEvent: PropTypes.func,
```
cvalarida, line 93: Where is this used? I don't see any call to a `trackEvent` function..

src/components/AdditionalInfo/AdditionalInfo.unit.spec.jsx
```diff
@@ -7,38 +7,145 @@ import AdditionalInfo from './AdditionalInfo.jsx';
...
-    return axeCheck(
-      <ExpandingGroup open={first.open}>
-        <div className="child1" />
-        <div className="child2" />
-      </ExpandingGroup>,
-    );
+
+  describe('analytics event', () => {
+    it('should include the correct expand action', () => {
+      const handleExpandAnalyticsEvent = e => {
```
cvalarida, line 49: I'd recommend only testing for one specific part of this object per test. For example, only ensure that the `e.detail.action === 'expand'` here.

Bonus points for hoisting this function out into the `describe()` callback scope, but no worries if that's not feasible.

src/components/AdditionalInfo/AdditionalInfo.unit.spec.jsx
```diff
@@ -7,38 +7,145 @@ import AdditionalInfo from './AdditionalInfo.jsx';
...
+          details: {
+            triggerText: 'This is test triggerText.',
+          },
+        });
+      };
+
+      global.document.body.addEventListener(
+        'component-library-analytics',
+        handleExpandAnalyticsEvent,
+      );
```
cvalarida, line 62: I'm back to being paranoid about what happens if, for some reason, the component fails to trigger the custom event. This `handleExpandAnalyticsEvent` wouldn't be called, and the `expect` in it would never be run. The test would then succeed when it should fail.

What do you think of something like this?
```js
const handleExpandAnalyticsEvent = sinon.spy(e => { expect(...); });
// Add the listener
// Do the thing
// Remove the listener
expect(handleExpandAnalyticsEvent.called).to.be.true;
```

This way, we not only make sure the event is what we expect it to be, but also that the event was fired at all.

Potentially, this could be its own separate test, but I like the idea of adding it into every test just in case. Since it ensures the validity of the test itself.

cvalarida (commented):
I think the only thing I'd like to see before approving is ensuring the validity of the tests by spying on the event handler. Future proof it against any changes in the event name, for example.

src/components/AdditionalInfo/AdditionalInfo.unit.spec.jsx
cvalarida: Ha! Just saw @bkjohnson had the same idea in [your other PR](https://github.com/department-of-veterans-affairs/component-library/pull/41#discussion_r581464447). :joy:

src/components/AdditionalInfo/AdditionalInfo.jsx
bsmartin-ep: Nope - an artifact of the first version of this.

jonwehausen (approved):
LGTM! Thank you sir!

src/components/AdditionalInfo/AdditionalInfo.unit.spec.jsx
bsmartin-ep: I hear Chris is a little busy @bkjohnson.  Mind taking a look at this when you have time?

```js
expect(
  handleAnalyticsEvent.calledWith(
    sinon.match.hasNested('detail.action', 'collapse'),
  ),
).to.be.true;
```

bkjohnson (approved):

PR #41, opened 2021-02-16
Add analytics tracking to Alert Box link clicks

bsmartin-ep:
Standardize data layer events and variables to track alert box components across www.va.gov

re:
https://github.com/department-of-veterans-affairs/va.gov-team/issues/14489


bkjohnson:
I'll hold off on reviewing this until #35 is merged, since both of these PRs share the new analytics helper function.

src/components/AlertBox/AlertBox.unit.spec.jsx
```diff
@@ -119,4 +119,47 @@ describe('<AlertBox />', () => {
         backgroundOnly
       />,
     ));
+
+  describe('analytics event', function () {
+    it('should be triggered when link in AlertBox is clicked', () => {
```
bkjohnson, line 124: Since the component is set up with the event handler on the wrapping div, can we have a test that also checks that the event is triggered when a link in the `headline` is clicked?

src/components/AlertBox/AlertBox.unit.spec.jsx
```diff
@@ -119,4 +119,47 @@ describe('<AlertBox />', () => {
...
+      );
+
+      const testLink = wrapper.find('a');
+      testLink.simulate('click');
+
+      global.document.body.removeEventListener(
+        'component-library-analytics',
+        handleAnalyticsEvent,
+      );
+    });
```
bkjohnson, line 163: Since the `expect` checking is done in the event handler, it isn't really obvious what we expect if we are just reading this portion of the test. Because of that, can you try to rewrite it so that it resembles:

```js
      const testLink = wrapper.find('a');
      testLink.simulate('click');
      
      expect(eventHandler.calledWith(eventArgs)).to.be.true)
```

This is assuming that `eventHandler` was set up previously with something like `sinon.spy()`.

src/components/AlertBox/AlertBox.unit.spec.jsx
bsmartin-ep: Done!

src/components/AlertBox/AlertBox.unit.spec.jsx
bsmartin-ep: Thanks for taking care of the eslint stuff this AM.

Let me know if this is what you were thinking as far as testing:

https://github.com/department-of-veterans-affairs/component-library/blob/8258a488b1bc3212bc0bed009396aea0c0382541/src/components/AlertBox/AlertBox.unit.spec.jsx

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/component-library",
-  "version": "2.0.0",
+  "version": "2.0.1",
```
bsmartin-ep, line 3: This feel like a patch / fix to you?  Or bump to 2.1.0?

src/components/AlertBox/AlertBox.unit.spec.jsx
```diff
@@ -119,4 +120,96 @@ describe('<AlertBox />', () => {
...
+      // Click link in content
+      const testLink = wrapper.find('.usa-alert-text a');
+      testLink.simulate('click');
+
+      global.document.body.removeEventListener(
+        'component-library-analytics',
+        handleAnalyticsEvent,
+      );
+
+      expect(handleAnalyticsEvent.called).to.be.true;
```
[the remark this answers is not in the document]
bsmartin-ep: Sure thing.  Here's what I just committed...

```js
const handleAnalyticsEvent = sinon.spy();

global.document.body.addEventListener(
  'component-library-analytics',
  handleAnalyticsEvent,
);

// Click link in headline
const testLink = wrapper.find('.usa-alert-heading a');
testLink.simulate('click');

global.document.body.removeEventListener(
  'component-library-analytics',
  handleAnalyticsEvent,
);

expect(
  handleAnalyticsEvent.calledWith(
    sinon.match.has('detail', {
      componentName: 'AlertBox',
      action: 'linkClick',
      details: {
        clickLabel: 'with a link',
        headline: HeadlineWithLink,
        status: 'info',
        backgroundOnly: true,
        closeable: false,
      },
    }),
  ),
).to.be.true;
```

src/components/AlertBox/AlertBox.unit.spec.jsx
```diff
@@ -119,4 +120,100 @@ describe('<AlertBox />', () => {
...
+      testLink.simulate('click');
+
+      global.document.body.removeEventListener(
+        'component-library-analytics',
+        handleAnalyticsEvent,
+      );
+
+      expect(
+        handleAnalyticsEvent.calledWith(
+          sinon.match.has('detail', {
```
bkjohnson, line 170: This is cool!

bkjohnson (approved):
:+1:

PR #44, opened 2021-02-19
Add testing to CI processes

bkjohnson:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/20051

This adds a GitHub action that will run `yarn test` in each of our PRs to `master`.

Once this is merged, I will open a similar PR to add this workflow to the `web-components` branch.

## Testing done

- Checked that GH actions run & pass

## Screenshots

A screenshot of this very PR :scream: 

image

Additionally, `build_and_test` is set up to be a required check:

image

## Acceptance criteria
- [x]  Branches which introduce regressions are blocked from being merged into master
- [x]  Tests are also run on the master branch after merging 

## Definition of done

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x, 12.x, 14.x, 15.x]
+        node-version: [14.x, 15.x]
         # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
 
     steps:
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -1,0 +1,30 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+
+name: Node.js CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [10.x, 12.x, 14.x, 15.x]
+        # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm ci
+    - run: npm run build --if-present
+    - run: npm test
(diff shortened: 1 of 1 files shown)


.github/workflows/node.js.yml
```diff
@@ -0,0 +1,30 @@
...
+        # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: yarn
+    - run: yarn build
```
cvalarida, line 29: From what I can tell, you don't need this to run the tests. What happens if you remove it (and rename the job to `run_unit_tests` or something)?

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,30 @@
...
+    branches: [ master ]
+
+jobs:
+  build_and_test:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [14.x, 15.x]
```
cvalarida, line 19: Do we need to run this on two different Node versions? What do we gain, seeing as we're not really testing anything in Node?

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,30 @@
...
+        # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: yarn
```
cvalarida, line 28: Of the 1:27 this takes, installing the dependencies takes a full minute. I wonder if there's a way to cache them or something. :thinking: 

To be clear, I'm definitely considering it out of scope for this PR. But if it's easy, that'd be awesome. Because you know we won't get around to making that improvement later. :sweat_smile:

cvalarida (approved):
This looks good. There are a few improvements I think we can make, though.

.github/workflows/node.js.yml
bkjohnson: Haha, I actually removed the build and then decided to add it back in for no particular reason. I know we don't need it for the tests, but I thought it would be nice to prevent something from getting merged if it would fail the build. We can add that later though.

.github/workflows/node.js.yml
bkjohnson: I don't think so - this was just how the default workflow was set up so I went with it. I'll make some changes.

.github/workflows/node.js.yml
bkjohnson: Caching is now implemented! We went from 1m31s without the cache down to 35s once the cache had been created :tada:

cvalarida: Nice! Was that the `uses: actions/cache@v2`?

PR #47, opened 2021-02-20
Lots of linting updates

cvalarida:
## Description
As @bsmartin-ep has been experiencing, our linting configuration needed some sprucing up. This is that sprucing.

Namely, what I did here is:
1. Lint the modules run in Node separately from the modules run in the browser
2. Move `.eslintrc.json` to `.eslintrc.js` so I can re-use some config in :point_up: 
3. Add some dependencies for :point_up:
4. Add some dependencies we were already using (the linter caught)
5. Ran the linter on all the files with `--fix`


.gitignore
```diff
@@ -1,2 +1,3 @@
 node_modules
 /*.js
+!.eslintrc.js
```
cvalarida, line 3: This will be fun when we go to remove all the built JS files with `yarn build`—can't just `rm *.js` anymore. :sigh:

config/babel.config.js
```diff
@@ -59,6 +58,7 @@ module.exports = function(api) {
             ],
           },
           useBuiltIns: 'entry',
+          corejs: '3.9',
```
cvalarida, line 61: We were getting a warning without this, so I figured I'd throw it in there.
image

.gitignore
bkjohnson: :sob:

bkjohnson (approved):
:taco:

PR #62, opened 2021-03-16
Add analytics event to ProgressBar and SegmentedProgressBar components

bsmartin-ep:
## Description
For https://github.com/department-of-veterans-affairs/va.gov-team/issues/19455

"Standardize data layer events and variables to track progress bars."

~## TODO: WRITE TESTS!~


bsmartin-ep:
@jonwehausen - refresh my memory, what sort of throttling did we want to do on the updates to `ProgressBar` and `SegmentedProgressBar`?

I'm thinking...

- Every update to `SegmentedProgressBar`
- A minimum of _0%_, _100%_ for `ProgressBar` so we can diff them in BigQuery.


Thoughts?

jonwehausen:
> sort of throttling did we want

@bsmartin-ep Thanks for the question here. 

- For segmented progress bars, Is there a way to understand how many segments there are before determining if we will collect? Thinking for example if there are 50 segments, we still say we'll collect every 10%? But if the segmented progress bar only has 5 we just collect every change. Does that make sense? 
- For standard progress bars, I'm right there with you I think 0% and 100% would be fine and we can just that in BQ as you mentioned for several pieces of valuable insight.

jonwehausen (approved):
Looks good! See comment below

src/components/ProgressBar/ProgressBar.jsx
```diff
@@ -1,29 +1,77 @@
...
-      aria-valuenow={percent}
-      aria-valuemin="0"
-      aria-valuemax="100"
-      tabIndex="0"
-      aria-label={label}
-    >
-      <div className="progress-bar-inner" style={{ width: `${percent}%` }} />
-    </div>
-  );
+class ProgressBar extends Component {
```
bkjohnson, line 5: Why was this turned into a class-based component? Did an effect hook not work?

src/components/SegmentedProgressBar/SegmentedProgressBar.jsx
```diff
@@ -1,34 +1,67 @@
...
-        <div
-          key={step}
-          className={`progress-segment ${
-            current > step ? 'progress-segment-complete' : ''
-          }`}
-        />
-      ))}
-    </div>
-  );
+export default class SegmentedProgressBar extends React.Component {
```
bkjohnson, line 12: Same comment here.

src/components/ProgressBar/ProgressBar.jsx
bsmartin-ep: Nope - I didn't see any examples of `useEffect` in the component-library, so just decided to switch to a class-based component to take advantage of the lifecycle. 

Want me to convert this over to a hook?

bkjohnson: Yeah, if it can be done with a hook that would be ideal :+1:

bsmartin-ep: Sure thing.  Look OK to you?

https://github.com/department-of-veterans-affairs/component-library/blob/bba37848873b57e8108bc886ff7934085836661a/src/components/ProgressBar/ProgressBar.jsx

src/components/SegmentedProgressBar/SegmentedProgressBar.jsx
bsmartin-ep: ..and this one...

https://github.com/department-of-veterans-affairs/component-library/blob/bba37848873b57e8108bc886ff7934085836661a/src/components/SegmentedProgressBar/SegmentedProgressBar.jsx

src/components/ProgressBar/ProgressBar.unit.spec.jsx
```diff
@@ -12,5 +13,91 @@ describe('<ProgressBar/>', () => {
     tree.unmount();
   });
 
-  it('should pass aXe check', () => axeCheck(<ProgressBar percent={35} label="aria label here"/>));
+  it('should pass aXe check', () =>
+    axeCheck(<ProgressBar percent={35} label="aria label here" />));
+
+  describe('analytics event', function () {
+    it('should be triggered when a ProgressBar is mounted at 0%', () => {
```
bkjohnson, line 20: Is it ok that the event will have the `change` action?

src/components/ProgressBar/ProgressBar.unit.spec.jsx
```diff
@@ -12,5 +13,91 @@ describe('<ProgressBar/>', () => {
...
+      const tree = mount(
+        <ProgressBar percent={42} label="This is my ProgressBar." />,
+      );
+
+      tree.setProps({ percent: 100 });
+
+      global.document.body.removeEventListener(
+        'component-library-analytics',
+        handleAnalyticsEvent,
+      );
```
bkjohnson, line 70: Why are we removing this event listener in the middle of a test shortly after adding it?

src/components/ProgressBar/ProgressBar.unit.spec.jsx
```diff
@@ -12,5 +13,91 @@ describe('<ProgressBar/>', () => {
...
+          sinon.match.has('detail', {
+            componentName: 'ProgressBar',
+            action: 'change',
+            details: {
+              percent: 42,
+              label: 'This is my ProgressBar.',
+            },
+          }),
+        ),
+      ).to.be.false;
```
bkjohnson, line 87: Can this be communicated with something like:

```js

expect(handleAnalyticsEvent.timesCalled).to.equal(1)

```

I'm sure the syntax is wrong, but something like that preceding the expectation with all of the detailed specifics seems like it may be easier to understand.

src/components/SegmentedProgressBar/SegmentedProgressBar.unit.spec.jsx
```diff
@@ -14,5 +15,42 @@ describe('<SegmentedProgressBar/>', () => {
   });
 
   it('should pass aXe check', () =>
-    axeCheck(<SegmentedProgressBar current={2} total={5} label="aria label here"/>));
+    axeCheck(
+      <SegmentedProgressBar current={2} total={5} label="aria label here" />,
+    ));
+
+  describe('analytics event', function () {
+    it('should be triggered when a SegmentedProgressBar is updated', () => {
```
bkjohnson, line 23: Does the event trigger on initial mount as well, i.e. when `current` is 0 in this case? If so, can that be included in the tests?

bkjohnson (commented):
A few questions, but thank you for making the changes using hooks!

src/components/ProgressBar/ProgressBar.unit.spec.jsx
bsmartin-ep: I was actually thinking about this too.   On the GTM side, we don't make a distinction between mounting at 0% and a change back to 0%.  I could change to `load` maybe, but it would probably still enter GA as a "change".

src/components/ProgressBar/ProgressBar.unit.spec.jsx
bsmartin-ep: Goood question.  This and the unmount have been dropped at the bottom of all existing tests.

src/components/ProgressBar/ProgressBar.unit.spec.jsx
bsmartin-ep: Updated to `calledOnceWith` (who knew?):

```
expect(
        handleAnalyticsEvent.calledOnceWith(
          sinon.match.has('detail', {
            componentName: 'ProgressBar',
            action: 'change',
            details: {
              percent: 100,
              label: 'This is my ProgressBar.',
            },
          }),
        ),
      ).to.be.true;
```

src/components/SegmentedProgressBar/SegmentedProgressBar.unit.spec.jsx
bsmartin-ep: Yes, indeed.  An oversight on my part.  Updated!

src/components/ProgressBar/ProgressBar.unit.spec.jsx
bsmartin-ep: Sorry - I should have mentioned that I would prefer to keep this as `change` if possible.    Maybe consider it the `change` from _undefined_ to _0_?    

¯\\\_(ツ)\_/¯

bsmartin-ep:
Can you re-review when you have a minute @bkjohnson?

bkjohnson (approved):

PR #74, opened 2021-04-07
Add an alert component

bkjohnson:
## Description

Part of https://github.com/department-of-veterans-affairs/va.gov-team/issues/22844

## Testing done

- Storybook :eyes: 
- e2e
- testing in `vets-website` with `yarn link`

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/src/components/va-alert/test/va-alert.e2e.ts
+++ b/src/components/va-alert/test/va-alert.e2e.ts
@@ -37,4 +37,38 @@ describe('va-alert', () => {
       </va-alert>
     `);
   });
+
+  it('wraps the headline in a header element', async () => {
+    const page = await newE2EPage();
+    await page.setContent('<va-alert headline="This is an alert"></va-alert>');
+
+    const body = await page.find('va-alert >>> div.body');
+
+    expect(body).toEqualHtml(`
+      <div class="body">
+        <h3>
+          This is an alert
+        </h3>
+        <slot></slot>
+      </div>
+    `);
+  });
+
+  it('allows the heading level to be changed', async () => {
+    const page = await newE2EPage();
+
+    await page.setContent(
+      '<va-alert headline="Changed the level" level="4"></va-alert>',
+    );
+    const body = await page.find('va-alert >>> div.body');
+
+    expect(body).toEqualHtml(`
+      <div class="body">
+        <h4>
+          Changed the level
+        </h4>
+        <slot></slot>
+      </div>
+    `);
+  });
 });
--- a/src/components/va-alert/test/va-alert.e2e.ts
+++ b/src/components/va-alert/test/va-alert.e2e.ts
@@ -1,0 +1,11 @@
+import { newE2EPage } from '@stencil/core/testing';
+
+describe('va-alert', () => {
+  it('renders', async () => {
+    const page = await newE2EPage();
+    await page.setContent('<va-alert></va-alert>');
+
+    const element = await page.find('va-alert');
+    expect(element).toHaveClass('hydrated');
+  });
+});
--- a/src/components/va-alert/test/va-alert.e2e.ts
+++ b/src/components/va-alert/test/va-alert.e2e.ts
@@ -3,9 +3,23 @@ import { newE2EPage } from '@stencil/core/testing';
 describe('va-alert', () => {
   it('renders', async () => {
     const page = await newE2EPage();
-    await page.setContent('<va-alert></va-alert>');
 
+    await page.setContent('<va-alert></va-alert>');
     const element = await page.find('va-alert');
-    expect(element).toHaveClass('hydrated');
+
+    expect(element).toEqualHtml(`
+      <va-alert class="hydrated">
+        <mock:shadow-root>
+          <div class="alert info">
+            <div class="body">
+              <slot></slot>
+            </div>
+          </div>
+          <button aria-label="Close notification" class="va-alert-close">
+            <i aria-label="Close icon" class="fa-times-circle fas"></i>
+          </button>
+        </mock:shadow-root>
+      </va-alert>
+    `);
   });
 });
--- a/src/components/va-alert/test/va-alert.e2e.ts
+++ b/src/components/va-alert/test/va-alert.e2e.ts
@@ -22,4 +22,19 @@ describe('va-alert', () => {
       </va-alert>
     `);
   });
+
+  it('renders an empty div with a "polite" aria-live tag when not visible', async () => {
+    const page = await newE2EPage();
+
+    await page.setContent('<va-alert visible="false"></va-alert>');
+    const element = await page.find('va-alert');
+
+    expect(element).toEqualHtml(`
+      <va-alert class="hydrated" visible="false">
+        <mock:shadow-root>
+          <div aria-live="polite"></div>
+        </mock:shadow-root>
+      </va-alert>
+    `);
+  });
 });
--- a/src/components/va-alert/va-alert.css
+++ b/src/components/va-alert/va-alert.css
@@ -1,0 +1,71 @@
+:host {
+  display: block;
+  background-color: var(--color-gray-lightest);
+  font-size: 2rem;
+}
+
+div.alert {
+  display: table;
+  /* padding-left: 0; */
+  /* padding-right: 0; */
+  padding: 3.2rem 6.4rem 3.2rem 2.4rem;
+  width: 100%;
+  vertical-align: middle;
+  border-left-style: solid;
+  border-left-width: 10px;
+}
+
+div.alert::before {
+  display: block;
+  width: auto;
+  background: none;
+  font-family: 'Font Awesome 5 Free';
+  font-size: 2rem;
+  margin-right: 1.6rem;
+  position: static;
+  font-weight: 900;
+}
+
+div.body {
+  padding-left: 0;
+  padding-right: 0;
+  display: table-cell;
+  vertical-align: middle;
+  width: 100%;
+}
+
+div.info::before {
+  content: '\F05A';
+}
+
+.info {
+  border-left-color: #00a6d2;
+}
+
+.va-alert-close {
+  right: 0;
+  top: 0;
+  position: absolute;
+  margin: 1.6rem;
+  width: auto;
+  color: #0071bb;
+  font-size: 2.25rem;
+  appearance: none;
+  border: 0;
+}
+
+i {
+  font-family: 'Font Awesome 5 Free';
+  font-style: normal;
+  font-weight: 900;
+  margin-right: 8px;
+  font-size: 0.75em;
+}
+i::before {
+  content: '\F057'; /* fa-times-circle*/
+}
+
+h2 {
+  padding: 0;
+  margin: 0;
+}
--- a/src/components/va-alert/va-alert.css
+++ b/src/components/va-alert/va-alert.css
@@ -1,13 +1,11 @@
 :host {
-  display: block;
+  display: flex;
   background-color: var(--color-gray-lightest);
   font-size: 2rem;
 }
 
 div.alert {
   display: table;
-  /* padding-left: 0; */
-  /* padding-right: 0; */
   padding: 3.2rem 6.4rem 3.2rem 2.4rem;
   width: 100%;
   vertical-align: middle;
--- a/src/components/va-alert/va-alert.css
+++ b/src/components/va-alert/va-alert.css
@@ -97,7 +97,8 @@ i::before {
   content: '\F057'; /* fa-times-circle*/
 }
 
-h2 {
+h3 {
+  line-height: 26px;
   padding: 0;
   margin: 0;
 }
--- a/src/components/va-alert/va-alert.css
+++ b/src/components/va-alert/va-alert.css
@@ -39,7 +39,7 @@ div.info::before {
 }
 
 .info {
-  border-left-color: #00a6d2;
+  border-left-color: var(--color-primary-alt-dark);
 }
 
 .va-alert-close {
@@ -48,7 +48,7 @@ div.info::before {
   position: absolute;
   margin: 1.6rem;
   width: auto;
-  color: #0071bb;
+  color: var(--color-primary);
   font-size: 2.25rem;
   appearance: none;
   border: 0;
--- a/src/components/va-alert/va-alert.css
+++ b/src/components/va-alert/va-alert.css
@@ -43,12 +43,35 @@ div.continue::before {
   color: var(--color-green);
 }
 
+div.success::before {
+  content: '\F00C';
+  color: var(--color-green);
+}
+
+div.warning::before {
+  content: '\F071';
+}
+
+div.error::before {
+  content: '\F06A';
+  color: var(--color-secondary-dark);
+}
+
 .info {
   border-left-color: var(--color-primary-alt-dark);
 }
 
-.continue {
+.continue,
+.success {
   border-left-color: var(--color-green);
+}
+
+.warning {
+  border-left-color: var(--color-gold);
+}
+
+.error {
+  border-left-color: var(--color-secondary-dark);
 }
 
 .va-alert-close {
--- a/src/components/va-alert/va-alert.css
+++ b/src/components/va-alert/va-alert.css
@@ -38,8 +38,17 @@ div.info::before {
   content: '\F05A';
 }
 
+div.continue::before {
+  content: '\F023';
+  color: var(--color-green);
+}
+
 .info {
   border-left-color: var(--color-primary-alt-dark);
+}
+
+.continue {
+  border-left-color: var(--color-green);
 }
 
 .va-alert-close {
--- a/src/components/va-alert/va-alert.css
+++ b/src/components/va-alert/va-alert.css
@@ -4,6 +4,11 @@
   font-size: 2rem;
 }
 
+:host([background-only='true']),
+:host([background-only='']) {
+  background-color: unset;
+}
+
 div.alert {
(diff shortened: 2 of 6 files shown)


src/components/va-alert/va-alert.css
```diff
@@ -0,0 +1,137 @@
+@import '../../mixins/accessibility.css';
+
+:host {
```
bkjohnson, line 3: A big chunk of the line count is coming from the classes in here for styling the different statuses.

src/components/va-alert/va-alert.tsx
```diff
@@ -0,0 +1,117 @@
...
+@Component({
+  tag: 'va-alert',
+  styleUrl: 'va-alert.css',
+  shadow: true,
+})
+export class VaAlert {
+  /* Determines the icon and border/background color.
+   * One of `info`, `error`, `success`, `warning`, or `continue`
+   */
+  @Prop() status: string = 'info';
```
bsmartin-ep, line 12: Huh - what to do in the absence of a `PropTypes.oneOf` equivalent?  Validation via `@Watch()` decorator?

src/components/va-alert/va-alert.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  /**
+   * Determines the level of the heading in the alert
+   */
+  @Prop() level: number = 3;
+
+  @Prop() closeBtnAriaLabel: string = 'Close notification';
+
+  @Prop() closeable: boolean = false;
+
```
bsmartin-ep, line 40: You thinking this scroll stuff is unnecessary?  I only see one current component (in _ContactInformationActionButtons.jsx_) that deviates from the default value.

```js
  /**
   * Defaults to 'start' but customizable.
   */
  scrollPosition: PropTypes.string,
```

src/components/va-alert/va-alert.tsx
bkjohnson: `Watch` is only on updates - it doesn't trigger on an initial render.

src/components/va-alert/va-alert.tsx
bkjohnson: I left it out because of the problematic smooth scroll behavior in the `CollapsiblePanel` component.

src/components/va-alert/test/va-alert.e2e.ts
```diff
@@ -0,0 +1,156 @@
...
+
+    const closeSpy = await page.spyOnEvent('close');
+
+    const button = await page.find('va-alert >>> button');
+    await button.click();
+
+    expect(closeSpy).toHaveReceivedEventTimes(1);
+  });
+
+  it('fires an analytics event when a link is clicked', async () => {
```
cvalarida, line 120: Beautiful!

src/components/va-alert/va-alert.tsx
```diff
@@ -0,0 +1,117 @@
...
+    if (!visible) return <div aria-live="polite" />;
+
+    return (
+      <Host>
+        <div class={classes}>
+          <i aria-hidden="true" role="img"></i>
+          <span class="sr-only">Alert: </span>
+          <div class="body" onClick={this.handleAlertBodyClick.bind(this)}>
+            {headline && !backgroundOnly && h(`h${level}`, null, headline)}
```
cvalarida, line 98: I'm not in love with this, but I see that's how it was used before. I'd like it to be used like:
```jsx
<va-alert-box>
  <h1>Alert!</h1>
  <p>Everything I had to say goes here</p>
</va-alert-box>
```
:man_shrugging: 

For now, that's probably out of scope for this PR, but _would_ that be a difficult thing? Something about multiple `slot`s, maybe..?

src/components/va-alert/va-alert.tsx
```diff
@@ -0,0 +1,117 @@
...
+    } = this;
+    const classes = `alert ${status} ${backgroundOnly ? 'bg-only' : ''}`;
+
+    if (!visible) return <div aria-live="polite" />;
+
+    return (
+      <Host>
+        <div class={classes}>
+          <i aria-hidden="true" role="img"></i>
+          <span class="sr-only">Alert: </span>
```
cvalarida, line 96: @1Copenut How will this affect screen readers? Specifically, if this `span.sr-only` comes before the heading and the screen reader user is navigating by heading.

And to be clear, while I definitely want to make sure we do the right thing, if we don't know what that is yet, I don't want it to block this PR. Adding this bit is an improvement over [what we've currently got](https://github.com/department-of-veterans-affairs/component-library/blob/master/src/components/AlertBox/AlertBox.jsx#L101-L114). (Unless, of course, you think it's actually detrimental, in which case we'll just remove it until we come up with the right answer.)

src/components/va-alert/va-alert.tsx
```diff
@@ -0,0 +1,117 @@
...
+          <div class="body" onClick={this.handleAlertBodyClick.bind(this)}>
+            {headline && !backgroundOnly && h(`h${level}`, null, headline)}
+            <div class="text">
+              <slot></slot>
+            </div>
+          </div>
+        </div>
+
+        {closeable && (
```
cvalarida, line 105: :heart_eyes: I love how this doesn't try to be clever. :smile: "You want this alert to be closeable? Just let me know."

src/components/va-alert/va-alert.tsx
```diff
@@ -0,0 +1,117 @@
...
+  @Event({
+    eventName: 'component-library-analytics',
+    composed: true,
+    bubbles: true,
+  })
+  componentLibraryAnalytics: EventEmitter;
+
+  private closeHandler(e: MouseEvent): void {
+    this.close.emit(e);
+  }
```
cvalarida, line 56: Previously, we were able to pass a callback that got ran when the alert box was closed. Is this intended to replace that?

If so, can you give an example of how somebody might do that in React? (Context: I can't think of a way that's especially convenient, so I'm wondering if we should _additionally_ accept a callback prop and both call that function and emit the event.)

cvalarida (commented):
So, uh...you missed my reviews while I was out, right? :sweat_smile: Most of my comments are on the prop documentation, but I also what to know about the `status` validation and a few other things.

src/global/variables.css
```diff
@@ -8,9 +8,19 @@
   --color-gray-light: #aeb0b5;
   --color-gray-lighter: #d6d7d9;
   --color-gray-lightest: #f1f1f1;
+  --color-green: #2e8540;
```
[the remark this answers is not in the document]
bkjohnson: I took them from the color palette page: https://design.va.gov/design/color-palette

src/components/va-alert/va-alert.tsx
```diff
@@ -0,0 +1,117 @@
...
+   * to the status - no icon or left border.
+   */
+  @Prop() backgroundOnly: boolean = false;
+
+  /**
+   * If true, doesn't fire the CustomEvent which can be used for analytics tracking
+   */
+  @Prop() disableAnalytics: boolean = false;
+
+  @Prop() visible: boolean = true;
```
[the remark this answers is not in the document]
bkjohnson: I'm not sold on it either - it is simply the current behavior :shrug:

src/components/va-alert/va-alert.tsx
bkjohnson: It wouldn't be too difficult - that would likely use [named slots](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot#named-slot).

src/components/va-alert/va-alert.tsx
bkjohnson: Just adding that this was requested by @joshkimux 

https://github.com/department-of-veterans-affairs/va.gov-team/issues/19674

src/components/va-alert/va-alert.tsx
bkjohnson: You can still pass a callback if the component is being used in a React app, but for static HTML pages we will have to set up the event handler the old fashioned way. The [documentation](https://stenciljs.com/docs/events#using-events-in-jsx) covers this in some detail.

So, in React we would have something like:

```jsx
<va-alert headline="Blah" onClose={e => console.log('close button clicked')}>Alert content</va-alert>
```

But in a static template, we wouldn't have the `onClose` in there and instead we would have a `document.addEventListener` call to listen to the `close` event in a `<script>` tag on the page.

src/components/va-alert/va-alert.tsx
bkjohnson: Given what we have, I am in favor of not trying to be clever here. The one thing I might like to do is add an export somewhere in `vets-website` like in `platform/constants` that contains all of the valid statuses so that it can be imported and used in React apps.

Here are some options we could go with regarding an invalid `status`:

1. Do nothing - if someone passes in an incorrect `status` prop, that isn't the responsibility of the DS team to fix
1. If the `status` is anything other than the pre-defined accepted ones, force it back into the default `info`
1. Leave `status` alone, but send out an error message to let a developer looking at the console know that something is wrong
1. Completely stop the render and throw an error

Option 2 could create annoying debug scenarios (imagine sending in `succes` instead of `success` and then wondering why you are seeing the styling for the `info` status). Option 4 would be pretty bad if it happened in production. Option 3 is pretty close to option 1, but I'm not convinced that it would have much of an impact, especially if the `status` is dynamically generated. I often see errors/warnings get ignored by developers.

Additionally, here is what the styling looks like if `status` is something unexpected. The black border and lack of an icon clearly doesn't match the documented component, so a developer would know that something is wrong.

image

src/components/va-alert/va-alert.tsx
bkjohnson: Here are some links to scrolling problems in other components:

- https://dsva.slack.com/archives/C01FNSTQD7H/p1617036059425700?thread_ts=1617024286.386200&cid=C01FNSTQD7H
- https://github.com/department-of-veterans-affairs/va.gov-team/issues/6461

My understanding is that it increases cognitive load, and also causes problems for users with motion sensitivity.

src/components/va-alert/va-alert.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  /**
+   * Headline for the alert. Won't show if `backgroundOnly` is true
+   */
+  @Prop() headline: string;
+
+  /**
+   * Determines the level of the heading in the alert
+   */
+  @Prop() level: number = 3;
```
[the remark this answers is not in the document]
bkjohnson: I think the better description is nice, but if we did this we would also have to change it in `<va-accordion>`. Additionally, camelCased props have to be converted to kebab case even in JSX, so we would have to make sure all developers understand that.

image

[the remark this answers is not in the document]
cvalarida: I'm not concerned about the kebab casing. We can't stick to only single-word prop names, so that's fine. I can make a ticket for updating the accordion component prop to match—`headingLevel`. That is, unless we wanna go the named slots route, which I think I like. Gets rid of two props (`headline` and `level`).

(I also find it mildly ironic that `onClose` is camel case in the screenshot there. :joy: )

src/components/va-alert/va-alert.tsx
cvalarida: Thanks for the links!

I know I hated the scrolling myself. It's an unexpected behavior and I didn't want it.

src/components/va-alert/va-alert.tsx
cvalarida: Okay, that makes sense for the static site. Awesome to hear that about the `onClose` prop! :+1: 

Let's make sure we add some documentation for this so folks know they can do that. If we add JSDocs to the `@Event`, will that show up in the Storybook docs page?

src/components/va-alert/va-alert.tsx
cvalarida: Ahh, I gotchu. :+1:

src/components/va-alert/va-alert.tsx
cvalarida: I...kind of love that, actually. It simplifies this component, adds flexibility to its use, and makes using it more semantic at the cost of having to understand a little about web component named slots and remember this particular slot name (which will be something clever like `headline`, right? :stuck_out_tongue: ).

Since it's not too much of a change, how do you feel about rolling that in here? Any concerns?

src/components/va-alert/test/va-alert.e2e.ts
```diff
@@ -0,0 +1,156 @@
...
+    await page.setContent('<va-alert>Alert</va-alert>');
+
+    const element = await page.find('va-alert');
+
+    let button = await page.find('va-alert >>> button');
+    expect(button).toBeNull();
+
+    element.setProperty('closeable', true);
+    await page.waitForChanges();
+    button = await page.find('va-alert >>> button');
```
cvalarida, line 101: So slick!

src/components/va-alert/va-alert.css
```diff
@@ -0,0 +1,142 @@
...
+}
+i.fa-times-circle::before {
+  content: '\F057'; /* fa-times-circle*/
+}
+
+h3 {
+  line-height: 26px;
+  padding: 0;
+  margin: 0;
+}
```
cvalarida, line 142: Well this seems oddly specific. Copy pasta?

cvalarida (approved):
This looks good. There's a lot of copy pasta in here with like the styling that inflates the diff somewhat. It was easy to review. Nicely done! :raised_hands: 

I'm not strongly opinionated about this, but I think I'd like to see that named slot for the heading. It may be a Bad Idea™ and I may regret it later, but it _seems_ like a nice way to simplify the component a little and do some standard, semantic web component stuff along the way.

src/components/va-alert/va-alert.css
bkjohnson: Yup - `uswds` is removing padding and margin from the headings:

https://github.com/uswds/uswds/blob/43f8f697e7924ccaae26654f815fcf4860a5ac15/src/stylesheets/components/_alerts.scss#L85-L88

src/components/va-alert/test/va-alert.e2e.ts
```diff
@@ -43,50 +44,12 @@ describe('va-alert', () => {
   it('passes an axe check', async () => {
     const page = await newE2EPage();
     await page.setContent(
-      `<va-alert headline="Alert">Alert content</va-alert>`,
+      `<va-alert><h3 slot="headline">Alert></h3>Alert content</va-alert>`,
```
cvalarida, line 47:
```suggestion
      `<va-alert><h3 slot="headline">Alert</h3>Alert content</va-alert>`,
```

src/components/va-alert/test/va-alert.e2e.ts
```diff
@@ -132,6 +95,7 @@ describe('va-alert', () => {
       action: 'linkClick',
       componentName: 'AlertBox',
       details: {
+        headline: 'This is an alert',
```
cvalarida, line 98: :raised_hands:

src/components/va-alert/va-alert.tsx
```diff
@@ -65,6 +66,19 @@ export class VaAlert {
   }
 
   private handleAlertBodyClick(e: MouseEvent): void {
+    let headlineText = null;
+
+    // This is the happy path, meaning the user isn't using IE11
+    try {
+      const children = this.el.shadowRoot.querySelector('slot').assignedNodes();
+      const headlineNode = children[0];
+      headlineText = headlineNode.textContent;
```
cvalarida, line 75: What if there is no `<h2 slot="headline">Foo</h2>`? Will this throw an exception? (And thus get caught below?)

src/components/va-alert/va-alert.tsx
```diff
@@ -65,6 +66,19 @@ export class VaAlert {
...
+    } catch (e) {
+      // This is where we handle the edge case of the user being on IE11
+      const children = this.el.shadowRoot.childNodes;
+      headlineText = children[0].textContent;
```
cvalarida, line 79: Above, we grab only the slotted element. This ensures the headline text is the text of the element that we _expect_ to be the headline. Here, we're just grabbing the first child of the alert, which may or may not be a headline.

What happens if that headline slot is left empty? Can we maybe do a `this.el.shadowRoot.querySelector('[slot="headline"]');`? If so, would that eliminate the need for a `try / catch` block?

cvalarida (commented):
It all looks great. The only concern I have left is what will happen in `handleAlertBodyClick` if that headline slot isn't filled.

src/components/va-alert/va-alert.tsx
bkjohnson: I'll see what happens with that

src/components/va-alert/test/va-alert.e2e.ts
bkjohnson: Those sneaky brackets, always running off `:shakesfist:`

src/components/va-alert/va-alert.tsx
bkjohnson: I get `null` from `    console.log(this.el.shadowRoot.querySelector('[slot="headline"]'));`. I tried it with the `<h4>` as well as when it was removed.

cvalarida: Interesting. I wonder if that's because it like removes the `slot` property when it actually slots it into the shadow DOM. Ah well, it was worth a shot. Thanks for humoring me. :smile:

cvalarida (approved):
Thanks for fixing up that try / catch! :raised_hands:

PR #84, opened 2021-05-05
Add Pagination web component

Mottie:
## Description

I wanted to learn how to use StencilJS, so I started messing around and created a Pagination web component.

2021-05-05 08_07_15-Components _ Pagination - Default ⋅ Storybook

I followed the design, but made a few improvements including:
- Replacing `showLastPage` prop with `edge` -  this shows _both_ the first and last pages. One of the serious problems I mentioned in [this ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/5962).
- Replacing `ariaLabelSuffix` with `linkAriaLabel` - a template string allowing for customization of the aria-label
- Replacing `maxPageListLength`  with `maxVisible` prop - this now counts the first, last and ellipsis shown within the pager. This does lead to an issue where the pager will start compressed while on the first page, and expand once the first page (edge) and associated ellipsis become visible while navigating through the pages
- Renaming `pages` to `total` since it's easy to overlook the differences 

## Testing done

Added an e2e test, but I couldn't figure out how to fix the "Page must have means to bypass repeated blocks" axe error?

<details><summary>Error</summary>

2021-05-05 07_45_30-va-accordion e2e ts - stenciljs (Workspace) - Visual Studio Code</details>

## Screenshots

<details><summary>Storybook with Pagination</summary>

2021-05-05 07_39_57-Components _ Pagination - Default ⋅ Storybook</details>

## Acceptance criteria
- [ ] It works?

## Definition of done


component-docs.d.ts
```diff
@@ -0,0 +1,121 @@
+
+/**
+ * This is an autogenerated file created by the Stencil compiler.
+ * DO NOT MODIFY IT MANUALLY
+ */
+export interface JsonDocs {
```
bkjohnson, line 6: I should add this to the `.gitignore` since we don't need to commit it.

component-docs.json
```diff
@@ -0,0 +1,503 @@
+{
```
bkjohnson, line 1: Same with this.

src/components/va-pagination/va-pagination.tsx
```diff
@@ -0,0 +1,306 @@
...
+   * @returns {JSX} next link
+   */
+  private addNextLink = ({ current, total }) => {
+    const nextPage = current + 1 <= total
+      ? current + 1
+      : null;
+    const content = (
+      <Fragment>
+        <span class="hide-narrow">Next</span>
+        {'\u00a0\u00a0\u203a'}
```
bkjohnson, line 184: What is this supposed to be?

src/components/va-pagination/va-pagination.tsx
```diff
@@ -0,0 +1,306 @@
...
+   * @param className - custom link class
+   */
+  private getLinkProps = ({
+    page,
+    current = '',
+    className = '',
+  }) => ({
+    href: '#',
+    class: className,
+    'data-page': page,
```
bkjohnson, line 135: Is this still necessary?

src/components/va-pagination/va-pagination.tsx
```diff
@@ -0,0 +1,306 @@
...
+    return (
+      <nav aria-label="pagination">
+        <ul>
+          {this.addPrevLink(opts)}
+          {this.buildPager(opts)}
+          {this.addNextLink(opts)}
+        </ul>
+      </nav>
+    );
+  };
```
bkjohnson, line 305: I'm curious to see if we could keep the JSX more pieced together rather than having to visually jump around to the right function to look for the markup and having to stitch it together in your head. What do you think of something like this:


```js

  render() {
    const { current } = this.checkValidation();
    const prevPage = current - 1 >= 1 ? current - 1 : null;
  
    return (
      <nav aria-label="pagination">
        <ul>
          <li>
            {/** Previous button **/}
            <a {...props} disabled={prevPage === null}>
              <span class="vads-u-visibility--screen-reader">Previous</span>
              <span aria-hidden="true">{'\u2039\u00a0\u00a0'}
                <span class="hide-narrow">Prev</span>
              </span>
            </a>
          </li>
          {this.buildPager(opts)}
          {this.addNextLink(opts)}
        </ul>
      </nav>
    );
  };
```
Looking at this right now I'm not actually sure if I'm sold - I do recognize that there is a balance, because a giant wall of dozens of lines of nesting JSX could also be hard to decipher. :shrug: 
  
  Also, it seems like whether or not we have a previous or next page only determines if an `<a>` gets included as a wrapper inside of the `<li>`. Do you think it would be possible to simplify some things by using a bool to set `disabled` on the anchor tag?

bkjohnson (commented):
I made these comments without looking at the comparable React component, so :pinching_hand:  :salt:

src/components/va-pagination/va-pagination.tsx
Mottie: The `\u00a0` are non-breaking white spaces and the `\u203a` is a `›` character. This is added by the formation `nav-paginate-next` class name, but I ended up not using that class because it adds negative margins to keep the spaces and symbol in the view (narrow screens). The same is true for the previous link.

src/components/va-pagination/va-pagination.tsx
Mottie: Oh, I can remove that. 👍🏽

src/components/va-pagination/va-pagination.tsx
Mottie: I started out that way, but the `<a>` tag doesn't have a disabled state - https://css-tricks.com/how-to-disable-links/#just-dont-do-it - I mean it is possible to set the `pointer-events` to `none` to disable it, but I think it still causes problems for screen reader users.

Either way, I'm open to the idea of moving the prev & next links into the `render` function if you think that makes it more readable.

Mottie:
I also forgot to mention that when I build the storybook page locally, the default values and "show code" don't render properly. Is that something that only happens in the local environment? And I could use some advice on making the e2e test work. Thanks!

src/components/va-pagination/va-pagination.tsx
```diff
@@ -0,0 +1,306 @@
...
+
+  /**
+   * Key down event handler - this allows the user to hold down the key to
+   * quickly navigate through the pages
+   * @param key - event.key
+   * @param page - page associated with the interaction
+   */
+  private handleKeyDown = (key: string, page: number) => {
+    const activate = ['Enter', ' ', 'Spacebar' /* IE11 */];
```
Mottie, line 115: I'm going to remove this based on this article - https://www.a11yproject.com/posts/2019-02-15-creating-valid-and-accessible-links/#toc_Focus-state-and-keyboard

kelsonic (approved):

src/components/va-pagination/va-pagination.e2e.ts
```diff
@@ -0,0 +1,44 @@
...
+import { axeCheck } from '../../testing/test-helpers';
+
+describe('va-pagination', () => {
+  it('renders', async () => {
+    const page = await newE2EPage();
+
+    await page.setContent('<va-pagination></va-pagination>');
+    const element = await page.find('va-pagination');
+    expect(element).toEqualHtml(`
+      <va-pagination page="1" class="sc-va-pagination-h sc-va-pagination-s hydrated">
```
bkjohnson, line 12: I know `hydrated` is a stencil thing, but I'm not sure where these other classes are coming from or why they are there.

src/components/va-pagination/va-pagination.tsx
```diff
@@ -0,0 +1,236 @@
...
+    );
+  };
+
+  /**
+   * Build page links
+   */
+  private buildLink = (index: number, current: number, total: number) => {
+    let page = index + 1;
+    if (page < 1) {
+      page = 1;
```
bkjohnson, line 145: It seems like this `if` would only happen when `index` is a negative - is that possible?

src/components/va-pagination/va-pagination.tsx
```diff
@@ -0,0 +1,236 @@
...
+      }
+      for (
+        index = Math.max(total - edges, end);
+        index < total;
+        index++
+      ) {
+        pages.push(this.buildLink(index, current, total));
+      }
+    }
+    return pages;
```
bkjohnson, line 219: The multiple nested conditionals and loops make this hard for me to follow. I'm personally a fan of this kind of pattern:

```jsx

myCollection.map(item => {
    <a>{item.whatever}</a>
});
```

Our React implementation does it like that...

https://github.com/department-of-veterans-affairs/component-library/blob/f3c3794991dfd82d20aa20fe4c7b1e9ca7de9a6f/src/components/Pagination/Pagination.jsx#L180-L196

_but_ it's backed up by something similar to what you have here:

https://github.com/department-of-veterans-affairs/component-library/blob/f3c3794991dfd82d20aa20fe4c7b1e9ca7de9a6f/src/components/Pagination/Pagination.jsx#L135-L161


I know that the markup has changed a bit, but do you think we could generate the pager portion with a `.map()` call?

bkjohnson (commented):
Some more comments! :sweat_smile:

pdavies88:
Closing because continued work was done in: #275 to complete the `va-pagination` web component

PR #90, opened 2021-05-10
Add crisis line modal story

bkjohnson:
## Description

This adds a crisis line modal to storybook so that we can use the iframe to replace what we currently have (see Screenshots).

## Testing done

Storybook :eyes: 

## Screenshots

These screenshots are of content on design.va.gov.

### Current

image

### Current - open

image

### With change

image

### With change - open

image

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,3 +1,5 @@
+const path = require('path');
+
 module.exports = {
   stories: ['../@(src|stories)/**/*.stories.@(js|jsx|ts|tsx|mdx)'],
   addons: [
@@ -6,4 +8,20 @@ module.exports = {
     '@storybook/addon-links',
     // '@whitespace/storybook-addon-html/register',
   ],
+
+  webpackFinal: async (config, { configType }) => {
+    // `configType` has a value of 'DEVELOPMENT' or 'PRODUCTION'
+    // You can change the configuration based on that.
+    // 'PRODUCTION' is used when building the static version of storybook.
+
+    // Make whatever fine-grained changes you need
+    config.module.rules.push({
+      test: /\.scss$/,
+      use: ['style-loader', 'css-loader', 'sass-loader'],
+      include: path.resolve(__dirname, '../'),
+    });
+
+    // Return the altered config
+    return config;
+  },
 };
--- a/.storybook/style.scss
+++ b/.storybook/style.scss
@@ -1,0 +1,7 @@
+// $fa-font-path: '~@fortawesome/fontawesome-free/webfonts';
+$formation-asset-path: '~@department-of-veterans-affairs/formation/assets';
+
+@import '@department-of-veterans-affairs/formation/sass/base/_b-mixins.scss';
+@import '@department-of-veterans-affairs/formation/sass/base/_b-variables.scss';
+@import '@department-of-veterans-affairs/formation/sass/base/_b-breakpoints.scss';
+@import '@department-of-veterans-affairs/formation/sass/site/_m-crisis-line.scss';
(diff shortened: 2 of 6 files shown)


.storybook/main.js
```diff
@@ -6,4 +8,20 @@ module.exports = {
...
+    // Make whatever fine-grained changes you need
+    config.module.rules.push({
+      test: /\.scss$/,
+      use: ['style-loader', 'css-loader', 'sass-loader'],
+      include: path.resolve(__dirname, '../'),
+    });
+
+    // Return the altered config
+    return config;
+  },
```
bkjohnson, line 26: I copied this from Storybook's docs:

https://storybook.js.org/docs/react/configure/webpack#extending-storybooks-webpack-config

.storybook/style.scss
```diff
@@ -0,0 +1,6 @@
+$formation-asset-path: '~@department-of-veterans-affairs/formation/assets';
+
+@import '@department-of-veterans-affairs/formation/sass/base/_b-mixins.scss';
+@import '@department-of-veterans-affairs/formation/sass/base/_b-variables.scss';
+@import '@department-of-veterans-affairs/formation/sass/base/_b-breakpoints.scss';
+@import '@department-of-veterans-affairs/formation/sass/site/_m-crisis-line.scss';
```
bkjohnson, line 6: All of these are necessary in order to get the `crisis-line` module to load.

src/components/Modal/Modal.stories.jsx
```diff
@@ -20,6 +20,41 @@ const Template = args => {
   );
 };
 
+const CrisisTemplate = args => {
```
bkjohnson, line 23: This template and the child content in `CrisisLineModal` are modeled after: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/6f11751e623e7d567c2f1b34f9176fdc224253d6/src/_components/html/crisis-line.html#L1

src/components/Modal/Modal.stories.jsx
```diff
@@ -20,6 +20,41 @@ const Template = args => {
...
+  const [visible, setVisible] = useState(args.visible);
+  const onClose = () => setVisible(!visible);
+  const openModal = () => setVisible(true);
+  return (
+    <>
+      <div
+        className="va-crisis-line"
+        style={{ position: 'static', height: '600px', maxWidth: 'none' }}
```
bkjohnson, line 31: `maxWidth` is set here to remove what the `va-crisis-line` CSS is adding. We can't actually remove it from the CSS because that is used in teamsite pages where the React component can't be used.

src/components/Modal/Modal.stories.jsx
```diff
@@ -93,3 +128,41 @@ HideCloseButton.args = {
...
 };
+
+export const CrisisLineModal = CrisisTemplate.bind({});
+CrisisLineModal.args = {
+  ...defaultArgs,
+  primaryButton: null,
+  secondaryButton: null,
+  title: 'Get help from Veterans Crisis Line',
+  children: (
+    <div className="va-crisis-panel" style={{ transform: 'none' }}>
```
bkjohnson, line 139: Same story here with the `transform` styling.

.storybook/style.scss
cvalarida: :scream:

cvalarida (approved):
Holy frijoles! Them's a lot of changes to the `yarn.lock` file. I don't see how you can get around it, so it's okay, but wow. :see_no_evil:

bkjohnson:
In the future we might be able to undo some of this if we also publish compiled CSS files for each module in `formation` or something like that, but for now :upside_down_face: .

PR #96, opened 2021-05-13
feat(TextInput): Add new ariaDescribedBy prop

dat-gcio:
## Description
Addresses the need for aria-describedby for TextInput.
[Ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/23616)
[Ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/23222)

## Testing done
Manual Storybook With axe
Unit Test added

## Screenshots
Only ariaDescribedBy:
image
Both error and ariaDescribedBy:
image
No error or ariaDescribedBy:
image

## Acceptance criteria
- [ ] Allow additional prop ariaDescribedBy to be passed to TextInput

## Definition of done
[Ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/23616)
[Ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/23222)

### DIFF
--- a/src/components/TextInput/TextInput.jsx
+++ b/src/components/TextInput/TextInput.jsx
@@ -57,14 +57,14 @@ class TextInput extends React.Component {
   }
 
   render() {
+    let ariaDescribedBy = this.props.ariaDescribedBy;
     // Calculate error state.
     let errorSpan = '';
     let maxCharacters;
-    let errorSpanId = undefined;
     let inputErrorClass = undefined;
     let labelErrorClass = undefined;
     if (this.props.errorMessage) {
-      errorSpanId = `${this.inputId}-error-message`;
+      const errorSpanId = `${this.inputId}-error-message`;
       errorSpan = (
         <span className="usa-input-error-message" role="alert" id={errorSpanId}>
           <span className="sr-only">Error</span> {this.props.errorMessage}
@@ -72,6 +72,9 @@ class TextInput extends React.Component {
       );
       inputErrorClass = 'usa-input-error';
       labelErrorClass = 'usa-input-error-label';
+      ariaDescribedBy = ariaDescribedBy
+        ? `${ariaDescribedBy} ${errorSpanId}`
+        : errorSpanId;
     }
 
     // Calculate max characters and display '(Max. XX characters)' when max is hit.
@@ -98,7 +101,7 @@ class TextInput extends React.Component {
         {errorSpan}
         <input
           className={this.props.additionalClass}
-          aria-describedby={errorSpanId}
+          aria-describedby={ariaDescribedBy}
           id={this.inputId}
           placeholder={this.props.placeholder}
           name={this.props.name}
@@ -160,6 +163,11 @@ TextInput.propTypes = {
    * Called when input value is changed
    */
   onValueChange: PropTypes.func.isRequired,
+  /**
+   * Sets the aria-describedby attribute
+   */
+  /* eslint-disable consistent-return */
+  ariaDescribedBy: PropTypes.string,
   /**
    * `<input>` type attribute
    */
--- a/src/components/TextInput/TextInput.unit.spec.jsx
+++ b/src/components/TextInput/TextInput.unit.spec.jsx
@@ -239,6 +239,61 @@ describe('<TextInput>', () => {
     return check;
   });
 
+  it('passes aXe check when error and ariaDescribedBy present', () => {
+    const check = axeCheck(
+      <TextInput
+        field={makeField('')}
+        label="test"
+        placeholder="Placeholder"
+        errorMessage="error"
+        onValueChange={value => value}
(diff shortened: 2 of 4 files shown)


oseasmoran73 (approved):

src/components/TextInput/TextInput.jsx
```diff
@@ -57,21 +57,24 @@ class TextInput extends React.Component {
...
       errorSpan = (
         <span className="usa-input-error-message" role="alert" id={errorSpanId}>
           <span className="sr-only">Error</span> {this.props.errorMessage}
         </span>
       );
       inputErrorClass = 'usa-input-error';
       labelErrorClass = 'usa-input-error-label';
+      ariaDescribedBy = ariaDescribedBy
+        ? `${ariaDescribedBy} ${errorSpanId}`
+        : errorSpanId;
```
cvalarida, line 77: I'm not sure I understand this. `ariaDescribedBy` could potentially be any DOM element. Appending the `${errorSpanId}` tightly couples this component to implementation details outside of this component.

Under what circumstances should the `aria-describedby` property be different than the error message rendered by this component? Or am I missing something?

src/components/TextInput/TextInput.unit.spec.jsx
```diff
@@ -239,6 +239,61 @@ describe('<TextInput>', () => {
     return check;
   });
 
+  it('passes aXe check when error and ariaDescribedBy present', () => {
```
cvalarida, line 242: Is it possible to test the inverse of this—that it fails the aXe check without the `ariaDescribedBy`?

cvalarida (commented):
I think I'm not quite following an underlying assumption here. I'd love if you could help me understand.

Also, I saw you had bumped the version number and then reverted that. You'll probably want to keep the version bump so we can publish this change and you can use it in your application quickly. (One of these days, we'll work on automatically bumping the version, but we're not there yet. :man_shrugging:)

dat-gcio:
> I think I'm not quite following an underlying assumption here. I'd love if you could help me understand.
> 
> Also, I saw you had bumped the version number and then reverted that. You'll probably want to keep the version bump so we can publish this change and you can use it in your application quickly. (One of these days, we'll work on automatically bumping the version, but we're not there yet. 🤷‍♂️)

In one of the recent updates to the 1010CG form, there was an additional description text that followed an input. It(the additional description) coincidentally also proceeds another input(a checkbox). It was advised by accessibility that we tie that description into the corresponding input via the aria-describedby attribute.
[Links to the findings issue](https://github.com/department-of-veterans-affairs/va.gov-team/issues/23222)
I jumped on this ticket/project just last week, but maybe @joshkimux has a better perspective on if my assumption is out of wack.

src/components/TextInput/TextInput.jsx
dat-gcio: The reply in the root thread has the background story, but there was a case where we had additional description that we wanted to tie for accessibility reasons into a TextInput. 

From what I can gather on the aria-describedby documentation, it can take in multiple identifies separated by spaces. I wanted to have any ariaDescribedBy proceed the errorSpanId because the error message/information would be read last, after all the additional descriptions.

Please let me know if any of that doesn't makes sense, or if you have any recommendations!

src/components/TextInput/TextInput.unit.spec.jsx
dat-gcio: I don't believe there is really anything programmatically mandatory about associating some element to another via the aria-describedby attribute. The element we tie in could be any random element.

src/components/TextInput/TextInput.jsx
cvalarida: > it can take in multiple identifies separated by spaces.

Ahhhh, so it does, [look at that](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute#value). TIL. :smile: That's not misleading at all. :stuck_out_tongue_closed_eyes:

src/components/TextInput/TextInput.unit.spec.jsx
```diff
@@ -239,6 +239,61 @@ describe('<TextInput>', () => {
...
+      <TextInput
+        field={makeField('')}
+        label="test"
+        onValueChange={value => value}
+        ariaDescribedBy={describedById}
+      />,
+    );
+
+    const attribute = wrapper.find('input').prop('aria-describedby');
+    expect(attribute).to.exist;
```
cvalarida, line 269:
```suggestion
```
If it passes the next `expect`, it exists. :man_shrugging:

cvalarida (approved):
I thought it `aria-describedby` took a CSS selector, but you're right—it's a list of IDs. That removes all concerns I had about tightly coupling the expected error message and the `aria-describedby` element. :+1: 

In that case, this looks great!

PR #97, opened 2021-05-14
Unmount components in analytics tests

bsmartin-ep:
## Description
Correct me if I'm wrong, but I should be unmounting these components even if they are shallow render.

## Acceptance criteria
- [ ]

## Definition of done


bkjohnson (approved):
Good call. I think we've run into problems in the past in other repos where not unmounting was causing some flakiness.

PR #99, opened 2021-05-26
Add va-select Web Component

bkjohnson:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/25147

## Testing done

- Local storybook browser testing
- e2e tests written

## Acceptance criteria
- [ ]

## Definition of done


src/assets/arrow-both.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 10 14"><g fill="#323a45" fill-rule="evenodd"><path d="M.62639628 5.99993896H9.3727389c.25339173.00130856.482416-.15074264.57956202-.38477624.09714598-.2340336.04312294-.50357616-.13670922-.6820964L5.4401815.1835435C5.32475606.06613116 5.1670083 0 5.00236114 0c-.16464716 0-.3223949.06613117-.43782037.1835435L.1835435 4.93306632C.06613116 5.0484918 0 5.20623953 0 5.3708867s.06613117.32239492.1835435.43782038c.1146444.1224934.27508006.19177244.44285278.19123188zM9.3727389 7.9999685H.6263963c-.1662653-.00316183-.32647608.06237894-.4428528.18116705C.0661312 8.296561 0 8.45430875 0 8.61895592s.06613117.32239492.1835435.43782037l4.38099726 4.7596802C4.67996622 13.9338688 4.83771396 14 5.00236113 14s.32239492-.0661312.43782038-.1835435l4.3754102-4.74961537c.17983216-.17852023.2338552-.44806278.13670922-.68209638-.09714602-.2340336-.3261703-.3860848-.579562-.38477625z"/></g></svg>
```
bkjohnson, line 1: Copied from USWDS: https://github.com/uswds/uswds/blob/v1.6.10/src/img/arrow-both.png

src/components/va-accordion/va-accordion.tsx
```diff
@@ -1,4 +1,14 @@
...
+  Component,
+  Element,
+  Event,
+  EventEmitter,
+  Host,
+  Listen,
+  Prop,
+  h,
+} from '@stencil/core';
+import { getSlottedNodes } from '../../utils/utils';
```
bkjohnson, line 11: This is the only real change to this file.

src/components/va-select/va-select.tsx
```diff
@@ -0,0 +1,141 @@
...
+      this.options = getSlottedNodes(this.el, 'option').map(
+        (node: HTMLOptionElement) => {
+          return (
+            <option value={node.value} selected={value === node.value}>
+              {node.text}
+            </option>
+          );
+        },
+      );
+  }
```
cvalarida, line 104: So if we dynamically update the options (slotted children), they'll get rendered to the light DOM _and_ be reflected in the state variable, is that right? Is there any way that this might cause a disagreement between what's rendered to the DOM and what the state variable holds?

src/components/va-select/va-select.tsx
```diff
@@ -0,0 +1,141 @@
...
+            {error}
+          </span>
+        )}
+        <slot></slot>
+        <select
+          aria-describedby={errorSpanId}
+          id="select"
+          name={name}
+          onKeyDown={this.handleKeyDown.bind(this)}
+          onChange={this.handleChange.bind(this)}
```
cvalarida, line 128: If we declare these instance methods as arrow functions, do we have to `.bind(this)`?

src/components/va-select/va-select.tsx
```diff
@@ -0,0 +1,141 @@
...
+} from '@stencil/core';
+import { getSlottedNodes } from '../../utils/utils';
+
+@Component({
+  tag: 'va-select',
+  styleUrl: 'va-select.css',
+  shadow: true,
+})
+export class VaSelect {
+  @Element() el!: any;
```
cvalarida, line 19: TypeScript n00b question: What's this `!`? The closest I've found is the [non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator), but this seems different. :thinking:

cvalarida (commented):
It looks good, but I had a few questions before I approve.

src/components/va-select/test/va-select.e2e.ts
```diff
@@ -0,0 +1,173 @@
...
+          <option value="">Please select an option</option>
+          <option value="foo">Foo</option>
+          <option value="bar">Bar</option>
+        </va-select>
+      `);
+
+    await axeCheck(page);
+  });
+
+  it('adds a span for an aria-live region', async () => {
```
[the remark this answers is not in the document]
bkjohnson: The current implementation only adds it if the prop is supplied:

https://github.com/department-of-veterans-affairs/component-library/blob/c53257cdc36c8d0fb17bfdf494d80ba42825e31c/src/components/Select/Select.jsx#L118-L126

src/components/va-select/va-select.tsx
bkjohnson: I think it might be the non-null assertion operator, but now I'm not sure how this ended up here. I thought I copy-pasted it from [the docs](https://stenciljs.com/docs/host-element#element-decorator), but they don't have the `!`. Maybe they did at one point and have since been changed :thinking: 

Anyway, looking at the docs again, I don't think it is necessary here, so I can remove it.

src/components/va-select/va-select.tsx
bkjohnson: Looks like we can!

src/utils/utils.ts
```diff
@@ -1,3 +1,30 @@
...
+  );
+}
+
+/**
+ * Get all of the slotted children in the root element that match `nodeName`
+ */
+export function getSlottedNodes(
+  root: HTMLElement,
+  nodeName: string,
+): Array<Node> {
```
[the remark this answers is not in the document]
bkjohnson: I'll try to spend a few more minutes and see if I can get a test to work this time, but I made some local changes just to confirm that this is an integral part of the component workings, and a bunch of tests failed when I changed the return value of this function. I'll spare you most of the verbose output, but this is what I got:

image


When I made these changes:

image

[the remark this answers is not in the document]
bkjohnson: Also while I (sort of) remember, I'll put this here in case I want to come back to it later when referring to why this was done. When I wrote this function originally as a helper for `<va-accordion>`, I couldn't unit test it because the mock object was missing some important things. It didn't seem worth the effort to manually mock out _everything_ required since what was provided was pretty bare.

src/components/va-select/test/va-select.e2e.ts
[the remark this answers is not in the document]
cvalarida: Okay, this is non-blocking, then, but the question still stands. :+1:

cvalarida (approved):

src/components/va-select/va-select.tsx
bkjohnson: The previous implementation (lifecycle method + conditional assignment) did _not_ render any updated nodes from the light DOM. However, the most recent handful of commits have fixed this by removing the conditional assignment and having the function be a handler for the `slotchange` event.

src/components/va-select/test/va-select.e2e.ts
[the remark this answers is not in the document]
1Copenut: I ran this through a screen reader (VoiceOver) and I don't believe we need this live region. Screen readers announce the selected option when users trigger the change.

src/components/va-select/va-select.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+  /**
+   * Error message to display. When defined, this indicates an error.
+   */
+  @Prop() error: string;
+
+  /**
+   * Optional text that is displayed to screen readers only.
+   */
+  @Prop() ariaLiveRegionText: string;
```
1Copenut, line 49: Unless there's a back support need for this prop, I'd ask if we can remove it.

1Copenut (changes requested):

src/components/va-select/va-select.tsx
bkjohnson: This is great news, I love simplifying things! It looks like it isn't used in any apps, so I'll go ahead and remove it.

image

src/components/va-select/test/va-select.e2e.ts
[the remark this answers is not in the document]
bkjohnson: Removed this in the latest commit: https://github.com/department-of-veterans-affairs/component-library/pull/99/commits/d32e14d863547300980b4e3ec982dd79d79a4c63

1Copenut (approved):
LGTM!

PR #101, opened 2021-05-28
Create codeql-analysis.yml

cvalarida:
## Description
Adds CodeQL for security scanning.
See [About CodeQL](https://codeql.github.com/docs/codeql-overview/about-codeql/) for more information.


cvalarida:
Bah. This is already a thing. I thought so, but was blind when I looked for it. :man_facepalming:

PR #104, opened 2021-06-03
Add aria-describedby to Date, Select & NumberInput

Mottie:
## Description

This PR adds an `aria-describedby` attribute to `Date`, `Select` and `NumberInput` components to allow screenreaders to target content related to form element.

This value should only be added when the target ID exists within the DOM.

Related:

- https://github.com/department-of-veterans-affairs/va.gov-team/issues/22569
- https://github.com/department-of-veterans-affairs/va.gov-team/issues/22612

## Testing done

Updated unit tests

## Screenshots

N/A

## Acceptance criteria
- [x] `aria-describedby` added to the `Date`, `Select` and `NumberInput`

## Definition of done

### DIFF
--- a/src/components/Date/Date.jsx
+++ b/src/components/Date/Date.jsx
@@ -27,7 +27,7 @@ class Date extends React.Component {
   }
 
   UNSAFE_componentWillMount() {
-    this.id = _.uniqueId('date-input-');
+    this.inputId = _.uniqueId('date-input-');
   }
 
   handleChange(path, update) {
--- a/src/components/Date/Date.jsx
+++ b/src/components/Date/Date.jsx
@@ -92,6 +92,10 @@ class Date extends React.Component {
       );
     }
 
+    const ariaDescribedby =
+      [errorSpanId, this.props.ariaDescribedby || ''].join(' ').trim() ||
+      undefined;
+
     return (
       <fieldset className={fieldsetClass}>
         <legend className="vads-u-font-size--base vads-u-font-weight--normal">
@@ -115,7 +119,7 @@ class Date extends React.Component {
                 onValueChange={update => {
                   this.handleChange('month', update);
                 }}
-                ariaDescribedby={this.props.ariaDescribedby}
+                ariaDescribedby={ariaDescribedby}
               />
             </div>
             <div className="form-datefield-day">
@@ -128,7 +132,7 @@ class Date extends React.Component {
                 onValueChange={update => {
                   this.handleChange('day', update);
                 }}
-                ariaDescribedby={this.props.ariaDescribedby}
+                ariaDescribedby={ariaDescribedby}
               />
             </div>
             <div className="usa-datefield usa-form-group usa-form-group-year">
@@ -143,7 +147,7 @@ class Date extends React.Component {
                 onValueChange={update => {
                   this.handleChange('year', update);
                 }}
-                ariaDescribedby={this.props.ariaDescribedby}
+                ariaDescribedby={ariaDescribedby}
               />
             </div>
           </div>
--- a/src/components/Date/Date.jsx
+++ b/src/components/Date/Date.jsx
@@ -2,7 +2,6 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import _ from 'lodash';
-import moment from 'moment';
 
 import Select from '../Select/Select';
 import NumberInput from '../NumberInput/NumberInput';
@@ -116,6 +115,7 @@ class Date extends React.Component {
                 onValueChange={update => {
                   this.handleChange('month', update);
                 }}
+                ariaDescribedby={this.props.ariaDescribedby}
               />
             </div>
             <div className="form-datefield-day">
@@ -128,6 +128,7 @@ class Date extends React.Component {
                 onValueChange={update => {
                   this.handleChange('day', update);
                 }}
+                ariaDescribedby={this.props.ariaDescribedby}
               />
             </div>
             <div className="usa-datefield usa-form-group usa-form-group-year">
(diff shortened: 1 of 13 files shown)


src/components/Date/Date.jsx
```diff
@@ -2,7 +2,6 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import _ from 'lodash';
-import moment from 'moment';
```
bkjohnson, line 5: :raised_hands:

src/components/Date/Date.stories.jsx
```diff
@@ -61,3 +57,9 @@ Error.args = {
     },
   },
 };
+
+export const AriaDescribedby = Template.bind({});
+AriaDescribedby.args = {
+  ...defaultArgs,
+  ariaDescribedby: 'some-id',
```
bkjohnson, line 64: Do we have `some-id` on the page?

src/components/Date/Date.unit.spec.jsx
```diff
@@ -204,8 +189,38 @@ describe('<Date>', () => {
...
+  });
+  it('should pass aXe check when multiple aria-describedby ids are set', () => {
+    const date = {
+      day: makeField(''),
+      month: makeField(''),
+      year: makeField('1890'),
+    };
+    date.year.dirty = true;
+    date.month.dirty = true;
+    date.day.dirty = true;
```
bkjohnson, line 220: I know that this is following the existing pattern of other tests, but you may be happy to know that we are going to be improving the form component API and getting rid of this awkwardness.

src/components/NumberInput/NumberInput.jsx
```diff
@@ -53,6 +53,10 @@ class NumberInput extends React.Component {
       requiredSpan = <span className="form-required-span">(*Required)</span>;
     }
 
+    const ariaDescribedby =
+      [errorSpanId, this.props.ariaDescribedby || ''].join(' ').trim() ||
```
bkjohnson, line 57: Why are we using `errorSpanId` here and not in the `Date` component?

src/components/NumberInput/NumberInput.unit.spec.jsx
```diff
@@ -167,12 +167,70 @@ describe('<NumberInput>', () => {
...
+      />,
+    );
+
+    // Ensure no error.
+    expect(tree.find('.usa-input-error')).to.have.lengthOf(0);
+
+    // No error means no aria-describedby to not confuse screen readers.
+    const inputs = tree.find('input');
+    expect(inputs).to.have.lengthOf(1);
+    expect(inputs.prop('aria-describedby')).to.equal('some-id');
```
bkjohnson, line 186: The comments here are confusing to me. We don't want an error, but that means that there shouldn't be an `aria-describedby`, yet we still expect to find it?

src/components/Select/Select.jsx
```diff
@@ -62,6 +62,10 @@ class Select extends React.Component {
       requiredSpan = <span className="form-required-span">(*Required)</span>;
     }
 
+    const ariaDescribedby =
+      [errorSpanId, this.props.ariaDescribedby || ''].join(' ').trim() ||
```
bkjohnson, line 66: We have the error id here as well :thinking:

bkjohnson (commented):
A couple of questions.

src/components/Date/Date.stories.jsx
Mottie: No, not in the stories. I wasn't sure how to add it to the stories page.

src/components/NumberInput/NumberInput.jsx
Mottie: The `Date` component doesn't handle errors, it lets the `Select` and `NumberInput` components handle them.

src/components/NumberInput/NumberInput.unit.spec.jsx
Mottie: Yeah, that comment is bad... what I'm trying to say is that the `aria-describedby` shouldn't be added as an attribute if there are no errors or no `ariaDescribedby` prop. In this case, there is no error but there is an `ariaDescribedby` prop, so it still adds the attribute with _only_ the id passed into it. I'll fix the comment.

src/components/Select/Select.jsx
Mottie: The `errorSpanId` is only set as a non-empty string when there is a validation error. The `aria-describedby` attributes does allow multiple ids as long as they are space separated.

bkjohnson: I think it's strange that we're including the `errorSpanId` in the `Select` and `NumberInput` components, but not in `Date`. It seems like the id is there:

https://github.com/department-of-veterans-affairs/component-library/blob/59f39ca75d652ecbaa0389b590366275e598cd6a/src/components/Date/Date.jsx#L88

Mottie: Hmm, you're right... I'll add the `errorSpanId` to the `ariaDescribedby` prop passed to the elements... on Monday when I work on it again 😸

src/components/Date/Date.jsx
```diff
@@ -28,7 +27,7 @@ class Date extends React.Component {
   }
 
   UNSAFE_componentWillMount() {
-    this.id = _.uniqueId('date-input-');
+    this.inputId = _.uniqueId('date-input-');
```
Mottie, line 30: Changed this to `inputId` since the `errorSpanId` is using this name instead of `id` (line 87) - found when writing unit tests and the `errorSpanId` included `undefined` in the id.

bkjohnson (approved):

PR #105, opened 2021-06-07
Add va-select to storybook

bkjohnson:
## Description

Related to #99 

## Testing done

Storybook :eyes: 

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,9 +37,9 @@ workflows:
   publish-dev:
     jobs:
       - publish-storybook:
-          filters:
-            branches:
-              only: master
+          # filters:
+          #   branches:
+          #     only: master
           bucket: dev-design.va.gov
           context:
             - Platform
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,9 +37,9 @@ workflows:
   publish-dev:
     jobs:
       - publish-storybook:
-          # filters:
-          #   branches:
-          #     only: master
+          filters:
+            branches:
+              only: master
           bucket: dev-design.va.gov
           context:
             - Platform
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "build": "node ./scripts/build.js",
     "build-storybook": "build-storybook",
     "prepublishOnly": "pinst --disable && node ./scripts/prepublish.js",
-    "postinstall": "husky install && cd node_modules/web-components && npm install --only=prod && npm run-script build",
+    "postinstall": "husky install && cd node_modules/web-components && yarn install --production && npm run-script build",
     "postpublish": "pinst --enable",
     "storybook": "start-storybook -p 6006",
     "test-debug": "karma start testing/karma.conf.js --browsers Chrome",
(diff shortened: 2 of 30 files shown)


.circleci/config.yml
```diff
@@ -37,9 +37,9 @@ workflows:
   publish-dev:
     jobs:
       - publish-storybook:
-          filters:
-            branches:
-              only: master
+          # filters:
+          #   branches:
+          #     only: master
```
cvalarida, line 42: Reminder: Uncomment these lines.

package.json
```diff
@@ -97,6 +97,6 @@
     "sass-loader": "^8.0.2",
     "sinon": "^9.2.2",
     "uswds": "1.6.10",
-    "web-components": "https://github.com/department-of-veterans-affairs/component-library.git#wc-v0.6.1"
+    "web-components": "https://github.com/department-of-veterans-affairs/component-library.git#wc-select"
```
cvalarida, line 100: Reminder: Use the tagged version.

yarn.lock
```diff
@@ -1579,9 +1579,9 @@
   integrity sha512-+iTbntw2IZPb/anVDbypzfQa+ay64MW0Zo8aJ8gZPWMMK6/OubMVb6lUPMagqjOPnmtauXnFCACVl3O7ogjeqQ==
 
 "@stencil/core@^2.0.1":
-  version "2.5.2"
-  resolved "https://registry.yarnpkg.com/@stencil/core/-/core-2.5.2.tgz#ad00d495ee37bbed4044524d59c7f22de15ab4a7"
-  integrity sha512-bgjPXkSzzg1WnTgVUm6m5ZzpKt602WmA/QljODAW1xVN40OHJdbGblzF/F6MFzqv2c5Cy30CB41arc8qADIdcQ==
+  version "2.6.0"
```
cvalarida, line 1582: I mean, I'm fine with this, but should this be included? I recommend undoing these changes, updating the `web-components` dependency version in the `package.json` to use the latest tagged version like it should, and `yarn install` again to make sure that the changes to lock file committed here really should be there.

stories/va-select.stories.jsx
```diff
@@ -0,0 +1,91 @@
...
+          setModifiedOptions([
+            ...modifiedOptions,
+            <option key="6" value="new">
+              Something new
+            </option>,
+          ]);
+        }}
+      >
+        Change the options
+      </button>
```
cvalarida, line 68: Let's move this to a story rather than the template. We want to demonstrate the behavior of changing the options dynamically, but we don't need to do that for every story. It kind of detracts from showcasing the component itself.

cvalarida (commented):
The stories part looks great. I recommend moving the dynamic updates to a new story, though. I'll give it my stamp of approval after you take out the changes you made for testing. :wink:

stories/va-select.stories.jsx
bkjohnson: Due to the state variables, I left this in the template but it's conditionally rendered based on the arguments. What do you think about that?

cvalarida (approved):

PR #107, opened 2021-06-16
Va alert box content

mdewey:
## Description
Addresses feedback from our designer in [this ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/25938). Our designer found that if no content is supplied, then there is extra space below the header. This would give the developers the option to hide the content div if not needed. 

I will update the Story on the main branch in a separate PR if/when this gets approved. 

## Testing done
- added two unit tests

## Screenshots
before:
Screen Shot 2021-06-16 at 10 48 56 AM

After:
Screen Shot 2021-06-16 at 10 48 09 AM

## Acceptance criteria
- [ ] White space is removed

Let me know if I need anything else :-)

### DIFF
--- a/src/components/va-alert/va-alert.tsx
+++ b/src/components/va-alert/va-alert.tsx
@@ -28,6 +28,11 @@ export class VaAlert {
    */
   @Prop() backgroundOnly: boolean = false;
 
+    /**
+   * If true, hides the content slot and displays with only a header
+   */
+  @Prop() hideContent: boolean = false;
+
   /**
    * If true, doesn't fire the CustomEvent which can be used for analytics tracking.
    */
@@ -106,11 +111,11 @@ export class VaAlert {
   }
 
   render() {
-    const { backgroundOnly, status, visible, closeable } = this;
+    const { backgroundOnly, status, visible, closeable , hideContent} = this;
     const classes = `alert ${status} ${backgroundOnly ? 'bg-only' : ''}`;
 
-    if (!visible) return <div aria-live="polite" />;
-
+    if (!visible) return <div aria-live="polite" />; 
+    
     return (
       <Host>
         <div class={classes}>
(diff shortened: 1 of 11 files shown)


src/components/va-alert/va-alert.tsx
```diff
@@ -106,21 +111,21 @@ export class VaAlert {
...
+    
     return (
       <Host>
         <div class={classes}>
           <i aria-hidden="true" role="img"></i>
           <span class="sr-only">Alert: </span>
           <div class="body" onClick={this.handleAlertBodyClick.bind(this)}>
             {!backgroundOnly && <slot name="headline"></slot>}
-            <div class="text">
+            {!hideContent && <div class="text">
```
bkjohnson, line 126: I think we can simplify this by not adding the extra `hideContent` prop. Since the `text` class only served as a target for the CSS, we can remove that div entirely and change how we apply the margin, like so:

image


Now, if I put this on a page:

```html
    <va-alert status="success">
      <h3 slot="headline">Congrats, you did it!</h3>
    </va-alert>
    <br />
    <va-alert status="success" background-only>
      <div>Some content</div>
    </va-alert>
 ```
 
 I get what we expect:
 
 
image

This implementation works in IE11 as well thanks to the CSS polyfills:

signal-2021-06-16-104548

bkjohnson (changes requested):

src/components/va-alert/va-alert.tsx
mdewey: That works for me @bkjohnson! I will update this PR to be just those changes :-)

bkjohnson (commented):
Make sure you remove `<div class="text">` wrapping the `<slot>`, otherwise the spacing change won't look correct on IE.

bkjohnson (approved):
Before you merge, do a `yarn version --patch` to bump the version. You can let me know when it's merged and I'll take care of tagging the release.

bkjohnson:
You're welcome to tag the release as well - just follow step 3 here: https://github.com/department-of-veterans-affairs/component-library#web-components

mdewey:
@bkjohnson  I tried to publish to npm, but I do not have access to npm.

bkjohnson:
We don't publish this branch to npm at the moment. I will take care of making the release.

PR #108, opened 2021-06-17
Va text input cv

cvalarida:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/25142

## Testing done
- [x] Automated tests written
- [x] Manually verified styles

## Screenshots
**Normal**
image

**Required**
image

**Error**
image

**Error with `<body>` highlighted**
Because the red bar on the left extends outside of the parent node.
image

## Acceptance criteria
See the test cases written for a list of AC.

### DIFF
--- a/src/components/va-text-input/test/va-text-input.e2e.ts
+++ b/src/components/va-text-input/test/va-text-input.e2e.ts
@@ -99,7 +99,22 @@ describe('va-text-input', () => {
     );
   });
 
-  it('adds adds a character limit with descriptive text', async () => {});
+  it('adds adds a character limit with descriptive text', async () => {
+    const page = await newE2EPage();
+    await page.setContent('<va-text-input maxlength="3" value="22"/>');
+
+    // Level-setting expectations
+    const inputEl = await page.find('va-text-input >>> input');
+    expect(await inputEl.getProperty('value')).toBe('22');
+    expect(await page.find('va-text-input >>> small')).toBeNull();
+
+    // Test the functionality
+    await inputEl.press('2');
+    expect(await inputEl.getProperty('value')).toBe('222');
+    expect((await page.find('va-text-input >>> small')).innerText).toContain(
+      '(Max. 3 characters)',
+    );
+  });
 
   it('passes unknown props to the input element in the shadow DOM', async () => {
     // This is primarily so we don't have to make a new prop for each aria-* attribute
--- a/src/components/va-text-input/test/va-text-input.e2e.ts
+++ b/src/components/va-text-input/test/va-text-input.e2e.ts
@@ -86,7 +86,20 @@ describe('va-text-input', () => {
 
   it('fires an analytics event (when?)', () => {});
 
-  it('adds placeholder text', async () => {});
+  it('adds placeholder text', async () => {
+    const page = await newE2EPage();
+    await page.setContent(
+      '<va-text-input placeholder="Enter your life story" />',
+    );
+
+    // Render the error message text
+    const inputEl = await page.find('va-text-input >>> input');
+    expect(inputEl.getAttribute('placeholder')).toContain(
+      'Enter your life story',
+    );
+  });
+
+  it('adds adds a character limit with descriptive text', async () => {});
 
   it('passes unknown props to the input element in the shadow DOM', async () => {
     // This is primarily so we don't have to make a new prop for each aria-* attribute
--- a/src/components/va-text-input/test/va-text-input.e2e.ts
+++ b/src/components/va-text-input/test/va-text-input.e2e.ts
@@ -38,8 +38,8 @@ describe('va-text-input', () => {
     await page.setContent('<va-text-input error="This is a mistake" />');
 
     // Render the error message text
-    const error = await page.find('va-text-input >>> input');
-    expect(error.getAttribute('aria-describedby')).toContain('error-message');
+    const inputEl = await page.find('va-text-input >>> input');
+    expect(inputEl.getAttribute('aria-describedby')).toContain('error-message');
   });
 
   it('appends to an existing aria-describedby for error message', async () => {
--- a/src/components/va-text-input/test/va-text-input.e2e.ts
+++ b/src/components/va-text-input/test/va-text-input.e2e.ts
@@ -11,7 +11,9 @@ describe('va-text-input', () => {
     expect(element).toEqualHtml(`
       <va-text-input class="hydrated">
         <mock:shadow-root>
-          <input id="inputField" type="text" />
+          <div class="input-wrapper">
+            <input id="inputField" type="text" />
+          </div>
         </mock:shadow-root>
       </va-text-input>
     `);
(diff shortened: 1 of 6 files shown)


src/components/va-text-input/va-text-input.tsx
```diff
@@ -0,0 +1,148 @@
...
+        (atts['aria-describedby'] || '') + ' error-message'
+      ).trim();
+    }
+    return (
+      <Host>
+        <div class={this.error ? 'input-wrapper input-error' : 'input-wrapper'}>
+          {this.label && (
+            <label htmlFor="inputField">
+              {this.label}{' '}
+              {this.required && <span class="required">(Required)</span>}
```
bkjohnson, line 103:
```suggestion
              {this.required && <span class="required">(*Required)</span>}
```

src/components/va-text-input/va-text-input.tsx
```diff
@@ -0,0 +1,148 @@
...
+  render() {
+    const atts = assembleAttributes(this.el.attributes);
+    if (this.error) {
+      atts['aria-describedby'] = (
+        (atts['aria-describedby'] || '') + ' error-message'
+      ).trim();
+    }
+    return (
+      <Host>
+        <div class={this.error ? 'input-wrapper input-error' : 'input-wrapper'}>
```
bkjohnson, line 99: I think you might be able to get rid of this wrapping div entirely. In the CSS you could use an attribute selector like this:

https://github.com/department-of-veterans-affairs/component-library/blob/d32e14d863547300980b4e3ec982dd79d79a4c63/src/components/va-select/va-select.css#L13

And move the `input-error` styles into that.

You could also replace `input-wrapper` with a `:host` selector in the CSS which might remove the need for this div.

src/components/va-text-input/va-text-input.tsx
```diff
@@ -0,0 +1,148 @@
...
+// input element. Note: The following attributes are deliberately left out so
+// they get passed to the input:
+//   - required
+//   - placeholder
+//   - maxlength
+//   - autocomplete
+//   - value
+const wcPropNames = ['label', 'error', 'disableAnalytics'];
+
+const assembleAttributes = (atts: NamedNodeMap) =>
```
bkjohnson, line 133: If we want to do this again, let's remember to move this into a `utils` folder or something like that and add some unit tests.

bkjohnson (commented):
Looks like there are a few failing tests.

src/components/va-text-input/va-text-input.tsx
cvalarida: Eh, for consistency, I probably should. I did deliberately leave it out because we'll be removing it later, but I'll leave it in for now. I don't know if there's a ticket for this yet, but this came up in a support channel request the other day: https://dsva.slack.com/archives/CBU0KDSB1/p1623354715388700

cvalarida:
Oh, heh. I fixed those tests, but never committed the fix. That's what's up there.

src/components/va-text-input/va-text-input.tsx
cvalarida: Yeah, we probably will want to do this again. That's a good call. To keep this one smaller, I'll leave it alone for now, but follow up with another PR to move this out and add tests.

src/components/va-text-input/va-text-input.tsx
cvalarida: Ahhhh, `:host([error])`. I like it! :shipit:

src/components/va-text-input/va-text-input.css
```diff
@@ -0,0 +1,58 @@
...
+    border-left: 4px solid var(--color-secondary-dark);
+    margin-top: 3rem;
+    padding-bottom: .8rem;
+    padding-left: 1.5rem;
+    padding-top: .8rem;
+    position: relative;
+    right: 1.9rem;
+}
+
+:host([error]) label, .input-error span {
```
bkjohnson, line 52: This `.input-error` seems like a leftover.

src/components/va-text-input/va-text-input.tsx
```diff
@@ -0,0 +1,146 @@
...
+   */
+  @Prop() disableAnalytics?: boolean;
+
+  /*
+   * The value for the input.
+   */
+  @Prop({ mutable: true }) value?: string = '';
+  // TODO: Make the value prop reflective. Currently, it isn't because it screws
+  // up the input behavior. For now, the only bug is that the changed value
+  // isn't reflected in the DOM on the web component.
```
bkjohnson, line 67: Should we link to the MDN docs where it says that an `<input` value isn't _meant_ to be reflective?

src/components/va-text-input/va-text-input.tsx
```diff
@@ -0,0 +1,146 @@
...
+  private handleChange = (e: Event) => {
+    const target = e.target as HTMLInputElement;
+    this.value = target.value;
+  };
+
+  private handleBlur = () => {
+    if (this.disableAnalytics) return;
+
+    this.componentLibraryAnalytics.emit({
+      componentName: 'va-text-input',
```
bkjohnson, line 81: :thinking: I don't know what the google analytics side of this looks like, but I've been making `componentName` the same as the React implementations. Maybe we should get clarification on if this is automatic or if something has to be manually created for each new `componentName`.

src/components/va-text-input/test/va-text-input.e2e.ts
```diff
@@ -0,0 +1,186 @@
...
+
+    // Test the functionality
+    await inputEl.press('2');
+    expect(await inputEl.getProperty('value')).toBe('222');
+    expect((await page.find('va-text-input >>> small')).innerText).toContain(
+      '(Max. 3 characters)',
+    );
+  });
+
+  it('passes unknown props to the input element in the shadow DOM', async () => {
```
bkjohnson, line 167: The `componentName` thing reminded me to take a gander at some props. We don't have a `name` prop specifically, but that little helper should automatically pass any additional props/attrs onto the `input`. However, _should_ we have a `name` prop? That may be useful when it comes time to get the data out of the form. We can do `myEl.value`, but we won't have anything to match it to.

bkjohnson (commented):
Some more thoughts!

src/components/va-text-input/va-text-input.css
cvalarida: Dang flabbit. I did a `:%s/\.input-error/:host([error])/` without giving it the `g` on the end to match all occurrences in a line. :roll_eyes: 

Thanks.

src/components/va-text-input/va-text-input.tsx
cvalarida: Ah, right. @bsmartin-ep Are you the one we talk to about this?

src/components/va-text-input/test/va-text-input.e2e.ts
cvalarida: I discovered that `inputEl.name` returned the name, but `vaTextInputEl.name` did not. `vaTextInputEl.getAttribute('name')` returned what we expect, though. :man_shrugging: 

For consistency (and perhaps clearer docs), I just added `name` as a prop. :+1:

bkjohnson (approved):
:shipit:

src/components/va-text-input/va-text-input.tsx
cvalarida: I just merged this, but we can change the `componentName` here if we need to in another PR.

bsmartin-ep: Sorry @cvalarida and @bkjohnson - was out all week.

> Maybe we should get clarification on if this is automatic or if something has to be manually created for each new componentName.

Yes indeed.  New components need to be [added to this lookup table](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/platform/site-wide/component-library-analytics-setup.js).  

Any objections to matching the react component name?  Is is inaccurate to use that for a web component?

PR #109, opened 2021-06-21
Resolve trim-newlines vulnerability

cvalarida:
## Description
Like it says on the tin, there was a vulnerability with the version of `trim-newlines` we were using, so this pins it to `^4.0.2`, which is not vulnerable to the RegExp DOS attack that `1.0.0` is.

`trim-newlines` is a dependency of `meow` which is a dependency of `node-sass`. I looked into updating `node-sass`, but it uses the same version of `meow` as the version we're on now, so that was a dead end.

## Testing done
None :see_no_evil:
If this doesn't work as expected, I'm not sure how we'd figure it out unless it breaks spectacularly. :shrug:

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "build": "node ./scripts/build.js",
     "build-storybook": "build-storybook",
     "prepublishOnly": "pinst --disable && node ./scripts/prepublish.js",
-    "postinstall": "husky install && cd node_modules/web-components && npm install --only=prod && npm run-script build",
+    "postinstall": "husky install && cd node_modules/web-components && yarn install --production && npm run-script build",
     "postpublish": "pinst --enable",
     "storybook": "start-storybook -p 6006",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -19,7 +19,7 @@
     "build": "node ./scripts/build.js",
     "build-storybook": "build-storybook",
     "prepublishOnly": "pinst --disable && node ./scripts/prepublish.js",
-    "postinstall": "husky install && cd node_modules/web-components && npm install --only=prod && npm run-script build",
+    "postinstall": "husky install && cd node_modules/web-components && yarn install --production && npm run-script build",
```
bkjohnson, line 22: For any future :detective:  - this is an unrelated fix.

bkjohnson (approved):
Since `node-sass` is something we're using in the build and not during runtime, I'm not sure how we would be affected, but:

PR #111, opened 2021-06-23
va-text-input stories

cvalarida:
## Description
Adds the stories for `va-text-input`.

After much struggling with getting the funky default value to work like I wanted to, I gave up and settled on adding every bloody prop to the `defaultArgs`. :roll_eyes: It's still a bit funky, but it's not _awful._

## Testing done
:eyes: 

## Screenshots
image
And more! But they're basically all like that. Very exciting.

### DIFF
--- a/stories/va-text-input.stories.jsx
+++ b/stories/va-text-input.stories.jsx
@@ -18,11 +18,12 @@ const defaultArgs = {
   name: 'my-input',
   label: 'My input',
   autocomplete: false,
-  'disable-analytics': true,
-  // Fun fact: This should really be false, but that passes required="false" to
-  // the web component, which correctly interprets this as a truthy value.
-  // See also:
+  // Fun fact: This should really be false, but that passes
+  // enable-analytics="false" to the web component, which correctly interprets
+  // this as a truthy value. See also:
   // https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attribute
+  'enable-analytics': null,
+  // Same story here
   required: null,
   error: null,
   maxlength: null,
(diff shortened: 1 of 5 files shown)


stories/va-text-input.stories.jsx
```diff
@@ -0,0 +1,66 @@
...
+const defaultArgs = {
+  name: 'my-input',
+  label: 'My input',
+  autocomplete: false,
+  'disable-analytics': true,
+  // Fun fact: This should really be false, but that passes required="false" to
+  // the web component, which correctly interprets this as a truthy value.
+  // See also:
+  // https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attribute
+  required: null,
```
bkjohnson, line 26: I think we can make some changes so that `required=false` will behave as the value would suggest. Is this specific to the CSS?

bkjohnson (approved):

bkjohnson (commented):
Actually, before we merge this we may want to mark `TextInput` as deprecated in storybook.

stories/va-text-input.stories.jsx
cvalarida: Well, the web component will see `this.required` as true, so it'll add the `span.required`. Not just CSS, no.

cvalarida: The trouble here is that Storybook adds the attribute to the `<va-text-input>` in the DOM when it shouldn't, according to that WHATWG spec.

bkjohnson: Oh, is this just a storybook problem? In other JSX would we be able to have `required={false}` and have it be ok?

cvalarida: Yeah, exactly. `required={false}` is okay, but `required="false"` is wrong.

PR #114, opened 2021-06-24
disableAnalytics -> enableAnalytics on va-text-input

cvalarida:
## Description
Changes the disableAnalytics prop to enableAnalytics so it doesn't fire the events by default. This matches the current behavior of `TextInput`.

## Testing done
Automated tests.

## Screenshots
N/A

## Acceptance criteria
- [x] The events don't fire by default
- [x] The events do fire when the `enableAnalytics` prop is passed to the component

### DIFF
--- a/src/components/va-text-input/test/va-text-input.e2e.ts
+++ b/src/components/va-text-input/test/va-text-input.e2e.ts
@@ -84,7 +84,7 @@ describe('va-text-input', () => {
     await axeCheck(page);
   });
 
-  it('fires an analytics event', async () => {
+  it('fires an analytics event when enableAnalytics is true', async () => {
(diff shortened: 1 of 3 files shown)


src/components/va-text-input/test/va-text-input.e2e.ts
```diff
@@ -87,7 +87,9 @@ describe('va-text-input', () => {
   it('fires an analytics event', async () => {
```
bkjohnson, line 87:
```suggestion
  it('fires an analytics event when enableAnalytics is true', async () => {
  ```

bkjohnson (approved):

bkjohnson (approved):

PR #118, opened 2021-07-08
Remove unknown attributes spread cv

cvalarida:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/27033

I also removed the `required` from the `input` on the `va-text-input` test. We may add it back in there, but it doesn't technically do that in the React version. I don't know whether or not it should, but I've asked our a11y specialists and will write up an issue to add it back in if we need it.

## Testing done
Unit tests pass.

### DIFF
--- a/src/components.d.ts
+++ b/src/components.d.ts
@@ -79,6 +79,7 @@ export namespace Components {
         "visible": boolean;
     }
     interface VaCheckbox {
+        "ariaDescribedby": string;
         /**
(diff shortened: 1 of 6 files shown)


bkjohnson (commented):
Shouldn't we also remove the function from `utils`?

cvalarida:
Ah! Good call. I meant to do that but forgot. :+1:

bkjohnson (approved):

PR #120, opened 2021-07-13
Add radio button web component

bkjohnson:
## Description
Closes <ticket>

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/src/components/va-radio-option/test/va-radio-option.e2e.ts
+++ b/src/components/va-radio-option/test/va-radio-option.e2e.ts
@@ -8,4 +8,15 @@ describe('va-radio-option', () => {
     const element = await page.find('va-radio-option');
     expect(element).toHaveClass('hydrated');
   });
+
+  it('sets checked to true based on prop', async () => {
+    const page = await newE2EPage();
+    await page.setContent(
+      '<va-radio-option checked label="An option"></va-radio-option>',
+    );
+
+    const inputEl = await page.find('va-radio-option >>> input');
+
+    expect(await inputEl.getProperty('checked')).toBeTruthy();
+  });
 });
--- a/src/components/va-radio-option/test/va-radio-option.e2e.ts
+++ b/src/components/va-radio-option/test/va-radio-option.e2e.ts
@@ -1,0 +1,11 @@
+import { newE2EPage } from '@stencil/core/testing';
+
+describe('va-radio-option', () => {
+  it('renders', async () => {
+    const page = await newE2EPage();
+    await page.setContent('<va-radio-option></va-radio-option>');
+
+    const element = await page.find('va-radio-option');
+    expect(element).toHaveClass('hydrated');
+  });
+});
--- a/src/components.d.ts
+++ b/src/components.d.ts
@@ -110,6 +110,16 @@ export namespace Components {
     }
     interface VaOnThisPage {
     }
+    interface VaRadioButton {
+        "label": string;
+        "required": boolean;
+    }
+    interface VaRadioOption {
+        "checked": boolean;
+        "label": string;
+        "name": string;
+        "value": string;
+    }
     interface VaSelect {
         /**
           * Whether or not to fire the analytics events
@@ -213,6 +223,18 @@ declare global {
         prototype: HTMLVaOnThisPageElement;
         new (): HTMLVaOnThisPageElement;
     };
+    interface HTMLVaRadioButtonElement extends Components.VaRadioButton, HTMLStencilElement {
+    }
+    var HTMLVaRadioButtonElement: {
+        prototype: HTMLVaRadioButtonElement;
+        new (): HTMLVaRadioButtonElement;
+    };
+    interface HTMLVaRadioOptionElement extends Components.VaRadioOption, HTMLStencilElement {
+    }
+    var HTMLVaRadioOptionElement: {
+        prototype: HTMLVaRadioOptionElement;
+        new (): HTMLVaRadioOptionElement;
+    };
     interface HTMLVaSelectElement extends Components.VaSelect, HTMLStencilElement {
     }
     var HTMLVaSelectElement: {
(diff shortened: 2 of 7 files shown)


src/components/va-radio/va-radio.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  @Listen('radioOptionSelected')
+  radioOptionSelectedHandler(event: CustomEvent): void {
+    const clickedItem = event.target as HTMLVaRadioOptionElement;
+
+    getSlottedNodes(this.el, 'va-radio-option')
+      .filter(item => item !== clickedItem)
+      .forEach(item => ((item as HTMLVaRadioOptionElement).checked = false));
+
+    clickedItem.checked = true;
```
bkjohnson, line 62: Because each `<input>` is within its own shadow DOM, we lose the benefit of the browser being able to automatically connect them by `name`. This ensure that when one option is checked, the others are unchecked.

bkjohnson (commented):
I will still need to add comments to the props, but this should be ready for a first look.

src/components/va-radio/test/va-radio.e2e.ts
```diff
@@ -0,0 +1,145 @@
...
+      <va-radio label="Regular label">
+        <va-radio-option label="First option" value="one"></va-radio-option>
+      </va-radio>
+      `);
+
+    const changeSpy = await page.spyOnEvent('vaValueChange');
+    const inputEl = await page.find('va-radio-option >>> input');
+    await inputEl.click();
+
+    expect(changeSpy).toHaveReceivedEvent();
```
cvalarida, line 127: Can you include the event details here?

src/components/va-radio/test/va-radio.e2e.ts
```diff
@@ -0,0 +1,145 @@
...
+  });
+
+  it('fires a vaKeyDown event when a key is pressed on focused button', async () => {
```
cvalarida, line 130: This is maybe more for my edification, but why does this need to happen? What will the `vaKeyDown` event be used for?

src/components/va-radio/va-radio.tsx
```diff
@@ -0,0 +1,107 @@
...
+  })
+  vaMouseDown: EventEmitter;
+
+  @Listen('radioOptionSelected')
+  radioOptionSelectedHandler(event: CustomEvent): void {
+    const clickedItem = event.target as HTMLVaRadioOptionElement;
+
+    getSlottedNodes(this.el, 'va-radio-option')
+      .filter(item => item !== clickedItem)
+      .forEach(item => ((item as HTMLVaRadioOptionElement).checked = false));
```
cvalarida, line 60: :thinking: Is there a simple way to either ensure each slotted node is a radio button _or_ filter the list of slotted nodes by type to only grab the radio buttons so we can be sure each item _is_ in fact an `HTMLVaRadioOptionElement`?

cvalarida (commented):
This looks great. I had a few questions before approval, though.

src/components/va-radio-option/test/va-radio-option.e2e.ts
```diff
@@ -0,0 +1,39 @@
...
+    const inputEl = await page.find('va-radio-option >>> input');
+
+    expect(inputEl.getAttribute('aria-describedby')).toEqual(null);
+
+    element.setProperty('checked', true);
+    await page.waitForChanges();
+
+    expect(inputEl.getAttribute('aria-describedby')).toEqual('test');
+  });
+});
```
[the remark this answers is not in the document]
bkjohnson: Sure.

src/components/va-radio/test/va-radio.e2e.ts
bkjohnson: I'm not really sure... I was just modeling this component after what we already have:

https://github.com/department-of-veterans-affairs/component-library/blob/d71c50001f14751acd5f2b4f1bda4301ef90f535/src/components/RadioButtons/RadioButtons.jsx#L226-L229

src/components/va-radio/va-radio.tsx
bkjohnson: I don't think we can put limits on which slotted nodes are _allowed_. The closest thing might be to have a `::slotted(*) { display: none;}` followed up by `::slotted(va-radio-option) { display: block;}`, but that's just for presentation.

Regarding the filtering by type, we might be able to do that here if I'm understanding what you're asking:

https://github.com/department-of-veterans-affairs/component-library/blob/4ee6bb1b0dfe33f029d13b8ec5cd10b61bffdcb3/src/utils/utils.ts#L27-L29

I'll see what a refactor looks like based on type as opposed to the node name.

cvalarida: :man_facepalming: I think what I'm actually asking is for `getSlottedNodes` to do what it already does. I totally didn't read the part where this _is_ getting all `va-radio-option`s. That's my bad. You're good here. :sweat_smile:

src/components/va-radio/test/va-radio.e2e.ts
bkjohnson: Putting this here for posterity after slacking.

---

Did some more digging. It looks like at the moment we only have one usage of `<RadioButtons>` with `onKeyDown`:

image

That usage looks like this:

https://github.com/department-of-veterans-affairs/vets-website/blob/b87df585fae073fa2ac4b7d3eb504cf15ea16339/src/applications/coronavirus-vaccination-expansion/containers/IntroductionPage.jsx#L83-L86

Which appears to be a copypaste from storybook:

image


So, I think it's safe to delete the `vaKeyDown` event.

src/components/va-radio/test/va-radio.e2e.ts
```diff
@@ -0,0 +1,129 @@
...
+      <va-radio label="Regular label">
+        <va-radio-option label="First option" value="one"></va-radio-option>
+      </va-radio>
+      `);
+
+    const changeSpy = await page.spyOnEvent('vaValueChange');
+    const inputEl = await page.find('va-radio-option >>> input');
+    await inputEl.click();
+
+    expect(changeSpy).toHaveReceivedEventDetail({ value: 'one' });
```
bkjohnson, line 127: This is based on the current handler receiving an object that looks like `{ value, dirty}`. I'm not returning a `dirty` key because based on observation and conversation, that's something that we've decided to get away from.

cvalarida (approved):
:shipit: Looks great!

PR #121, opened 2021-07-15
Build during the prepare script

cvalarida:
## Description
This builds the library when it's [installed via a git URL](https://docs.npmjs.com/cli/v7/using-npm/scripts#life-cycle-scripts). With this addition, we won't have to require the downstream project to have that awful `postinstall` hook. 

## Testing done
Tested this on a local project installing this branch as a dependency via a [git URL](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#git-urls-as-dependencies).

## Screenshots
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,8 @@
     "generate": "stencil generate",
     "storybook": "start-storybook -p 6006",
     "serve": "stencil build --dev --watch --serve",
-    "build-storybook": "build-storybook"
+    "build-storybook": "build-storybook",
+    "prepare": "npm run build"
   },
   "dependencies": {
     "@stencil/core": "^2.0.1",


package.json
```diff
@@ -23,7 +23,8 @@
     "generate": "stencil generate",
     "storybook": "start-storybook -p 6006",
     "serve": "stencil build --dev --watch --serve",
-    "build-storybook": "build-storybook"
+    "build-storybook": "build-storybook",
+    "prepare": "npm run build"
```
bkjohnson, line 27: Are we going to have problems here using `npm` instead of `yarn`?

bkjohnson (approved):
Yay for no more awful `postinstall`!

package.json
cvalarida: No, I don't think so. This just runs the script. :man_shrugging:

bkjohnson: I think I was remembering that `npm link` that we were working with and how it _insisted_ on doing an `npm install` :sweat_smile:

PR #124, opened 2021-07-21
Allow multiple accordion items open by default

bkjohnson:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/27198

## Testing done

Storybook testing & stencil dev server

Also, E2E tests

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,8 @@
   "unpkg": "dist/component-library/component-library.esm.js",
   "files": [
     "dist/",
-    "loader/"
+    "loader/",
+    "tsconfig.json"
   ],
   "scripts": {
     "build": "stencil build --docs-json component-docs.json",
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "files": [
     "dist/",
     "loader/",
-    "tsconfig.json"
+    "component-docs.json"
   ],
   "scripts": {
     "build": "stencil build --docs-json component-docs.json",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -12,7 +12,8 @@
   "unpkg": "dist/component-library/component-library.esm.js",
   "files": [
     "dist/",
-    "loader/"
+    "loader/",
+    "component-docs.json"
```
bkjohnson, line 16: We need this file so that Storybook can properly create some documentation

src/components/va-accordion/va-accordion.e2e.ts
```diff
@@ -29,10 +29,10 @@ describe('va-accordion', () => {
     await axeCheck(page);
   });
 
-  it('closes one item when another opens', async () => {
+  it('closes one item when another opens if `multi` is false', async () => {
     const page = await newE2EPage();
     await page.setContent(`
-      <va-accordion>
+      <va-accordion multi="false">
```
cvalarida, line 35: I typically prefer truthy props. So rather than `multi="false"`, I'd recommend `open-single` or something.

cvalarida (approved):
This looks fine. I'd prefer to not have true-by-default props overridden by explicitly passing false, but I think that's just a personal preference. I don't foresee any issue with this. :shipit:

src/components/va-accordion/va-accordion.e2e.ts
bkjohnson: That seems reasonable - I'll make the change.

cvalarida (approved):

PR #129, opened 2021-07-27
Add section heading to va-accordion web component

bsmartin-ep:
## Description
For https://github.com/department-of-veterans-affairs/va.gov-team/issues/27847

> The transition from us-accordion to the va-accordion has revealed a gap in the latter.
> 
> We need to add the 'section-label' as a component parameter (for analytics only) and back out the dataLayer injection Ryan Leahy added to vets-website.

## Testing done

Added "Accordion Section Heading" story.

## Screenshots

image

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/src/components/va-accordion/va-accordion.e2e.ts
+++ b/src/components/va-accordion/va-accordion.e2e.ts
@@ -103,6 +103,7 @@ describe('va-accordion', () => {
         header: "First item",
         subheader: "First subheader",
         level: 2,
+        sectionHeading: null,
       },
     });
   });
--- a/src/components/va-accordion/va-accordion.tsx
+++ b/src/components/va-accordion/va-accordion.tsx
@@ -68,6 +68,8 @@ export class VaAccordion {
           header: clickedItem.header,
           subheader: clickedItem.subheader,
           level: clickedItem.level,
+          sectionHeading: this.sectionheading,
+          bordered: this.bordered
         },
       };
       this.componentLibraryAnalytics.emit(detail);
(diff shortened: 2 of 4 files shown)


src/components/va-accordion/va-accordion.e2e.ts
```diff
@@ -103,6 +103,7 @@ describe('va-accordion', () => {
         header: "First item",
         subheader: "First subheader",
         level: 2,
+        sectionHeading: null,
```
bkjohnson, line 106: Can we also have a test showing that the prop gets mapped the the event correctly?

bsmartin-ep: Sure thing.

src/components/va-accordion/va-accordion.tsx
```diff
@@ -68,6 +68,8 @@ export class VaAccordion {
           header: clickedItem.header,
           subheader: clickedItem.subheader,
           level: clickedItem.level,
+          sectionHeading: this.sectionHeading,
+          bordered: this.bordered
```
bkjohnson, line 72: Is including `bordered` here intentional?

bsmartin-ep: Weird.  It was not.  Removed!

bkjohnson (approved):
If this is semi-urgent you'll want to increase the version so we can tag it with a release. Also, since this is so minor and it doesn't involve anything visual, I don't think we need to have a storybook story for this in the main branch (what you have here is fine for testing).

bkjohnson (approved):

PR #131, opened 2021-08-06
Create react bindings

bkjohnson:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/27691

This creates React bindings for our Web Components which can then be imported and used. The main reason for this is so we can get the `onEvent` prop syntax that engineers are familiar with from working in React. With Web Components alone, we can't do that inside React.

## Testing done

Used `yarn link` to point `vets-website` to my local version of `web-components`, then tested using this code:

image

`change` gets logged anytime I typed into the `VaTextInput`

## Screenshots

In the `Components` devtools tab, I can see the `VaTextInput`:

image

And in the inspector, we see the `<va-text-input>`:

image

## Acceptance criteria
- [ ]

## Definition of done


tsconfig.bindings.json
```diff
@@ -0,0 +1,25 @@
...
+    "noImplicitReturns": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "outDir": "bindings",
+    "removeComments": false,
+    "sourceMap": true,
+    "jsx": "react",
+    "target": "es2015"
+  },
+  "include": ["react-bindings/**/*.ts", "react-bindings/**/*.tsx"],
```
bkjohnson, line 22: This entire file was copied from a template, and the only changes are removing the `exclude`, changing the directories in `include`, and changing the `outDir`.

https://github.com/ionic-team/stencil-ds-plugins-demo/blob/04f0108046c02ca494dd6551b1dd822d16b87b8d/packages/component-library-react/tsconfig.json#L22-L25

stencil.config.ts
```diff
@@ -24,6 +25,11 @@ export const config: Config = {
     slotChildNodesFix: true,
   },
   outputTargets: [
+    reactOutputTarget({
+      componentCorePackage: '../dist/types',
```
bkjohnson, line 29: This is named in order to satisfy a generated import, and because we are not using a monorepo. If we set the value to `component-library`, this is the error we get when we compile the bindings:

image

The built file `dist/types/index.d.ts` has the export we need:

image

stencil.config.ts
```diff
@@ -24,6 +25,11 @@ export const config: Config = {
...
+      proxiesFile: './react-bindings/components.ts',
```
bkjohnson, line 30: This is the output for the generated _source_ typescript React bindings. We still need to compile them to JS in order for them to be used elsewhere.

stencil.config.ts
```diff
@@ -24,6 +25,11 @@ export const config: Config = {
...
+      includeDefineCustomElements: false,
```
bkjohnson, line 31: I'm going against the default here because:

1. We are already [calling `defineCustomElements` in `vets-website`](https://github.com/department-of-veterans-affairs/vets-website/blob/294c69152bdf6728207091e18564b83b7dae7184/src/platform/site-wide/wc-loader.js#L5-L7) so that the plain Web Components will work.
1. Related to my earlier comment on the `componentCorePackage` - if this is true, then we try to import from `${componentCorePackage}/loader`, and `dist` and `loader` are siblings rather than parent/child

package.json
```diff
@@ -17,6 +17,7 @@
   ],
   "scripts": {
     "build": "stencil build --docs-json component-docs.json",
+    "build-bindings": "tsc -p tsconfig.bindings.json",
```
bkjohnson, line 20: Modeled after the scripts in the [sample template](
https://github.com/ionic-team/stencil-ds-plugins-demo/blob/04f0108046c02ca494dd6551b1dd822d16b87b8d/packages/component-library-react/package.json#L12-L15).

bkjohnson (commented):
I'm not sold on the `bindings` and `react-bindings` names (they were placeholders during testing attempts) - open to any and all suggestions!

We'll probably also want to add `build-bindings` to the `prepare` script.

react-bindings.js
```diff
@@ -0,0 +1 @@
+export * from './generated/bindings/components';
```
bkjohnson, line 1: Having this means an improved developer experience:

```js
import { VaTextInput } from 'web-components/react-bindings';
```

stencil.config.ts
```diff
@@ -24,6 +25,11 @@ export const config: Config = {
     slotChildNodesFix: true,
   },
   outputTargets: [
+    reactOutputTarget({
+      componentCorePackage: '../../dist/types',
```
bkjohnson, line 29: This is named in order to satisfy a generated import, and because we are not using a monorepo. If we set the value to `component-library`, this is the error we get when we compile the bindings:

image

The built file `dist/types/index.d.ts` has the export we need:

image

stencil.config.ts
```diff
@@ -24,6 +25,11 @@ export const config: Config = {
...
+      proxiesFile: './generated/react-bindings/components.ts',
```
bkjohnson, line 30: This is the output for the generated _source_ typescript React bindings. We still need to compile them to JS in order for them to be used elsewhere.

tsconfig.bindings.json
```diff
@@ -0,0 +1,28 @@
...
+    "outDir": "generated/bindings",
+    "removeComments": false,
+    "sourceMap": true,
+    "jsx": "react",
+    "target": "es2015"
+  },
+  "include": [
+    "generated/react-bindings/**/*.ts",
+    "generated/react-bindings/**/*.tsx"
+  ],
```
bkjohnson, line 25: This entire file was copied from a template, and the only changes are removing the `exclude`, changing the directories in `include`, and changing the `outDir`.

https://github.com/ionic-team/stencil-ds-plugins-demo/blob/04f0108046c02ca494dd6551b1dd822d16b87b8d/packages/component-library-react/tsconfig.json#L22-L25

bkjohnson (commented):
Re-applying some outdated comments so they show up in the `Files` tab.

cvalarida (commented):
I haven't tested it, but this looks great to me. Is there any way we can write tests for these bindings? Like, just a smoke test to make sure the build script is working as intended.

I think we'll also want to add `npm run build-bindings` to the `prepare` script.

Alternatively, change the `build` script to something like `"build-wc": "stencil build --docs-json component-docs.json"` and then make a new `"build": "npm run build-wc && npm run build-bindings"`

bkjohnson:
I'm not sure what a test for this might look like. Having a test to make sure _each_ React component exists sounds like a chore to keep updated.

Would a test that verifies that _something_ is being exported from the generated `components` file be what you're looking for?

cvalarida:
Yeah, exactly. Just to make sure the build is working as expected. If it's infeasible for a quick discovery spike, I'd say don't worry about it for now. I'd just like to verify the build if it's possible (simply).

cvalarida (approved):

PR #134, opened 2021-08-10
Scroll va-accordion-item into view when not in viewport

Rleahy22:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/28506

See slack issue: https://dsva.slack.com/archives/C52CL1PKQ/p1628617238133100
@RLHecht found an issue where accordion items with large chunks of text in the content block will expand outside of the viewport. This is a result of the large amount of vertical space taken up by the content so when a previously expanded item is collapsed and the new item is expanded the browser viewport stays at the same y coordinate while the accordion item is now displayed higher up in the viewport. This change scrolls the expanded item into view only in these cases where it would otherwise not be visible. See https://caniuse.com/scrollintoview

## Testing done
Stroybook

## Screenshots
https://user-images.githubusercontent.com/3144003/128927458-2f781d49-d95f-40d2-901f-aac965ebf17b.mov

## Acceptance criteria
- [ ]

## Definition of done


kelsonic (approved):
LGTM! 🎉

src/components/va-accordion/va-accordion.tsx
```diff
@@ -74,6 +74,21 @@ export class VaAccordion {
...
+
+  isScrolledIntoView(el: Element) {
+    const elemTop = el?.getBoundingClientRect().top;
+
+    if (!elemTop && elemTop !== 0) {
+      return false;
+    }
+
+    // Only partially || completely visible elements return true
+    return elemTop >= 0 && elemTop <= window.innerHeight;
```
bkjohnson, line 91: I'm not sure what the a11y impact here might be, so I'm going to let @1Copenut weigh in.

src/components/va-accordion/fixtures/va-accordion.fixture.json
```diff
@@ -0,0 +1,7 @@
+{
```
bkjohnson, line 1: Just as an FYI, these stories won't appear on design.va.gov/storybook. If you want them to, you'll have to make another PR based off of `master` and add the stories there :upside_down_face:

src/components/va-accordion/va-accordion.tsx
Rleahy22: I'm happy to hear what @1Copenut has to say, and I will note that we are not changing focus so I would not envision any a11y impacts.

src/components/va-accordion/va-accordion.tsx
```diff
@@ -74,6 +74,21 @@ export class VaAccordion {
     }
 
     clickedItem.setAttribute('open', !prevAttr);
+
+    if (!this.isScrolledIntoView(clickedItem)) {
+      clickedItem.scrollIntoView();
```
1Copenut, line 79: The default scrolling behavior uses smooth scroll, which isn't a great experience for users who might be triggered by the motion. Some users may experience dizziness or nausea. Could we do something to prevent this? Thinking of polling `window.matchMedia` ?

https://since1979.dev/respecting-prefers-reduced-motion-with-javascript-and-react/

```javascript
const mediaQuery = window.matchMedia("(prefers-reduced-motion: reduce)");

if (!this.isScrolledIntoView(clickedItem) && window.match) {
  clickedItem.scrollIntoView({ behavior: auto });
}

if (!this.isScrolledIntoView(clickedItem) && !window.match) {
  clickedItem.scrollIntoView();
}
```

Rleahy22: I believe the default behavior is `auto`, and not `smooth` so `clickedItem.scrollIntoView({ behavior: auto });` and `clickedItem.scrollIntoView();` are equivalent. See: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#parameters

1Copenut: @Rleahy22 Yep, I was trying to multi-task and misread the MDN docs. Please disregard my comment, and thank you for looking into it!

PR #136, opened 2021-08-17
Change how React bindings are built

bkjohnson:
## Description
This is a followup for #131. We needed the bindings to be includes in the `files` property in `package.json` so that they remain after installation.

The directory structure has also been reworked slightly so that inside of the `react-bindings` directory we have an `index.js` file and an `index.d.ts` file after building the bindings. This allows typescript _and_ javascript projects to import from `component-library/react-bindings`

## Testing done

Local using https://github.com/department-of-veterans-affairs/formulate/pull/16

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/stencil.config.ts
+++ b/stencil.config.ts
@@ -26,8 +26,8 @@ export const config: Config = {
   },
   outputTargets: [
     reactOutputTarget({
-      componentCorePackage: '../../dist/types',
-      proxiesFile: './generated/react-bindings/components.ts',
+      componentCorePackage: '../dist/types',
+      proxiesFile: './react-bindings/index.ts',
       includeDefineCustomElements: false,
     }),
     {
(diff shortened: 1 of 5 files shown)


stencil.config.ts
```diff
@@ -26,8 +26,8 @@ export const config: Config = {
   },
   outputTargets: [
     reactOutputTarget({
-      componentCorePackage: '../../dist/types',
-      proxiesFile: './generated/react-bindings/components.ts',
+      componentCorePackage: '../dist/types',
```
cvalarida, line 29: What's this relative to? (What happens if it's wrong?)

cvalarida (approved):
Looks great. I just had one question to make sure I understand what's going on.

stencil.config.ts
bkjohnson: It's relative to the proxies file.

image

If it's wrong then we get an error: https://github.com/department-of-veterans-affairs/component-library/pull/131#discussion_r684465784

PR #137, opened 2021-08-18
Issue 204 - Alert Boxes - Role alert MUST announce itself to assistiv…

noahgelman:
## Note
I cannot force `h1-6` tags to announce when a user passes that HTML into the alert box for use as a heading. The content will announce as normal however. If the user must have the title of the alert box announce they can pass the header in as a `p` or `div` and use the `heading-level-x` classes to style it and then it'll announce too.

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/204

## Testing done
Tested in Storybook

## Acceptance criteria
- [ ] On load, the alert content is announced if the status is set to 'error'

## Definition of done
- [ ] Screen readers will announce the error on page load


src/components/va-alert/test/va-alert.e2e.ts
```diff
@@ -13,7 +13,6 @@ describe('va-alert', () => {
         <mock:shadow-root>
           <div class="alert info">
             <i aria-hidden="true" role="img"></i>
-            <span class="sr-only">Alert:</span>
```
noahgelman, line 16: In now announces as an alert since it has the correct role in place. Removing this to prevent double announcing the "Alert" text

src/components/va-alert/va-alert.tsx
```diff
@@ -108,14 +108,15 @@ export class VaAlert {
...
+    const ariaLive = status === 'error' ? 'assertive' : null;
 
     if (!visible) return <div aria-live="polite" />;
     
     return (
       <Host>
-        <div class={classes}>
+        <div role={role} aria-live={ariaLive} class={classes}>
           <i aria-hidden="true" role="img"></i>
-          <span class="sr-only">Alert: </span>
```
noahgelman, line 118: In now announces as an alert since it has the correct role in place. Removing this to prevent double announcing the "Alert" text

src/components/va-alert/test/va-alert.e2e.ts
```diff
@@ -140,4 +139,15 @@ describe('va-alert', () => {
...
 
+  it('has the correct accessibility attributes when in an error state', async () => {
+    const page = await newE2EPage();
+    await page.setContent(
+      '<va-alert status="error"><h4 slot="headline">This is an alert</h4><div>This is the alert content</div>',
+    );
+
+    const element = await page.find('va-alert >>> .alert');
+
+    expect(element).toEqualAttribute('role', 'alert');
```
1Copenut, line 150: What do you think about adding a line to check for the `aria-live="assertive"` too? I know it's pedantic and picky, so add it if you feel compelled, ignore if not.

```javascript
expect(element).toEqualAttribute('aria-live', 'assertive');
```

src/components/va-alert/va-alert.tsx
1Copenut: I like the way you're adding role and aria-live attributes when needed, ignoring when not.

1Copenut (approved):
LGTM! I left one comment for a check you could add to the unit test, not a blocking item.

src/components/va-alert/test/va-alert.e2e.ts
noahgelman: You're right, I actually had that in before I was testing another change. I'll add it back in

PR #138, opened 2021-08-18
Bump react focus lock version

jbalboni:
## Description
This bumps the version of react-focus-lock being used to the current version, 2.5.2. I also ended up bumping react-focus-on to the next patch version, because it looks clearer in the yarn.lock file that we're always using the 2.5.2 version of react-focus-lock.

There is a [performance issue](https://github.com/theKashey/react-focus-lock/issues/125) with this library and jsdom, that was fixed in 2.5.1. Upgrading this dependency cuts about 30 seconds off of the runtime of the VAOS tests, from 94s down to around 67s. Presumably it would also improve unit test speed across vets-website, since this issue affects any test that uses a modal.

## Testing done
Ran VAOS vets-website unit tests

## Acceptance criteria
- [ ] Tests pass

## Definition of done


bkjohnson (approved):
A lock file update looks good

PR #144, opened 2021-08-23
Automatically build package and attatch to release

### LINKED ISSUE #142: Build & include binaries with each Web Component release
In [`formulate`](https://github.com/department-of-veterans-affairs/formulate) we have upgraded to Yarn 2 which has a few enhancements. One bug we've noticed is that using the `web-components` dependency from a git URL causes problems with the checksums in the lock file since the `prepare` script does the build on each machine. In order to avoid this, we can use a prebuilt gzipped tarball instead of a git URL for a specific release.

## Technical direction

`yarn pack` will automatically create a tarball based on what is listed in the `files` property in `package.json`. There is also a GitHub Action that we may be able to use to [upload release artifacts](https://github.com/marketplace/actions/github-action-publish-binaries) for each release.

## Acceptance Criteria

- [ ] Prebuilt packages are automatically uploaded to each Release

bkjohnson:
## Description
Closes #142.

This supercedes #143. In order for the action to run when a Release is tagged based on the `web-components` branch, this action will also need to be on that branch. Trying to change the base in the former PR messed things up, presumably due to no shared history.

We are doing this because using a `.git` url and building the package locally on different machines results in different checksums in the lockfile for yarn 2. Building the package remotely and attaching it to the release will solve this problem until we are able to actually unify `master` and `web-components`.

## Testing done

I monitored the `Actions` tab and made releases to trigger the action

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/.github/workflows/add-release-files.yml
+++ b/.github/workflows/add-release-files.yml
@@ -21,6 +21,7 @@ jobs:
           node-version: 14.x
       - run: yarn
       - run: yarn build
+      - run: yarn build-bindings
       - run: yarn pack
       - run: ls
       - name: Upload the artifacts
--- a/.github/workflows/add-release-files.yml
+++ b/.github/workflows/add-release-files.yml
@@ -1,0 +1,31 @@
+on:
+  release:
+    types: [created, published, edited]
+name: Handle Release
+jobs:
+  generate:
+    name: Create release-artifacts
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the web-components branch
+        uses: actions/checkout@v2
+        with:
+          ref: wc-automate-release-files
+      - uses: actions/cache@v2
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
+      - name: Use Node.js 14.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14.x
+      - run: yarn
+      - run: yarn build
+      - run: yarn pack
+      - run: ls
+      - name: Upload the artifacts
+        uses: skx/github-action-publish-binaries@release-0.15
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          args: '*.tgz'
--- a/.github/workflows/add-release-files.yml
+++ b/.github/workflows/add-release-files.yml
@@ -1,6 +1,6 @@
 on:
   release:
-    types: [created, published, edited]
+    types: [published]
 name: Handle Release
 jobs:
   generate:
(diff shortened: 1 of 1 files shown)


.github/workflows/add-release-files.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Use Node.js 14.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14.x
+      - run: yarn
+      - run: yarn build
+      - run: yarn build-bindings
+      - run: yarn pack
+      - name: Upload the artifacts
+        uses: skx/github-action-publish-binaries@release-0.15
```
bkjohnson, line 27: `release-0.15` is the latest. I didn't want to use master like the [sample](https://github.com/marketplace/actions/github-action-publish-binaries#sample-configuration) suggests because some new push could break our workflow.

.github/workflows/add-release-files.yml
```diff
@@ -0,0 +1,31 @@
...
+          path: '**/node_modules'
+          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
+      - name: Use Node.js 14.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14.x
+      - run: yarn
+      - run: yarn build
+      - run: yarn build-bindings
+      - run: yarn pack
```
bkjohnson, line 25: The first three are pretty standard. `yarn pack` creates a gzipped tarball with all of the folders/files from `package.json.files` + other package files (License, readme, package.json) inside of a nested folder. This nested folder is important so that the structure will be preserved when yarn/npm unpacks it.

.github/workflows/add-release-files.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Upload the artifacts
+        uses: skx/github-action-publish-binaries@release-0.15
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          args: 'web-components-*.tgz'
```
bkjohnson, line 31: This name is based on the package name.

.github/workflows/add-release-files.yml
```diff
@@ -0,0 +1,31 @@
...
+name: Handle Release
+jobs:
+  generate:
+    name: Create release-artifacts
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the web-components branch
+        uses: actions/checkout@v2
+        with:
+          ref: web-components
```
bkjohnson, line 13: We want to make sure `web-components` is checked out rather than `master`.

bkjohnson:
I'm merging without review because:

1. This isn't a functional change to the library
1. I won't be able to get an approving review for 1 week
1. Not having this means that people will have to update their lock file for `formulate` if they are doing development each time after someone else has merged to master (and had to deal with the same lockfile problems)

PR #148, opened 2021-09-01
29551: Update error message placement for <Checkbox /> component

craigwheeler:
## Description
Closes [#29551](https://app.zenhub.com/workspaces/vsa---debt-607736a6c8b7e2001084e3ab/issues/department-of-veterans-affairs/va.gov-team/29551)

The error rendered for the `<Checkbox />` component is currently is under the input, this change renders the error above the input. Refer to this issue for details about this change: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/535

## Screenshots
Components  Checkbox - Error ⋅ Storybook 2021-09-01 18-01-20

## Acceptance criteria
- [ ] Error messages should appear consistently regardless of how many checkboxes are in a group - under the label, and above the first input (like we do for a checkbox group).
- [ ] Design system guidance recommends consistent placement of error messages.

## Definition of done


cvalarida:
I'm good with the change, but now that the error message is above the input, we'll need to adjust the top margin of the `errorSpan` too so it doesn't take up more space than it should.

cvalarida:
:thinking: Actually, I think I'm looking at the wrong thing. The top margin I'm seeing is for the description and isn't affected by this change (but is still obnoxious :stuck_out_tongue_winking_eye:).

@GnatalieH What do you think about the spacing between `Error message` and the check box input? I feel like it should be a little bigger, but I'd like to get a :sparkles: Designer Opinion™. :sparkles: :sweat_smile:

craigwheeler:
Yep you're correct @cvalarida the extra margin on the top is a little obnoxious 😁  ...  @GnatalieH just let me know if we wanna change it I can open a separate pr with veteran-facing-services-tools to adjust the margin [here](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/8e5f11181b62ff29aed70de50e52f10250ce20eb/packages/formation/sass/base/_va.scss#L1428)

Removing the top margin seems to work:
```
.label-above-checkbox {
    display: block;
    margin: 0 0 1rem;
}
```
Components  Checkbox - Error ⋅ Storybook 2021-09-01 19-04-57


Screen Shot 2021-09-01 at 6 56 00 PM

GnatalieH:
@craigwheeler and @cvalarida I am wondering what a long error message might look like. We definitely want the error message to appear closer to the input (checkbox in this case) than the label, so I'm not 100% convinced, yet, that we should increase the spacing.

GnatalieH:
This example from Storybook of an error message above a radio button group looks perfect to me in terms of spacing between the description and the error message and spacing between the error message and group of radio button options:

image

Is the excessive top margin, above the description, a new issue that has occurred because we changed the placement of the error message? We should decrease that top margin, for sure. Thanks.

craigwheeler:
Okay sounds good i'll verify that `<Checkbox/>` is using the same spacing between the error and label text as `<CheckboxGroup/>` and update if needed. As far as the top margin on the label I just realized that by removing it will improve the look of the error, however, we are still removing 30px of margin which could affect the UI in places where this component is used that may rely on that extra spacing so not sure we wanna do this but i'll defer to you guys to make the call? @cvalarida @GnatalieH

craigwheeler:
Hey @GnatalieH 👋  ... so verified that both `<Checkbox/>` and `<CheckboxGroup/>` are using that same spacing between error message and checkbox and since were not adjusting the top margin on the label it doesn't appear that anymore changes are needed of course if we wanna tweak anything lemme know otherwise we should be good to merge?

Components  Checkbox - Error ⋅ Storybook 2021-09-01 18-01-20

Components  CheckboxGroup - Error ⋅ Storybook 2021-09-02 16-19-24

GnatalieH:
@craigwheeler looks good to me. @cvalarida cool to merge?

PR #151, opened 2021-09-03
Add <th> with scope='row' to Table component for screen readers

ddzz:
## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/489

## Testing done
Manual testing.

## Acceptance criteria
- [ ] Table component rows appear as described in the ticket.

### DIFF
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -6,7 +6,12 @@ const borderClasses =
   'vads-u-border-top--0 vads-u-border-right--0 vads-u-border-left--0 vads-u-font-family--sans vads-u-padding--0 vads-u-padding-y--0p5 medium-screen:vads-u-padding--2';
 const rowPaddingClass = 'vads-u-padding-y--2';
 
-const cellProps = ({ alignLeft, alignRight, label }, index, rowIndex) => {
+const cellProps = (
+  { alignLeft, alignRight, label },
+  index,
+  rowIndex,
+  scope = null,
+) => {
   return {
     'data-index': index,
     className: classNames(borderClasses, {
@@ -16,13 +21,7 @@ const cellProps = ({ alignLeft, alignRight, label }, index, rowIndex) => {
     'data-label': label,
     key: `${rowIndex}-${label}`,
     role: 'cell',
-  };
-};
-
-const thProps = (field, index, rowIndex) => {
-  return {
-    ...cellProps(field, index, rowIndex),
-    scope: 'row',
+    scope: scope,
   };
 };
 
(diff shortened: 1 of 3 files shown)


CBonade (approved):

src/components/Table/Table.jsx
```diff
@@ -48,20 +48,36 @@ function Table(props) {
...
+                <th
+                  data-index={index}
+                  className={classNames(borderClasses, {
+                    'vads-u-text-align--left': field.alignLeft,
+                    'medium-screen:vads-u-text-align--right': field.alignRight,
+                  })}
+                  data-label={field.label}
+                  key={`${rowIndex}-${field.label}`}
+                  role="cell"
+                  scope="row"
```
mpelzsherman, line 62: There's a lot of repetition here.  What do you think about creating an object with the common key/value pairs, and using the spread operator to add to each element? e.g.

```jsx
const cellProps = {
  'data-index': { index },
  className: {
    classNames(borderClasses, {
      'vads-u-text-align--left': field.alignLeft,
      'medium-screen:vads-u-text-align--right': field.alignRight,
    }),
  },
  'data-label': { field.label },
  ...etc.
};

const thProps = {
  ...cellProps,
  scope: 'row',
}


{fields.map((field, index) =>
  index === 0 ? (
    <th {...thProps} />
  ) : (
    <td {...cellProps} />
```

ddzz: Thanks for reviewing! I updated the PR with this approach. I had to use functions to create the objects.

mpelzsherman (approved):

cvalarida (commented):
The change to the table component looks good, but can you add a test to check that the first element in a `tr` is a `th` with `scope="row"`?

Also, I'm really confused about the change to that existing test. :sweat_smile:

src/components/Table/Table.unit.spec.jsx
```diff
@@ -35,7 +35,8 @@ describe('<Table />', () => {
     );
 
     expect(wrapper.find('thead')).to.have.lengthOf(1);
-    expect(wrapper.find('th')).to.have.lengthOf(6); // Once for each field
+    expect(wrapper.find('th')).to.have.lengthOf(15);
+    expect(wrapper.find('td')).to.have.lengthOf(45);
```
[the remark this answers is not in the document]
ddzz: There are 60 total cells in the table based on the test data.

ddzz:
> The change to the table component looks good, but can you add a test to check that the first element in a `tr` is a `th` with `scope="row"`?
> 
> Also, I'm really confused about the change to that existing test. 😅

Thanks for the review! Specs have been updated.

cvalarida (approved):
Thanks for the updated tests! :shipit:

PR #154, opened 2021-09-03
Increase distance between stacked accordions

ddzz:
## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/314

## Testing done
Manual testing

## Screenshots
Before:
Screen Shot 2021-09-03 at 4 53 58 PM

After:
Screen Shot 2021-09-03 at 4 54 59 PM

## Acceptance criteria
- [ ] Margin difference can be seen in Storybook.

### DIFF
--- a/src/components/va-accordion/va-accordion-item.css
+++ b/src/components/va-accordion/va-accordion-item.css
@@ -8,6 +8,10 @@
   margin-bottom: 22px;
 }
 
+:host(:last-child) {
+  margin-bottom: 0.5rem;
+}
+
 :host(:not([open])) #content,
 :host([open='false']) #content {
   display: none;
--- a/src/components/va-accordion/va-accordion-item.css
+++ b/src/components/va-accordion/va-accordion-item.css
@@ -1,8 +1,8 @@
 @import '../../mixins/focusable.css';
 
-:host {
+:host(:not(:last-child)) {
   display: block;
-  margin-bottom: 0.5rem;
+  margin-bottom: 22px;
 }
 
 :host(:not([open])) #content,
--- a/src/components/va-accordion/va-accordion-item.css
+++ b/src/components/va-accordion/va-accordion-item.css
@@ -1,7 +1,10 @@
 @import '../../mixins/focusable.css';
 
-:host(:not(:last-child)) {
+:host {
   display: block;
+}
+
+:host(:not(:last-child)) {
   margin-bottom: 22px;
 }
 


cvalarida:
@GnatalieH Look good to you?

GnatalieH:
@cvalarida what was the impetus for increasing the spacing between elements? In my experience, accordion items are stacked more like the "before" image, with tighter spacing. It makes the two items look unrelated with extra padding, I think.

pjhill:
> @cvalarida what was the impetus for increasing the spacing between elements? In my experience, accordion items are stacked more like the "before" image, with tighter spacing. It makes the two items look unrelated with extra padding, I think.

This change stems from the accessibility concerned summarized in this [ticket ](https://github.com/department-of-veterans-affairs/component-library/pull/154#issuecomment-914353254)as -- 

> There isn't a specific recommendation in WCAG for distance between tap targets, and I haven't been able to find recommendations on this from reliable sources online. @jenstrickland shared that Apple Interaction Design Best Practices used to recommend 22px spacing between touch targets, to help those with mobility considerations. That seems reasonable to me, and based on my quick-n-dirty assessment, also aligns with the distance between tap targets on GOV.uk (on mobile).

GnatalieH:
@pjhill is this the pattern you're referring to on gov.uk? 

image

pjhill:
> @pjhill is this the pattern you're referring to on gov.uk?

Thanks for the example! I'm actually not entirely sure what was intended by that original comment. My original reply had an incorrect link to the originating ticket. [This is the correct link](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/314). They did call out mobile viewports in particular. Another example I was just viewing on gov.uk is -- 
image

GnatalieH:
@pjhill thank you for showing me that example. A big difference between our accordion pattern and gov.uk is that they use a white background header and we use gray. So, there's no distinguishable space *between* accordion items to the user on gov.uk -- they all look like objects in the same group. I don't think increasing the white space between our accordion headers will have the same effect. 

What would, say, five items in an accordion (collapsed as well as in an expanded state) look like with the proposed 22px spacing between accordion headers? I would lean on increasing the height of the gray headers and modifying the touch target area within the header to mitigate the user tapping on the wrong item, if that's our concern. I worry we're breaking the gestalt law of proximity if we increase the spacing between accordion headers by 22 px. Thanks for working with me on this.

pjhill:
> @pjhill thank you for showing me that example. A big difference between our accordion pattern and gov.uk is that they use a white background header and we use gray. So, there's no distinguishable space _between_ accordion items to the user on gov.uk -- they all look like objects in the same group. I don't think increasing the white space between our accordion headers will have the same effect.
> 
> What would, say, five items in an accordion (collapsed as well as in an expanded state) look like with the proposed 22px spacing between accordion headers? I would lean on increasing the height of the gray headers and modifying the touch target area within the header to mitigate the user tapping on the wrong item, if that's our concern. I worry we're breaking the gestalt law of proximity if we increase the spacing between accordion headers by 22 px. Thanks for working with me on this.

These also sound like excellent points to me! @cvalarida , should we update / comment on the [originating request](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/314)? Does this need a design system council review?

cvalarida:
> should we update / comment on the originating request?

:thinking: Let's make a comment there once we come up with the final answer, but we can hold off for now.

>Does this need a design system council review?

I'm inclined to say @GnatalieH's expertise is plenty, but I'll double-check to be sure. If we do need a council approval, we can start [a discussion](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/discussions/categories/design-system-council-topics) for it. :smile:

cvalarida:
Confirmed! Whatever @GnatalieH recommends is golden. No council decision necessary.

So! What's your official recommendation, Natalie?

GnatalieH:
@cvalarida I will need to do a little research in order to make my final recommendation. What is the current margin spacing between accordion headers ("before" image)? I don't see this component in Storybook. I also have some technical questions about how we can possibly increase the height of the accordion headers to achieve the suggested margin of 22px between target areas.

cvalarida:
You can find the accordion in Storybook at the [`<va-accordion>` story](https://design.va.gov/storybook/?path=/docs/components-va-accordion--default).

Looks like there's a 5px margin between accordion items right now.
image

bkjohnson:
Looks like it's 5px: 

image

cvalarida:
We can also increase the target area by increasing the `<button>` padding.
image

GnatalieH:
Thank you @cvalarida and @bkjohnson for those examples. I totally missed the components that have the "va" prefix at the bottom of the Storybook. Is there a reason they are named differently than the other components?

Are we following the specs that USWDS uses for the [Accordion component](https://designsystem.digital.gov/components/accordion/)? I can't see any margin specs when I inspect an Accordion header. I think we should increase the margin between the items, possibly, to match that of USWDS, and then also increase the padding to increase the target area. Do y'all have any time to discuss this today so we can knock it out together?

cvalarida:
Sounds like we have some more questions on the impetus behind this change before we can approve this PR. Unfortunately, we're pretty backed up right now, so we'll loop back around next week when we can.

pjhill:
> Sounds like we have some more questions on the impetus behind this change before we can approve this PR. Unfortunately, we're pretty backed up right now, so we'll loop back around next week when we can.

@cvalarida, it's been a few weeks. Is there some information gathering that _Testing Tools_ can do for this ticket?

GnatalieH:
@pjhill perhaps I can answer your question. Can we increase the margin/white space between the accordion headers to 8 px and increase the padding above and below the target area in the accordion header to achieve the 22 px total between target areas we are aiming for? I believe the minimum target area should be 25 px in height (for mobile consideration). @cvalarida do we need to increase the height of the accordion header, or can we just decrease the target area height?

PR #156, opened 2021-09-07
Add list role to Breadcrumbs for screen readers

ddzz:
## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/240

## Testing done
Tested with VoiceOver in Storybook.

## Acceptance criteria
- [ ] Screen reader announces number of breadcrumb items.


1Copenut (approved):
LGTM! In a strange twist of fate, Safari latest seems to have fixed this bug. But we're going to keep it because not every use is on the latest version, and it's not causing interference.

bkjohnson (approved):

PR #158, opened 2021-09-08
Add aria hidden to alert box to prevent screen reader tab stop

CBonade:
## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/242

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/src/components/va-alert/va-alert.tsx
+++ b/src/components/va-alert/va-alert.tsx
@@ -144,7 +144,7 @@ export class VaAlert {
             aria-label={this.closeBtnAriaLabel}
             onClick={this.closeHandler.bind(this)}
           >
-            <i class="fa-times-circle" aria-label="Close icon" />
+            <i aria-hidden="true" class="fas fa-times-circle" role="presentation" />
           </button>
         )}
       </Host>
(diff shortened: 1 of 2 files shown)


noahgelman (approved):
LGTM

src/components/va-alert/va-alert.tsx
```diff
@@ -144,7 +144,7 @@ export class VaAlert {
             aria-label={this.closeBtnAriaLabel}
             onClick={this.closeHandler.bind(this)}
           >
-            <i class="fa-times-circle" aria-label="Close icon" />
+            <i aria-hidden="true" class="fas fa-times-circle" role="presentation" />
```
bkjohnson, line 147: I don't think the `fas` class is doing anything, so it should be removed. What appear to be fontawesome classes are coming from here:

https://github.com/department-of-veterans-affairs/component-library/blob/cfffd5588044c7c2f37a5bae0be625aa50f8c1af/src/components/va-alert/va-alert.css#L141-L143

bkjohnson (changes requested):

1Copenut (changes requested):
👍 LGTM!

src/components/va-alert/va-alert.tsx
CBonade: Should be all good here! Thanks for the feedback.

bkjohnson (approved):

PR #161, opened 2021-09-10
SimpleDate

cvalarida:
`SimpleDate` pulls out the visual elements of the `Date` into its own component. The `Date` field then uses it and layers the validation on top.

The change was so Formulate can control the validation.

Along the way, I also added `onBlur` to `Date`, `NumberInput`, and `Select` so we can call the validation functions we want when the field is blurred.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "@storybook/addon-links": "6.1.21",
     "@storybook/react": "6.1.21",
     "@testing-library/react": "^12.0.0",
+    "@testing-library/user-event": "^13.2.1",
     "@whitespace/storybook-addon-html": "^2.0.1",
     "babel-loader": "^8.1.0",
     "babel-plugin-lodash": "^3.3.4",
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "@storybook/addon-essentials": "6.1.21",
     "@storybook/addon-links": "6.1.21",
     "@storybook/react": "6.1.21",
+    "@testing-library/react": "^12.0.0",
     "@whitespace/storybook-addon-html": "^2.0.1",
     "babel-loader": "^8.1.0",
     "babel-plugin-lodash": "^3.3.4",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/component-library",
-  "version": "3.0.3",
+  "version": "3.1.0",
   "description": "VA.gov component library in React",
   "keywords": [
     "react",
--- a/src/components/Date/Date.unit.spec.jsx
+++ b/src/components/Date/Date.unit.spec.jsx
@@ -1,38 +1,13 @@
 import React from 'react';
 // import SkinDeep from 'skin-deep';
 import { expect } from 'chai';
-import { shallow } from 'enzyme';
-import { axeCheck } from '../../helpers/test-helpers';
+import { render } from '@testing-library/react';
 import Date from './Date';
 import { makeField } from '../../helpers/fields.js';
 
 import { minYear, maxYear } from '../../helpers/validations';
 
 describe('<Date>', () => {
-  it('wraps input elements in a fieldset', () => {
-    const date = {
-      day: makeField(1),
-      month: makeField(12),
-      year: makeField(2010),
-    };
-    const tree = shallow(<Date date={date} onValueChange={() => {}} />);
-    expect(tree.find('fieldset')).to.have.lengthOf(1);
-    const legend = tree.find('legend');
-    expect(legend).to.have.lengthOf(1);
-    expect(legend.text()).to.equal('Date of birth');
-    tree.unmount();
-  });
-  it('renders input elements', () => {
-    const date = {
-      day: makeField(1),
-      month: makeField(12),
-      year: makeField(2010),
-    };
-    const tree = shallow(<Date date={date} onValueChange={() => {}} />);
-    expect(tree.find('NumberInput')).to.have.lengthOf(1);
-    expect(tree.find('Select')).to.have.lengthOf(2);
-    tree.unmount();
-  });
   it('displays required message', () => {
     const date = {
       day: makeField(''),
@@ -43,16 +18,13 @@ describe('<Date>', () => {
     date.month.dirty = true;
     date.day.dirty = true;
 
-    const tree = shallow(
+    const { getByText } = render(
       <Date required date={date} onValueChange={() => {}} />,
     );
 
-    expect(tree.find('.usa-input-error').exists()).to.be.true;
-    expect(tree.find('.usa-input-error-message').text()).to.equal(
-      'Error Please provide a response',
-    );
-    tree.unmount();
+    getByText(/Please provide a response/);
   });
+
   it('displays invalid message', () => {
     const date = {
       day: makeField(''),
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/component-library",
-  "version": "3.0.3",
+  "version": "3.1.0",
```
cvalarida, line 3: Bumping the minor version because the `SelectInput`, `NumberInput`, and `Date` all accept an `onBlur`. Also we have a whole new `SimpleDate`.

Arguably, this could be a major version bump because of the "breaking" change `NumberInput`—it no longer marks a field as dirty when the field is blurred. Just when it's changed.

src/components/Date/SimpleDate.jsx
```diff
@@ -0,0 +1,211 @@
...
+          </div>
+        </div>
+      </div>
+    </fieldset>
+  );
+};
+
+const dateFragment = PropTypes.shape({
+  value: PropTypes.string.isRequired,
+  dirty: PropTypes.bool.isRequired,
```
cvalarida, line 148: This is never even used, but it's there in the old version, so :man_shrugging: 

Subject to refactoring later.

src/components/NumberInput/NumberInput.jsx
```diff
@@ -27,7 +27,7 @@ class NumberInput extends React.Component {
   }
 
   handleBlur() {
-    this.props.onValueChange(makeField(this.props.field.value, true));
+    if (this.props.onBlur) this.props.onBlur();
```
bkjohnson, line 30: Do we need this check here? Can the handler just be inlined, like:

```jsx
onBlur={this.props.onBlur}
```

src/components/Date/SimpleDate.unit.spec.jsx
```diff
@@ -0,0 +1,213 @@
+import React from 'react';
+// import SkinDeep from 'skin-deep';
```
bkjohnson, line 2: Did you mean to leave this comment?

src/components/Date/SimpleDate.jsx
bkjohnson: `:old-man-yells-at-cloud:`

src/components/Date/SimpleDate.jsx
```diff
@@ -0,0 +1,211 @@
...
+  label,
+  name,
+  required,
+  ariaDescribedby,
+  minYear,
+  maxYear,
+  onValueChange,
+  onBlur,
+}) => {
+  const inputId = useRef(_.uniqueId('date-input-'));
```
bkjohnson, line 41: There was a recent PR (in `vets-website` I think) where we replaced all top-level lodash imports with their more specific counterparts, like:

```js
import { uniqueId } from 'lodash'
```

I think it would be good to do that here as well.

src/components/Date/SimpleDate.jsx
```diff
@@ -0,0 +1,211 @@
...
+  const { day, month, year } = date;
+
+  let daysForSelectedMonth = [];
+  if (month.value) {
+    daysForSelectedMonth = days[month.value];
+  }
```
bkjohnson, line 47: I see that this is copy-pasted from the old `Date`, but would this work? I like that it would let us use a `const` so a reader knows that the value won't change.

```js
const daysForSelectedMonth = days?[month.value] || [];
```

src/components/Date/SimpleDate.jsx
```diff
@@ -0,0 +1,211 @@
...
+        ![
+          `#${inputId.current} [name="${name}Month"]`,
+          `#${inputId.current} [name="${name}Day"]`,
+          `#${inputId.current} [name="${name}Year"]`,
+        ].some(
+          selector =>
+            document.querySelector(selector) === document.activeElement,
+        )
+      )
+        onBlur();
```
bkjohnson, line 79: Should we set a default function for the `onBlur` prop in order to do this?

src/components/Date/Date.unit.spec.jsx
```diff
@@ -1,38 +1,13 @@
 import React from 'react';
 // import SkinDeep from 'skin-deep';
 import { expect } from 'chai';
-import { shallow } from 'enzyme';
-import { axeCheck } from '../../helpers/test-helpers';
+import { render } from '@testing-library/react';
```
bkjohnson, line 4: Long live RTL! :joy:

src/components/Date/Date.unit.spec.jsx
```diff
@@ -43,16 +18,13 @@ describe('<Date>', () => {
...
+    const { getByText } = render(
       <Date required date={date} onValueChange={() => {}} />,
     );
 
-    expect(tree.find('.usa-input-error').exists()).to.be.true;
-    expect(tree.find('.usa-input-error-message').text()).to.equal(
-      'Error Please provide a response',
-    );
-    tree.unmount();
+    getByText(/Please provide a response/);
```
bkjohnson, line 25: I think that the `getBy*` functions throw an error if the thing isn't found, but do you think it makes sense to also have an `expect` for these tests?

bkjohnson (commented):
A few questions

src/components/NumberInput/NumberInput.jsx
cvalarida: Oh good call. I had it in here because I initially was keeping the `onValueChange` bit, but since I got rid of it, inlining makes more sense. :+1:

src/components/Date/SimpleDate.jsx
cvalarida: I'm not sure if we have optional chaining enabled in our Babel setup in this repo, but I can give it a shot! :+1: This looks cleaner to me.

But for what it's worth, as far as arrays (and Objects and Maps and honestly anything other than primitives) go, we still have no assurance that the value won't change. We just know the variable won't be reassigned. `const` != `immutable`.

src/components/Date/Date.unit.spec.jsx
cvalarida: :fist_raised:

src/components/Date/Date.unit.spec.jsx
cvalarida: I was going back and forth. Having an `expect` in there is probably clearer, even if it's not _cleaner._ :man_shrugging:

src/components/Date/SimpleDate.jsx
cvalarida: I went with this to preserve the same functionality if `month.value` exists but is not a key of `days`:
```js
  const daysForSelectedMonth = month.value ? days[month.value] : [];
```

src/components/Date/SimpleDate.jsx
cvalarida: Oh, that's a good call. If `onBlur` isn't a function, this is going to throw an error. That error wouldn't be blocking anything since it's not evaluated as part of any other function's call stack thanks to `setTimeout`, but it's still no good.

src/components/Date/SimpleDate.jsx
bkjohnson: Yeah, good point about non-primitives. At least it indicates that it won't be reassigned.

bkjohnson (approved):

PR #170, opened 2021-09-16
Add Back to top Web Component

bkjohnson:
## Description
Part of https://github.com/department-of-veterans-affairs/va.gov-team/issues/29785

This adds a `<va-back-to-top>` Web Component. This will replace our current "Back to top" button. It solves some problems with the old version:

- The button doesn't need to pay attention to the footer, and it will stop once it reaches its "dock"
- The dock ensures that space is reserved for the button
    - The button will no longer appear on the same "line" as pagination buttons, etc.
- When the button is visible & undocked, its width is dynamically set to be that of its parent

## Testing done

Local dev server, storybook, and E2E tests

## Screenshots

See #171 

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/src/components/va-back-to-top/test/va-back-to-top.e2e.ts
+++ b/src/components/va-back-to-top/test/va-back-to-top.e2e.ts
@@ -121,8 +121,9 @@ describe('va-back-to-top', () => {
   it('passes an axe check when revealed', async () => {
     const page = await pageSetup();
     const button = await page.find('va-back-to-top >>> button');
+    const pastRevealPoint = 300;
 
-    await page.mouse.wheel({ deltaY: 300 });
+    await page.mouse.wheel({ deltaY: pastRevealPoint });
     await page.waitForChanges();
 
     expect(button).toHaveClass('reveal');
@@ -132,10 +133,13 @@ describe('va-back-to-top', () => {
 
   it('passes an axe check when docked', async () => {
     const page = await pageSetup();
+    const wrapper = await page.find('va-back-to-top >>> div');
+    const pastPlaceholder = 1000;
 
-    await page.mouse.wheel({ deltaY: 1400 });
+    await page.mouse.wheel({ deltaY: pastPlaceholder });
     await page.waitForChanges();
 
+    expect(wrapper).toHaveClass('docked');
     // The color contrast error appears to be a false negative
     await axeCheck(page, ['color-contrast']);
   });
--- a/src/components/va-back-to-top/test/va-back-to-top.e2e.ts
+++ b/src/components/va-back-to-top/test/va-back-to-top.e2e.ts
@@ -1,0 +1,11 @@
+import { newE2EPage } from '@stencil/core/testing';
+
+describe('va-back-to-top', () => {
+  it('renders', async () => {
+    const page = await newE2EPage();
+    await page.setContent('<va-back-to-top></va-back-to-top>');
+
+    const element = await page.find('va-back-to-top');
+    expect(element).toHaveClass('hydrated');
+  });
+});
--- a/src/components/va-back-to-top/test/va-back-to-top.e2e.ts
+++ b/src/components/va-back-to-top/test/va-back-to-top.e2e.ts
@@ -83,9 +83,16 @@ describe('va-back-to-top', () => {
   it('stays docked even with really long footers', async () => {
     const page = await pageSetup();
     const wrapper = await page.find('va-back-to-top >>> div');
-    const pastDock = 1300; // Placeholder section is 1000px tall & viewport is 200
+    const pastPlaceholder = 1000;
+    // Placeholder section is 1000px tall. Accounting for the header,
+    // 2000px should put us well past the dock
+    const pastDock = 500;
 
-    // Scroll past the dock
+    // Scroll to the dock to give IntersectionObserver time to trigger
+    await page.mouse.wheel({ deltaY: pastPlaceholder });
+    await page.waitForChanges();
+
+    // Scroll past dock
     await page.mouse.wheel({ deltaY: pastDock });
     await page.waitForChanges();
 
--- a/src/components/va-back-to-top/test/va-back-to-top.e2e.ts
+++ b/src/components/va-back-to-top/test/va-back-to-top.e2e.ts
@@ -1,81 +1,34 @@
 import { newE2EPage } from '@stencil/core/testing';
 import { axeCheck } from '../../../testing/test-helpers';
 
-const pageContent = `
-      <h1>The top</h1>
-      <p>Lorem ipsum</p>
-      <p>Some more</p>
-      <p>Lorem ipsum</p>
-      <p>Lorem ipsum</p>
-      <p>Lorem ipsum</p>
-      <p>Lorem ipsum</p>
-      <p>Lorem ipsum</p>
-      <p>Lorem ipsum</p>
-      <p>Lorem ipsum</p>
-      <p>Lorem ipsum</p>
-      <p>Lorem ipsum</p>
-      <p>Lorem ipsum</p>
-      <p>Lorem ipsum</p>
-      <p>Lorem ipsum</p>
-      <p>Lorem ipsum</p>
-      <p>Lorem ipsum</p>
-      <p>Lorem ipsum</p>
-      <p>Lorem ipsum</p>
-      <p>Lorem ipsum</p>
-      <p>Lorem ipsum</p>
-      <p>Lorem ipsum</p>
-      <p>Lorem ipsum</p>
-      <p>Lorem ipsum</p>
-      <p>Lorem ipsum</p>
-      <p>Lorem ipsum</p>
-      <p>Some more</p>
-      <p>Some more</p>
-      <p>Some more</p>
-      <p>Some more</p>
-      <p>Some more</p>
-      <p>Some more</p>
-      <p>Some more</p>
-      <p>Some more</p>
-      <p>Some more</p>
-      <p>Some more</p>
-      <p>Some more</p>
-      <p>Some more</p>
-      <p>Some more</p>
-      <p>Some more</p>
-      <p>Some more</p>
-      <p>Some more</p>
-      <p>Some more</p>
-      <p>
-        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
-        tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim
-        veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea
-        commodo consequat. Duis aute irure dolor in reprehenderit in voluptate
-        velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint
-        occaecat cupidatat non proident, sunt in culpa qui officia deserunt
-        mollit anim id est laborum.
-      </p>
-      <p>
-        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
-        tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim
-        veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea
-        commodo consequat. Duis aute irure dolor in reprehenderit in voluptate
-        velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint
-        occaecat cupidatat non proident, sunt in culpa qui officia deserunt
-        mollit anim id est laborum.
-      </p>
-      <p>Some more</p>
-      <p>
-        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
-        tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim
-        veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea
-        commodo consequat. Duis aute irure dolor in reprehenderit in voluptate
-        velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint
-        occaecat cupidatat non proident, sunt in culpa qui officia deserunt
-        mollit anim id est laborum.
-      </p>
-      <p>Some more</p>
-      <p>Some more</p>
-      `;
+const pageStyles = `
+:root { --reveal-breakpoint: 200px}
+div {
+  margin: 0;
+  padding: 0;
+  height: 1000px;
+}
+footer {
+  height: 500px;
+}
+`;
+
+const pageSetup = async () => {
+  const page = await newE2EPage();
+  await page.setViewport({ width: 800, height: 200 });
+  await page.setContent(`
+    <main>
+    <h1>The top</h1>
+    <div>Placeholder</div>
+    <va-back-to-top></va-back-to-top>
+    </main>
+    <footer>The footer</footer>
+    `);
+
+  await page.addStyleTag({ content: pageStyles });
+
+  return page;
+};
 
 describe('va-back-to-top', () => {
   it('renders', async () => {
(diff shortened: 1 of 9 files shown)


src/components/va-back-to-top/va-back-to-top.css
```diff
@@ -0,0 +1,101 @@
+@import '../../../node_modules/animate.css/source/sliding_entrances/slideInUp.css';
+@import '../../../node_modules/animate.css/source/sliding_exits/slideOutDown.css';
```
bkjohnson, line 2: I don't like importing like this, but I couldn't get `postcss-import` to work. Regardless, these get inlined anyway.

At some point I'd like to use postcss to help with our CSS.

src/components/va-back-to-top/va-back-to-top.css
```diff
@@ -0,0 +1,101 @@
...
+  font-size: 2rem;
+}
+i {
+  font-family: 'Font Awesome 5 Free';
+  font-style: normal;
+  font-weight: 900;
+}
+i::before {
+  content: '\F062'; /* fa-arrow-up*/
+}
```
bkjohnson, line 84: Related to the previous comment. I don't want to `@import` things from fontawesome just yet, because it isn't granular. Anywhere that fontawesome is imported, the `<style>` for the component would have every single fontawesome class. Some postcss should be able to remove all unused classes, and once that is in place I'd like to improve our component CSS.

src/mixins/buttons.css
```diff
@@ -0,0 +1,16 @@
+@import './focusable.css';
+button {
```
bkjohnson, line 2: This is copied from `uswds` with slight changes.

https://github.com/uswds/uswds/blob/c4c9410d54b537c6112037a36ed289d8b7e8940a/src/stylesheets/elements/_buttons.scss#L37-L51

src/components/va-back-to-top/va-back-to-top.tsx
```diff
@@ -0,0 +1,78 @@
+import { Component, Element, Host, State, h } from '@stencil/core';
+import classnames from 'classnames';
+
+import 'intersection-observer';
```
bkjohnson, line 4: This is a polyfill for IE11.

src/components/va-back-to-top/va-back-to-top.tsx
```diff
@@ -0,0 +1,78 @@
...
+@Component({
+  tag: 'va-back-to-top',
+  styleUrl: 'va-back-to-top.css',
+  shadow: true,
+})
+export class VaBackToTop {
+  revealPixel!: HTMLSpanElement;
+
+  @State() hasHitBreakpoint = false;
+  @State() isDocked = false;
```
bkjohnson, line 15: These control the component's behaviors: revealing and docking

src/components/va-back-to-top/va-back-to-top.tsx
```diff
@@ -0,0 +1,78 @@
...
+      el.setAttribute('tabindex', '-1');
+      el.focus();
+
+      // Cleanup the tabindex on blur.
+      el.addEventListener('blur', () => el.removeAttribute('tabindex'));
+    }
+
+    // Scroll to the top.
+    return window.scrollTo(0, 0);
+  }
```
bkjohnson, line 53: Copied from our current version:

https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/6fc5f410037d867f998ace61a1398c6b802cea00/packages/formation/js/back-to-top.js#L6-L20

src/components/va-back-to-top/va-back-to-top.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  render() {
+    const undockedWidth = this.el.getBoundingClientRect().width;
+    this.el.style.setProperty('--undocked-width', `${undockedWidth}px`);
```
bkjohnson, line 57: This ensures that when the button is revealed but not docked, its width is the same as its parent.

src/components/va-back-to-top/va-back-to-top.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+    return (
+      <Host>
+        <span
+          class="reveal-point"
+          ref={el => (this.revealPixel = el as HTMLSpanElement)}
+        ></span>
```
bkjohnson, line 64: When the viewport has scrolled enough to where this DOM node is above it, the button is revealed. This is different from the [current implementation](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/6fc5f410037d867f998ace61a1398c6b802cea00/packages/formation/js/back-to-top.js#L37-L38) which compares `window.scrollY` against a magic number of 600.

bkjohnson (commented):
I feel like visual screenshot tests would be appropriate for this component, but that's kind of experimental, so :shrug:

src/components/va-back-to-top/test/va-back-to-top.e2e.ts
```diff
@@ -0,0 +1,220 @@
...
+    <va-back-to-top></va-back-to-top>
+    <footer style="height: 100px">The footer</footer>
+    </main>
+    `);
+
+    await page.mouse.wheel({ deltaY: 1400 });
+    await page.waitForChanges();
+
+    await axeCheck(page);
+  });
```
bkjohnson, line 219: I might be getting false negatives here - I'd like to confirm that if possible. 

Here are the failures when not skipped:

image


Here are some axe tests I did in the browser - I even disabled `formation` on the off chance it was adding some color contrast that the tests weren't getting.

No color contrast errors when revealed:

image

No color contrast errors when docked:

image

src/components/va-back-to-top/test/va-back-to-top.e2e.ts
```diff
@@ -0,0 +1,220 @@
...
+    <footer style="height: 100px">The footer</footer>
+    </main>
+    `);
+
+    await page.mouse.wheel({ deltaY: 1400 });
+    await page.waitForChanges();
+
+    // The color contrast error appears to be a false negative
+    await axeCheck(page, ['color-contrast']);
+  });
```
bkjohnson, line 219: I might be getting false negatives here - I'd like to confirm that if possible. 

Here are the failures when not skipped:

image


Here are some axe tests I did in the browser - I even disabled `formation` on the off chance it was adding some color contrast that the tests weren't getting.

No color contrast errors when revealed:

image

No color contrast errors when docked:

image

cvalarida: I wonder if the color contrast issue is the dark background of the button over a block of text. Are you able to consistently reproduce this error? If so, what happens if you replace that `pageContent` with a single element with a specific `height`? (The idea here is that we're ensuring the button is over blank space and not text.)

If it _is_ a problem with the button over text, we'll have to address that.

src/components/va-back-to-top/va-back-to-top.css
cvalarida: I like them, but are these animations necessary? Are they accessible? cc @noahgelman

src/components/va-back-to-top/va-back-to-top.css
cvalarida: Might need to specify font awesome (or `fortawesome`) as a peer dependency.

src/components/va-back-to-top/va-back-to-top.css
```diff
@@ -0,0 +1,101 @@
+@import '../../../node_modules/animate.css/source/sliding_entrances/slideInUp.css';
+@import '../../../node_modules/animate.css/source/sliding_exits/slideOutDown.css';
+@import '../../mixins/buttons.css';
+:host {
```
cvalarida, line 4: This is a _lot_ of CSS for this little button. Did you copy it from somewhere, or craft it by hand?

If copied, is there anything we can take out?

src/components/va-back-to-top/va-back-to-top.tsx
cvalarida: What'll happen for non-IE browsers? Will it still load up this polyfill?

src/components/va-back-to-top/va-back-to-top.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+import 'intersection-observer';
+
+@Component({
+  tag: 'va-back-to-top',
+  styleUrl: 'va-back-to-top.css',
+  shadow: true,
+})
+export class VaBackToTop {
+  revealPixel!: HTMLSpanElement;
```
cvalarida, line 12: Remind me again what the `!` is for? Non-nullable property?

src/components/va-back-to-top/va-back-to-top.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  @State() hasHitBreakpoint = false;
```
cvalarida, line 14: Alternative names targeting the intent of this state variable: `revealed`, `visible`, `shouldReveal`

src/components/va-back-to-top/va-back-to-top.tsx
```diff
@@ -0,0 +1,78 @@
...
+  @Element() el: HTMLElement;
+
+  componentDidLoad() {
+    // Setup observer to handle docking behavior
+    const dockObserver = new IntersectionObserver(
+      entries => {
+        // The second half of the || ensures that the button
+        // stays docked even if the viewport is scrolled below the dock
+        this.isDocked =
+          entries[0].isIntersecting || entries[0].boundingClientRect.y < 0;
```
cvalarida, line 26: :+1:

src/components/va-back-to-top/va-back-to-top.tsx
cvalarida: It kind of assumes that the `h1` is in the viewport when we scroll to `0, 0`, but :man_shrugging:

src/components/va-back-to-top/va-back-to-top.tsx
cvalarida: Smells like a good code comment for posterity. :wink: I also recommend mentioning how we're using the `--undocked-width` in the `style` since that's not a native _thing._

src/components/va-back-to-top/va-back-to-top.tsx
cvalarida: I recommend documenting how this component works. Probably in a JSDoc on the `@Component()`. It's clever, which means folks who aren't privy to how it works right now might struggle figuring it out later.

cvalarida (commented):
This looks generally good, but...I had _many_ comments. :sweat_smile:

src/components/va-back-to-top/va-back-to-top.tsx
bkjohnson: Non-IE browsers will exit early instead of actually loading the polyfill:

https://github.com/w3c/IntersectionObserver/blob/cac7ad29d28629f8d25affda853eda3440b05f97/polyfill/intersection-observer.js#L17-L18

src/components/va-back-to-top/va-back-to-top.tsx
bkjohnson: Yup, that's exactly it.

https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#non-null-assertion-operator-postfix-

src/components/va-back-to-top/va-back-to-top.tsx
bkjohnson: What do you mean it isn't a native thing? I'm setting it [in a standard way](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#values_in_javascript), and CSS custom properties are already used elsewhere.

I'll add the code comment saying that this is what restricts the width :+1:

src/components/va-back-to-top/va-back-to-top.css
bkjohnson: Well, they aren't _necessary_. @GnatalieH said she was ok with them as part of this redesign though, since the same results are in the current version (even though this implementation is different).

src/components/va-back-to-top/va-back-to-top.css
bkjohnson: I've taken a bit more out, but at this point I think it's as minimal as it can get.

src/components/va-back-to-top/test/va-back-to-top.e2e.ts
bkjohnson: I can reproduce it consistently, and it happens even with the page looking like this:

```html
    <main>
    <h1>The top</h1>
    <div>Placeholder</div>
    <va-back-to-top></va-back-to-top>
    </main>
    <footer>The footer</footer>
 ```

File: src/components/va-back-to-top/test/va-back-to-top.e2e.ts [hunk 15]
```diff
@@ -1,81 +1,34 @@
...
-        velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint
-        occaecat cupidatat non proident, sunt in culpa qui officia deserunt
-        mollit anim id est laborum.
-      </p>
-      <p>Some more</p>
-      <p>Some more</p>
-      `;
+const pageStyles = `
+:root { --reveal-breakpoint: 200px}
+div {
```
cvalarida, line 6: I'd like to see this given an ID, but this is good. :+1:

src/components/va-back-to-top/test/va-back-to-top.e2e.ts
```diff
@@ -87,82 +40,75 @@ describe('va-back-to-top', () => {
...
-    const page = await newE2EPage();
-    await page.setContent(`
-    <main>
-    ${pageContent}
-    <va-back-to-top></va-back-to-top>
-    </main>
-    `);
+    const page = await pageSetup();
     const button = await page.find('va-back-to-top >>> button');
+    const notQuitePastRevealPoint = 199;
```
cvalarida, line 45: :joy: :100: 

Still, bonus points for using a `revealPoint` variable to declare both `--reveal-breakpoint` in the style and this one here. Just to make sure they're really tied together.

src/components/va-back-to-top/test/va-back-to-top.e2e.ts
```diff
@@ -87,82 +40,75 @@ describe('va-back-to-top', () => {
...
-    await page.setContent(`
-    <main>
-    ${pageContent}
-    <va-back-to-top></va-back-to-top>
-    <footer style="height: 100px">The footer</footer>
-    </main>
-    `);
+    const page = await pageSetup();
     const wrapper = await page.find('va-back-to-top >>> div');
+    const pastPlaceholder = 1000;
```
cvalarida, line 69: Same here: If we had a `revealPoint` variable, we could set this to `revealPoint + 1000` or whatever.

src/components/va-back-to-top/va-back-to-top.tsx
```diff
@@ -3,6 +3,16 @@ import classnames from 'classnames';
...
 
+/**
+ * This component has three main behaviors:
+ * 1. Going to the top of the page when the button is clicked
+ * 2. "Revealing" after scrolling down far enough
+ * 3. "Docking" when the dock becomes visible
+ *
+ * The span.reveal-point determines when the reveal happens.
+ * It is positioned on the page in an absolute manner, and the
+ * button will reveal when the span is above the viewport.
```
cvalarida, line 14:
```suggestion
 * It uses `position: absolute` so the button will reveal when
 * the `span` is above the viewport.
```

This doc comment is awesome. Thanks!

src/components/va-back-to-top/va-back-to-top.tsx
cvalarida: Right right, what I'm saying is that `--undocked-width` is a CSS custom property, not a standard CSS property. So a bit of explanation might be good for what we're using that custom property for.

What you've got now is great. :+1:

cvalarida (approved):
Looks great! I'd still like to see fewer magic numbers in the tests, but this is good enough for me now. :+1: Thanks! :raised_hands:

src/components/va-back-to-top/va-back-to-top.css
joshkimux: @cvalarida , I can chime in here as Noah is out sick at the moment-- animations could put folks with vestibular disorders at risk, so we'd recommend either (a) not including them in the redesign or (b) using the reduced motion css media query so it only activates when folks have an explicit preference for motion set

joshkimux: This ticket on [accessibility for smooth scrolling](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/312) might be helpful

joshkimux: I'd recommend the first as (a) it's the easiest and (b) it's less distracting to users 🙏

PR #173, opened 2021-09-17
Create featured content web component

holdenhinkle:
## Description
Closes <ticket>

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/src/components/va-featured-content/va-featured-content.css
+++ b/src/components/va-featured-content/va-featured-content.css
@@ -1,0 +1,3 @@
+:host {
+  display: block;
+}
--- a/src/components/va-featured-content/va-featured-content.tsx
+++ b/src/components/va-featured-content/va-featured-content.tsx
@@ -1,0 +1,18 @@
+import { Component, Host, h } from '@stencil/core';
+
+@Component({
+  tag: 'va-featured-content',
+  styleUrl: 'va-featured-content.css',
+  shadow: true,
+})
+export class VaFeaturedContent {
+
+  render() {
+    return (
+      <Host>
+        <slot></slot>
+      </Host>
+    );
+  }
+
+}
--- a/src/components/va-featured-content/va-featured-content.tsx
+++ b/src/components/va-featured-content/va-featured-content.tsx
@@ -1,4 +1,9 @@
-import { Component, Host, h } from '@stencil/core';
+import {
+  Component,
+  Host,
+  Prop,
+  h,
+} from '@stencil/core';
 
 @Component({
   tag: 'va-featured-content',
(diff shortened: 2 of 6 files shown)


CBonade (approved):
LGTM!

src/components/va-featured-content/va-featured-content.tsx
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * Sets the level for the HTML section heading elements.
+   * Valid values: 3, 4
+   */
+  @Prop() level: number;
+
+  /**
+   * Sets the text for the section heading.
+   */
+  @Prop() header: string;
```
bkjohnson, line 23: Instead of having to use two props, we can use a named slot like this:

https://github.com/department-of-veterans-affairs/component-library/blob/ccd5f49b327edfbfb814153be5a75b4e45257161/src/components/va-alert/va-alert.tsx#L144

This would allow the component to be used like this:

```html
<va-featured-content>
   <h3 slot="headline">Feature headline</h3>
   <div>Body text</div>
</va-featured-content>
```

src/components/va-featured-content/va-featured-content.css
```diff
@@ -0,0 +1,12 @@
+@import '../../mixins/accessibility.css';
+
+:host {
+  display: block;
+}
+
+.feature {
+  background: #e1f3f8;
```
bkjohnson, line 8: You can use the CSS variable `var(--color-primary-alt-lightest)` for this.

src/components/va-featured-content/va-featured-content.css
```diff
@@ -0,0 +1,12 @@
+@import '../../mixins/accessibility.css';
```
bkjohnson, line 1: I don't think we need this. Right now that file just contains the `sr-only` class, and it doesn't look like it's being used.

bkjohnson (commented):
Some initial comments.

src/components/va-featured-content/va-featured-content.css
holdenhinkle: Thanks. Updated.

src/components/va-featured-content/va-featured-content.css
holdenhinkle: Thanks. Updated.

src/components/va-featured-content/va-featured-content.tsx
holdenhinkle: @bkjohnson I read somewhere, that the Featured Content headline can only be `h3` or `h4`. Should we trust that only `h3` or `h4` headlines will be passed to that slot or should I add some sort of error handling like I did for the `level` prop?

See `src/components/va-featured-content/va-featured-content.tsx`:
```
    if (this.level !== 3 && this.level !== 4) {
      throw new Error('Featured Content header level must be 3 or 4');
    }
```

Thanks!

holdenhinkle: I removed the props and updated the tests.

bkjohnson: I'm not too worried about that being a problem right now. If it _is_ something we want to enforce, we can solve for that later.

bkjohnson (commented):
This looks like a minor thing, but I think we should move some "global" styles in here to ensure that child content in a feature will always look like it's supposed to. Right now in `formation`, certain nodes that are children of a `.feature` get special styling:

https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/12a43d63e40131b5bf4b350e118d9f6a43a6d31d/packages/formation/sass/base/_va.scss#L394-L426

Right now, we are getting some styling that looks right:

image

But it's not related to `.feature` at all:

https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/12a43d63e40131b5bf4b350e118d9f6a43a6d31d/packages/formation/sass/base/_va.scss#L113-L117

As an example, we can use `::slotted` to style the slotted content:

image

image


With this, even if the global style is removed, the component styling will still remain.

bkjohnson (approved):

PR #175, opened 2021-09-20
Updated css to change the halo focus color

noahgelman:
## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/359

Also related to this PR: https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/pull/684

## Definition of done
- [ ] Form elements have updated focus color

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "web-components",
-  "version": "0.9.6",
+  "version": "0.9.7",
   "description": "Stencil Component Starter",
   "main": "dist/index.cjs.js",
   "module": "dist/custom-elements/index.js",
--- a/src/components/va-checkbox/va-checkbox.css
+++ b/src/components/va-checkbox/va-checkbox.css
@@ -62,7 +62,7 @@ input:checked+label::before {
 }
 
 input:not([disabled]):focus+label:before {
-  outline: 2px solid var(--color-gold-light);
+  outline: 2px solid var(--color-olive);
   outline-offset: 2px;
 }
 
--- a/src/global/variables.css
+++ b/src/global/variables.css
@@ -13,6 +13,7 @@
   --color-gray-lightest: #f1f1f1;
   --color-green: #2e8540;
   --color-green-lightest: #e7f4e4;
+  --color-olive: #988530;
   --color-gold: #fdb81e;
   --color-gold-light: #f9c642;
   --color-gold-lightest: #fff1d2;
--- a/src/mixins/focusable.css
+++ b/src/mixins/focusable.css
@@ -3,6 +3,6 @@
 button:not([disabled]):focus,
 select:not([disabled]):focus,
 a:not([disabled]):focus {
-  outline: 2px solid var(--color-gold-light);
+  outline: 2px solid var(--color-olive);
   outline-offset: 2px;
 }


bkjohnson (commented):
I left a comment on the other PR before checking this one :sweat_smile: 

This is assuming that the web components will always be on a light background, correct? I _think_ that might be the case currently, but it won't always be true once more apps/static pages get migrated to use web components. For example, on mobile the footer uses accordions on a dark background:

image

Do we need to solve for someway of determining which color to use right now?

pjhill:
> I left a comment on the other PR before checking this one 😅
> 
> This is assuming that the web components will always be on a light background, correct? I _think_ that might be the case currently, but it won't always be true once more apps/static pages get migrated to use web components. For example, on mobile the footer uses accordions on a dark background:
> > 
> Do we need to solve for someway of determining which color to use right now?

This makes sense. While we are working on this, let's also address the need to have different colors depending on dark/light backgrounds. Is there an existing pattern for determining the light/dark background color of an element?

[Here's ](https://awik.io/determine-color-bright-dark-using-javascript/)an interesting discussion.

bkjohnson:
I can imagine a possible solution for having web components choose their own focus color based on their parent's background.

I'm not aware of an existing pattern for this. Maybe we could use the [CSS style declaration interface](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration) and implement a threshold comparison like in your link:

```js

const val = el.style.getPropertyValue('background-color')

const backgroundType = hsp(val) > someThreshold : 'light' : 'dark'
```

A web component could then use this to set a CSS custom property that would only affect itself and its descendants.

noahgelman:
@rtwell Have you looked into this?

noahgelman:
@bkjohnson Can you look into this for approval? Would like to move to stage

caw310:
@k80bowman , could you look at this since Brooks is out for a few days?

k80bowman:
> @k80bowman , could you look at this since Brooks is out for a few days?

I can take a look at it first thing tomorrow morning

k80bowman:
@noahgelman the change itself looks fine to me, but since this PR was created, this repo has been updated such that the web-components branch is no longer used (I believe). I think this change will have to be made in the `master` branch instead of the web-components branch, but if you want to wait until @bkjohnson gets back to confirm, that's fine. If you make the change on the `master branch` inside the `web-components` directory, the version numbers will need to be updated inside the `web-components` directory and inside the `core` directory.

bkjohnson:
Katy is right - a new PR will have to be made based off of `master`.

k80bowman:
@noahgelman are you going to be able to prioritize transferring this work to a new PR sometime soon?

noahgelman:
@k80bowman Yeah, I'm going to close this and open a new one

k80bowman:
@noahgelman awesome, thank you!

bkjohnson:
Closing this PR. We will work on implementing a double box-shadow design [based on this blog post](https://adhoc.team/2022/02/08/creating-focus-style-for-themable-design-system/). This was [discussed](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/359#issuecomment-1051110667) in a design system council meeting.

PR #179, opened 2021-09-22
Fixed icon accessibility error in AlertBox

timwright12:
## Description
This was throwing an error in CI nightwatch for a misplaces aria attribute.

```
 src/site/tests/home/00-required.e2e.spec   ✖ Failed [fail]: (
          Description:  Elements must only use allowed ARIA attributes
          Impact:       serious
          Rule ID:      aria-allowed-attr
          URL:          [n/a]
          DOM node:     
<i class="fas fa-times-circle" aria-label="Close icon"></i>
i[aria-label="Close\ icon"]
          More help:    https://dequeuniversity.com/rules/axe/4.3/aria-allowed-attr?application=axeAPI
        )
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/component-library",
-  "version": "3.1.1",
+  "version": "3.1.2",
   "description": "VA.gov component library in React",
   "keywords": [
     "react",
--- a/src/components/AlertBox/AlertBox.jsx
+++ b/src/components/AlertBox/AlertBox.jsx
@@ -93,7 +93,7 @@ class AlertBox extends Component {
         aria-label={this.props.closeBtnAriaLabel}
         onClick={this.props.onCloseAlert}
       >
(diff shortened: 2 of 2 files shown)


kelsonic (approved):

bkjohnson:
Is the only failure happening in ` src/site/tests/home/00-required.e2e.spec`? It looks like this issue is fixed in the web component version:

https://github.com/department-of-veterans-affairs/component-library/blob/5e152b7777cf20c1437c4469e1f6bb2690b2a4de/src/components/va-alert/va-alert.tsx#L147

timwright12:
@bkjohnson yeah just on that test. Does the template need to be updated then? cc @kelsonic

bkjohnson:
I've got two related PRs out to hopefully address this:

- Add the fullwidth alert variant as an option to `<va-alert>`: #181
- Use `<va-alert>` with the `fullwidth` prop in the `<Banner>` component: #180

PR #180, opened 2021-09-22
Use fullwidth alert component in Banner

bkjohnson:
## Description

This is meant to address an a11y issue with the `<AlertBox>` component @timwright12 mentioned in https://github.com/department-of-veterans-affairs/component-library/pull/179. In an effort to further the migration away from React components and towards Web Components, the `<Banner>` will use `<va-alert>` instead of its deprecated React counterpart.

If this gets merged, we should be able to delete [this file](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/a50f7304cfbb9824b75fd8af215769f1cf595c67/packages/formation/sass/modules/_m-emergency-banner.scss#L1).

## Testing done

Storybook :eyes: 

## Acceptance criteria
- [ ]

## Definition of done
- [ ] Documentation has been updated, if applicable
- [ ] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)

PR #181, opened 2021-09-22
Add full-width variant for `<va-alert>`

bkjohnson:
## Description

This is to be used in https://github.com/department-of-veterans-affairs/component-library/pull/180

Implements a fullwidth variant for `<va-alert>` as found on https://design.va.gov/components/alertboxes#full-width-alerts

image

## Testing done

Local storybook :eyes: 

## Screenshots

image

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/src/components/va-alert/va-alert.css
+++ b/src/components/va-alert/va-alert.css
@@ -26,12 +26,12 @@ div.alert {
   box-sizing: border-box;
 }
 
-div.alert:not(.fullwidth):not(.bg-only) {
+div.alert:not(.full-width):not(.bg-only) {
   border-left-style: solid;
   border-left-width: 10px;
 }
-div.alert.fullwidth.info,
-div.alert.fullwidth.warning {
+div.alert.full-width.info,
+div.alert.full-width.warning {
   border-top-style: solid;
   border-top-width: 10px;
 }
--- a/src/components/va-alert/va-alert.css
+++ b/src/components/va-alert/va-alert.css
@@ -23,9 +23,16 @@ div.alert {
   padding: 3.2rem 6.4rem 3.2rem 2.4rem;
   width: 100%;
   vertical-align: middle;
+  box-sizing: border-box;
+}
+
+div.alert:not(.fullwidth):not(.bg-only) {
   border-left-style: solid;
   border-left-width: 10px;
-  box-sizing: border-box;
+}
+div.alert.fullwidth {
+  border-top-style: solid;
+  border-top-width: 10px;
 }
 
 div.alert > i::before {
@@ -80,7 +87,7 @@ div.bg-only {
 }
 
 .info {
-  border-left-color: var(--color-primary-alt-dark);
+  border-color: var(--color-primary-alt-dark);
 }
 .info.bg-only {
   background-color: var(--color-primary-alt-lightest);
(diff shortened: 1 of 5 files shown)


src/components/va-alert/va-alert.tsx
```diff
@@ -121,8 +127,10 @@ export class VaAlert {
   }
 
   render() {
-    const { backgroundOnly, status, visible, closeable } = this;
-    const classes = `alert ${status} ${backgroundOnly ? 'bg-only' : ''}`;
+    const { backgroundOnly, status, fullwidth, visible, closeable } = this;
+    const classes = `alert ${status} ${fullwidth ? 'fullwidth' : ''} ${
```
bkjohnson, line 131: The `fullwidth` prop means we get a `fullwidth` class on the wrapping `<div>`

src/components/va-alert/va-alert.css
```diff
@@ -23,9 +23,17 @@ div.alert {
...
+
+div.alert:not(.fullwidth):not(.bg-only) {
   border-left-style: solid;
   border-left-width: 10px;
-  box-sizing: border-box;
+}
+div.alert.fullwidth.info,
+div.alert.fullwidth.warning {
+  border-top-style: solid;
+  border-top-width: 10px;
```
bkjohnson, line 36: The `.info` and `.warning` are low effort ways of enforcing this bullet point from [the docs](https://design.va.gov/components/alertboxes#more-about-full-width-alerts):

> Only available in `info` or `warning` variants.

I don't think we need to spend a lot of of time right now coming up with some strategy involving throwing errors or ensuring that certain props are mutually exclusive.

cvalarida: So we're defining a misuse of the component here as using `fullwidth` with an alert type of anything other than `info` or `warning`, right? If the component is misused, what should happen? What does happen?

src/components/va-alert/va-alert.tsx
```diff
@@ -48,6 +48,12 @@ export class VaAlert {
    */
   @Prop() closeable: boolean = false;
 
+  /**
+   * If true, the alert will be full width.
+   * Should be for emergency communication only.
+   */
+  @Prop() fullwidth: boolean = false;
```
cvalarida, line 55: :thinking: If we label this as `fullWidth`, does the use then become `<va-alert full-width>`? (Hyphenated instead of two words smushed together?)

cvalarida (approved):
There might be some unexpected behavior in here, and I think I like `full-width` as an attribute name better than `fullwidth`, but this generally looks good to me. :+1:

src/components/va-alert/va-alert.css
bkjohnson: As far as what should happen goes, I don't know :shrug:. We could log an error message and hope that a dev sees it, but they might not. They would be even less likely to see it if the props are dynamic in any way. We could also not render anything if the props are incompatible with each other. However, what _does_ happen is we get no border at all. Here is an example of what happens with `status="continue"`:

image

src/components/va-alert/va-alert.tsx
bkjohnson: Yeah, I'm pretty sure. I'll switch to that.

bkjohnson: https://github.com/department-of-veterans-affairs/component-library/pull/181/commits/fd1ba62f67af64455b2faa3d76c57fa8506a5506

image

PR #182, opened 2021-09-22
Add event listener and role to additional info component

CBonade:
## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/246

I was a little unclear here on exactly what was the goal of this ticket, so please let me know if more or something entirely different needs to be done here.

## Testing done
Tested in storybook and when focused, the spacebar collapsed the info box as expected.

### DIFF
--- a/src/components/AdditionalInfo/AdditionalInfo.jsx
+++ b/src/components/AdditionalInfo/AdditionalInfo.jsx
@@ -11,21 +11,11 @@ export default class AdditionalInfo extends React.Component {
     this.expandedContentId = uniqueId('tooltip-');
     this.state = { open: false };
   }
-  componentDidMount() {
-    this.setEventListener(this.expandedContentId, 'click');
-    this.setEventListener(this.expandedContentId, 'keydown');
-  }
 
-  setEventListener = (elemId, eventHandler) => {
-    const elem = document.getElementById(`${elemId}-button`);
-    if (elem) {
-      elem.addEventListener(eventHandler, event => {
-        const key = event.key || event.keyCode;
-        if (key === '' || key === 32 || eventHandler === 'click') {
-          event.preventDefault();
-          this.toggle();
-        }
-      });
+  handleKeyDown = event => {
+    if (event.key === '' || event.keyCode === 32) {
+      event.preventDefault();
+      this.toggle();
     }
   };
 
@@ -72,6 +62,8 @@ export default class AdditionalInfo extends React.Component {
         aria-expanded={this.state.open ? 'true' : 'false'}
         aria-controls={this.expandedContentId}
         onClick={this.toggle}
+        onKeyDown={this.handleKeyDown}
+        tabIndex="0"
       >
         <TagName className="additional-info-title">
           {triggerText}
--- a/src/components/AdditionalInfo/AdditionalInfo.jsx
+++ b/src/components/AdditionalInfo/AdditionalInfo.jsx
@@ -10,7 +10,22 @@ export default class AdditionalInfo extends React.Component {
     super(props);
     this.expandedContentId = uniqueId('tooltip-');
     this.state = { open: false };
+    this.setEventListener(this.expandedContentId, 'click');
+    this.setEventListener(this.expandedContentId, 'keydown');
   }
+
+  setEventListener = (elemId, eventHandler) => {
+    const elem = document.getElementById(`${elemId}-button`);
+    if (elem) {
+      elem.addEventListener(eventHandler, event => {
+        const key = event.key || event.keyCode;
+        if (key === '' || key === 32 || eventHandler === 'click') {
+          event.preventDefault();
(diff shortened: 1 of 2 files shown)


src/components/AdditionalInfo/AdditionalInfo.jsx
```diff
@@ -48,8 +65,9 @@ export default class AdditionalInfo extends React.Component {
     });
 
     const trigger = (
-      <button
-        type="button"
+      <a
```
bkjohnson, line 68: > conditionally use that component as an `<a>` tag with `role="button"`

I don't know what the condition _should_ be but it seems missing.

src/components/AdditionalInfo/AdditionalInfo.jsx
```diff
@@ -11,6 +11,23 @@ export default class AdditionalInfo extends React.Component {
...
+    if (elem) {
+      elem.addEventListener(eventHandler, event => {
+        const key = event.key || event.keyCode;
+        if (key === '' || key === 32 || eventHandler === 'click') {
+          event.preventDefault();
+          this.toggle();
+        }
+      });
+    }
+  };
```
bkjohnson, line 30: This looks a bit complicated. If possible I'd rather something like `onClick={toggleHandler}` and `onKeydown={toggleHandler}` in the JSX, and then define the handler function.

bkjohnson (commented):
Tagging @noahgelman since he might have a better idea about the conditional `<a>` mentioned in the ticket.

src/components/AdditionalInfo/AdditionalInfo.jsx
CBonade: I had spoken with @noahgelman prior and he said that his interpretation was that the conditional was based on whether there was expandable content or not. However, as I started working through that I couldn't think of a situation where we wouldn't have expandable content, unless it was future planning to have an alternate implementation of these. After discussing with our team engineering lead we couldn't figure out what the intended conditional was either, so I was going to defer to you all for what the intended outcome was here. Whatever is decided, I'm happy to implement.

src/components/AdditionalInfo/AdditionalInfo.jsx
CBonade: Not sure why I didn't do it this way in the first place. Think I got too deep into being focused on the codepen example trying to figure out the conditional that I stuck too close to it. Have pushed an update to this that tested well in storybook preview.

src/components/AdditionalInfo/AdditionalInfo.jsx
```diff
@@ -12,6 +12,13 @@ export default class AdditionalInfo extends React.Component {
     this.state = { open: false };
   }
 
+  handleKeyDown = event => {
+    if (event.key === '' || event.keyCode === 32) {
```
bkjohnson, line 16: Should the `key` be a space instead of an empty string?

```suggestion
    if (event.key === ' ' || event.keyCode === 32) {
```

Also while I was reading docs, I came across [this one](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode)/ It looks like `keyCode` is deprecated. Can we use `code` instead?

src/components/AdditionalInfo/AdditionalInfo.jsx
```diff
@@ -48,18 +55,21 @@ export default class AdditionalInfo extends React.Component {
     });
 
     const trigger = (
-      <button
-        type="button"
+      <a
+        role="button"
+        id={`${this.expandedContentId}-button`}
```
bkjohnson, line 60: What does setting the `id` give us?

bkjohnson (commented):
I'm a bit fuzzy on why it's better to have an `<a>` that is acting like a button rather than a link, but it seems like that isn't a problem.

Also, can you add a test that checks for the expand to happen on the appropriate keyboard events?

src/components/AdditionalInfo/AdditionalInfo.jsx
CBonade: This was a remnant from the overcomplicated event listener that I removed - I unfortunately forgot to remove this. This is gone now.

src/components/AdditionalInfo/AdditionalInfo.jsx
CBonade: Just ditched the keyCode entirely and went with your recommendation of key with a blank space (I didn't see the option to commit here, so pushed it up with my other changes).

CBonade:
> I'm a bit fuzzy on why it's better to have an `<a>` that is acting like a button rather than a link, but it seems like that isn't a problem.
> 
> Also, can you add a test that checks for the expand to happen on the appropriate keyboard events?

Unit test written - opens and closes the element while checking the state in both cases.

bkjohnson (approved):
Looks great!

PR #183, opened 2021-09-23
Copy over customEvent polyfill from vets-website to allow storybook t…

CBonade:
…o function in IE11

## Description
This addresses the master branch portion of concerns listed in https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/401.

A separate PR will need to be opened for the web components portion. This borrows the polyfill from vets-website that accounts for the `CustomEvent` ES6 function and stops it from breaking in `analytics.js` while testing storybook in IE11 on `master`.
## Testing done

Tested numerous components in IE11 on my PC and no issues that I was able to find.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -14,6 +14,10 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
+      - uses: actions/cache@v2
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
       - name: Install dependencies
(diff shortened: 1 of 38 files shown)


bkjohnson (commented):
Since this is just for storybook, would this work if you put it in this file?

https://github.com/department-of-veterans-affairs/component-library/blob/master/.storybook/preview.js

CBonade:
> Since this is just for storybook, would this work if you put it in this file?
> 
> https://github.com/department-of-veterans-affairs/component-library/blob/master/.storybook/preview.js

Went ahead and moved this and it works all the same. Ready to go!

bkjohnson (approved):

PR #185, opened 2021-09-27
Add sidebar styles to va-accordion

Rleahy22:
## Description
https://github.com/department-of-veterans-affairs/va.gov-team/issues/28922

We migrated most uses of `usa-accordion` to `<va-accordion>` but we were not able to do so with the sidebar-nav as the styles were too different. This closes that gap.

## Testing done
local via storybook

## Screenshots
Screen Shot 2021-09-27 at 3 20 17 PM

## Acceptance criteria
- [ ]

## Definition of done


cvalarida (commented):
The code all looks great to me, but I'm wondering two things:
- Does it make more sense for the `<va-accordion>` to have the `isSidebar` or the `<va-accordion-item>`? I'm leaning more toward the accordion itself. Did you try that? The web component style encapsulation might get a bit funky, but we might could make it work still.
- @GnatalieH what do you think of the design? This should be the same as is what's in the wild right now, so I _assume_ it's okay, but what say you?

Rleahy22:
> The code all looks great to me, but I'm wondering two things:
> 
> * Does it make more sense for the `<va-accordion>` to have the `isSidebar` or the `<va-accordion-item>`? I'm leaning more toward the accordion itself. Did you try that? The web component style encapsulation might get a bit funky, but we might could make it work still.
> * @GnatalieH what do you think of the design? This should be the same as is what's in the wild right now, so I _assume_ it's okay, but what say you?

@cvalarida yeah I definitely agree with that, but the styles needed to be applied based on `va-accordion-item` and not `va-accordion`: https://github.com/department-of-veterans-affairs/component-library/pull/185/files#diff-f7e0ba5886671b1b6e4a8e1e9c98a76a3f2aac69b66b4c3f5fe4f827fcf60d23
I'm not sure how to make that work.

GnatalieH:
Thanks for tagging me @cvalarida, I saw this issue come through but had forgotten to comment. I'd say that according to our documentation on design.va.gov, the Sidenav component should only be used as supplementary navigation item on the _side_ of a page. I don't think it was intended to be nested in another component. If there is a need for nesting content within an accordion, I'm 1. wondering what the use case is, and 2. would suggest using the Accordion to achieve the same thing. 

CC: @Rleahy22

Rleahy22:
> Thanks for tagging me @cvalarida, I saw this issue come through but had forgotten to comment. I'd say that according to our documentation on design.va.gov, the Sidenav component should only be used as supplementary navigation item on the _side_ of a page. I don't think it was intended to be nested in another component. If there is a need for nesting content within an accordion, I'm 1. wondering what the use case is, and 2. would suggest using the Accordion to achieve the same thing.
> 
> CC: @Rleahy22

That's a great point. The only real driver behind this change was consistency in moving all `usa-accordions` over to `va-accordion` but if we're kind of shoehorning it into `va-accordion` then it's probably not worth it as the sidebar nav works as is. I'm kind of thinking I will just close this pr and leave the sidebar-nav in content-build as it is, unless there's strong opinion to the contrary.

PR #186, opened 2021-09-29
Limit the body width to 1000px on full-width

bkjohnson:
## Description

It was noticed that some recent changes (#180, #181) resulted in the homepage banner being too wide in `vets-website`. This PR should fix that by limiting the body content to a maximum width of 1000px, while allowing the alert itself to extend across the entire page.

[Slack thread with additional context](https://dsva.slack.com/archives/C52CL1PKQ/p1632926162017600)

## Testing done

Looking at storybook and comparing it to va.gov

## Screenshots

image

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/src/components/va-alert/test/va-alert.e2e.ts
+++ b/src/components/va-alert/test/va-alert.e2e.ts
@@ -11,13 +11,11 @@ describe('va-alert', () => {
     expect(element).toEqualHtml(`
       <va-alert class="hydrated">
         <mock:shadow-root>
-          <div>
-            <div class="alert info">
-              <i aria-hidden="true" role="img"></i>
-              <div class="body">
-                <slot name="headline"></slot>
-                <slot></slot>
-              </div>
+          <div class="alert info">
+            <i aria-hidden="true" role="img"></i>
+            <div class="body">
+              <slot name="headline"></slot>
+              <slot></slot>
             </div>
           </div>
         </mock:shadow-root>
--- a/src/components/va-alert/test/va-alert.e2e.ts
+++ b/src/components/va-alert/test/va-alert.e2e.ts
@@ -11,11 +11,13 @@ describe('va-alert', () => {
     expect(element).toEqualHtml(`
       <va-alert class="hydrated">
         <mock:shadow-root>
-          <div class="alert info">
-            <i aria-hidden="true" role="img"></i>
-            <div class="body">
-              <slot name="headline"></slot>
-              <slot></slot>
+          <div>
+            <div class="alert info">
+              <i aria-hidden="true" role="img"></i>
+              <div class="body">
+                <slot name="headline"></slot>
+                <slot></slot>
+              </div>
             </div>
           </div>
         </mock:shadow-root>
--- a/src/components/va-alert/test/va-alert.e2e.ts
+++ b/src/components/va-alert/test/va-alert.e2e.ts
@@ -25,6 +25,15 @@ describe('va-alert', () => {
     `);
   });
 
+  it('renders the right markup for the full-width variant', async () => {
+    const page = await newE2EPage();
+    await page.setContent('<va-alert full-width></va-alert>');
+
+    const wrapper = await page.find('va-alert >>> div');
+
+    expect(wrapper).toHaveClass('full-width');
+  });
+
   it('renders an empty div with a "polite" aria-live tag when not visible', async () => {
     const page = await newE2EPage();
 
--- a/src/components/va-alert/test/va-alert.e2e.ts
+++ b/src/components/va-alert/test/va-alert.e2e.ts
@@ -23,15 +23,6 @@ describe('va-alert', () => {
     `);
   });
 
-  it('renders the right markup for the full-width variant', async () => {
-    const page = await newE2EPage();
-    await page.setContent('<va-alert full-width></va-alert>');
-
-    const wrapper = await page.find('va-alert >>> div');
-
-    expect(wrapper).toHaveClass('full-width');
-  });
-
   it('renders an empty div with a "polite" aria-live tag when not visible', async () => {
     const page = await newE2EPage();
 
--- a/src/components/va-alert/va-alert.css
+++ b/src/components/va-alert/va-alert.css
@@ -30,7 +30,6 @@ div:not(.full-width) div.alert:not(.bg-only) {
   border-left-style: solid;
   border-left-width: 10px;
 }
-div.full-width,
 div.full-width {
   border-top-style: solid;
   border-top-width: 10px;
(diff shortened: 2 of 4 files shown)


src/components/va-alert/va-alert.tsx
```diff
@@ -128,33 +128,37 @@ export class VaAlert {
...
     if (!visible) return <div aria-live="polite" />;
 
     return (
       <Host>
-        <div role={role} aria-live={ariaLive} class={classes}>
-          <i aria-hidden="true" role="img"></i>
-          <div class="body" onClick={this.handleAlertBodyClick.bind(this)}>
-            {!backgroundOnly && <slot name="headline"></slot>}
-            <slot></slot>
+        <div class={fullWidth ? 'full-width' : ''}>
```
bkjohnson, line 139: The simplest way to limit the body width while letting the existing styles apply to the rest of the markup was to create a wrapping div. This div will be at full width, which means the top border is also full width. But this also allows the CSS to limit the width of the inner div.

src/components/va-alert/va-alert.css
```diff
@@ -26,16 +26,21 @@ div.alert {
...
-div.alert.full-width.warning {
+div.full-width,
+div.full-width {
   border-top-style: solid;
   border-top-width: 10px;
 }
 
+div.full-width div.alert {
+  max-width: 1000px;
+  margin: 0 auto;
```
bkjohnson, line 41: This is based on what I saw on va.gov:

image

src/components/va-alert/test/va-alert.e2e.ts
```diff
@@ -11,18 +11,29 @@ describe('va-alert', () => {
...
+                <slot></slot>
+              </div>
             </div>
           </div>
         </mock:shadow-root>
       </va-alert>
     `);
   });
 
+  it('renders the right markup for the full-width variant', async () => {
```
cvalarida, line 28: Define "the right markup" in the test name.

src/components/va-alert/va-alert.css
```diff
@@ -26,16 +26,21 @@ div.alert {
...
 
-div.alert:not(.full-width):not(.bg-only) {
+div:not(.full-width) div.alert:not(.bg-only) {
   border-left-style: solid;
   border-left-width: 10px;
 }
-div.alert.full-width.info,
-div.alert.full-width.warning {
+div.full-width,
+div.full-width {
```
cvalarida, line 34:
```suggestion
div.full-width {
```
Does this not limit the full-width banner to `info` and `warning` types now?

src/components/va-alert/va-alert.tsx
cvalarida: Did you choose to not use the `host:` selector because the changes it would require in the CSS would be pretty extensive?

cvalarida (approved):

src/components/va-alert/va-alert.css
bkjohnson: Ah, I need to make an additional CSS change. I had intended to move this down towards the bottom but didn't finish implementing it.

src/components/va-alert/va-alert.tsx
bkjohnson: Originally I was just trying to do this quickly, thinking that I could get the fix out before the daily deploy. Now that I have more time I'll actually give that a shot :sweat_smile:

src/components/va-alert/va-alert.css
bkjohnson: Here it is: https://github.com/department-of-veterans-affairs/component-library/pull/186/commits/d137e5fab64900ef37feca6d51a63fcd883436c7

src/components/va-alert/va-alert.css
```diff
@@ -11,6 +11,33 @@
   background-color: unset;
 }
 
+:host([full-width='true']),
+:host([full-width='']) {
+  border-top-style: solid;
+  border-top-width: 10px;
+}
```
bkjohnson, line 18: If the component has the `full-width` prop, give it a top border. No additional markup required.

src/components/va-alert/va-alert.css
```diff
@@ -11,6 +11,33 @@
...
+:host([status='info']) {
+  border-color: var(--color-primary-alt-dark);
+}
+:host([status='error']) {
+  border-color: var(--color-secondary-dark);
+}
+:host([status='continue']),
+:host([status='success']) {
+  border-color: var(--color-green);
+}
```
bkjohnson, line 32: This benefits non `full-width` alerts as well. These replace the individual pieces of CSS down below:

```css
.[$STATUS] {
    border-color: var(--status-color);
}
```

This is possible due to the `border-color: inherit`.

src/components/va-alert/va-alert.css
```diff
@@ -11,6 +11,33 @@
...
+
+:host([full-width='']) div.alert,
+:host([full-width='true']) div.alert {
+  border-left: none;
+  max-width: 1000px;
+  margin: 0 auto;
+}
```
bkjohnson, line 39: If we're a `full-width` alert, remove the left border from the inner div(*) and limit the width appropriately.

*I didn't want this PR to get too out of scope and have it change the original implementation of the left border. There may be a reason why the border is on an inner div as opposed to `:host`.

src/components/va-alert/va-alert.css
```diff
@@ -26,15 +53,11 @@ div.alert {
   box-sizing: border-box;
 }
 
-div.alert:not(.full-width):not(.bg-only) {
+div.alert:not(.bg-only) {
+  border-color: inherit;
```
bkjohnson, line 57: We aren't adding an additional `full-width` class, so the `:not` selector can go. The `border-color` change is to support the new pattern for status colors - defining them at the host.

src/components/va-alert/va-alert.tsx
```diff
@@ -127,10 +127,8 @@ export class VaAlert {
   }
 
   render() {
-    const { backgroundOnly, status, fullWidth, visible, closeable } = this;
-    const classes = `alert ${status} ${fullWidth ? 'full-width' : ''} ${
-      backgroundOnly ? 'bg-only' : ''
-    }`;
+    const { backgroundOnly, status, visible, closeable } = this;
+    const classes = `alert ${status} ${backgroundOnly ? 'bg-only' : ''}`;
```
bkjohnson, line 131: This is what we had before the changes in #181:

https://github.com/department-of-veterans-affairs/component-library/blob/3694481bca2490adc534effed971bd53dc1806dc/src/components/va-alert/va-alert.tsx#L124-L125

bkjohnson (commented):
The solution is now based in CSS.

src/components/va-alert/va-alert.tsx
bkjohnson: I've changed it now so that it's based around `:host`.

cvalarida (approved):

PR #188, opened 2021-09-29
Provide property to optimize va-text-input fields for numeric input

acrollet:
## Description

This PR is in support of [va.gov-team#30129 ](department-of-veterans-affairs/va.gov-team#30129). With this PR, you may specify that input should be numeric, causing the number keypad to be shown on mobile:

`<va-text-input numeric-input />`

When this attribute is set, `inputmode` is set to `numeric` and the `pattern` attribute is set to `[0-9]*`. This is the approach suggested in this excellent post: [Why the GOV.UK Design System team changed the input type for numbers](https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/) 

Additionally, it aligns with the [accessibility guidance we've received](https://dsva.slack.com/archives/C022AC2STBM/p1632408155062000?thread_ts=1632337879.046900&cid=C022AC2STBM).

This is my first PR to `component-library` so I much appreciate any guidance/help!

## Testing done

Tested locally with stencil using this html:

```html
<va-text-input
  label="Last 4 digits of your Social Security number"
  maxlength=4
  name="last-4-ssn"
  numeric-input
  required
/>
```

## Acceptance criteria

- [ ] a VA-sanctioned approach for displaying the number keypad on text input fields is available

## Definition of done

### DIFF
--- a/src/components/va-text-input/test/va-text-input.e2e.ts
+++ b/src/components/va-text-input/test/va-text-input.e2e.ts
@@ -11,7 +11,7 @@ describe('va-text-input', () => {
     expect(element).toEqualHtml(`
       <va-text-input class="hydrated">
         <mock:shadow-root>
-          <input id="inputField" type="text" />
+          <input id="inputField" inputmode="" pattern="" type="text">
         </mock:shadow-root>
       </va-text-input>
     `);
@@ -64,7 +64,7 @@ describe('va-text-input', () => {
       <va-text-input class="hydrated" label="This is a field" required="">
         <mock:shadow-root>
           <label for="inputField">This is a field <span class="required">(*Required)</span></label>
-          <input id="inputField" type="text" />
+          <input id="inputField" inputmode="" pattern="" type="text">
         </mock:shadow-root>
       </va-text-input>
     `);
@@ -160,7 +160,7 @@ describe('va-text-input', () => {
     );
   });
 
-  it('adds adds a character limit with descriptive text', async () => {
+  it('adds a character limit with descriptive text', async () => {
     const page = await newE2EPage();
     await page.setContent('<va-text-input maxlength="3" value="22"/>');
 
@@ -175,5 +175,48 @@ describe('va-text-input', () => {
     expect((await page.find('va-text-input >>> small')).innerText).toContain(
       '(Max. 3 characters)',
     );
+  });
+
+  it('adds a field optimized for numeric input', async () => {
+    const page = await newE2EPage();
+    await page.setContent('<va-text-input numeric-input />');
+
+    // Level-setting expectations
+    const inputEl = await page.find('va-text-input >>> input');
+    expect(inputEl.getAttribute('inputmode')).toBe('numeric');
+    expect(inputEl.getAttribute('pattern')).toBe('[0-9]*');
+
+    // Test the functionality
+    await inputEl.press('2');
+    expect(await inputEl.getProperty('value')).toBe('2');
+  });
+
+  it('allows manually setting the inputmode attribute', async () => {
+    const page = await newE2EPage();
+    await page.setContent('<va-text-input inputmode="decimal" />');
+
+    // Level-setting expectations
+    const inputEl = await page.find('va-text-input >>> input');
+    expect(inputEl.getAttribute('inputmode')).toBe('decimal');
+
+    // Test the functionality
+    await inputEl.press('2');
+    await inputEl.press('.');
+    await inputEl.press('0');
+    expect(await inputEl.getProperty('value')).toBe('2.0');
+  });
+
+  it('allows manually setting the pattern attribute', async () => {
+    const page = await newE2EPage();
+    await page.setContent('<va-text-input pattern="[a-z]*" />');
+
+    // Level-setting expectations
+    const inputEl = await page.find('va-text-input >>> input');
+    expect(inputEl.getAttribute('pattern')).toBe('[a-z]*');
+
+    // Test the functionality
+    await inputEl.press('v');
+    await inputEl.press('a');
+    expect(await inputEl.getProperty('value')).toBe('va');
   });
 });
--- a/src/components/va-text-input/va-text-input.tsx
+++ b/src/components/va-text-input/va-text-input.tsx
@@ -27,6 +27,11 @@ export class VaTextInput {
   @Prop() error?: string | HTMLElement;
 
   /**
+   * Optimize the field for numeric input.
+   */
+  @Prop() numericInput?: boolean;
+
+  /**
    * Set the input to required and render the (Required) text.
    */
   @Prop() required?: boolean;
@@ -35,6 +40,16 @@ export class VaTextInput {
    * Placeholder text to show in the input field.
    */
   @Prop() placeholder?: string;
+
+  /**
+   * The inputmode attribute.
+   */
+  @Prop() inputmode?: string = '';
+
+  /**
+   * The pattern attribute.
+   */
+  @Prop() pattern?: string = '';
 
   /**
    * The maximum number of characters allowed in the input.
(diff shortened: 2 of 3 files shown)


cvalarida (commented):
So....I have a lot of thoughts here. :sweat_smile:

I'll start with a clarification: This PR does a couple more things than just add `numeric-input` to the props list. It also adds `inputmode` and `pattern`.

**TL;DR:** Scroll down to _Conclusion._ :sweat_smile: I talked to myself for a while to figure out what the best answer is here.

# Expanding the scope
Currently, the purpose of this component is to capture only text input. Expanding the scope to also include numbers seems okay, but I'm debating whether or not it would be better as a separate component simply for a clear separation of concerns. :thinking: We'd then have like a `<va-text-input>` and `<va-number-input>`.

## What's next?
Following this addition through to the next logical conclusion, though, we'd expand the scope further into other input types—but just the ones that capture text, so not `type="checkbox"` or `type="radio"` or anything like that.

The input types we may eventually need include:
- `email`
- `search`
- `tel`
- `url`
- `number`

## What does the best implementation look like?
This time next year, when we have totally new engineers working with the component library, I want it to be clear, easy to use, and easy to understand. With no weird warbles in our API because of historical reasons. So what does the best implementation look like?

The options I see are either make a web component for each type, or to continue to add to this one web component to accommodate all the input types.

### Lots of web components
By making a new web component for each type, we either start from scratch on each one, or have each input type component render a common `<va-input>` (or something) that handles all the common logic, styles, event handling (with some possible caveats because of the shadow DOM), etc.

So: `<va-number-input>` renders `<va-input>`. `va-text-input` exposes a subset of `va-input`'s API with potentially some quality of life improvements such as adding a `decimal` prop which modifies the `inputmode` and `pattern` passed to the `va-input`.

If we go down this path, this is the preferred way for maintainability.

### One component to rule them all
If we make `<va-text-input>` the one component which handles all input types, that seems like it'd be fine, honestly. There aren't that many input types, and they all seem pretty lightweight. If implementation gets messy, I'm confident there are ways we can refactor it to keep it conceptually clean.

What I'm most concerned about with this approach is a clean API. Providing a new boolean prop for each new input type gets messy fast. Instead, what I propose we do is mirror the `input` API (unfortunately) and provide a `type` and `inputmode`, ignoring `pattern` for now. (I don't want conflicting validation error messages, and I'd need to confer with a11y specialists to know what the UX is with both. There are edge cases I'd need to take the time to suss out.)

Later, we may also need to add input type-specific props such as `min`, `max`, and `step` because the native `<input>`'s API is a mess, but for now, I'd like to keep it simple.

# Conclusion
That was a lot of words for me to figure out what I'd like to see here. :sweat_smile: Had to talk it through with myself, I guess.

**Let's go with _One component to rule them all_ by mirroring the `<input>` API.** This means removing the `numericInput` prop in favor of `type`. But let's also make sure that `type` is of a supported type. If it's unsupported, we'll ignore it (which will default to `text`).

src/components/va-text-input/test/va-text-input.e2e.ts
```diff
@@ -176,4 +176,48 @@ describe('va-text-input', () => {
...
+    await page.setContent('<va-text-input inputmode="numeric" />');
+
+    // Level-setting expectations
+    const inputEl = await page.find('va-text-input >>> input');
+    expect(inputEl.getAttribute('inputmode')).toBe('numeric');
+
+    // Test the functionality
+    await inputEl.press('2');
+    await inputEl.press('0');
+    expect(await inputEl.getProperty('value')).toBe('20');
```
cvalarida, line 191: :thinking: I'm kind of torn on these tests. On the one hand, it's great to make sure it behaves as expected. I almost want you to add the failure mode to the tests as well. (Typing invalid data doesn't work.) I think instead, though, I'd prefer to leave this level of testing alone. That is, what we're doing here isn't testing that our web component behaves as expected, but that the underlying _browser_ behaves as expected on native elements (because we're interfacing directly with the `<input>` here).

How do you feel about not typing anything into the `<input>` and instead have the following tests?
1. Iterate over an allow-list for valid `type`s, expecting that the `type` attribute is set correctly for each
2. Check that an invalid `type` sets the `type` attribute on the `<input>` element to `text`
3. Iterate over an allow-list for valid `inputmode`s, expecting that the `inputmode` attribute is set correctly for each
4. Check that an invalid `inputmode` sets the `inputmode` attribute on the `<input>` element to whatever the default is (nothing?)

src/components/va-text-input/va-text-input.tsx
```diff
@@ -120,6 +130,15 @@ export class VaTextInput {
     const describedBy =
       `${this.ariaDescribedby} ${this.error ? 'error-message' : ''}`.trim() ||
       null; // Null so we don't add the attribute if we have an empty string
+    const inputMode =
+      this.inputmode ?
+      this.inputmode :
+      null; // Null so we don't add the attribute if we have an empty string
+    const type =
+      ['email', 'number', 'search', 'tel', 'text', 'url'].includes(this.type) ?
```
cvalarida, line 138: Can you pull this out into a named variable for clarity? Maybe `allowedInputTypes` or `textInputTypes` or something.

src/components/va-text-input/va-text-input.tsx
```diff
@@ -120,6 +130,15 @@ export class VaTextInput {
     const describedBy =
       `${this.ariaDescribedby} ${this.error ? 'error-message' : ''}`.trim() ||
       null; // Null so we don't add the attribute if we have an empty string
+    const inputMode =
+      this.inputmode ?
+      this.inputmode :
+      null; // Null so we don't add the attribute if we have an empty string
```
cvalarida, line 136: Should this also be filtered by an allowed list?

cvalarida (approved):
I'd love to make a development-only `console.error` when we use invalid `type`s, but I don't know how to shake that out of the prod build, so :man_shrugging: .

Anyhow, this looks good. I have some idea about tests and stuff, but they're pretty minor, so I'm giving this the ol' :+1: .

Thanks for doing this! :bow:

src/components/va-text-input/va-text-input.tsx
acrollet: I went back and forth about this, but decided not to filter because we're not trying to restrict the allowed values in the same way as with `type`, (Unless perhaps we wanted to disallow passing `none`?) and browsers default to using `text` if an invalid value is passed.  That said I don't have strong feelings about it if you'd like to see it there.

src/components/va-text-input/test/va-text-input.e2e.ts
acrollet: > That is, what we're doing here isn't testing that our web component behaves as expected, but that the underlying browser behaves as expected on native elements

That makes sense, I'll update the tests :-)

acrollet: Changed to add #s 1-3, not doing #4 unless we decide to explicitly restrict allowed values, see the other comment. :-)

acrollet:
> I'd love to make a development-only console.error when we use invalid types

This seemed worth figuring out, this now shows when running with `--dev`:

Screen Shot 2021-10-04 at 4 25 41 PM

cvalarida:
Hey, that's awesome! :raised_hands:  Where did you find the documentation for `Build`?

acrollet:
> Hey, that's awesome! 🙌 Where did you find the documentation for `Build`?

It's not documented afaik, I found it in the Stencil source on the principle of "_someone_ must have wanted to do this already..."

cvalarida:
Nice spelunking. :tada: 

I'm not sure if it'll get shaken out of the prod build, but we can cross that bridge later. If we need to, we can just remove that `if` statement. :man_shrugging: 

Thanks again!

cvalarida:
:thinking: Not sure why the action didn't run...or how to manually get it to run on the latest commit. I looked at the [history](https://github.com/department-of-veterans-affairs/component-library/actions/workflows/node.js.yml?query=branch:acrollet-30129-va-text-input-numeric), but :man_shrugging: .

Maybe try pushing up another commit? :confused:

acrollet:
@cvalarida yeah, that's a weird failure :-\ Anyway an empty commit did the trick.

PR #191, opened 2021-09-30
Create ProcessList component

ddzz:
## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/247

## Testing done
Manual testing in Storybook.

## Acceptance criteria
- [ ] Manual testing.
- [ ] CI passes.


src/components/ProcessList/ProcessList.jsx
```diff
@@ -0,0 +1,32 @@
...
+    super(props);
+  }
+
+  render() {
+    const { children } = this.props;
+
+    return (
+      <div className="process schemaform-process">
+        <ol role="list">
+          {children.map((child, index) => (
```
cvalarida, line 20: What happens if there's only one child passed to this component? (Would be weird for a process list, but :man_shrugging:) We might want to use [`React.Children.map`](https://reactjs.org/docs/react-api.html#reactchildrenmap) instead.

src/components/ProcessList/ProcessList.unit.spec.jsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import { shallow } from 'enzyme';
```
cvalarida, line 2: Generally speaking, using `@testing-library/react` is preferred.

cvalarida (approved):
I had a couple of minor comments, but nothing to stop me from giving this the ol' :+1: .

src/components/ProcessList/ProcessList.jsx
ddzz: Thanks for reviewing. Good catch, nothing renders when there's only one child. It's working with `React.Children`.

PR #197, opened 2021-10-15
Integrate web-components branch, Pt. 2

bkjohnson:
## Description

Part of https://github.com/department-of-veterans-affairs/va.gov-team/issues/30694

A lot of the diff is from the `yarn.lock` file, and most of the files are just renames with the import location changed. See my [review comment](https://github.com/department-of-veterans-affairs/component-library/pull/197#pullrequestreview-783753819) for more detail.

## Testing done

Storybook, CI, & local building & testing

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/core/src/main.js
+++ b/packages/core/src/main.js
@@ -1,7 +1,6 @@
 import '@department-of-veterans-affairs/web-components/dist/component-library/component-library.css';
-import webComponentDocs from '@department-of-veterans-affairs/web-components/component-docs';
 
 export * from '@department-of-veterans-affairs/react-components';
 export * from '@department-of-veterans-affairs/web-components';
 export * from '@department-of-veterans-affairs/web-components/loader';
-export {webComponentDocs};
+export {default as webComponentDocs} from '@department-of-veterans-affairs/web-components/component-docs';


.github/workflows/chromatic.yml
```diff
@@ -18,13 +18,14 @@ jobs:
...
+      - run: yarn workspace @department-of-veterans-affairs/web-components run build
         # 👇 Adds Chromatic as a step in the workflow
       - name: Publish to Chromatic
-        uses: chromaui/action@v1
-        # Chromatic GitHub Action options
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          # 👇 Chromatic projectToken, refer to the manage page to obtain it.
-          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+        run: npx chromatic --project-token=${{ secrets.CHROMATIC_PROJECT_TOKEN }}
```
bkjohnson, line 30: I've tried lots of different things, but I haven't gotten this action to work yet under the new repo structure. However, we can still deploy to chromatic from the command line, so all isn't lost.

.github/workflows/node.js.yml
```diff
@@ -10,22 +10,18 @@ on:
...
         node-version: 14.x
-    - run: yarn
-    - run: cd packages/web-components && yarn
-    - run: cd packages/web-components && yarn build
-    - run: cd packages/react-components && yarn
-    - run: cd packages/react-components && yarn test
+    - run: yarn workspaces foreach run install
+    - run: yarn workspace @department-of-veterans-affairs/react-components run build
+    - run: yarn workspace @department-of-veterans-affairs/web-components run build
+    - run: yarn workspaces foreach run test
```
bkjohnson, line 27: This lets us run the tests, and they pass. However, without the cache things don't seem to get installed properly. I'd like to address this in a following PR.

packages/core/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "license": "MIT",
+  "scripts": {
+    "build": "webpack"
+  },
+  "main": "dist/main-bundle.js",
+  "module": "src/main.js",
+  "dependencies": {
+    "@department-of-veterans-affairs/react-components": "workspace:*",
+    "@department-of-veterans-affairs/web-components": "workspace:*"
+  },
```
bkjohnson, line 14: Getting these two packages bundled as one is the main goal of this PR. The webpack config makes everything a named export, and in a followup PR I'd like to modify it (possibly with the copy plugin) to move the babel-ified files from `react-components` into the build output here so we don't immediately have to change the `vets-website` imports that look like this:

```js
import AdditionalInfo from '@department-of-veterans-affairs/component-library/AdditionalInfo';
```

packages/core/src/main.js
```diff
@@ -0,0 +1,7 @@
+import '@department-of-veterans-affairs/web-components/dist/component-library/component-library.css';
+import webComponentDocs from '@department-of-veterans-affairs/web-components/component-docs';
+
+export * from '@department-of-veterans-affairs/react-components';
+export * from '@department-of-veterans-affairs/web-components';
+export * from '@department-of-veterans-affairs/web-components/loader';
+export {webComponentDocs};
```
bkjohnson, line 7: This is what the `core` bundle will contain:

1. The css for the web components that defines all of the top-level CSS variables
1. The JSON for web component docs that gets used in storybook
1. The files exported by `index.js` in `react-components`
1. The exports from `web-components/dist`
1. The exports from `web-components/loader` so that we get `applyPolyfills`.

packages/core/webpack.config.js
```diff
@@ -0,0 +1,45 @@
...
+          MiniCssExtractPlugin.loader, // instead of style-loader
+          'css-loader'
+        ]
+      }
+    ],
+  },
+  resolve: {
+    alias: {
+      '@department-of-veterans-affairs/web-components/react-bindings':
+        path.resolve(__dirname, '../web-components/'),
```
bkjohnson, line 31: Fairly straightforward webpack config. This alias is specifically for the `Banner` component where we are using the `VaAlert` react binding.

packages/core/webpack.config.js
```diff
@@ -0,0 +1,45 @@
...
+          loader: "babel-loader",
+        },
+      },
+      {
+        test: /\.css$/,
+        use: [
+          MiniCssExtractPlugin.loader, // instead of style-loader
+          'css-loader'
+        ]
+      }
```
bkjohnson, line 25: This allows us to have the CSS file from `web-components` in our export.

packages/react-components/index.js
```diff
@@ -0,0 +1,88 @@
+import AcceptTermsPrompt from './AcceptTermsPrompt';
```
bkjohnson, line 1: This index file exports all of our built components so that `core` has easy access using:

```js
export * from '@department-of-veterans-affairs/react-components';
```

Also made possible by the `files` property in the `package.json`

packages/react-components/package.json
```diff
@@ -9,6 +9,9 @@
     "va",
     "vads"
   ],
+  "files": [
+    "index.js"
+  ],
```
bkjohnson, line 14: Allows `core` to import the index file easily.

packages/react-components/package.json
```diff
@@ -62,14 +63,8 @@
     "@babel/preset-react": "^7.14.5",
     "@department-of-veterans-affairs/formation": "^6.17.3",
     "@department-of-veterans-affairs/web-components": "workspace:packages/web-components",
-    "@storybook/addon-a11y": "6.1.21",
-    "@storybook/addon-actions": "6.1.21",
-    "@storybook/addon-essentials": "6.1.21",
-    "@storybook/addon-links": "6.1.21",
-    "@storybook/react": "6.1.21",
```
bkjohnson, line 69: All these storybook dependencies were moved to the `storybook` package.

packages/react-components/scripts/build.js
```diff
@@ -8,7 +8,7 @@ const path = require('path');
 
 console.log('Starting component-library build');
 console.log('Cleaning old build');
-rimraf.sync('./*.js');
+rimraf.sync('./[.]*!(index).js');
```
bkjohnson, line 11: We don't want the build script to delete that `index.js` file that exports all of the components.

packages/react-components/src/components/Telephone/Telephone.jsx
```diff
@@ -1,9 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import { CONTACTS } from './contacts';
+import { CONTACTS, contactsMap } from './contacts';
+import { patternsMap } from './patterns';
 
-export { CONTACTS };
+export { CONTACTS, contactsMap, patternsMap };
```
bkjohnson, line 7: Since we moved the stories into their own package, I had to do a bit of refactoring here to make sure we still have access to the same exports.

packages/storybook/.storybook/preview.js
```diff
@@ -1,10 +1,10 @@
 import '@department-of-veterans-affairs/formation/dist/formation.min.css';
 import './style.scss';
 import '@department-of-veterans-affairs/formation/dist/formation';
-import { withHTML } from '@whitespace/storybook-addon-html/react';
+import {withHTML} from '@whitespace/storybook-addon-html/react';
 
-import 'web-components/dist/component-library/component-library.css';
-import { applyPolyfills, defineCustomElements } from 'web-components/loader';
+import '@department-of-veterans-affairs/component-library/dist/main.css';
+import {applyPolyfills, defineCustomElements} from '@department-of-veterans-affairs/component-library';
```
bkjohnson, line 7: The imports changed now that we're getting from the `core` export.

packages/storybook/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@department-of-veterans-affairs/storybook",
+  "packageManager": "yarn@3.0.2",
+  "scripts": {
+    "storybook": "start-storybook -p 6006",
+    "build-storybook": "build-storybook"
+  },
+  "dependencies": {
+    "@department-of-veterans-affairs/component-library": "workspace:packages/core",
```
bkjohnson, line 9: `storybook` is its own package since it contains stories for `react-components` _and_ `web-components`. Leaving it in `react-components` didn't make much sense now that we have a monorepo.

packages/storybook/stories/Telephone.stories.jsx
```diff
@@ -6,12 +6,10 @@ import {
...
   ArgsTable,
   Stories,
+  PRIMARY_STORY,
 } from '@storybook/addon-docs/blocks';
 
-import Telephone, { CONTACTS, PATTERNS } from './Telephone';
-import { contactsMap } from './contacts';
-import { patternsMap } from './patterns';
-import Table from '../Table/Table';
+import {Telephone, PHONE_CONTACTS, PHONE_CONTACTS_MAP, PHONE_PATTERNS, PHONE_PATTERNS_MAP, Table} from '@department-of-veterans-affairs/component-library';
```
bkjohnson, line 12: Since the story file is no longer a sibling to the original component, I had to make sure that we exported those consts from the sibling files in order to access them from storybook.

packages/storybook/stories/Telephone.stories.jsx
```diff
@@ -55,7 +53,7 @@ const Page = () => (
     <Subtitle />
     <Description />
     <Primary />
-    <ArgsTable />
+    <ArgsTable story={PRIMARY_STORY}/>
```
bkjohnson, line 56: This change is due to a storybook version bump, and I based it on [current documentation](https://storybook.js.org/docs/react/writing-docs/docs-page#remixing-docspage-using-doc-blocks). Without it I wasn't seeing the args table.

image

packages/storybook/stories/wc-helpers.jsx
```diff
@@ -10,13 +10,13 @@ import {
   PRIMARY_STORY,
 } from '@storybook/addon-docs/blocks';
 
-import componentDocs from 'web-components/component-docs.json';
+import {webComponentDocs} from '@department-of-veterans-affairs/component-library';
```
bkjohnson, line 13: The `core` webpack config bundles the JSON in with the exports, rather than leaving it as its own file.

packages/core/src/main.js
cvalarida: Are we importing `webComponentDocs` and then exporting it separately because we want to re-export it as a named export?

I guess my question is if we could do this:
```js
export { default as webComponentDocs } from '@department-of-veterans-affairs/web-components/component-docs';
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export

packages/core/webpack.config.js
cvalarida: Is the plan to remove this alias later? Why not import the react bindings from `@department-of-veterans-affairs/web-components`, then?

packages/react-components/index.js
cvalarida: Oof. This should really be part of the build process. That is, we should have a `src/index.js` and use a proper bundler. But hey. Future enhancements. :man_shrugging: I know that's kind of out of scope for this PR.

packages/react-components/package.json
cvalarida: Bonus points for also adding a `module` property.

It's an unofficial property, but [Rollup documents it](https://rollupjs.org/guide/en/#publishing-es-modules).

packages/storybook/.storybook/preview.js
cvalarida: Also a bit of code formatting changes. Might look into how Prettier is set up in this new monorepo.

packages/core/src/main.js
bkjohnson: I was getting some syntax errors with how I was trying to import and export it at once, but I'll try that line and see what happens.

packages/react-components/index.js
bkjohnson: Yeah... our options were to rework the build process and standardize it, possibly combining the react components and web components into the same package at the root level, or keep things separate and go with a monorepo to combine them, and the second option seemed like a lighter touch. I definitely don't _love_ the current build process for `react-components`, but it seems easier to setup a monorepo and to deal with that build later.

packages/core/webpack.config.js
bkjohnson: I'd like to remove this by way of converting `Banner` to a web component, and that would eliminate this awkward dependency situation.

>  Why not import the react bindings from @department-of-veterans-affairs/web-components

I think that would require further modification to the `web-components` build, and I don't know what all would be required for that. Currently, we have this which gives us that folder to begin with:

https://github.com/department-of-veterans-affairs/component-library/blob/15d606e04ea012c1fb8242bd390103a67010b0e6/packages/web-components/stencil.config.ts#L28-L32

And we're including it in our `files` property in `package.json`, but these other properties mean that `react-bindings` isn't included at the root level:

https://github.com/department-of-veterans-affairs/component-library/blob/15d606e04ea012c1fb8242bd390103a67010b0e6/packages/web-components/package.json#L5-L8

We might be able to achieve a named import at the root level if we had another entry file that rexported the index in `dist` and `react-bindings` and then had that file be the `main`, but that's adding more complexity.

PR #208, opened 2021-10-28
Increase accordion item margin

bkjohnson:
## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/314

This is a follow up to #154 to merge the changes into `master`. The previous PR had begun when `web-components` was still being used.

I met with @GnatalieH and we decided to just increase the space between accordion items slightly. The original proposed 22px gap resulted in the items not looking like they were associated with each other. A compromise of reducing the target area while maintaining the current background color seemed like it would be non-intuitive.

If user research shows that Veterans are having problems with this design that get solved by increasing the margin, then we can reconsider.

## Testing done
N/A

### Before

image

### After

image

## Acceptance criteria
- [ ]

## Definition of done


GnatalieH (approved):
Looks great, thanks!

PR #209, opened 2021-10-29
Remove my-component

cvalarida:
## Description
Removes `my-component`. Because we don't actually want it.


bkjohnson (approved):
:tada: :tada: :tada:

PR #210, opened 2021-10-29
Update README.md with publishing instructions

bkjohnson:
## Description

Now that https://github.com/department-of-veterans-affairs/component-library/pull/204 is live, we need to document some basic instructions on how to get changes published as a package.

## Testing done

N/A

## Acceptance criteria
- [ ]

## Definition of done


k80bowman (approved):

pdavies88 (approved):
Do we want to have any additional notes on what warrants a Major, Minor, and Patch release? I can just see someone publishing and not knowing which number to change in the package.json and just picking one at random which might not be the best.

bkjohnson:
I've added some examples of which changes correspond to which version type. I also added badges to show the status of the most recent publish attempt and the latest version on npm. If a publish were to fail, the red "x" might be harder to find since I don't think it would be applied to the most recent commit.

PR #214, opened 2021-11-05
Add accessibility to va-radio and va-radio-option

andresriveratoro:
## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/562

## Testing done

Storybook

## Screenshots

N/A

## Acceptance criteria
- [ ] Radio options are accessible using `TAB`, `Arrow Up`, `Arrow Down`, `Arrow Left`, `Arrow Right`, and `Space` for navigating through the radio group and making selections

## Definition of done


noahgelman:
I've spoken to @andresriveratoro about some coding issues in the radio button component in storybook. I will not be approving this ticket till those other issues are fixed and I can re-review:

The current issues I see at the moment are:
The radio inputs need to have the same name to group them together in the same radio group
The id's for each input need to be unique. Current they're both just "input"
The label for each input needs to have its' "for" attribute value updated to the unique id of its' related radio input.
The radio inputs need their focus styling applied when focused

k80bowman:
@noahgelman that is really great feedback on this component, thank you. I would like to keep this PR scoped to the request in the [ticket](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/562), however. Would it be possible for you to create a new issue with your requests and provide detailed information about the HTML changes you are requesting?

noahgelman:
@k80bowman Sure. I'll create a ticket. As a note, these are not accessibility issues. It's basic semantic HTML the developer should be aware of.

k80bowman:
@noahgelman thank you! If you send me a link to the ticket when you create it I'll make sure we get it prioritized. These are definitely issues we would like to get fixed.

noahgelman:
@k80bowman @andresriveratoro https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/585

k80bowman:
I tried this out and I'm not sure it works exactly how we want it to. Here is a video I made of tabbing through the component in Firefox. I was only able to tab to the first radio button

https://user-images.githubusercontent.com/12739849/142659247-4816a0fd-e201-4af6-bccd-26319c5a522e.mov

And here is a video I made of using the arrow keys to access the component. I had to tab in, but then I was able to navigate between the radio buttons using the arrow keys. As I went back and forth, however, the buttons are selected only using the arrow keys.

https://user-images.githubusercontent.com/12739849/142659697-bdd174b5-c108-4bc3-88d7-28ffad8a3c94.mov

If I use the arrow keys to get to the second radio button, then switch to the tab key I can only tab to the second radio button.

I tried it in Safari and Chrome and had the same experience.

I also checked voiceover with Safari just to be sure that wasn't affected and it does seem to be working OK. I did notice that when I enter the web area focus goes directly to the first radio option and not the label. I'm not sure if that has to do with your work, but it's definitely something to take a look at.

k80bowman (changes requested):
See comment

andresriveratoro:
> I tried this out and I'm not sure it works exactly how we want it to. Here is a video I made of tabbing through the component in Firefox. I was only able to tab to the first radio button
> 
>  Screen.Recording.2021-11-19.at.11.29.04.AM.mov 
> And here is a video I made of using the arrow keys to access the component. I had to tab in, but then I was able to navigate between the radio buttons using the arrow keys. As I went back and forth, however, the buttons are selected only using the arrow keys.
> 
>  Screen.Recording.2021-11-19.at.11.31.05.AM.mov 
> If I use the arrow keys to get to the second radio button, then switch to the tab key I can only tab to the second radio button.
> 
> I tried it in Safari and Chrome and had the same experience.
> 
> I also checked voiceover with Safari just to be sure that wasn't affected and it does seem to be working OK. I did notice that when I enter the web area focus goes directly to the first radio option and not the label. I'm not sure if that has to do with your work, but it's definitely something to take a look at.

This behavior is how it should work according to https://www.w3.org/TR/wai-aria-practices-1.1/#radiobutton under Keyboard Interaction -> For Radio Groups Not Contained in a Toolbar and implemented using https://www.w3.org/TR/wai-aria-practices-1.1/examples/radio/radio-1/radio-1.html as a reference. Were you able to select a button with the space key once tabbed in?

k80bowman:
Oh interesting... thank you for the reference! Yes, I was able to select a button with the space key once I tabbed in.

k80bowman (approved):
Great job!

PR #221, opened 2021-11-12
Add additional info web component

bkjohnson:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/30136

This adds `<va-additional-info>`

`tagName` exists on `<AdditionalInfo>` but appears to be an unused prop, so I'm not porting it over:

image

I'm also not planning on exposing an external click event like we've done with other comopnents, because it appears that the `onClick` handlers for the current `<AdditionalInfo>` are there to fire an analytics event. More info in https://github.com/department-of-veterans-affairs/va.gov-team/issues/33086

## Testing done

- E2E tests
- Storybook :eyes: 

## Screenshots

Peek 2021-12-07 13-39

## Acceptance criteria
- [ ]

## Definition of done


packages/storybook/stories/AdditionalInfo.stories.jsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
-import {AdditionalInfo} from '@department-of-veterans-affairs/component-library';
+import { AdditionalInfo } from '@department-of-veterans-affairs/component-library';
 
 export default {
-  title: 'Components/AdditionalInfo',
+  title: 'Components/AdditionalInfo (deprecated)',
```
bkjohnson, line 5: Doing this will change the URL for the story that's [used on the component documentation page](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/7744fe91a1501f07622f7a04930b88c35ae305df/src/_components/additional-info.md?plain=1#L14), so we may want to not include this or just immediately update design.va.gov to point to the Web Component story.

packages/web-components/src/components/va-additional-info/va-additional-info.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+    this.open = !this.open;
+  }
+
+  handleKeydown(event): void {
+    if (event.key === ' ' || event.key === 'Enter') {
+      event.preventDefault();
+      this.toggleOpen();
+    }
+  }
```
bkjohnson, line 56: Expanding on `Enter` is an enhancement over current behavior:

https://github.com/department-of-veterans-affairs/component-library/blob/4a7c522fe2f21d8ba0e7b0bea795fe0d959a2a0c/packages/react-components/src/components/AdditionalInfo/AdditionalInfo.jsx#L15-L20

packages/web-components/src/components/va-additional-info/va-additional-info.css
```diff
@@ -0,0 +1,57 @@
+@import '../../mixins/links.css';
+:host {
+  display: block;
+}
+a[aria-expanded='true'] ~ #info {
+  padding-left: calc(2rem - 7px);
+  border-left: 7px solid transparent;
+  border-left-color: var(--color-primary-alt-light);
+}
```
bkjohnson, line 9: Applying the left border to the expanding text only is a bug fix for: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/577

k80bowman (commented):
This looks good to me, but I do have a question (I looked, but I couldn't find an answer).

When I use this component with voiceover in Safari, even when the accordion is not expanded, voiceover moves through the elements inside the accordion. It's been a while since I tested accordion components and I don't remember, is this how it is supposed to work?

https://user-images.githubusercontent.com/12739849/144869424-3061ff18-e284-48b6-bbc6-c452f508a87b.mov

k80bowman:
I checked a current version of this component on the [site](https://www.va.gov/education/about-gi-bill-benefits/how-to-use-benefits/vettec-high-tech-program/) and it looks like voiceover does not go down inside the component if it is collapsed. It only goes down when it is open. I made another video so you can see it in action.

https://user-images.githubusercontent.com/12739849/144905198-58f08a2e-9abf-41b8-a957-d052d3964a99.mov

packages/web-components/src/components/va-additional-info/va-additional-info.css
```diff
@@ -0,0 +1,61 @@
...
+
+#info.closed {
+  visibility: hidden;
+}
```
bkjohnson, line 13: This fixes the problem that @k80bowman noticed. The reason it was happening with this web component and not with the older React component is because the React implementation wasn't rendering the children when the component was "closed":

https://github.com/department-of-veterans-affairs/component-library/blob/4a7c522fe2f21d8ba0e7b0bea795fe0d959a2a0c/packages/react-components/src/components/ExpandingGroup/ExpandingGroup.jsx#L37-L40

But with the Web Component implementation, that child content is rendered by default into the `<slot>`. `visibility: hidden` [removes the node and its child nodes from the accessibility tree](https://developer.mozilla.org/en-US/docs/Web/CSS/visibility#accessibility_concerns):

> Using a `visibility` value of `hidden` on an element will remove it from the accessibility tree. This will cause the element and all its descendant elements to no longer be announced by screen reading technology.

k80bowman: This fix works great, thank you!

k80bowman (approved):

k80bowman (approved):

PR #223, opened 2021-11-17
Isolate host styling in "Back to top"

bkjohnson:
## Description
Related to https://github.com/department-of-veterans-affairs/va.gov-team/issues/31200

`<va-back-to-top>` is designed to be the last element in the main content area. Sometimes the main content area is styled with `usa-width-three-fourths` - a `uswds` class, and part of that styling is to [remove the margin from the last child](https://github.com/uswds/uswds/blob/c4c9410d54b537c6112037a36ed289d8b7e8940a/src/stylesheets/elements/_typography.scss#L227-L231). This PR adds "insulation" to `<va-back-top>` by adding an internal element that we can style as if it were the host. The `uswds` styles aren't able to have an effect this way with `:last-child` or `:first-child`.

This will unblock https://github.com/department-of-veterans-affairs/content-build/pull/759

## Testing done

Storybook :eyes: 

### No margin from `:host`, `uswds` styles are _technically_ applying

image

### The actual margin for the component is applied to the internal `#marginWrapper`

image

## Acceptance criteria
- [x] `<va-back-to-top>` doesn't have its margin removed by `uswds` styles

## Definition of done

### DIFF
--- a/packages/storybook/stories/va-back-to-top.stories.jsx
+++ b/packages/storybook/stories/va-back-to-top.stories.jsx
@@ -9,24 +9,28 @@ export default {
 
 const Template = () => {
   return (
-    <main className="vads-l-col--12 medium-screen:vads-l-col--8">
-      <h1>The top</h1>
-      {[...Array(10)].map(i => (
-        <div
-          key={i}
-          style={{
-            height: '300px',
-            background: 'whitesmoke',
-            margin: '2em 0',
-            display: 'flex',
-            alignItems: 'center',
-            justifyContent: 'center',
-          }}
-        >
-          Scroll down
+    <>
+      <div className="usa-grid usa-grid-full">
+        <div className="usa-width-three-fourths">
+          <h1>The top</h1>
+          {[...Array(10)].map(i => (
+            <div
+              key={i}
+              style={{
+                height: '300px',
+                background: 'whitesmoke',
+                margin: '2em 0',
+                display: 'flex',
+                alignItems: 'center',
+                justifyContent: 'center',
+              }}
+            >
+              Scroll down
+            </div>
+          ))}
+          <va-back-to-top />
         </div>
-      ))}
-      <va-back-to-top />
+      </div>
       <footer
         style={{
           height: '500px',
@@ -38,7 +42,7 @@ const Template = () => {
       >
         The footer
       </footer>
-    </main>
+    </>
   );
 };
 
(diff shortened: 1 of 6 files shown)


packages/storybook/stories/va-back-to-top.stories.jsx
```diff
@@ -9,24 +9,28 @@ export default {
...
-            margin: '2em 0',
-            display: 'flex',
-            alignItems: 'center',
-            justifyContent: 'center',
-          }}
-        >
-          Scroll down
+    <>
+      <div className="usa-grid usa-grid-full">
+        <div className="usa-width-three-fourths">
```
bkjohnson, line 14: This class/markup combination emulates the situations that cause `<va-back-to-top>`'s margin to be removed as the last or first child.

packages/storybook/stories/va-back-to-top.stories.jsx
```diff
@@ -9,24 +9,28 @@ export default {
...
+                margin: '2em 0',
+                display: 'flex',
+                alignItems: 'center',
+                justifyContent: 'center',
+              }}
+            >
+              Scroll down
+            </div>
+          ))}
+          <va-back-to-top />
```
bkjohnson, line 31: We moved the footer out of the main content container which fixes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/583.

This also makes `<va-back-to-top>` the last child, so that we can make sure the styling change applies.

packages/web-components/src/components/va-back-to-top/va-back-to-top.css
```diff
@@ -1,15 +1,16 @@
 @import '../../../../../node_modules/animate.css/source/sliding_entrances/slideInUp.css';
 @import '../../../../../node_modules/animate.css/source/sliding_exits/slideOutDown.css';
 @import '../../mixins/buttons.css';
-:host {
+
+#marginWrapper {
```
bkjohnson, line 5: Using this extra wrapping element gives us the isolation we need to ensure the margin doesn't get removed by the higher specificity CSS from the global styles.

[From some documentation on styling + Shadow DOM](https://developers.google.com/web/fundamentals/web-components/shadowdom):

> One gotcha with `:host` is that rules in the parent page have higher specificity than `:host` rules defined in the element. That is, outside styles win.

k80bowman:
I like this solution better, thank you. Looks like there's a test that needs to be fixed, but once that's done I'm ready to approve. I do think we should look into this issue more, though, and figure out if there is anything preemptive we can do to prevent it from happening again.

bkjohnson:
:thinking: It looks like the wrapper may be causing some flakiness - I'll look into it.

k80bowman (approved):

PR #224, opened 2021-11-19
fix: add missing enableAnalytics arg

andresriveratoro:
## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/587

## Testing done

Storybook

## Screenshots

Screen Shot 2021-11-22 at 10 36 43
Screen Shot 2021-11-22 at 10 36 53

## Acceptance criteria
- [ ] va-radio story is using a boolean prop named enableAnalytics and can be toggled between true and false.

## Definition of done


k80bowman (approved):
Looks good!

packages/storybook/stories/va-radio.stories.jsx
```diff
@@ -23,6 +23,7 @@ const Template = args => {
 };
 
 const defaultArgs = {
+  enableAnalytics: false,
```
bkjohnson, line 26: I think this may need to be `kebab-cased` to actually apply, like this:

https://github.com/department-of-veterans-affairs/component-library/blob/e63bb48b0063d07093d6329d9d289b352d3a31aa/packages/storybook/stories/va-loading-indicator.stories.jsx#L39

andresriveratoro: It looks like we have the option of using camel case in JSX, but HTML must be kebab case. I think it's best if the casing is consistent across our web components. Good catch!

andresriveratoro:
Screenshots have been updated to the latest changes.

packages/storybook/stories/va-radio.stories.jsx
```diff
@@ -13,19 +13,30 @@ export default {
...
       <va-radio-option label="Option two" name="two" value="2" />
     </va-radio>
   );
 };
 
 const defaultArgs = {
-  label: 'This is a label',
-  required: false,
-  error: null,
+  'enable-analytics': false,
```
bkjohnson, line 36: This gives us an extra arg:

image

packages/storybook/stories/va-radio.stories.jsx
```diff
@@ -13,19 +13,30 @@ export default {
   subcomponents: componentStructure(radioItem),
 };
 
-const Template = args => {
+const Template = ({
+  'enable-analytics': enableAnalytics,
```
bkjohnson, line 17: I'm not sure if renaming the prop here is going to do what we want.

packages/storybook/stories/va-radio.stories.jsx
andresriveratoro: Good catch. The latest changes use the components prop attribute instead of prop name.

packages/storybook/stories/va-radio.stories.jsx
andresriveratoro: The latest changes addresses this. The web component helper propStructure is now using the prop attr instead of prop name.

andresriveratoro:
If these changes make it to master, stories that are using camel casing will have to be updated or there will be duplicate props that differ in casing.

bkjohnson:
Since this change would mean that other files need to get updated as well, do we have a list of which ones? Also, I wonder if we could avoid that by leaving the kebabCased names in the `defaultArgs` and doing the switch inside of the `Template`.

andresriveratoro:
> Since this change would mean that other files need to get updated as well, do we have a list of which ones? Also, I wonder if we could avoid that by leaving the kebabCased names in the `defaultArgs` and doing the switch inside of the `Template`.

Before the change:
If we update defaultArgs to be in kebab case and make changes inside of `Template`, we will still receive the camel cased props in Storybook docs because of how we're generating ArgTypes: `Default.argTypes = propStructure(radioDocs);`. This will result in "duplicate" props with different casing.

After the change:
ArgTypes are now returned in kebab case. Any keys used in `defaultArgs` that are not returned from ArgTypes are treated as a separate prop resulting in "duplicate" props with different casing. This change also requires updating defaultArgs to be in kebab case and the prop aliasing in `Template`.

Web component stories affected by this change:
`<va-alert>`
`<va-checkbox>`
`<va-select>`

bkjohnson (approved):

PR #226, opened 2021-11-23
32229 accordion header api slot

pdavies88:
## Description
Closes [https://app.zenhub.com/workspaces/vsp-design-system-5f8de67192551b0012ebb802/issues/department-of-veterans-affairs/va.gov-team/32229](https://app.zenhub.com/workspaces/vsp-design-system-5f8de67192551b0012ebb802/issues/department-of-veterans-affairs/va.gov-team/32229)

## Testing done
Screen Shot 2021-11-23 at 11 19 37 AM

## Screenshots
Screen Shot 2021-11-23 at 11 20 19 AM

## Acceptance criteria
- As a user I should be able to pass a level and header prop to the `va-accordion-item` to set the wrapping heading tag and text of an accordion item or I should be able to use the `slot` methodology of other components and have that determine the text and heading tag of the accordion

## Definition of done


packages/web-components/src/components/va-accordion/va-accordion-item.tsx
```diff
@@ -35,28 +40,41 @@ export class VaAccordionItem {
...
 
+  @State() slotHeader: string = null;
+
+  @State() slotTag: string = null;
+
   private toggleOpen(e: MouseEvent): void {
     this.accordionItemToggled.emit(e);
   }
 
+  private populateStateValues() {
```
bkjohnson, line 51: Can we also have a comment saying why we're using a `slotchange` handler?

packages/web-components/src/components/va-accordion/va-accordion-item.tsx
```diff
@@ -35,28 +40,41 @@ export class VaAccordionItem {
...
+    getSlottedNodes(this.el, null).map(
```
bkjohnson, line 52: Since we don't care about the returned value, we could probably use `.forEach` instead of `.map`

packages/web-components/src/utils/utils.ts
```diff
@@ -26,9 +27,9 @@ export function getSlottedNodes(
     ? root.shadowRoot.querySelector('slot').assignedNodes()
     : root.shadowRoot.childNodes;
 
-  return Array.from(children).filter(
+  return nodeName !== null ? Array.from(children).filter(
```
bkjohnson, line 30: I don't think it needs to be done in this PR since `getSlottedNodes` doesn't currently have any unit tests, but now that we're using it more and making changes to it, we probably want a ticket for adding test coverage.

packages/web-components/src/components/va-accordion/va-accordion-item.e2e.ts
```diff
@@ -72,6 +105,22 @@ describe('va-accordion-item', () => {
     expect(button.getAttribute('aria-expanded')).toEqual('true');
   });
 
+  it('slot usage sets aria-expanded to true based on prop', async () => {
```
bkjohnson, line 108: We might not need this one since it doesn't look like it's related to the header changes.

bkjohnson (commented):
Looks pretty good, just some minor comments. Could you also add a test to `<va-accordion>` that uses the new `<h* slot="headline">The header</h*>` pattern? Specifically, one to make sure that the `header` text and `level` get into the analytics event:

https://github.com/department-of-veterans-affairs/component-library/blob/0dce5c07ad2fae4c2d9fe07230f8f969208d40d4/packages/web-components/src/components/va-accordion/va-accordion.e2e.ts#L99-L105

packages/web-components/src/components/va-accordion/va-accordion-item.tsx
pdavies88: Updated

packages/web-components/src/components/va-accordion/va-accordion-item.tsx
```diff
@@ -35,28 +40,41 @@ export class VaAccordionItem {
    */
   @Prop() level: number = 2;
 
+  @State() slotHeader: string = null;
+
+  @State() slotTag: string = null;
```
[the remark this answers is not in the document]
pdavies88: Updated with comments and tests on `<va-accordion>`

packages/web-components/src/components/va-accordion/va-accordion-item.e2e.ts
pdavies88: Removed with most recent commit

packages/web-components/src/components/va-accordion/va-accordion-item.tsx
pdavies88: Updated to forEach, also with the `populateStateValues` function we are setting that on `<slot name="headline" onSlotchange={() => this.populateStateValues()}/>` so I don't think we'd have to worry about the filter because the element which nodes it's looping through is just that `<h*>` tag and not the parent where we might have a conflict with multiple slots during the loop

packages/web-components/src/components/va-accordion/va-accordion.tsx
```diff
@@ -112,6 +120,18 @@ export class VaAccordion {
...
+   * Provides context of the header text to the Accordion Item
+   */
+  @State() slotHeader: string = null;
+
+  /**
+   * Local State for slot=headline replacement of props (level).
+   * Provides context of the level to the Accordion Item
+   */
+  @State() slotTag: number = null;
+
```
bkjohnson, line 134: I don't think these need to be `State` variables. Changing them would trigger a rerender, and these are referring to individual pieces of each accordion item.

packages/web-components/src/components/va-accordion/va-accordion.tsx
```diff
@@ -52,6 +53,13 @@ export class VaAccordion {
       event.detail.currentTarget.parentNode.parentNode.host || // if we are on IE, `.host` won't be there
       event.detail.currentTarget.parentNode.parentNode;
     // Close the other items if this accordion isn't multi-selectable
+    
+    // Usage for slot to provide context to analytics for header and level
+    getSlottedNodes(clickedItem, null).map(
+      (node: HTMLSlotElement) => {
+        this.slotHeader = node.innerHTML
+        this.slotTag = parseInt(node.tagName.toLowerCase().split('')[1])
+    })
```
bkjohnson, line 62: Instead, we could just put the mapped `getSlottedNodes` return values into some local variables. Something like:

```js
const [headerText, headerLevel] = getSlottedNodes(clickedItem, null).map(
    (node: HTMLSlotElement) => {
        const text = node.innerHTML
        const level = parseInt(/*blah blah */)
        return [text, level]
    })
```

(I haven't tested this)

And then use `headerText` and `headerLevel` down in the analytics details instead of the state variables.

bkjohnson (commented):
One more comment about the analytics/state stuff in the `<va-accordion>` wrapper, but everything else is :+1:

packages/web-components/src/components/va-accordion/va-accordion.tsx
```diff
@@ -52,6 +52,16 @@ export class VaAccordion {
       event.detail.currentTarget.parentNode.parentNode.host || // if we are on IE, `.host` won't be there
       event.detail.currentTarget.parentNode.parentNode;
     // Close the other items if this accordion isn't multi-selectable
+    
+    // Usage for slot to provide context to analytics for header and level
+    let slotHeader
+    let slotTag
```
bkjohnson, line 58: I think the `slotTag` should be renamed to include `level` somehow since that's what it represents.

bkjohnson (approved):
One possible name change, I'm approving so you can go ahead and merge when you're able.

Also, this seems like a feature enhancement to me, so having the `enhancement` label on it may be good so that it works well with our automated release notes.

packages/web-components/src/components/va-accordion/va-accordion.tsx
pdavies88: Updated to variable names to headerText and headerLevel as you had in the previous comment

PR #227, opened 2021-11-24
Add progress bar web component

andresriveratoro:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/32866

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-progress-bar/va-progress-bar.tsx
+++ b/packages/web-components/src/components/va-progress-bar/va-progress-bar.tsx
@@ -1,0 +1,64 @@
+import { Component, Event, EventEmitter, Host, Prop, h } from '@stencil/core';
+
+@Component({
+  tag: 'va-progress-bar',
+  styleUrl: 'va-progress-bar.css',
+  shadow: true,
+})
+export class VaProgressBar {
+  /**
+   * Whether or not an analytics event will be fired.
+   */
+  @Prop() enableAnalytics: boolean;
+
+  /**
+   * The text label for the progress bar.
+   */
+  @Prop() label: string;
+
+  /**
+   * Percent of progress made. 0 to 100.
+   */
+  @Prop() percent: number;
+
+  @Event({
+    eventName: 'component-library-analytics',
+    composed: true,
+    bubbles: true,
+  })
+  componentLibraryAnalytics: EventEmitter;
+
+  componentDidRender() {
+    if (this.enableAnalytics && (this.percent === 0 || this.percent === 100)) {
+      this.componentLibraryAnalytics.emit({
+        componentName: 'va-progress-bar',
+        action: 'change',
+        details: {
+          label: this.label,
+          percent: this.percent,
+        },
+      });
+    }
+  }
+
+  render() {
+    return (
+      <Host>
+        <div
+          aria-label={this.label || 'Working'}
+          aria-valuemax="100"
+          aria-valuemin="0"
+          aria-valuenow={this.percent}
+          class="progress-bar"
+          tabindex="0"
+          role="progressbar"
+        >
+          <div
+            class="progress-bar-inner"
+            style={{ width: `${this.percent}%` }}
+          />
+        </div>
+      </Host>
+    );
+  }
+}
--- a/packages/web-components/src/components/va-progress-bar/va-progress-bar.tsx
+++ b/packages/web-components/src/components/va-progress-bar/va-progress-bar.tsx
@@ -14,7 +14,7 @@ export class VaProgressBar {
   /**
    * The text label for the progress bar.
    */
-  @Prop() label: string;
+  @Prop() label: string = 'Working';
 
   /**
    * Percent of progress made. 0 to 100.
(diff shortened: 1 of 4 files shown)


pdavies88 (approved):
Looks great to me when testing it

k80bowman (approved):
This seems good to me as well. @bkjohnson any comments?

packages/web-components/src/components/va-progress-bar/test/va-progress-bar.e2e.ts
```diff
@@ -0,0 +1,75 @@
...
+    `);
+  });
+
+  it('passes an axe check', async () => {
+    const page = await newE2EPage();
+    await page.setContent('<va-progress-bar percent="35"></va-progress-bar>');
+    await axeCheck(page);
+  });
+
+  it('fires an analytics event when enable-analytics is set and percent is set to 0', async () => {
```
bkjohnson, line 38: It might be nice if these tests included the event details as a way of documenting what is included.

bkjohnson (approved):
Looks good! Only one minor comment about testing.

andresriveratoro:
Updated the aria label default value to use percent if a label isn't provided and updated CSS to prevent inner bar from growing over 100% if given a percent value over 100.

packages/web-components/src/components/va-progress-bar/va-progress-bar.tsx
```diff
@@ -0,0 +1,65 @@
...
+        details: {
+          label: this.label,
+          percent: this.percent,
+        },
+      });
+    }
+  }
+
+  render() {
+    const ariaLabel = this.label ?? `${this.percent}%`;
```
bkjohnson, line 45: Looks like this is causing a test to fail.

Also, instead of nullish coalescing, when we're going for default values can we set that on the property itself, or use destructuring with default assignment?

andresriveratoro: I've pushed up a fix for the tests. I think the default value that we want for label is the value of the percent prop, but we expose a label prop to allow a developer to override it.

andresriveratoro: How about this?

```
  componentWillLoad() {
    if (!this.label) this.label = `${this.percent}`;
  }
```

andresriveratoro: I'm against the previous componentWillLoad idea after implementing it locally because it's adding unnecessary complexity.

bkjohnson: What about something like this, but "complete" instead of "Working"? This diff is a bit awkward because I moved `percent` above `label` :upside_down_face: 

image

andresriveratoro: It looks like this approach doesn't work.
Screen Shot 2021-12-09 at 09 47 43

andresriveratoro: I've updated it to use destructuring with default assignment.

andresriveratoro:
One of the tests was updated to skip until we discover how to properly test lifecycle methods on component initialization.

packages/web-components/src/components/va-progress-bar/va-progress-bar.tsx
```diff
@@ -12,14 +20,19 @@ export class VaProgressBar {
...
-   * Percent of progress made. 0 to 100.
+   * The text label for the progress bar.
    */
-  @Prop() percent: number;
+  @Prop({ mutable: true }) label: string = `${this.percent}% complete`;
+
+  @Watch('percent')
+  updateLabel(newPercent: number) {
+    this.label = `${newPercent}% complete`;
+  }
```
bkjohnson, line 35: Having a mutable prop with  `@Watch` looks like it might override a custom `label` value set by an engineer.

packages/web-components/src/components/va-progress-bar/va-progress-bar.tsx
```diff
@@ -34,15 +47,15 @@ export class VaProgressBar {
         componentName: 'va-progress-bar',
         action: 'change',
         details: {
-          label: this.label || `${this.percent}% complete`,
+          label: this.label,
```
bkjohnson, line 50: Is the `@Watch` decorator what's making sure that this label is defined?

packages/web-components/src/components/va-progress-bar/va-progress-bar.tsx
andresriveratoro: You're correct. This isn't going to work.

packages/web-components/src/components/va-progress-bar/va-progress-bar.tsx
andresriveratoro: I added the watch decorator to make sure the analytics receives the correct details if/when the percent prop is updated, but like you mentioned above, this won't work if a developer sets their own label because it'll override it.

bkjohnson (approved):

PR #229, opened 2021-11-29
Add option to show icon in background-only alert for <va-alert>

andresriveratoro:
## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/586

## Testing done

Storybook

## Screenshots

Screen Shot 2021-12-01 at 16 56 59

Screen Shot 2021-12-01 at 16 56 26

## Acceptance criteria
- [x] Add option to show icon in background-only alert for `<va-alert>`
## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-alert/test/va-alert.e2e.ts
+++ b/packages/web-components/src/components/va-alert/test/va-alert.e2e.ts
@@ -11,7 +11,7 @@ describe('va-alert', () => {
     expect(element).toEqualHtml(`
       <va-alert class="hydrated">
         <mock:shadow-root>
-          <div class="alert hide-icon info">
+          <div class="alert info">
             <i aria-hidden="true" role="img"></i>
             <div class="body">
               <slot name="headline"></slot>
--- a/packages/web-components/src/components/va-alert/test/va-alert.e2e.ts
+++ b/packages/web-components/src/components/va-alert/test/va-alert.e2e.ts
@@ -11,7 +11,7 @@ describe('va-alert', () => {
     expect(element).toEqualHtml(`
       <va-alert class="hydrated">
         <mock:shadow-root>
-          <div class="alert info">
+          <div class="alert hide-icon info">
             <i aria-hidden="true" role="img"></i>
             <div class="body">
               <slot name="headline"></slot>
--- a/packages/web-components/src/components/va-alert/va-alert.tsx
+++ b/packages/web-components/src/components/va-alert/va-alert.tsx
@@ -7,6 +7,7 @@ import {
   Prop,
   h,
 } from '@stencil/core';
+import classnames from 'classnames';
 
 @Component({
   tag: 'va-alert',
@@ -134,9 +135,10 @@ export class VaAlert {
 
   render() {
     const { backgroundOnly, status, visible, closeable, showIcon } = this;
-    const classes = `alert ${status} ${backgroundOnly ? 'bg-only' : ''} ${
-      backgroundOnly ? (showIcon ? '' : 'hide-icon') : ''
-    }`;
+    const classes = classnames('alert', status, {
+      'bg-only': backgroundOnly,
+      'hide-icon': backgroundOnly && !showIcon,
+    });
     const role = status === 'error' ? 'alert' : null;
     const ariaLive = status === 'error' ? 'assertive' : null;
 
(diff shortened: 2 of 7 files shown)


pdavies88 (approved):
Looks good to me when testing

bkjohnson (commented):
This looks good to me as well, though I would try to get clarification from @GnatalieH about if the background-only alert should _always_ have an icon, or if there should be a way of opting in or out.

packages/web-components/src/components/va-alert/test/va-alert.e2e.ts
```diff
@@ -11,7 +11,7 @@ describe('va-alert', () => {
     expect(element).toEqualHtml(`
       <va-alert class="hydrated">
         <mock:shadow-root>
-          <div class="alert info">
+          <div class="alert hide-icon info">
```
bkjohnson, line 14: Can we make sure to only add this class if `background-only` is truthy?

packages/web-components/src/components/va-alert/va-alert.tsx
```diff
@@ -28,6 +28,12 @@ export class VaAlert {
    */
   @Prop() backgroundOnly: boolean = false;
 
+  /**
+   * This only takes effect when background-only is true. If true, the background-only alert will
+   * include an icon.
+   */
+  @Prop() showIconInBackgroundOnly: boolean = false;
```
bkjohnson, line 35: I would like for this prop to be shorter if possible, but I'm not sure about what would be better. It might help if we had explicit answers to this question:

Do we want `backgroundOnly` alerts to have icons by default?
  - This changes the default behavior, and 10 alerts would be impacted in `vets-website` (none in `content-build`
    image

Based on the change it seems like the design allows for a `backgroundOnly` alert to have an icon or not. Liz's comments on the original issue suggest that the preference among most teams _might_ be for the icon to already be there. If this is the case and we want to encourage that, it may make sense to switch the default behavior to include an icon, and give teams a way to opt out with a prop like `text-only` or something.

bkjohnson (commented):
I think we should get a few more answers before going forward with this so that:

- the component has the default behavior that we want to encourage
- the component interface is as simple & easy to use as possible for teams

packages/web-components/src/components/va-alert/test/va-alert.e2e.ts
andresriveratoro: Done. Good catch!

packages/web-components/src/components/va-alert/va-alert.tsx
andresriveratoro: If teams have a preference for using an icon in background-only alerts, then I agree with you that we should change the default behavior to display the icon. Should I go ahead and make these changes or should this question be directed to someone else?

GnatalieH: I actually think it would be good not to include the icon by default, as most of the time this alert is used without an icon. That could change in time, though.

packages/web-components/src/components/va-alert/va-alert.tsx
```diff
@@ -127,8 +133,16 @@ export class VaAlert {
...
+    const {
+      backgroundOnly,
+      status,
+      visible,
+      closeable,
+      showIconInBackgroundOnly,
+    } = this;
+    const classes = `alert ${status} ${backgroundOnly ? 'bg-only' : ''} ${
+      backgroundOnly ? (showIconInBackgroundOnly ? '' : 'hide-icon') : ''
+    }`;
```
bkjohnson: This is getting a bit harder to read with the addition of more ternaries and template strings. Could we use the `classnames` dependency for this, maybe like:

```js
const classes = classnames(
    'alert',
    status, 
    {
      'bg-only': backgroundOnly,
      'hide-icon': backgroundOnly && showIconInBackgroundOnly
    })
```

packages/web-components/src/components/va-alert/va-alert.tsx
andresriveratoro: I reached out to Kevin, and he said the default behavior should be to not display the icon unless a developer decides otherwise. I've updated the prop name to `showIcon`.

bkjohnson (approved):
:+1:

PR #230, opened 2021-11-30
32860 Segmented Progress Bar Web Component

pdavies88:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/32860

## Testing done

Screen Shot 2021-11-30 at 8 48 50 AM

## Screenshots

Screen Shot 2021-11-30 at 8 49 15 AM
Screen Shot 2021-11-30 at 8 49 42 AM

## Acceptance criteria
- [X] `<va-segmented-progress-bar>` is written
- [X] E2E tests are written
- [X] Storybook is updated

## Definition of done


packages/storybook/stories/va-segmented-progress-bar.stories.jsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { getWebComponentDocs, propStructure } from './wc-helpers';
+
+const segmentedProgressBarDocs = getWebComponentDocs('va-segmented-progress-bar');
+
+export default {
+  title: 'Components/va-segmented-progress-bar',
+};
+
+const Template = ({ 'enable-analytics': enableAnalytics, current, total, 'aria-label': ariaLabel }) => (
```
k80bowman, line 10: Not sure where you would want to put it, but could we add a top and bottom margin in here somewhere similar to what we have in the [stories.jsx](https://github.com/department-of-veterans-affairs/component-library/blob/7ac1818bfe1800703feb2238ce8e78c732e34542/packages/storybook/stories/SegmentedProgressBar.stories.jsx#L10) file for the React segmented progress bar?

When I pop the story for this component into its own window it's very hard to see:
image

pdavies88: Updated to give it breathing room in Storybook and added a comment about why we are doing this here

k80bowman: Looks great, thank you!

k80bowman (approved):

packages/web-components/src/components/va-segmented-progress-bar/va-segmented-progress-bar.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * The total number of segments in the progress bar
+     */
+    @Prop() total: number;
+
+    /**
+    * An override for the default aria label.
+    */
+    @Prop() ariaLabel: string = '';
```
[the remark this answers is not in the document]
pdavies88: Switched it over to label

packages/web-components/src/components/va-segmented-progress-bar/va-segmented-progress-bar.tsx
```diff
@@ -0,0 +1,73 @@
+import { Component, Event, EventEmitter, Host, Prop, h } from '@stencil/core';
+import { range } from 'lodash';
```
bkjohnson, line 2: I'm a bit surprised that this isn't causing any issues since `lodash` isn't defined as a dependency in the web-components manifest:

image


Maybe there's some magic going on to make it available through the monorepo since it's used in `react-components`. Anyway, I think we can [use some native JS](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_range) for this. And I know that the existing React `SegmentedProgressBar` is using lodash, but I don't think we need to carry that over unless absolutely necessary.

We already have polyfills for `Array.from` in IE11, so that shouldn't be an issue.

pdavies88: Updated

packages/web-components/src/components/va-accordion/va-accordion-item.e2e.ts
```diff
@@ -169,6 +169,6 @@ describe('va-accordion-item', () => {
     const accordionItemToggled = await page.spyOnEvent('accordionItemToggled');
     await page.waitForChanges();
 
-    expect(await accordionItemToggled).toHaveReceivedEventTimes(1);
+    await expect(accordionItemToggled).toHaveReceivedEventTimes(1);
```
bkjohnson, line 172: I don't think we need this `await` here.

packages/web-components/src/components/va-segmented-progress-bar/va-segmented-progress-bar.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * The total number of segments in the progress bar
+     */
+    @Prop() total: number;
+
+    /**
+    * An override for the default aria label.
+    */
+    @Prop() label: string = '';
```
bkjohnson, line 27: Does this need to default to an empty string? Can we instead set the `Step ${this.current} of ${this.total}` here?

packages/web-components/src/components/va-segmented-progress-bar/va-segmented-progress-bar.tsx
```diff
@@ -0,0 +1,73 @@
...
+                details: {
+                    current: this.current,
+                    total: this.total,
+                },
+            });
+        }
+    }
+
+    render() {
+        const label = this.label || `Step ${this.current} of ${this.total}`
```
bkjohnson, line 50: Alternatively, we could assign a default value here using something like:

```js
const {current, total, label = `Step ${current} of ${total}`} = this
```

Or at least, I think that's what the syntax would be.

bkjohnson (commented):
A couple of comments about the default value of a prop. If possible, I'd like to assign the default value in the initial prop definition, since the first place I'd look for any default values if I came in to work on this component 6 months in the future.

packages/web-components/src/components/va-segmented-progress-bar/va-segmented-progress-bar.tsx
pdavies88: Screen Shot 2021-12-09 at 10 54 26 AM
Sadly we cannot going to follow what Andres did in the Progress Bar PR #227

packages/web-components/src/components/va-segmented-progress-bar/va-segmented-progress-bar.tsx
pdavies88: I can destructure the current and total off of this but doing the default label does not work. In the current syntax you are suggesting the label becomes an empty string and does not default to Step X of Y
Screen Shot 2021-12-09 at 11 07 37 AM

pdavies88: @bkjohnson Ultimately I can make the code look like this:
// Label needs to use input or fallback and have instructions on correct one to use
// Needs to use initialized values as destructured values are causing data inconsistency
const label = this.label || Step ${this.current} of ${this.total}
const {current, total} = this

The other suggestions such as const {current, total, label = Step ${current} of ${total}} = this or 
const {current, total} = this and then const label = this.label || Step ${current} of ${total}
Result in the image I shared above with the label field being blank instead of showing Step X of Y

PR #233, opened 2021-12-01
Add telephone Web Component

bkjohnson:
## Description

Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/30137

The React `<Telephone>` implementation is a bit different than what I'm doing here, so I'll talk about what I've changed/left out.

### `pattern` prop

The React component allowed a `pattern` prop to be passed so that a developer could choose how to format the number:

1. default
2. N11-style
3. a number for international use (+1 at the beginning)
4. any other way you want

We can cover the first two cases by seeing if the length of the `contact` prop is 3 or 10, and a boolean prop can be used for the third case. I've only found one example of `pattern` being used in a custom way:

https://github.com/department-of-veterans-affairs/vets-website/blob/f3ee0b52ddbaf311f64740fa3ab1d6803845793c/src/applications/claims-status/components/appeals-v2/AppealHelpSidebar.jsx#L15

This is a vanity number and this particular use might [go against our guidance](https://design.va.gov/content-style-guide/dates-and-numbers#phone-numbers):

> We don’t use vanity phone numbers in body copy, as it adds visual noise and is not helpful to screen readers. We use and **hyperlink only the numeric phone number** in body copy.

The exception is:

> In marketing or promotional components, use the format: 877-222-VETS (8387) **and hyperlink the complete number including the parenthetical**.

image

Since it seems like a help sidebar might not be considered a marketing or promotional component, I haven't added any functionality that would allow for vanity numbers/formatting.

### Children

The React component also allowed children to be passed in. Out of 140 files that import & use this component, this feature is only used 3 times:

- [`src/applications/check-in/pre-check-in/components/Footer.jsx`](https://github.com/department-of-veterans-affairs/vets-website/blob/f3ee0b52ddbaf311f64740fa3ab1d6803845793c/src/applications/check-in/pre-check-in/components/Footer.jsx#L28)
- [`src/applications/health-care-questionnaire/shared/components/footer/GetHelpFooter.jsx`](https://github.com/department-of-veterans-affairs/vets-website/blob/f3ee0b52ddbaf311f64740fa3ab1d6803845793c/src/applications/health-care-questionnaire/shared/components/footer/GetHelpFooter.jsx#L107)
- [`src/applications/facility-locator/components/search-results-items/common/LocationPhoneLink.jsx`](https://github.com/department-of-veterans-affairs/vets-website/blob/f3ee0b52ddbaf311f64740fa3ab1d6803845793c/src/applications/facility-locator/components/search-results-items/common/LocationPhoneLink.jsx#L44)

Since it is rarely used, I'm opting to not port that functionality over for now in order to ensure that `<va-telephone>` isn't used to display arbitrary content. We may be able to address those instances and see if it truly is required at a later date.

### AriaDescribedby

Rather than having a pass-through prop for `aria-describedby` [like we have for the React component](https://github.com/department-of-veterans-affairs/component-library/blob/4a7c522fe2f21d8ba0e7b0bea795fe0d959a2a0c/packages/react-components/src/components/Telephone/Telephone.jsx#L219-L222), I'm hoping that developers can set the aria attribute on the host component and have it work as expected.

## Testing done

- E2E tests
- Unit tests for static helper functions
- Storybook :eyes: 

## Screenshots

image

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-telephone/test/va-telephone.spec.tsx
+++ b/packages/web-components/src/components/va-telephone/test/va-telephone.spec.tsx
@@ -2,6 +2,7 @@ import { VaTelephone } from '../va-telephone';
 
 describe('formatPhoneNumber', () => {
   const contact = '8885551234';
+  const N11 = '911';
   const extension = 123;
   it('formats a contact number with no extension', () => {
     expect(VaTelephone.formatPhoneNumber(contact, null)).toBe('888-555-1234');
@@ -11,6 +12,18 @@ describe('formatPhoneNumber', () => {
     expect(VaTelephone.formatPhoneNumber(contact, extension)).toBe(
       '888-555-1234, ext. 123',
     );
+  });
+
+  it('formats a 3 digit contact number', () => {
+    expect(VaTelephone.formatPhoneNumber(N11, null)).toBe('911');
+  });
+
+  it('does not use extension for 3 digit contact', () => {
+    expect(VaTelephone.formatPhoneNumber(N11, extension)).toBe('911');
+  });
+
+  it('does not use international formatting for 3 digit contact', () => {
+    expect(VaTelephone.formatPhoneNumber(N11, extension, true)).toBe('911');
   });
 });
 
--- a/packages/web-components/src/components/va-telephone/test/va-telephone.spec.tsx
+++ b/packages/web-components/src/components/va-telephone/test/va-telephone.spec.tsx
@@ -1,0 +1,36 @@
+import { VaTelephone } from '../va-telephone';
+
+describe('formatPhoneNumber', () => {
+  const contact = '8885551234';
+  const extension = 123;
+  it('formats a contact number with no extension', () => {
+    expect(VaTelephone.formatPhoneNumber(contact, null)).toBe('888-555-1234');
+  });
+
+  it('formats a contact number with extension', () => {
+    expect(VaTelephone.formatPhoneNumber(contact, extension)).toBe(
+      '888-555-1234, ext. 123',
+    );
+  });
+});
+
+describe('formatTelLabel', () => {
+  const contact = '8885551234';
+  const extension = 123;
+  const formattedNumber = VaTelephone.formatPhoneNumber(contact, null);
+  const formattedNumberWithExt = VaTelephone.formatPhoneNumber(
+    contact,
+    extension,
+  );
+  it('formats a phone number into an assistive tech label', () => {
+    expect(VaTelephone.formatTelLabel(formattedNumber)).toBe(
+      '8 8 8. 5 5 5. 1 2 3 4',
+    );
+  });
+
+  it('formats a phone number with extension into an assistive tech label', () => {
+    expect(VaTelephone.formatTelLabel(formattedNumberWithExt)).toBe(
+      '8 8 8. 5 5 5. 1 2 3 4. extension. 1 2 3',
+    );
+  });
+});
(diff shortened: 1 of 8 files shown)


packages/web-components/src/components/va-telephone/test/va-telephone.spec.tsx
```diff
@@ -0,0 +1,66 @@
+import { VaTelephone } from '../va-telephone';
+
+describe('formatPhoneNumber', () => {
```
bkjohnson, line 3: These unit tests are because I wanted to test the static helper functions. I left them as static functions rather than pulling them out into a separate file because:

- They aren't used anywhere else
- They're pretty specific to each other/the telephone component
    - i.e., expecting `ext` in the string for creating an aria label.

I also made them static just because I would rather pass in explicit arguments instead of relying on class properties.

packages/web-components/src/components/va-telephone/va-telephone.tsx
```diff
@@ -0,0 +1,125 @@
...
+@Component({
+  tag: 'va-telephone',
+  styleUrl: 'va-telephone.css',
+  shadow: true,
+})
+export class VaTelephone {
+  /**
+   * 3 or 10 digit string representing the contact number
+   */
+  @Prop() contact!: string;
```
bkjohnson, line 19: The `!` means that the prop is required. This doesn't really change anything, but it affects the output in `component-docs.json` which we use in Storybook.

packages/web-components/src/components/va-telephone/va-telephone.tsx
```diff
@@ -0,0 +1,124 @@
...
+   * Prepends a "+1" to the formatted number.
+   */
+  @Prop() international: boolean = false;
+
+  @Event({
+    eventName: 'component-library-analytics',
+    composed: true,
+    bubbles: true,
+  })
+  componentLibraryAnalytics: EventEmitter;
```
bkjohnson, line 42: This is completely new. The React component had an `onClick` prop that [would get set in the component](https://github.com/department-of-veterans-affairs/component-library/blob/4a7c522fe2f21d8ba0e7b0bea795fe0d959a2a0c/packages/react-components/src/components/Telephone/Telephone.jsx#L170). This is only used in one file, and it is exclusively for analytics event tracking.

image


https://github.com/department-of-veterans-affairs/vets-website/blob/74b6c6bdc4fe99f55a41d9dc3db7eef58b118e67/src/applications/claims-status/components/StemAskVAQuestions.jsx#L11-L16

bkjohnson (commented):
Let me know if I should add more test coverage. I was under the impression that an `extension` isn't supposed to be used with a 3 digit `contact` - if that is the case I can make it explicit in the tests.

packages/web-components/src/components/va-telephone/va-telephone.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * Optional phone number extension
+   */
+  @Prop() extension: number;
+
+  /**
+   * Indicates if the phone number can be clicked or not
+   */
+  @Prop() notClickable: boolean = false;
```
bkjohnson, line 29: There's some interesting discussion around the naming of this prop in [this slack thread](https://dsva.slack.com/archives/C5HP4GN3F/p1638466045185900).

pdavies88 (approved):
Tested locally worked correctly in storybook

k80bowman (approved):
Looks great! Thank you for all the comments explaining things, that's awesome. 🏆

PR #235, opened 2021-12-03
Update stories to use kebab case for props

andresriveratoro:
## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/588

## Testing done

Storybook

## Screenshots

va-alert:
Screen Shot 2021-12-03 at 17 44 34
va-text-input:
Screen Shot 2021-12-03 at 17 45 27
va-select:
Screen Shot 2021-12-03 at 17 45 16
Screen Shot 2021-12-03 at 17 44 57
Screen Shot 2021-12-03 at 17 44 47

## Acceptance criteria
- [ ] Story props for `<va-alert>`, `<va-checkbox>`, `<va-loading-indicator>`, `<va-select>` and `<va-text-input>` are updated to use kebab case

## Definition of done

### DIFF
--- a/packages/storybook/stories/va-checkbox.stories.jsx
+++ b/packages/storybook/stories/va-checkbox.stories.jsx
@@ -10,15 +10,31 @@ export default {
 };
 
 const defaultArgs = {
-  label: 'This is a cleverly-labelled checkbox',
-  checked: null,
-  error: null,
-  required: null,
-  description: null,
-  enableAnalytics: false,
+  'label': 'This is a cleverly-labelled checkbox',
+  'checked': false,
+  'error': null,
+  'required': false,
+  'description': null,
+  'enable-analytics': false,
 };
 
-const Template = props => <va-checkbox {...props} />;
+const Template = ({
+  checked,
+  description,
+  'enable-analytics': enableAnalytics,
+  error,
+  label,
+  required,
+}) => (
+  <va-checkbox
+    checked={checked}
+    description={description}
+    enable-analytics={enableAnalytics}
+    error={error}
+    label={label}
+    required={required}
+  />
+);
 
 export const Default = Template.bind({});
 Default.args = { ...defaultArgs };
--- a/packages/storybook/stories/va-text-input.stories.jsx
+++ b/packages/storybook/stories/va-text-input.stories.jsx
@@ -12,12 +12,7 @@ const defaultArgs = {
   'name': 'my-input',
   'label': 'My input',
   'autocomplete': false,
-  // Fun fact: This should really be false, but that passes
-  // enable-analytics="false" to the web component, which correctly interprets
-  // this as a truthy value. See also:
-  // https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attribute
   'enable-analytics': false,
-  // Same story here
   'required': false,
   'error': null,
   'maxlength': null,
--- a/packages/storybook/stories/va-text-input.stories.jsx
+++ b/packages/storybook/stories/va-text-input.stories.jsx
@@ -9,24 +9,46 @@ export default {
 };
 
 const defaultArgs = {
-  name: 'my-input',
-  label: 'My input',
-  autocomplete: false,
+  'name': 'my-input',
+  'label': 'My input',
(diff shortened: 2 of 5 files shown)


packages/storybook/stories/va-text-input.stories.jsx
```diff
@@ -9,24 +9,46 @@ export default {
...
-  autocomplete: false,
+  'name': 'my-input',
+  'label': 'My input',
+  'autocomplete': false,
   // Fun fact: This should really be false, but that passes
   // enable-analytics="false" to the web component, which correctly interprets
   // this as a truthy value. See also:
   // https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attribute
-  'enable-analytics': null,
+  'enable-analytics': false,
```
k80bowman, line 19: given the note above, should 'enable-analytics' and 'required' stay `null`?

packages/storybook/stories/va-checkbox.stories.jsx
```diff
@@ -10,15 +10,31 @@ export default {
...
-  label: 'This is a cleverly-labelled checkbox',
-  checked: null,
-  error: null,
-  required: null,
-  description: null,
-  enableAnalytics: false,
+  'label': 'This is a cleverly-labelled checkbox',
+  'checked': false,
+  'error': null,
+  'required': false,
```
k80bowman, line 16: Have you checked if it is appropriate to change the 'checked' and 'required' values to false? Or should they stay null?

packages/storybook/stories/va-text-input.stories.jsx
andresriveratoro: The comments are correct in that a falsy value should not be displayed on the DOM at all. The reason the prop was updated to false is because Storybook correctly displays how it should be used when clicking on 'Show code' for a component. 

Note: If you inspect the DOM, it shows up as `<va-component enable-analytics="false">`. This is wrong and it's an issue affecting Stencil.

https://github.com/ionic-team/stencil-ds-output-targets/issues/123
https://github.com/ionic-team/stencil-ds-output-targets/pull/149/files

All this change is doing is making sure `enable-analytics` does not show up as `enable-analytics={null}` when false. It is correctly omitted when given a value of false.

packages/storybook/stories/va-checkbox.stories.jsx
andresriveratoro: See the comment above

packages/storybook/stories/va-text-input.stories.jsx
pdavies88: Checked locally and confirmed value does not display on element when false inside of Storybook "Show Code"

pdavies88 (approved):
Checked locally and components are working correctly

packages/storybook/stories/va-text-input.stories.jsx
k80bowman: Ok, that all makes sense. But, if we are going to have them as 'false' here, we should probably update the note so that it better represents the current situation. Is this something that FE devs will need to be aware of?

andresriveratoro: I've removed the comments since they no longer apply to how it's being used here. https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attribute is still relevant and devs should be aware of it, but I'm not sure if we have to take action because it's general HTML knowledge. What do you think?

We can and should verify Storybook's "Show Code" examples accurately represent how a developer should use each component. This means that when creating or updating a Storybook story, verify the component is receiving the correct data and data types for props. In the case of a boolean prop with a value of false, verify it is omitted when looking at the code in "Show Code".

Using the alert web component with a value of false for the boolean prop named background-only:
Screen Shot 2021-12-06 at 17 02 35

Using the alert web component with a value of true for the boolean prop named background-only:
Screen Shot 2021-12-06 at 17 02 49

k80bowman: I am good with this as long as it is clear how developers should use this component. Thank you!

k80bowman (approved):

PR #239, opened 2021-12-08
feat: add va-breadcrumbs web component

andresriveratoro:
## Description
Closes <ticket>

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.css
+++ b/packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.css
@@ -1,0 +1,118 @@
+:host {
+  display: block;
+}
+
+ul {
+  padding: 0 0 0 1.5em;
+  list-style: square;
+}
+
+/* .row {
+  margin: 0 auto;
+  max-width: 62.5em;
+}
+
+.columns {
+  width: 100%;
+} */
+
+.va-nav-breadcrumbs {
+  background: var(--color-white);
+  color: var(--color-link-default);
+  font-size: inherit;
+  padding: 1em 0;
+}
+
+li {
+  display: inline-block;
+  margin: 0.25em 0;
+  padding: 0.25em 0 0.25em 0;
+  vertical-align: middle;
+}
+
+li:after {
+  content: ' › ';
+  display: inline-block;
+  padding: 0 0.35em;
+}
+
+li:last-child:after {
+  content: '';
+}
+
+li.active {
+  font-weight: bold;
+  padding: 0.3em 0;
+}
+
+li.active:after {
+  content: '';
+}
+
+a {
+  color: var(--color-link-default);
+  display: inline-block;
+  padding: 2px;
+}
+
+a:hover {
+  /* background: $color-link-default-hover; */
+}
+
+a[aria-current='page'] {
+  color: var(--color-gray-dark);
+  cursor: default;
+  font-weight: bold;
+  pointer-events: none;
+  text-decoration: none;
+}
+
+.va-nav-breadcrumbs-list {
+  float: none;
+  position: relative;
+}
+
+/* Mobile class declaration for React apps, manual override */
+
+.va-nav-breadcrumbs--mobile li {
+  display: none;
+}
+
+.va-nav-breadcrumbs--mobile li:nth-last-child(2) {
+  display: inline-block;
+}
+
+.va-nav-breadcrumbs--mobile li:nth-last-child(2)::before {
+  content: ' \2039 ';
+  display: inline-block;
+  padding: 0 0.15em;
+}
+
+.va-nav-breadcrumbs--mobile li:nth-last-child(2)::after {
+  content: '';
+  display: none;
+  padding: 0;
+}
+
+/* Activate the mobile breadcrumb at $small-screen breakpoint */
+@media screen and (max-width: 481px) {
+  .va-nav-breadcrumbs li {
+    display: none;
+  }
+
+  .va-nav-breadcrumbs li:nth-last-child(2) {
+    display: inline-block;
+  }
+
+  .va-nav-breadcrumbs li:nth-last-child(2)::before {
+    content: ' \2039 ';
+    display: inline-block;
+    padding: 0 0.15em;
+  }
+
+  .va-nav-breadcrumbs li:nth-last-child(2)::after {
+    content: '';
+    display: none;
+    padding: 0;
+  }
+}
--- a/packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
+++ b/packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
@@ -1,0 +1,113 @@
+import {
+  Component,
+  Element,
+  Event,
+  EventEmitter,
+  Host,
+  h,
+  Prop,
+} from '@stencil/core';
+import classnames from 'classnames';
+
+@Component({
+  tag: 'va-breadcrumbs',
+  styleUrl: 'va-breadcrumbs.css',
+  shadow: true,
+})
+export class VaBreadcrumbs {
+  @Element() el: HTMLElement;
+
+  /**
+   * Adds an aria-label attribute to the <nav /> element.
+   */
+  @Prop() label: string = 'Breadcrumb';
+
+  /**
+   * Analytics tracking function(s) will not be called
+   */
+  @Prop() disableAnalytics: boolean;
+
+  /**
+   * Adds a custom id attribute to the UL element
+   */
+  @Prop() listId: string;
+
+  /**
+   * Adds CSS class `.va-nav-breadcrumbs--mobile` to the
+   * NAV element. The mobile breadcrumb will always
+   * be displayed while mobileFirstProp is True.
+   */
+  @Prop() mobileFirstProp: boolean;
+
+  /**
+   * Adds a custom id attribute to the NAV element
+   */
+  @Prop() navId: string;
+
+  @Event({
+    eventName: 'component-library-analytics',
+    composed: true,
+    bubbles: true,
+  })
+  componentLibraryAnalytics: EventEmitter;
+
+  private fireAnalyticsEvent(event) {
+    if (!this.disableAnalytics) {
+      // If it's a link being clicked, dispatch an analytics event
+      if (event.target?.tagName === 'A') {
+        const details = {
+          componentName: 'va-breadcrumbs',
+          action: 'linkClick',
+          details: {
+            clickLabel: event.target.innerText.trim(),
+            clickLevel: parseInt(event.target.dataset.index, 10) + 1,
+            // totalLevels: this.el.shadowRoot
+            //   .querySelector('slot')
+            //   .assignedNodes(),
+            mobileFirstProp: this.mobileFirstProp,
+          },
+        };
+        this.componentLibraryAnalytics.emit(details);
+      }
+    }
+  }
+
+  componentDidLoad() {
+    const nodes = this.el.shadowRoot.querySelector('slot').assignedNodes();
+    if (!nodes?.length) return;
+
+    // TODO: Wrap each child with <li>
+    nodes.forEach((node: HTMLSlotElement, index: number) => {
+      node.setAttribute('data-index', `${index}`);
+      if (index === nodes.length - 1) {
+        node.setAttribute('aria-current', 'page');
+      }
+    });
+  }
+
+  render() {
+    const { label, navId, listId, mobileFirstProp } = this;
+
+    return (
+      <Host>
+        <nav
+          aria-label={label}
+          class={classnames('va-nav-breadcrumbs', {
+            'va-nav-breadcrumbs--mobile': mobileFirstProp,
+          })}
+          data-mobile-first={mobileFirstProp}
+          id={navId}
+        >
+          <ul
+            class="row va-nav-breadcrumbs-list columns"
+            role="list"
+            id={listId}
+            onClick={this.fireAnalyticsEvent}
+          >
+            <slot></slot>
+          </ul>
+        </nav>
+      </Host>
+    );
+  }
+}
--- a/packages/storybook/stories/va-breadcrumbs.stories.jsx
+++ b/packages/storybook/stories/va-breadcrumbs.stories.jsx
@@ -1,0 +1,48 @@
+import React from 'react';
+import { getWebComponentDocs, propStructure } from './wc-helpers';
+
+const breadcrumbsDocs = getWebComponentDocs('va-breadcrumbs');
+
+export default {
+  title: 'Components/va-breadcrumbs',
+};
+
+const Template = ({
+  label,
+  'disable-analytics': disableAnalytics,
+  'list-id': listId,
+  'mobile-first-prop': mobileFirstProp,
+  'nav-id': navId,
+}) => (
+  <va-breadcrumbs
+    label={label}
+    disable-analytics={disableAnalytics}
+    list-id={listId}
+    mobile-first-prop={mobileFirstProp}
+    nav-id={navId}
+  >
+    <a href="#">Home</a>
+    <a href="#">Level One</a>
+    <a href="#">Level Two</a>
+  </va-breadcrumbs>
+);
+
+const defaultArgs = {
+  'label': 'Breadcrumb',
+  'disable-analytics': false,
+  'list-id': 'breadcrumbs-list-id',
+  'mobile-first-prop': false,
+  'nav-id': 'breadcrumbs-nav-id',
+};
+
+export const Default = Template.bind({});
+Default.args = {
+  ...defaultArgs,
+};
+Default.argTypes = propStructure(breadcrumbsDocs);
+
+export const MobileFirst = Template.bind({});
+MobileFirst.args = {
+  ...defaultArgs,
+  'mobile-first-prop': true,
+};
--- a/packages/web-components/src/components.d.ts
+++ b/packages/web-components/src/components.d.ts
@@ -78,6 +78,28 @@ export namespace Components {
     }
     interface VaBackToTop {
     }
+    interface VaBreadcrumbs {
+        /**
+          * Analytics tracking function(s) will not be called
+         */
+        "disableAnalytics": boolean;
+        /**
+          * Adds an aria-label attribute to the <nav /> element.
+         */
+        "label": string;
+        /**
+          * Adds a custom id attribute to the UL element
+         */
+        "listId": string;
+        /**
+          * Adds CSS class `.va-nav-breadcrumbs--mobile` to the NAV element. The mobile breadcrumb will always be displayed while mobileFirstProp is True.
+         */
+        "mobileFirstProp": boolean;
+        /**
+          * Adds a custom id attribute to the NAV element
+         */
+        "navId": string;
+    }
     interface VaCheckbox {
         /**
           * The aria-describedby attribute for the <intput> in the shadow DOM.
@@ -269,6 +291,12 @@ declare global {
         prototype: HTMLVaBackToTopElement;
         new (): HTMLVaBackToTopElement;
     };
+    interface HTMLVaBreadcrumbsElement extends Components.VaBreadcrumbs, HTMLStencilElement {
+    }
+    var HTMLVaBreadcrumbsElement: {
+        prototype: HTMLVaBreadcrumbsElement;
+        new (): HTMLVaBreadcrumbsElement;
+    };
     interface HTMLVaCheckboxElement extends Components.VaCheckbox, HTMLStencilElement {
     }
     var HTMLVaCheckboxElement: {
@@ -322,6 +350,7 @@ declare global {
         "va-accordion-item": HTMLVaAccordionItemElement;
         "va-alert": HTMLVaAlertElement;
         "va-back-to-top": HTMLVaBackToTopElement;
+        "va-breadcrumbs": HTMLVaBreadcrumbsElement;
         "va-checkbox": HTMLVaCheckboxElement;
         "va-featured-content": HTMLVaFeaturedContentElement;
         "va-loading-indicator": HTMLVaLoadingIndicatorElement;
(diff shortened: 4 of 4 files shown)


packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  @Event({
+    eventName: 'component-library-analytics',
+    composed: true,
+    bubbles: true,
+  })
+  componentLibraryAnalytics: EventEmitter;
+
+  private fireAnalyticsEvent(event: MouseEvent): void {
+    if (!this.disableAnalytics) {
```
pdavies88, line 58: So I know we do enable analytics on all the other components. Seems like when this was set up as a react component they reversed it for some reason. Was thinking we should change this logic to be enable analytics instead to keep the pattern the same as the other web components

bkjohnson: Some components have `disableEnalytics`, others have `enableAnalytics`. In general, the ones with `enableAnalytics` have the analytics event off by default, because those components are things like form controls, so there were concerns about PII being sent. Components like this have the event fire by default because those concerns don't apply, but `disableAnalytics` is provided in case teams want to change that default behavior.

Here you can see which of the React component have which default:

image


There are certain ones like `LoadingIndicator` which have the default behavior of not firing an event just because they add noise.

packages/web-components/src/components/va-breadcrumbs/test/va-breadcrumbs.e2e.ts
```diff
@@ -0,0 +1,79 @@
+import { newE2EPage } from '@stencil/core/testing';
+import { axeCheck } from '../../../testing/test-helpers';
+
+describe('va-breadcrumbs', () => {
```
bkjohnson, line 4: Can we some tests to cover that `<li>` and `<a>` tags are rendered appropriately?

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  /**
+   * Analytics tracking function(s) will not be called
+   */
+  @Prop() disableAnalytics: boolean;
+
+  /**
+   * Adds a custom id attribute to the UL element
+   */
+  @Prop() listId: string;
```
bkjohnson, line 34: I'm not sure if we need this. I know it's on the current `<Breadcrumbs>` react component, but it doesn't appear to be used.

image

Also, since the `<ul>` for this component is inside the shadow DOM, I don't think that certain things like `aria-describedby` would be able to successfully point to a custom ID inside of the component.

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
```diff
@@ -0,0 +1,126 @@
...
+   * Adds CSS class `.va-nav-breadcrumbs--mobile` to the
+   * NAV element. The mobile breadcrumb will always
+   * be displayed while mobileFirstProp is True.
+   */
+  @Prop() mobileFirstProp: boolean = false;
+
+  /**
+   * Adds a custom id attribute to the NAV element
+   */
+  @Prop() navId: string;
```
bkjohnson, line 46: We also might not need this.

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
```diff
@@ -0,0 +1,126 @@
...
+      const target = event.target as HTMLElement;
+      // If it's a link being clicked, dispatch an analytics event
+      if (target?.tagName === 'A') {
+        const details = {
+          componentName: 'va-breadcrumbs',
+          action: 'linkClick',
+          details: {
+            clickLabel: target.innerText.trim(),
+            clickLevel: parseInt(target.dataset.index, 10) + 1,
+            totalLevels: getSlottedNodes(this.el, 'a').length,
```
bkjohnson, line 68: Since we call this down in the `render` and assign the mapped result to the state variable, we may be able to use that here and not call the function again:

```js
            totalLevels: this.breadcrumbs.length,
```

I haven't tested this though.

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
```diff
@@ -0,0 +1,126 @@
...
+        );
+      }
+
+      return (
+        <li>
+          <a data-index={`${index}`} href={node.attributes['href'].value}>
+            {node.innerText}
+          </a>
+        </li>
+      );
```
bkjohnson, line 103: If the only difference here is rendering `aria-current="page"`, we may be able to combine these two and use a ternary, like:

```jsx
    this.breadcrumbs = nodes.map((node: HTMLAnchorElement, index: number) => (
        <li>
          <a
            aria-current={index === nodes.length - 1 ? "page" : undefined}
            data-index={index}
            href={node.attributes['href'].value}
          >
            {node.innerText}
          </a>
        </li>
      )}
```

I would expect that having an `undefined` attribute value would result in React not rendering it to the DOM. We may need a bit of trial and error to be sure.

bkjohnson (commented):
A few initial thoughts.

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
```diff
@@ -0,0 +1,126 @@
...
+            totalLevels: getSlottedNodes(this.el, 'a').length,
+            mobileFirstProp: this.mobileFirstProp,
+          },
+        };
+        this.componentLibraryAnalytics.emit(details);
+      }
+    }
+  }
+
+  componentDidLoad() {
```
pdavies88, line 77: When I ran `yarn test` locally I was receiving the following console warning
Screen Shot 2021-12-10 at 2 26 22 PM
I was wondering if `componentDidLoad` is the best place to do the `getSlottedNodes` function.
I know in `va-accordion-item` https://github.com/department-of-veterans-affairs/component-library/blob/8e89df6d262e5ee7cfa2c31355b49cb4f7cb18e0/packages/web-components/src/components/va-accordion/va-accordion-item.tsx#L105 we are using the `onSlotchange` to fire the event which builds the html in the correct format. Wondering if that is the best approach here as well @bkjohnson @andresriveratoro let me know if you think that is the best approach here and if you are also getting that console warn when running the test suite locally

andresriveratoro: I saw that console.warn too and tried using different lifecycle methods to see if I could get rid of it, but it didn't work as intended (nothing displays for children). `componentDidLoad` seems like the correct lifecycle for this use case. I need to be able to get the slot content passed in, but that's not available until after the component has rendered once.

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.css
```diff
@@ -0,0 +1,112 @@
...
+
+li:last-child::after {
+  content: '';
+}
+
+a {
+  background-color: transparent;
+  color: var(--color-link-default);
+  display: inline-block;
+  padding: 2px;
```
k80bowman, line 49: Could we change this to rem or does it need to be px?

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
```diff
@@ -0,0 +1,126 @@
...
+    const nodes = getSlottedNodes(this.el, 'a');
+    // This renders slot content outside of a slot.
+    // This is a workaround for the limitations of styling slotted nested elements.
+    // See https://stackoverflow.com/questions/61626493/slotted-css-selector-for-nested-children-in-shadowdom-slot/61631668
+    this.breadcrumbs = nodes.map((node: HTMLAnchorElement, index: number) => {
+      if (index === nodes.length - 1) {
```
k80bowman, line 83: Could you please add a test that covers this logic?

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  render() {
+    const { label, navId, listId, mobileFirstProp } = this;
+
+    return (
+      <Host>
+        <nav
+          aria-label={label}
+          class={mobileFirstProp ? 'va-nav-breadcrumbs--mobile' : undefined}
```
k80bowman, line 114: Could you please also add a test that covers this logic?

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
andresriveratoro: The 'render' test was updated to include the results of this logic - dev passes in anchor tags and it displays them wrapped in `<li>`. Is that okay or do you want another test for this?

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
andresriveratoro: Oh, I see. Thanks for that explanation. I'm going to look into it.

k80bowman:
I checked this using voiceover and while it correctly calls out each breadcrumb, it announces each of the little arrow icons with "you are currently on a text element". I've added a video here so you can see.


https://user-images.githubusercontent.com/12739849/145844874-638f4eb1-0411-4150-9534-75b9e21819d0.mov



The React breadcrumb component does this as well, but it would be worth reaching out to Courtney to ask if this is something it should be doing or if those icons should be ignored by voiceover.

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
k80bowman: I think we need a test for what happens if the dev doesn't pass in anchor tags as well.

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
andresriveratoro: I've updated it to render breadcrumbs `onSlotchange` instead of `componentDidLoad` for consistency, but I prefer lifecycle methods because they're predictable and seem more correct. I view it similarly to using React's `useRef` vs. something like running `document.getElementById()` inside a lifecycle method. What do you all think? @pdavies88 @k80bowman @bkjohnson

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.css
```diff
@@ -0,0 +1,112 @@
...
+  content: ' \2039 ';
+  display: inline-block;
+  padding: 0 0.24rem;
+}
+
+.va-nav-breadcrumbs--mobile li:nth-last-child(2)::after {
+  content: '';
+  display: none;
+  padding: 0;
+}
```
bkjohnson, line 90: In my previous comment, I linked to an old commit that had this [markdown](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/commit/5ed871aebcf9819ed6743257335b6dee82ee315a#diff-2daf03bb495ad970807e273b85e4b2ccd5bd84d0fa012895def22d21eb4de5d1R40):

> Passing prop `mobileFirstProp` to the `<Breadcrumbs />` component will override `state.mobileShow` and present the back by one mobile breadcrumb link on initial render or subsequent renders. This prop could be used in cases where developers want absolute control over which breadcrumb UI to show, instead of letting the user's viewport width toggle the correct view.

@k80bowman, is this something we want to bring over into the web component? It doesn't appear to be used:

image

Our [usage guidance](https://design.va.gov/components/breadcrumbs#usage) seems like it doesn't encourage this mobile styling in desktop, stating:

> On desktop, the breadcrumb path should always start with a link to the home page of the site and end with the current page as the last segment.

bkjohnson (commented):
`mobileFirstProp` exists on the React version of the component, but I'm wondering if it's just tech debt that we want to leave out with the web component version.

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
k80bowman: I took a look at some other web-components breadcrumbs components just to get an idea of how to resolve this issue. I found [this one](https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/breadcrumbs/breadcrumbs.tsx) from Ionic that looks like it is using Stencil's `componentWillLoad` lifecycle method instead of `componentDidLoad` and `onSlotchange`. Is that something you considered?

andresriveratoro: I tried other lifecycle methods including `componentWillLoad`, and it wouldn't display the breadcrumbs because it couldn't find them. The array length of slotted nodes returned 0.

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.css
k80bowman: If it is not being used and our usage guidance doesn't recommend this pattern, then I would say we should not bring it over. We can always add it back if people start requesting it.

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
k80bowman: Fun. I, too, prefer the lifecycle methods, but we also don't want extra re-renders. Are there any structural changes you can think of we could make to this component to make it possible to use a lifecycle method? If not, we may have to stick with `onSlotchange`.

andresriveratoro: Using the Ionic component you linked as a reference, I was able to get it working inside `componentWillLoad` by targeting the host element and using `querySelectorAll` for anchor elements instead of targeting the shadow root and the slot (`getSlottedNodes`). The component no longer triggers a re-render. 

`shadowRoot` is null until after the component has loaded. This is why getSlottedNodes was returning an empty array.

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.css
andresriveratoro: Removed mobileFirstProp

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  render() {
+    const { label, mobileFirstProp } = this;
+
+    return (
+      <Host>
+        <nav
+          aria-label={label}
+          class={mobileFirstProp ? 'va-nav-breadcrumbs--mobile' : undefined}
+          data-mobile-first={mobileFirstProp}
```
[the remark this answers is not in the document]
andresriveratoro: Removed mobileFirstProp and data-mobile-first

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
k80bowman: Does this feel like a better workaround to you?

andresriveratoro: I think this is a great solution for this component and one that we could apply in other components where similar logic is required and when we have a specific element or element type to target.

k80bowman: Fantastic! Thank you so much for digging in and figuring it out!

k80bowman (approved):
This looks great!

pdavies88 (approved):

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.css
bkjohnson, line 90: Is it safe to delete this now?

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
```diff
@@ -0,0 +1,96 @@
...
+    if (!this.disableAnalytics) {
+      const target = event.target as HTMLElement;
+      // If it's a link being clicked, dispatch an analytics event
+      if (target?.tagName === 'A') {
+        const details = {
+          componentName: 'va-breadcrumbs',
+          action: 'linkClick',
+          details: {
+            clickLabel: target.innerText.trim(),
+            clickLevel: parseInt(target.dataset.index, 10) + 1,
```
bkjohnson, line 49: I think we might be able to get away with not setting `data-index` and rely on some [array functionality](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex), maybe like:

```js
clickLabel: this.breadcrumbs.findIndex(node => node.innerText === target.innerText) + 1,
```

(I haven't tested this)

That does add some looping, but it prevents stuff being added to the DOM which could become tech debt in the future and puzzle developers (like the `mobileFirstProp` thing).

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
```diff
@@ -0,0 +1,96 @@
...
+    const anchorNodes = Array.from(this.el.querySelectorAll('a'));
+    this.breadcrumbs = anchorNodes.map(
+      (node: HTMLAnchorElement, index: number) => {
+        return (
+          <li>
+            <a
+              aria-current={
+                index === anchorNodes.length - 1 ? 'page' : undefined
+              }
+              data-index={`${index}`}
```
bkjohnson, line 71: This might be unnecessary based on my previous comment.

bkjohnson (commented):
A few minor comments

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
bkjohnson: This suggestion could also be problematic if two breadcrumb pieces had the same text which of course _shouldn't_ happen, but you never know. @k80bowman might have some thoughts on what to do.

k80bowman: I do like the idea of not setting data index. But I think we need to be clear on what kind of logic we are expecting here. It also might be worth breaking that logic out into its own function. Let me see if I understand this correctly...

For analytics, they need a clickLevel value that maps to the index of whatever breadcrumb is clicked on in the array of breadcrumbs (using 1-index, not 0-index). The way we have it now, we are attaching a data-index value to the breadcrumb link element that the analytics function can then retrieve and assign to clickLevel.

Brooks, your suggestion is to not add the data-index value to the breadcrumb link element and instead use `findIndex` to locate the index for the node that is clicked on by comparing the inner text from the target node to the array of breadcrumb text values.

If the concern is that using `innerText` could be problematic because of repeat breadcrumb names, could we use a different value that is less likely to repeat? I wonder if href would be better.

k80bowman: And if we use href, I would definitely recommend breaking this logic out into its own function. Something like

```
const getClickLevel = (breadcrumbs, target) => {
  const targetHref = target.attributes['href'].value;
  const index = breadcrumbs.findIndex(node => node.href === targetHref;
  return index + 1;
}
```

This way we can make it very clear what we are doing and hopefully prevent any future confusion.

k80bowman: same caveat as Brooks on this code, though. I have not tested it.

bkjohnson: Comparing `href` as a way of finding the index does sound better. We may also be able to compare the `target` and `node` (or whatever we name the param for the iterator) objects directly, but I'm not sure if that would do what we want.

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.css
andresriveratoro: I missed this. Thanks!

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
k80bowman: Comparing `target` and `node` sounds like a good thing to try first. But then if that doesn't do what you want, then maybe try comparing `href`s.

andresriveratoro: I tried the following and I'm getting the correct value for click level, but one of my tests fails due to never receiving an event. If I change it to what is currently pushed up, the test passes.
```javascript
  private getClickLevel(target: HTMLAnchorElement) {
    const anchorNodes = Array.from(this.el.querySelectorAll('a'));
    const index = anchorNodes.findIndex((node: HTMLAnchorElement) =>
      node.isEqualNode(target),
    );
    return index + 1;
  }
```

andresriveratoro: I managed to get it working, but I'm having some weird behavior when modifying the test to check for the last anchor element. There are 3 anchorElements, but the test fails because it never receives an event if I select anchorElements[2] and call click on it.

This works:
```javascript
    const anchorElements = await page.findAll('va-breadcrumbs >>> li > a');
    await anchorElements[1].click();
```

This doesn't work:
```javascript
    const anchorElements = await page.findAll('va-breadcrumbs >>> li > a');
    await anchorElements[2].click();
```

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  @Event({
+    eventName: 'component-library-analytics',
+    composed: true,
+    bubbles: true,
+  })
+  componentLibraryAnalytics: EventEmitter;
+
+  private getClickLevel(target: HTMLAnchorElement) {
+    const anchorNodes = Array.from(this.el.querySelectorAll('a'));
```
bkjohnson, line 40: Do we need to have `this.el.querySelectorAll('a')` here or could we rely on the already computed `this.breadcrumbs` state variable?

andresriveratoro: The first one is `this.el.querySelectorAll('a')` and the second array is `this.breadcrumbs`. I went with the querySelector because it made it easier to compare to the target.

Screen Shot 2022-01-05 at 11 55 41
Screen Shot 2022-01-05 at 11 55 52

bkjohnson (commented):
This looks good, we can probably schedule a staging review if we don't have one already.

PR #243, opened 2021-12-13
Update MaintenanceBanner to use native date functions

andresriveratoro:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/23523

## Testing done
Unit tests

## Screenshots
N/A

## Acceptance criteria
- [x] Update MaintenanceBanner to use native date functions

## Definition of done

### DIFF
--- a/packages/react-components/package.json
+++ b/packages/react-components/package.json
@@ -32,8 +32,6 @@
   "dependencies": {
     "classnames": "^2.2.6",
     "core-js": "^3.9.0",
-    "date-fns": "^2.17.0",
-    "date-fns-tz": "^1.1.1",
     "lodash": "^4.17.15",
     "moment": "^2.23.0",
     "number-to-words": "^1.2.4",
--- a/packages/react-components/src/helpers/format-date.js
+++ b/packages/react-components/src/helpers/format-date.js
@@ -19,7 +19,7 @@ export const isDateBefore = (date1, date2) => {
   return dateObj1.getTime() < dateObj2.getTime();
 };
 
-export const isDateToday = (date1, date2) => {
+export const isDateSameDay = (date1, date2) => {
   const dateObj1 = convertToDate(date1);
   const dateObj2 = convertToDate(date2);
   if (!dateObj1 || !dateObj2) return;
--- a/packages/react-components/src/helpers/format-date.js
+++ b/packages/react-components/src/helpers/format-date.js
@@ -1,53 +1,45 @@
-export const convertToDate = date => {
-  if (!date || !(date instanceof Date || typeof date === 'number')) return;
-  return new Date(date);
-};
-
 export const isDateAfter = (date1, date2) => {
-  const dateObj1 = convertToDate(date1);
-  const dateObj2 = convertToDate(date2);
-  if (!dateObj1 || !dateObj2) return;
+  if (!date1 || !(date1 instanceof Date) || !date2 || !(date2 instanceof Date))
+    return;
 
-  return dateObj1.getTime() > dateObj2.getTime();
+  return date1.getTime() > date2.getTime();
 };
 
 export const isDateBefore = (date1, date2) => {
-  const dateObj1 = convertToDate(date1);
-  const dateObj2 = convertToDate(date2);
-  if (!dateObj1 || !dateObj2) return;
+  if (!date1 || !(date1 instanceof Date) || !date2 || !(date2 instanceof Date))
+    return;
 
-  return dateObj1.getTime() < dateObj2.getTime();
+  return date1.getTime() < date2.getTime();
 };
 
 export const isDateSameDay = (date1, date2) => {
-  const dateObj1 = convertToDate(date1);
-  const dateObj2 = convertToDate(date2);
-  if (!dateObj1 || !dateObj2) return;
+  if (!date1 || !(date1 instanceof Date) || !date2 || !(date2 instanceof Date))
+    return;
 
-  dateObj1.setHours(0, 0, 0, 0);
-  dateObj2.setHours(0, 0, 0, 0);
-
-  return dateObj1.getTime() === dateObj2.getTime();
+  return (
+    date1.getFullYear() === date2.getFullYear() &&
+    date1.getMonth() === date2.getMonth() &&
+    date1.getDay() === date2.getDay()
+  );
 };
 
 export const formatDate = (date, pattern, timezone) => {
-  const dateObj = convertToDate(date);
-  if (!dateObj) return;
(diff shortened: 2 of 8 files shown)


packages/react-components/package.json
```diff
@@ -32,8 +32,6 @@
   "dependencies": {
     "classnames": "^2.2.6",
     "core-js": "^3.9.0",
-    "date-fns": "^2.17.0",
-    "date-fns-tz": "^1.1.1",
```
k80bowman, line 36: 🎉

packages/react-components/src/helpers/format-date.js
```diff
@@ -0,0 +1,55 @@
+export const convertToDate = date => {
+  if (!date || !(date instanceof Date || typeof date === 'number')) return;
+  return new Date(date);
+};
```
bkjohnson, line 4: Would we need this if we always used `Date` objects? Maybe a better question is - can/should we ensure that we are always using `Date` objects (without this function) as opposed to numbers?

packages/react-components/src/helpers/format-date.js
```diff
@@ -0,0 +1,55 @@
...
+
+export const isDateBefore = (date1, date2) => {
+  const dateObj1 = convertToDate(date1);
+  const dateObj2 = convertToDate(date2);
+  if (!dateObj1 || !dateObj2) return;
+
+  return dateObj1.getTime() < dateObj2.getTime();
+};
+
+export const isDateToday = (date1, date2) => {
```
bkjohnson, line 22: This function isn't checking a date against `Date.now()`, so I think it would make more sense if the name more closely described the actual comparison that's happening.

packages/react-components/src/helpers/format-date.js
andresriveratoro: I think we should continue to convert to a new Date object because it removes the risk of mutating the original Date object when using native methods such as `setHours`. `typeof date === 'number'` was added to allow some flexibility in case `Date.now()` was used, but we can replace it with `new Date()`.

bkjohnson: If we're only using `setHours` in the `isDateSameDay` function, could we avoid the risk of modifying the original with something like this?

```js
return (
  date1.getFullYear() === date2.getFullYear() &&
  date1.getMonth() === date2.getMonth() &&
  date1.getDay() === date2.getDay());
```

andresriveratoro: That works beautifully. Updated to remove creation of new Date objects and `isDateSameDay` with this suggestion.

packages/react-components/src/helpers/format-date.unit.spec.js
```diff
@@ -0,0 +1,63 @@
...
+  isDateAfter,
+  isDateBefore,
+  isDateSameDay,
+} from './format-date';
+
+describe('date functions', () => {
+  it('formats date to DayOfWeek, MonthName DayOfMonth, Year at Hour:Minutes [AM|PM]', () => {
+    const date = formatDate(new Date('December 17, 1995 03:24:00'));
+    expect(date).to.equal('Sunday, December 17, 1995 at 3:24 AM');
+  });
```
bkjohnson, line 13: Date logic is complicated. When I run this locally (on the West coast), it creates a date for 3:24 AM _local time_, then formats it to eastern time, so 3 hours ahead - 6:24 AM.

image

In order to avoid this, we may want to create a `new Date()` using a UTC string w/ timezone offset. Another option could be creating a date from a UTC timestamp.

packages/react-components/src/helpers/format-date.js
```diff
@@ -0,0 +1,47 @@
...
+    return;
+
+  return (
+    date1.getFullYear() === date2.getFullYear() &&
+    date1.getMonth() === date2.getMonth() &&
+    date1.getDay() === date2.getDay()
+  );
+};
+
+export const formatDate = (date, pattern, timezone) => {
```
bkjohnson, line 26: Could we do this and not have to use an `||` operator down below?

```suggestion
export const formatDate = (date, pattern, timeZone = 'America/New_York') => {
```

packages/react-components/src/helpers/format-date.js
```diff
@@ -0,0 +1,47 @@
...
+  if (pattern === 'timeShort') {
+    return date.toLocaleString('en-US', { timeStyle: 'short', timeZone });
+  }
+
+  // Example: Tuesday, June 9, 2020 at 10:00 AM
+  return date.toLocaleString('en-US', {
+    dateStyle: 'full',
+    timeStyle: 'short',
+    timeZone,
+  });
```
bkjohnson, line 46: Having a custom string control the conditional logic for which of the 3 return values we get seems like it might lead to mistakes in the future. What do you thinking about changing the exports so we have something like:

- `formatTime`
- `formatDate`
- `formatDateShort`

And we could have one be the "default", with the other two using it internally:

```js

/**
 * Accepts an optional `options` object matching `Intl.DateTimeFormat` standards.
 * Returns a string with date & time by default.
 * Ex: Tuesday, June 9, 2020 at 10:00 AM
 */
export const formatDate = (
  date,
  options = { dateStyle: 'full', timeStyle: 'short'},
  timeZone = 'America/New_York') => {
    date.toLocaleString('en-US', {...options, timeZone})
}

/**
 * Date only, no time.
 * Ex: Tuesday, June 9, 2020
 */
export const formatDateShort = (date, timezone) => formatDate(date, {dateStyle: 'full'}, timezone)

/**
 * Time only, no date.
 * Ex: 10:00 AM
 */
export const formatTime = (date, timezone) => formatDate(date, {timeStyle: 'short'}, timezone)
```

I haven't tested this, it's just pure speculation right now :sweat_smile:. I'm not totally sold on it either, I'm just seeing what comes to mind.

bkjohnson (commented):
Just me asking some questions and having no answers :upside_down_face:

bkjohnson (approved):
This looks OK to me. Also, there shouldn't be any issues with us expecting the props to be `Date` objects. [This](https://github.com/department-of-veterans-affairs/vets-website/blob/e86c3dd537c5f76e6c3668b785614c07c0e92bbe/src/platform/site-wide/banners/components/MaintenanceBanner/index.js#L7-L19) is the only time that `MaintenanceBanner` is used, and its [date props are `Date` objects](https://github.com/department-of-veterans-affairs/vets-website/blob/addtl-info-transformer/src/platform/site-wide/banners/config/maintenanceBanner.js#L5-L6) rather than numbers.

PR #244, opened 2021-12-15
33252 getSlottedNodes Unit Tests

pdavies88:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/33252

## Testing done and Screenshots of Tests
Tests are in the utils.spec.ts file
Screen Shot 2021-12-16 at 7 49 10 PM

## Acceptance criteria
- [X] We have unit tests for getSlottedNodes

## Definition of done

### DIFF
--- a/packages/web-components/src/utils/utils.spec.ts
+++ b/packages/web-components/src/utils/utils.spec.ts
@@ -78,18 +78,4 @@ describe('getSlottedNodes()', () => {
     const slottedNodes = getSlottedNodes(defaultElement, 'a');
     expect(slottedNodes).toEqual([{ nodeName: 'A' }]);
   });
-
-  it('childNodes empty array', async () => {
-    const mockObject : NodeListOf<Element> = document.querySelectorAll('.mockObject') 
-    window.customElements.define('custom-element', CustomElement);
-
-    const defaultElement = document.createElement('custom-element');
-
-    jest
-      .spyOn(defaultElement.shadowRoot, 'childNodes', 'get')
-      .mockReturnValue(mockObject);
-
-    const slottedNodes = getSlottedNodes(defaultElement, null);
-    expect(slottedNodes).toEqual([]);
-  });
 });
--- a/packages/web-components/src/utils/utils.spec.ts
+++ b/packages/web-components/src/utils/utils.spec.ts
@@ -1,4 +1,4 @@
-import { format } from './utils';
+import { format, getSlottedNodes } from './utils';
 
 describe('format', () => {
   it('returns empty string for no names defined', () => {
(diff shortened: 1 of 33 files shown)


k80bowman:
Could you add a test for what happens when hasShadowDOM is false?

pdavies88:
Updated suite to include check for light dom vs shadow dom compatibility

k80bowman (approved):
This looks great, thank you!

packages/web-components/src/utils/utils.spec.ts
```diff
@@ -14,6 +14,139 @@ describe('format', () => {
...
+        this.attachShadow({ mode: 'open' });
+        this.shadowRoot.innerHTML = `<h6>ABC</h6><div>DEF</div><p>GHI</p>`;
+      }
+    }
+
+    window.customElements.define('custom-element', CustomElement);
+
+    const defaultElement = document.createElement('custom-element');
+    const slottedNodes = getSlottedNodes(defaultElement, null);
+    expect(await slottedNodes[2].nodeName).toEqual('P');
```
bkjohnson, line 38: `getSlottedNodes` isn't returning a Promise, so I don't think we need any of these `await`s.

packages/web-components/src/utils/utils.ts
```diff
@@ -18,19 +18,21 @@ export function getSlottedNodes(
   // A `<style>` element and a `<slot>` element
   // Chrome, Firefox, Safari, Edge - literally every modern browser will use shadowDOM.
   // This is purely for IE compatibility
-  const hasShadowDOM =
-    Array.from(root.shadowRoot.childNodes).filter(
-      (node: any) => node.tagName === 'SLOT',
-    ).length > 0;
+  if (root.shadowRoot) {
```
bkjohnson, line 21: Why is this being added? It looks like this wouldn't return anything if `root.shadowRoot` is falsey, which I'm not sure is what we want.

pdavies88: Katy wanted a test for the usage of this function in the light dom vs shadow dom (the last test focus on this). Right now if we don't include that wrapper for the function when being used with the light dom it will throw an error on `root.shadowRoot.childNodes` because the shadow dom doesn't exist. So therefore I added this to protect against someone using it for the light dom and because the use case of this function should only be for the shadow dom

bkjohnson: Light dom and shadow dom aren't mutually exclusive - a single component can make use of both. If we've got a web component that's using the shadow dom and it has a `<slot>` inside, then it is expecting to receive something from the light dom - a piece of HTML that is external to the component (and also not part of the component's shadow dom).

The `hasShadowDom` check is there so that `getSlottedNodes` will still return an array even in browsers that don't support shadow DOM functionality, like IE11. This `if (root.shadowRoot)` change will break components that rely on `getSlottedNodes` when viewed in IE11. For example, when you click a `<va-accordion-item>` it won't expand to reveal its slotted content.

pdavies88: Sounds good let me remove and adjust the tests

andresriveratoro (approved):

packages/web-components/src/utils/utils.spec.ts
```diff
@@ -14,6 +14,112 @@ describe('format', () => {
...
+        this.shadowRoot.innerHTML = `<h6>ABC</h6><div>DEF</div><p>GHI</p>`;
+      }
+    }
+
+    window.customElements.define('custom-element', CustomElement);
+
+    const defaultElement = document.createElement('custom-element');
+    const slottedNodes = getSlottedNodes(defaultElement, 'p');
+    expect(slottedNodes[0].nodeName).toEqual('P');
+  });
```
bkjohnson, line 56: It looks like these two tests both result in `hasShadowDom` being false simply because `<slot>` isn't used in the custom element definition. This results in nothing being in the filtered array:

https://github.com/department-of-veterans-affairs/component-library/blob/309940ca635092368f1da82a8329048b8819ebf9/packages/web-components/src/utils/utils.ts#L21-L24

However, the point of using `getSlottedNodes` is when we know that a component definition has `<slot>`. If we want to have a test for `hasShadowDom` being false, I think we should still use a `<slot>` but have a mock that is closer to what we would get from IE11. Maybe we mock `childNodes` to return an empty array on the `shadowRoot`, for example.

bkjohnson (commented):
I think there's a bit more work to do.

bkjohnson (commented):
I talked with @k80bowman about testing `getSlottedNodes` when `isShadowDom` is false, and for now the decision is to not go out of our way to test that particular branch of conditional logic. Since [Jest supports shadowDOM](https://github.com/testing-library/jest-dom/issues/340#issuecomment-807348708), it's difficult to create a situation _without_ shadowDOM so that we can get a good test for IE. I think our options would be one of these:

- recreate whatever the shadowDOM polyfill is doing in IE instead of using the `CustomElement` as-is that we wrote for the tests
- modify `getSlottedNodes` to be more accommodating to a specific test, maybe by checking an environment variable or an optional argument
- create something that doesn't represent a real-world situation in order to "trick" the test into passing

Instead of going with one of those options, we can rely on browser testing anytime we make a component-library dependency update in `vets-website`. A good smoke test for IE is to find (or create) an `open-single` accordion and click on some of the items and make sure that the others get closed. This should work as a smoke test because the accordion wrapper uses `getSlottedNodes` to get the items and appropriately set the `open` prop when its click handler receives an event.

https://github.com/department-of-veterans-affairs/component-library/blob/309940ca635092368f1da82a8329048b8819ebf9/packages/web-components/src/components/va-accordion/va-accordion.tsx#L65-L69

If `getSlottedNodes` isn't working, then we shouldn't see the expected behavior from `open-single`.

bkjohnson (approved):

PR #246, opened 2021-12-17
WIP: 33479 va accordion item e2e test

pdavies88:
## Description
Closes department-of-veterans-affairs/vets-design-system-documentation#728

## Testing done / Screenshots
Screen Shot 2021-12-17 at 5 49 40 PM

## Acceptance criteria
- [X] Test is fixed and no longer fails intermittently

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-accordion/va-accordion-item.e2e.ts
+++ b/packages/web-components/src/components/va-accordion/va-accordion-item.e2e.ts
@@ -174,4 +174,22 @@ describe('va-accordion-item', () => {
 
     expect(accordionItemToggled).toHaveReceivedEventTimes(1);
   });
+
+  it('does not fire a click event when the location hash does not match the accordion id', async () => {
+    const page = await newE2EPage({
+      html: '<va-accordion-item id="randomId" header="The header">Content inside</va-accordion-item>',
+    });
+    const accordionItemToggled = await page.spyOnEvent('accordionItemToggled');
+    
+    await page.goto('http://localhost:3333/#testing', {waitUntil: 'domcontentloaded'});
+    const elId = (await page.find('va-accordion-item')).id
+    const idPageMatch = page.url().includes(elId)
+    const button = await page.find('va-accordion-item >>> button');
+    if(idPageMatch){
+      await button.click();
+    }
+    await page.waitForChanges();
+
+    expect(accordionItemToggled).toHaveReceivedEventTimes(0);
+  });
 });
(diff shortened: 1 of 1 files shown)


packages/web-components/src/components/va-accordion/va-accordion-item.e2e.ts
```diff
@@ -157,16 +157,19 @@ describe('va-accordion-item', () => {
...
-    );
+  it('fires a click event when the location hash matches the accordion id', async () => {
+    const page = await newE2EPage({
+      html: '<va-accordion-item id="testing" header="The header">Content inside</va-accordion-item>',
+    });
     const accordionItemToggled = await page.spyOnEvent('accordionItemToggled');
+    
+    await page.goto('http://localhost:3333/#testing', {waitUntil: 'domcontentloaded'});
+    const elId = (await page.find('va-accordion-item')).id
+    const idPageMatch = page.url().includes(elId)
```
bkjohnson, line 168: Is there a possibility in this test that the url wouldn't include the element id?

pdavies88: The url with the element id is the reason the test exists. It's to test that if the url has the element id in it then we should trigger a click event on the matching elements button when the page loads so that the accordion slot is opened for the matching id. Here's the code that's doing that: https://github.com/department-of-veterans-affairs/component-library/blob/0fa1727d55cf0f447004d68df02490cdec31b42b/packages/web-components/src/components/va-accordion/va-accordion-item.tsx#L76

k80bowman: I'd like to see an additional test for 'it does not fire a click event when the location hash does not match the accordion id'.

bkjohnson: It's been a while, but I think what I was getting at here is if we need the `idPageMatch` conditional at all, or if we need the button click. I think the point behind this test is to see that the accordion is opening _without_ a click, based on the contents of the url hash parameter

pdavies88: @k80bowman I added an additional test for 'it does not fire a click event when the location hash does not match the accordion id'

pdavies88:
Pending waiting for response from Stencil Team: https://github.com/ionic-team/stencil/issues/1740

micahchiang:
Commenting so I remember that this still needs to be done and has a corresponding [ticket in zenhub](https://app.zenhub.com/workspaces/platform-design-system-5f8de67192551b0012ebb802/issues/gh/department-of-veterans-affairs/vets-design-system-documentation/728)

PR #254, opened 2021-12-29
Fix core bundling for IE 11

andresriveratoro:
## Description
Closes part of https://github.com/department-of-veterans-affairs/va.gov-team/issues/34425

## Acceptance criteria
- [ ] Resolve IE11 issues with the core build

## Definition of done


packages/core/src/main.js
```diff
@@ -2,4 +2,7 @@ import '@department-of-veterans-affairs/web-components/dist/component-library/co
 
 export * from '@department-of-veterans-affairs/react-components';
 export * from '@department-of-veterans-affairs/web-components';
-export { applyPolyfills } from '@department-of-veterans-affairs/web-components/loader';
+export {
+  applyPolyfills,
+  defineCustomElements,
```
bkjohnson, line 7: Can you elaborate on this change? In https://github.com/department-of-veterans-affairs/component-library/pull/218 we switched from implicitly exporting everything (including `defineCustomElements`) to only explicitly exporting `applyPolyfills` in order to fix a webpack 5 issue in `vets-website`.

image

If the named export doesn't cause the same issue that the star export did then this is fine.

packages/core/webpack.config.js
```diff
@@ -13,7 +13,6 @@ module.exports = {
     rules: [
       {
         test: /\.js$/,
-        exclude: /node_modules/,
```
bkjohnson, line 16: Does this mean that we're now using babel to rebuild all of our npm dependencies?

bkjohnson (commented):
I was able to make some changes and publish prereleases and eventually get things to work like we want on IE11: https://github.com/department-of-veterans-affairs/vets-website/pull/19908

With this PR I think we'll need to reset the versions back like we're ready to publish a real version. Once we double check everything and feel comfortable we should be good to go.

bkjohnson (approved):
We confirmed that IE11 worked with the `vets-website` changes using the changes in the prerelease.

andresriveratoro:
Thank you for your work on this, Brooks.

PR #256, opened 2022-01-04
Change modal header to h1

bkjohnson:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/29215

This is an a11y fix to the `<Modal>` React component. It replaces the `<h3>` with an `<h1>` in order to improve the semantics for assistive tech.

## Testing done

- Unit testing

I also did an axe test on a modified story that had an `<h1>` already on the page in addition to the `<h1>` that would be in the modal.

<details>
<summary>H1 on page and in modal</summary>

image

</details>

As [this axe rule page notes](https://dequeuniversity.com/rules/axe/4.3/page-has-heading-one):

> Generally, it is a best practice to ensure that the beginning of a page's main content starts with a h1 element, **and also to ensure that the page contains only one h1 element.**

However, that page also says:

> This rule finds all elements which match the following selector and verifies that there is **at least one**: h1:not([role]), [role="heading"][aria-level="1"]

So, we don't have an automated test that will let us know if an `<h1>` in a modal conflicts with the `<h1>` already on the page. We will rely on manual testing from an a11y expert.

## Screenshots

<details>
<summary>Before</summary>

image

</details>

### After

image

## Acceptance criteria
- [x] Change modal h3 to an h1 & adjust header size
- [x] Add unit/a11y tests

## Definition of done

### DIFF
--- a/packages/storybook/stories/Modal.stories.jsx
+++ b/packages/storybook/stories/Modal.stories.jsx
@@ -1,5 +1,5 @@
 import React, { useState } from 'react';
-import {Modal} from '@department-of-veterans-affairs/component-library';
+import { Modal } from '@department-of-veterans-affairs/component-library';
 
 export default {
   title: 'Components/Modal',
@@ -12,6 +12,7 @@ const Template = args => {
   const openModal = () => setVisible(true);
   return (
     <div style={{ height: '500px' }}>
+      <h1>Testing h1 heading</h1>
       <button onClick={openModal}>Click here to open modal</button>
       <Modal {...args} visible={visible} onClose={onClose}>
         {args.children}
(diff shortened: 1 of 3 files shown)


packages/storybook/stories/Modal.stories.jsx
```diff
@@ -12,6 +12,7 @@ const Template = args => {
   const openModal = () => setVisible(true);
   return (
     <div style={{ height: '500px' }}>
+      <h1>Testing h1 heading</h1>
```
bkjohnson, line 15: This was added so @CourtneyChambers77 could make sure that two `<h1>`s on the same page didn't pose a problem when one of them was in the modal.

pdavies88 (approved):

PR #259, opened 2022-01-07
Update OMBInfo property description

andresriveratoro:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/34524

## Testing done
N/A

## Screenshots
Screen Shot 2022-01-12 at 18 34 57

## Acceptance criteria
- [x] Update description for OMBInfo's resBurden property

## Definition of done

### DIFF
--- a/packages/react-components/src/components/OMBInfo/OMBInfo.jsx
+++ b/packages/react-components/src/components/OMBInfo/OMBInfo.jsx
@@ -18,10 +18,10 @@ class OMBInfo extends React.Component {
     this.setState({ modalOpen: false });
   };
 
-  modalContents = minutes => (
+  modalContents = (minutes, hideResBurdenSection) => (
     <div>
       <h3 id={`${this.id}-title`}>Privacy Act Statement</h3>
-      {minutes && (
+      {minutes && !hideResBurdenSection && (
         <p>
           <strong>Respondent Burden:</strong> We need this information to
           determine your eligibility for education benefits (38 U.S.C. 3471).
@@ -70,7 +70,7 @@ class OMBInfo extends React.Component {
   );
 
   render() {
-    const { resBurden, ombNumber, expDate } = this.props;
+    const { resBurden, ombNumber, expDate, hideResBurdenSection } = this.props;
 
     return (
       <div className="omb-info">
@@ -97,7 +97,7 @@ class OMBInfo extends React.Component {
           contents={
             this.props.children
               ? this.props.children
-              : this.modalContents(resBurden)
+              : this.modalContents(resBurden, hideResBurdenSection)
           }
           id={this.id}
           visible={this.state.modalOpen}
@@ -123,6 +123,11 @@ OMBInfo.propTypes = {
    * Form expiration date.
    */
   expDate: PropTypes.string.isRequired,
+
+  /**
+   * Hides Respondent Burden section of the Privacy Act Statement
+   */
+  hideResBurdenSection: PropTypes.bool,
 
   /**
    * Child elements (content) of modal when displayed
--- a/packages/react-components/src/components/OMBInfo/OMBInfo.jsx
+++ b/packages/react-components/src/components/OMBInfo/OMBInfo.jsx
@@ -18,10 +18,10 @@ class OMBInfo extends React.Component {
     this.setState({ modalOpen: false });
   };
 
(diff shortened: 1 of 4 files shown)


packages/react-components/src/components/OMBInfo/OMBInfo.jsx
```diff
@@ -18,10 +18,10 @@ class OMBInfo extends React.Component {
     this.setState({ modalOpen: false });
   };
 
-  modalContents = minutes => (
+  modalContents = (minutes, hideResBurdenSection) => (
     <div>
       <h3 id={`${this.id}-title`}>Privacy Act Statement</h3>
-      {minutes && (
+      {minutes && !hideResBurdenSection && (
```
bkjohnson, line 24: Can this already be accomplished by just not passing in a `resBurden` prop?

packages/react-components/src/components/OMBInfo/OMBInfo.jsx
```diff
@@ -97,7 +97,7 @@ class OMBInfo extends React.Component {
           contents={
             this.props.children
               ? this.props.children
-              : this.modalContents(resBurden)
+              : this.modalContents(resBurden, hideResBurdenSection)
```
bkjohnson, line 100: `resBurden` is the minutes arg, so unless that prop is set and as long as we don't default it to something truthy, I think this might do what we want.

bkjohnson (commented):
It looks like we have a [story already showing how to do this](https://design.va.gov/storybook/?path=/docs/components-ombinfo--default#without-response-burden), and it isn't passing in `resBurden`:

image

Do we still need this PR?

packages/react-components/src/components/OMBInfo/OMBInfo.jsx
k80bowman: This is interesting, because unless I scrolled down in the Storybook story, I would not have assumed that this is how that prop works. And I didn't see any other documentation on this component. So while yes, that prop does what we need it to do, it's unclear. Is there a way that we can clarify this and make it more intuitive for our frontend teams?

bkjohnson: We could add a page for `OMBInfo` on design.va.gov/components that describes use cases, or we could add to the prop description which may be helpful:

image


Maybe:

> Leave empty to remove any mention of a response burden

k80bowman: I think we should probably do both.

packages/react-components/src/components/OMBInfo/OMBInfo.mdx
```diff
@@ -24,6 +24,7 @@ import OMBInfo from '@department-of-veterans-affairs/component-library/OMBInfo'
 |resBurden| String or Number | The number of minutes the form is expected to take. If left falsy, this will not show.| No|
 |ombNumber | String | The form's OMB (Office of Management and Budget) control number | No |
 | expDate | Date as a string | This should be a formated date, in the format of `MMMM DD YYYY`| Yes |
+| hideResBurdenSection | Boolean | Hides Respondent Burden section of the Privacy Act Statement | No |
```
bkjohnson, line 27: This can probably be removed.

packages/storybook/stories/OMBInfo.stories.jsx
```diff
@@ -1,17 +1,18 @@
...
 };
 
-const Template = (args) => <OMBInfo {...args} />;
+const Template = args => <OMBInfo {...args} />;
 
 const defaultArgs = {
   resBurden: 120,
   ombNumber: '12-3456',
   expDate: '12/31/2077',
+  hideResBurdenSection: false,
```
bkjohnson, line 15: This as well.

bkjohnson (approved):
Looks good to me. Do we want another ticket for adding OMBInfo component documentation to design.va.gov?

caw310:
@bkjohnson @k80bowman , yes, we should have another ticket to add guidance.

k80bowman:
Created [documentation ticket](https://app.zenhub.com/workspaces/vsp-design-system-5f8de67192551b0012ebb802/issues/department-of-veterans-affairs/va.gov-team/35397)

PR #260, opened 2022-01-07
34419 Remove Lodash and convert to Native JS

pdavies88:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/34419

## Testing done
Screen Shot 2022-01-07 at 2 20 42 PM

## Screenshots
Chrome:
Screen Shot 2022-01-07 at 2 12 06 PM
Screen Shot 2022-01-07 at 2 15 01 PM

Firefox:
Screen Shot 2022-01-07 at 2 15 30 PM
Screen Shot 2022-01-07 at 2 15 41 PM

Safari:
Screen Shot 2022-01-07 at 2 15 57 PM
Screen Shot 2022-01-07 at 2 16 58 PM

Edge
Screenshot (8)
Screenshot (9)

## Acceptance criteria
- [X] lodash is removed from react-components in component-library
- [X] All lodash functions are converted to native JS
- [] Tested and confirmed working in all supported browsers: https://depo-platform-documentation.scrollhelp.site/developer-docs/Browser-Support-Policy.1847951366.html
* Need merge in PR before testing on IE11

## Definition of done


k80bowman:
Could we call this file something else, or distribute these functions to other helper files if appropriate? I'm not keen on having a file named after a utility we no longer use. It's going to lose context over time.

packages/react-components/src/helpers/utilities.js
```diff
@@ -0,0 +1,48 @@
+export const uniqueId = (
+  counter =>
+  (str = '') =>
+    `${str}${++counter}`
+)(0);
+
+export const isUndefined = val => val === undefined;
```
bkjohnson, line 7: Since this is so simple and we're only using it once I think it can just be inlined.

packages/react-components/src/helpers/utilities.js
```diff
@@ -0,0 +1,48 @@
+export const uniqueId = (
+  counter =>
+  (str = '') =>
+    `${str}${++counter}`
+)(0);
```
bkjohnson, line 5: I'm a little wary about this, but if we're going to have it I think we should at least have unit tests for it.

bkjohnson (commented):
There's some technical debt going on that might not be obvious at first glance.

packages/react-components/src/components/ProgressButton/ProgressButton.jsx
```diff
@@ -9,7 +9,7 @@ import classNames from 'classnames';
 
 class ProgressButton extends React.Component {
   UNSAFE_componentWillMount() {
-    this.id = uniqueId();
+    this.id = uniqueId('continueButton-');
```
bkjohnson, line 12: Why is this changing in some places and not in others?

pdavies88: Most places we used the prefix and in two files we didnt which was arbitrary and was making the id read 1-continueButton. So I adjusted it so that all files would use the prefix option when using uniqueId to keep the methodology consistent

bkjohnson: This is the kind of thing that can cause unintended side-effects. I'd also like to avoid it to keep the scope of this PR limited to removing lodash, rather than refactoring things that result in DOM changes.

[Here's an example of how this change could cause more work to be done](https://github.com/department-of-veterans-affairs/vets-website/blob/94b21a0f8db202254d94347b6fcae5198debc566/src/applications/health-care-questionnaire/list/tests/e2e/04.questionnaire.happy.path.cypress.spec.js#L31). Any CSS or JS could be targeting an id with the pattern `#N-continueButton`, and we would break those styles/code/tests.

pdavies88:
Confirmed on IE11:
Screenshot (11)
Screenshot (10)

PR #262, opened 2022-01-10
Add prop for res burden feature name

andresriveratoro:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/34523

## Testing done
Screen Shot 2022-01-10 at 10 40 09

## Screenshots

Screen Shot 2022-01-10 at 10 36 25
Screen Shot 2022-01-10 at 10 34 26
Screen Shot 2022-01-10 at 10 34 17

## Acceptance criteria
- [x] Add ability to use custom respondent burden name in the Privacy Act Statement Respondent Burden section
- [x] Add Storybook story
- [x] Add tests

## Definition of done

### DIFF
--- a/packages/react-components/src/components/OMBInfo/OMBInfo.jsx
+++ b/packages/react-components/src/components/OMBInfo/OMBInfo.jsx
@@ -114,7 +114,7 @@ OMBInfo.propTypes = {
   resBurden: PropTypes.number,
 
   /**
-   * The name of the respondent burden feature. This is displayed inside the Privacy Act Statement
+   * The name of the respondent burden feature. This is displayed in the Privacy Act Statement
    * Respondent Burden section.
    */
   resBurdenName: PropTypes.string,
--- a/packages/react-components/src/components/OMBInfo/OMBInfo.jsx
+++ b/packages/react-components/src/components/OMBInfo/OMBInfo.jsx
@@ -51,7 +51,7 @@ class OMBInfo extends React.Component {
         58VA21/22/28, Compensation, Pension, Education, and Vocational
         Rehabilitation and Employment Records - VA, and published in the Federal
         Register. Your obligation to respond is required to obtain or retain
-        education benefits. Giving us your SSN account information is voluntary.
+        {benefitType}. Giving us your SSN account information is voluntary.
         Refusal to provide your SSN by itself will not result in the denial of
         benefits. The VA will not deny an individual benefits for refusing to
         provide his or her SSN unless the disclosure of the SSN is required by a
--- a/packages/react-components/src/components/OMBInfo/OMBInfo.jsx
+++ b/packages/react-components/src/components/OMBInfo/OMBInfo.jsx
@@ -18,24 +18,23 @@ class OMBInfo extends React.Component {
     this.setState({ modalOpen: false });
   };
 
-  modalContents = minutes => (
+  modalContents = (minutes, resBurdenName = 'education benefits') => (
     <div>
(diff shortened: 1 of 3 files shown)


packages/react-components/src/components/OMBInfo/OMBInfo.jsx
```diff
@@ -18,24 +18,23 @@ class OMBInfo extends React.Component {
     this.setState({ modalOpen: false });
   };
 
-  modalContents = minutes => (
+  modalContents = (minutes, resBurdenName = 'education benefits') => (
```
bkjohnson, line 21: I'm wondering if it would be better to have a more generic default, like "benefits". That would be a slight change to 27 `<OMBInfo>` components in `vets-website` once the component-library version got updated, but I don't think it would be a bad change.

packages/react-components/src/components/OMBInfo/OMBInfo.jsx
```diff
@@ -114,6 +113,12 @@ OMBInfo.propTypes = {
    */
   resBurden: PropTypes.number,
 
+  /**
+   * The name of the respondent burden feature. This is displayed in the Privacy Act Statement
+   * Respondent Burden section.
+   */
+  resBurdenName: PropTypes.string,
```
bkjohnson, line 120: What do you think of changing `resBurdenName` to `benefitType`?

bkjohnson (commented):
Since this is a class component, I think if we set the prop defaults like this we can have Storybook pick up on it as well:

https://github.com/department-of-veterans-affairs/component-library/blob/45fdcde32a89592c7adfba6beaa38a7b3a768691/packages/react-components/src/components/Modal/Modal.jsx#L267-L271

packages/react-components/src/components/OMBInfo/OMBInfo.jsx
andresriveratoro: I think that's a great suggestion. Who can I reach out to that can make this type of decision or can we? A more generic default would provide a lot of value as teams wouldn't have to update their OMBInfo components unless they want to override it.

bkjohnson: I think our team can make that decision as long as we make sure that it's captured in the release notes.

k80bowman: I agree, I think we can make this decision, but we need to reach out to the team responsible and make sure they are aware and work with them on updating their code. And in fact, it might be better if we just update their code and they review it. So once this gets merged, I'd say go ahead and do a release and then submit a PR in vets-website that bumps the component-library version and updates all uses of this component so that they still read "education benefits"

packages/react-components/src/components/OMBInfo/OMBInfo.jsx
```diff
@@ -130,4 +135,8 @@ OMBInfo.propTypes = {
   children: PropTypes.node,
 };
 
+OMBInfo.defaultProps = {
+  benefitType: 'benefits',
```
bkjohnson, line 139: :tada: 

image

bkjohnson (commented):
@k80bowman should we use `{benefitType}` in the Privacy Act Notice as well?

https://github.com/department-of-veterans-affairs/component-library/blob/45fdcde32a89592c7adfba6beaa38a7b3a768691/packages/react-components/src/components/OMBInfo/OMBInfo.jsx#L55

k80bowman:
Ohh good catch! Yes, we should use benefitType in the Privacy Act Notice.

bkjohnson (approved):

PR #264, opened 2022-01-14
Fix Pagination keyboard navigation

andresriveratoro:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/35291

https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/pull/724 contains the styling updates

## Testing done
Updated tests

## Screenshots

https://user-images.githubusercontent.com/36863582/149957536-1cbc26b5-69da-4051-85a2-3225bf8e0d37.mov

## Acceptance criteria
- [x] Pagination elements are keyboard accessible (focusable and TAB)

## Definition of done


packages/react-components/src/components/Pagination/Pagination.jsx
```diff
@@ -105,17 +109,20 @@ class Pagination extends Component {
     let lastPage;
     if (showLastPage && currentPage < totalPages - maxPageListLength + 1) {
       lastPage = (
-        <span>
-          <a aria-label="...">...</a>
-          <a
+        <>
+          <button aria-label="..." type="button">
```
andresriveratoro, line 113: This feels wrong as a button or anchor because it's just text.

packages/react-components/src/components/Pagination/Pagination.jsx
```diff
@@ -62,15 +62,17 @@ class Pagination extends Component {
...
           aria-label={`Next page ${this.props.ariaLabelSuffix}`}
           onClick={() => {
             this.onPageSelect(this.props.page + 1, 'nav-paginate-next');
           }}
           onKeyDown={e => this.handleKeyDown(e, this.props.page + 1)}
+          type="button"
         >
           Next
-        </a>
+          <span aria-hidden></span>
```
andresriveratoro, line 74: These are used to display the icons using `::after`

andresriveratoro:
I wanted to update the HTML structure to use a nav, ul and li for buttons, but I'm not sure if that work should be part of this PR. What do you think?

andresriveratoro:
I've left some comments for the web component at https://github.com/department-of-veterans-affairs/va.gov-team/issues/35098.

packages/react-components/src/components/Pagination/Pagination.jsx
bkjohnson: :thinking: Does it need to be a button or would a span work? When I click the `...` in storybook I don't see anything happen, so we may not need a clickable element.

bkjohnson (commented):
Would putting [`tabindex="0"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex) on existing anchor elements solve the problem? That might mean we don't have to make CSS changes.

> `tabindex="0"` means that the element should be focusable in sequential keyboard navigation, after any positive tabindex values and its order is defined by the document's source order.

andresriveratoro:
`tabindex="0"` is an option that doesn't require CSS changes, but it would require code changes to handle keyboard events for selecting the anchors.

packages/react-components/src/components/Pagination/Pagination.jsx
andresriveratoro: I brought it up because it seemed like it could result in a possible design change for the [web component ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/35098). We can update it to be a focusable span and apply the same styling (including when focused and hover) as the rest of the anchors (now buttons) to limit the scope of this PR. 

**Should it be focusable?** I can see an argument being made for screen readers. If it should be focusable, we should consider making the ellipsis visually different from the other buttons when focused/hovering. It feels wrong to me when something looks like a button but does nothing, especially next to real buttons that look exactly the same. We should also consider updating its aria-label value. Using VoiceOver, "..." isn't read by this screen reader and may not be providing the value we think it should.

bkjohnson:
I added `tabindex="0"` to the anchor elements locally and I can tab through and focus the elements:

Peek 2022-01-18 16-31

Would we need event handlers for keys other than `Tab` that we get automatically with a button?

andresriveratoro:
I stand corrected. I had tried this elsewhere and wasn't getting this behavior because it was missing the onKeyDown handler. We can get rid of all the CSS changes and HTML changes besides adding `tabindex="0"` to the anchors and adding an onKeyDown to the last button.

bkjohnson:
Since the original issue is about keyboard navigation rather than the semantics of anchors vs. buttons, I'd prefer to just solve for the keyboard navigation. If `tabindex="0"` gets us 90% of the way there and we have to add a new `keydown` handler for the last bit, that's what I'd recommend we go with.

andresriveratoro:
Sounds good. I'll update it to that. Thanks!

bkjohnson (commented):
Looks good! Could we also get a unit test for tabbing and selecting with the space key?

bkjohnson (approved):

PR #268, opened 2022-01-20
Prevent text overflowing container in va-accordion-item

andresriveratoro:
## Description
The previous CSS changes made to va-accordion-item in #252 and #267 have introduced defects and are reverted in this PR.  The fix for the original issue is included.

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/602

## Screenshots
Screen Shot 2022-01-19 at 20 38 14

Here you can see a child container with width 100% that contains children containers of different widths (400px, 100%). Each container has a string long enough to require a horizontal scrollbar to view unless wrapped. There is no horizontal bar.
Screen Shot 2022-01-19 at 20 47 44

## Acceptance criteria
- [x] va-accordion-item should wrap content inside (preventing a horizontal scrollbar)

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-accordion/va-accordion-item.css
+++ b/packages/web-components/src/components/va-accordion/va-accordion-item.css
@@ -61,8 +61,7 @@ button:hover {
   border-left: var(--item-border);
   border-right: var(--item-border);
   border-bottom: var(--item-border);
-  display: flex;
-  flex-wrap: wrap;
+  overflow-wrap: break-word;
 }
 
 button[aria-expanded='false'] {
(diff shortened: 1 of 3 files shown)


packages/web-components/src/components/va-accordion/va-accordion-item.css
```diff
@@ -61,8 +61,7 @@ button:hover {
   border-left: var(--item-border);
   border-right: var(--item-border);
   border-bottom: var(--item-border);
-  display: flex;
-  flex-wrap: wrap;
```
andresriveratoro, line 65: This was not a good solution for the problem because it created defects when children content didn't have a width of 100%. This caused elements like `<p>` to be displayed inline instead of at block level like previously. Using this previous solution would have required us to make unnecessary updates to account for new children elements.

pdavies88 (approved):

k80bowman (approved):

PR #270, opened 2022-01-22
Add web component events documentation on Storybook

andresriveratoro:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/34746

## Testing done
N/A

## Screenshots
Screen Shot 2022-01-24 at 22 28 27
Screen Shot 2022-01-24 at 22 29 05

## Acceptance criteria
- [x] Update Storybook with information about using React bindings

## Definition of done

### DIFF
--- a/packages/storybook/stories/va-alert.stories.jsx
+++ b/packages/storybook/stories/va-alert.stories.jsx
@@ -16,22 +16,13 @@ export default {
         disable: true,
       },
     },
-    onclose: {
-      description:
-        'If closeable is true, this event is triggered when an alert is closed.',
-      table: {
-        defaultValue: {
-          detail: undefined,
-        },
-      },
-      control: false,
-    },
   },
   parameters: {
     docs: {
       description: {
         component: `Use a heading element with an attribute named slot and a value of "headline" to control what is displayed for the alert's headline. 
-        Any children passed into this component without a parent slot "headline" will render in the alert's body.`,
+        Any children passed into this component without a parent slot "headline" will render in the alert's body. 
+        This component has 1 event: onClose and onVaComponentDidLoad. If you need to use these events in React applications, please use the bindings import for this web component.`,
       },
     },
   },
@@ -47,7 +38,6 @@ const defaultArgs = {
   'closeable': false,
   'full-width': false,
   'headline': <h3 slot="headline">Alert headline</h3>,
-  'onclose': undefined,
 };
 
 const Template = ({
--- a/packages/storybook/stories/va-alert.stories.jsx
+++ b/packages/storybook/stories/va-alert.stories.jsx
@@ -17,6 +17,16 @@ export default {
         disable: true,
       },
     },
+    close: {
+      description:
+        'If closeable is true, this event is triggered when an alert is closed.',
+      table: {
+        defaultValue: {
+          detail: undefined,
+        },
+      },
+      control: false,
+    },
   },
   parameters: {
     docs: {
--- a/packages/storybook/stories/va-alert.stories.jsx
+++ b/packages/storybook/stories/va-alert.stories.jsx
@@ -18,16 +18,6 @@ export default {
         disable: true,
       },
     },
-    closeEvent: {
-      description:
-        'If closeable is true, this event is triggered when an alert is closed.',
-      table: {
-        defaultValue: {
-          detail: undefined,
-        },
-      },
-      control: false,
-    },
   },
   parameters: {
     docs: {
--- a/packages/storybook/stories/va-alert.stories.jsx
+++ b/packages/storybook/stories/va-alert.stories.jsx
@@ -21,8 +21,7 @@ export default {
     docs: {
       description: {
         component: `Use a heading element with an attribute named slot and a value of "headline" to control what is displayed for the alert's headline. 
-        Any children passed into this component without a parent slot "headline" will render in the alert's body. 
-        This component has 1 event: onClose and onVaComponentDidLoad. If you need to use these events in React applications, please use the bindings import for this web component.`,
+        Any children passed into this component without a parent slot "headline" will render in the alert's body. This component has 1 event: onClose and onVaComponentDidLoad. If you need to use these events in React applications, please use the bindings import for this web component. See https://design.va.gov/documentation/developers#react-applications.`,
       },
     },
   },
--- a/packages/storybook/stories/va-alert.stories.jsx
+++ b/packages/storybook/stories/va-alert.stories.jsx
@@ -1,5 +1,6 @@
 /* eslint-disable react/prop-types */
 import React from 'react';
+import { EVENTS_DESCRIPTION } from './events';
 import { getWebComponentDocs, propStructure } from './wc-helpers';
 
 const alertDocs = getWebComponentDocs('va-alert');
@@ -20,8 +21,11 @@ export default {
   parameters: {
     docs: {
       description: {
-        component: `Use a heading element with an attribute named slot and a value of "headline" to control what is displayed for the alert's headline. 
-        Any children passed into this component without a parent slot "headline" will render in the alert's body. This component has 1 event: onClose and onVaComponentDidLoad. If you need to use these events in React applications, please use the bindings import for this web component. See https://design.va.gov/documentation/developers#react-applications.`,
+        component:
+          `Use a heading element with an attribute named slot and a value of "headline" to control what is displayed for the alert's headline. 
+        Any children passed into this component without a parent slot "headline" will render in the alert's body. 
+        This component has 2 events: close and vaComponentDidLoad.` +
+          EVENTS_DESCRIPTION,
       },
     },
   },
--- a/packages/storybook/stories/va-alert.stories.jsx
+++ b/packages/storybook/stories/va-alert.stories.jsx
@@ -1,6 +1,6 @@
 /* eslint-disable react/prop-types */
 import React from 'react';
-import { EVENTS_DESCRIPTION } from './events';
+import { generateEventsDescription } from './events';
 import { getWebComponentDocs, propStructure } from './wc-helpers';
 
 const alertDocs = getWebComponentDocs('va-alert');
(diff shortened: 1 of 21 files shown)


packages/storybook/stories/va-alert.stories.jsx
```diff
@@ -16,22 +16,12 @@ export default {
...
-      },
-      control: false,
-    },
   },
   parameters: {
     docs: {
       description: {
         component: `Use a heading element with an attribute named slot and a value of "headline" to control what is displayed for the alert's headline. 
-        Any children passed into this component without a parent slot "headline" will render in the alert's body.`,
+        Any children passed into this component without a parent slot "headline" will render in the alert's body. This component has 1 event: onClose and onVaComponentDidLoad. If you need to use these events in React applications, please use the bindings import for this web component. See https://design.va.gov/documentation/developers#react-applications.`,
```
pdavies88, line 24: @andresriveratoro think a typo - onClose and onVaComponentDidLoad 2 events

pdavies88:
Do we want a blurbs for `accordionItemToggled: EventEmitter` and  `radioOptionSelected: EventEmitter` or since they are child events we don't need to add them in?

pdavies88:
Also not sure if there is anything to do about this but I randomly adds a dash below the text in the new reducer to query and display the text for the events:
Screen Shot 2022-01-24 at 4 38 33 PM

packages/storybook/stories/va-alert.stories.jsx
andresriveratoro: Good catch, thanks!

andresriveratoro:
> Do we want a blurbs for `accordionItemToggled: EventEmitter` and `radioOptionSelected: EventEmitter` or since they are child events we don't need to add them in?

I think it's worth adding these in to offer additional flexibility to developers. I'll add them in. 👍

andresriveratoro:
> Also not sure if there is anything to do about this but I randomly adds a dash below the text in the new reducer to query and display the text for the events: Screen Shot 2022-01-24 at 4 38 33 PM

I'll see what I can do about removing these.

andresriveratoro:
https://github.com/department-of-veterans-affairs/vets-design-system-documentation/pull/630 should be merged before this.

pdavies88 (approved):

k80bowman (approved):

packages/storybook/stories/wc-helpers.jsx
```diff
@@ -54,6 +70,11 @@ export const propStructure = comp => {
     };
     return propObj;
   }, {});
+
+  const events = getEventObj(comp.events);
+  const listeners = getEventObj(comp.listeners);
+
+  return { ...props, ...events, ...listeners };
```
bkjohnson, line 77: Instead of manually listing the events for each component, can we use this data?

packages/storybook/stories/va-alert.stories.jsx
```diff
@@ -30,8 +31,11 @@ export default {
   parameters: {
     docs: {
       description: {
-        component: `Use a heading element with an attribute named slot and a value of "headline" to control what is displayed for the alert's headline. 
-        Any children passed into this component without a parent slot "headline" will render in the alert's body.`,
+        component:
+          `Use a heading element with an attribute named slot and a value of "headline" to control what is displayed for the alert's headline. 
+        Any children passed into this component without a parent slot "headline" will render in the alert's body. 
+        This component has 2 events: close and vaComponentDidLoad.` +
```
bkjohnson, line 37: And should we also be calling out the analytics event not necessarily as something people should have to handle, but just to inform them about it?

image


image

bkjohnson (commented):
Do we want to call attention to the global handler for `component-library-analytics`, either here or on design.va.gov?

https://github.com/department-of-veterans-affairs/vets-website/blob/84986db1357884e47b6a79a883a77a6db4172209/src/platform/site-wide/component-library-analytics-setup.js#L104-L107

GnatalieH:
I just wanted to say I'm stoked seeing more documentation for devs going onto Storybook! Yay!

packages/storybook/stories/wc-helpers.jsx
andresriveratoro: That's a great idea. Updated.

packages/storybook/stories/va-alert.stories.jsx
andresriveratoro: I removed the filter for component-library-analytics, so all events are displayed now. I left it out initially, but I think it's okay to display for the sake of completeness.

packages/storybook/stories/wc-helpers.jsx
bkjohnson: Ah, now that I'm taking a closer look at this I see that we're putting the events in with the props. I'm not sure that's very clear, but others may feel differently.

bkjohnson (commented):
I think the text in the blue box is fine, though we might want to improve that later. But I'm not sure if putting the event names into the prop table is great because it isn't obvious what's going on. Within that table you can use the controls to toggle booleans or change string values and see it reflected in the rendered story, so to me the events feel a bit out of place there. We might be setting `bordered`, but we aren't going to be setting `accordionItemToggled`.

image


I feel like having them in their own separate table like this would make it more obvious that these can be handled with an `on*` prop:

image

We're doing something similar in the [Telephone story](https://dev-design.va.gov/storybook/?path=/docs/components-telephone--default) that we may be able to copy, but I don't think it needs to be _exactly_ like that.

image

packages/storybook/stories/wc-helpers.jsx
andresriveratoro: I think you are correct that this could create confusion and that it should be easily identifiable as something separate like in the Telephone example. 

I want to call out that `docs-json` doesn't include description details for listeners like it does for events.
```js
export interface JsonDocsListener {
  event: string;
  target?: string;
  capture: boolean;
  passive: boolean;
}
```

```js
export interface JsonDocsEvent {
  event: string;
  bubbles: boolean;
  cancelable: boolean;
  composed: boolean;
  docs: string;
  docsTags: JsonDocsTag[];
  deprecation?: string;
  detail: string;
}
```

I had to manually override it through Storybook argTypes to add descriptions. I think we might be able to add a JSDoc comment in each component for listeners that gets added to a component's `docsTags`, and use that to return a modified docs using https://stenciljs.com/docs/docs-custom. I haven't tried it yet, but I think this is what we want?

andresriveratoro:
Screen Shot 2022-01-31 at 21 55 13
Screen Shot 2022-01-31 at 21 55 28

This is how it looks like with the latest changes.

andresriveratoro:
For events documentation to generate automatically and display properly on Storybook:
#### Each event must have comments with a description.
```js
  /**
   * The event used to track usage of the component. This is emitted when an
   * input value changes and enableAnalytics is true.
   */
  @Event({
    eventName: 'component-library-analytics',
    composed: true,
    bubbles: true,
  })
  componentLibraryAnalytics: EventEmitter;
```
#### If the event is considered a listener event, you must use JSDoc Tags.  If there are multiple listener events for a component, they must all be listed within the same comment.
```jsx
(...)
/**
 * @vaChange The event emitted when the input value changes.
 */
(...)

  @Listen('vaChange')
  vaChangeHandler(event: CustomEvent): void {
    const clickedItem = event.target as HTMLVaCheckboxElement;
    if (this.enableAnalytics) this.fireAnalyticsEvent(clickedItem.label);
  }
```

#### Example with multiple listener events:
```jsx
/**
 * @vaChange The event emitted when the input value changes.
 * @anotherListener Write a description for me please.
 */
```

packages/web-components/src/components/va-accordion/va-accordion.tsx
```diff
@@ -27,6 +27,10 @@ import { getSlottedNodes } from '../../utils/utils';
  * - Allow users to click anywhere in the header area to expand or collapse the content; a larger target is easier to manipulate.
  * - Make sure interactive elements within the collapsible region are far enough from the headers that users don’t accidentally trigger a collapse. (The exact distance depends on the device.
  */
+
+/**
+ * @accordionItemToggled This event is fired when an accordion item is opened or closed
```
bkjohnson, line 32: This looks like it does bubble up out of the wrapper so I think it's fine to advertise it like this. However, I noticed in the subcomponent arg table it isn't ending up in the stacked `Events` table:

image

I don't think this should block a merge, but if this can be like a 20 minute fix it would be nice to include it. If `subcomponents` work differently and there's nothing we can do about it then that's fine :shrug:

packages/web-components/src/components/va-checkbox-group/va-checkbox-group.tsx
```diff
@@ -9,6 +9,9 @@ import {
   h,
 } from '@stencil/core';
 
+/**
+ * @vaChange The event emitted when the input value changes.
```
bkjohnson, line 13: Are these JSDoc decorators on the `@Component` how stencil builds up its `listeners` object/array in the generated JSON?

bkjohnson (approved):

pdavies88:
Super cool thanks for all the hard work on this definitely will make things easier moving forward

packages/web-components/src/components/va-accordion/va-accordion.tsx
andresriveratoro: It looks like this might not be possible for subcomponents. I found this: https://github.com/storybookjs/storybook/issues/14710

packages/web-components/src/components/va-checkbox-group/va-checkbox-group.tsx
andresriveratoro: These JSDoc tags get added to the `docsTags` array inside the object generated by Stencil `component-docs.json`. There is nothing in the docsTags that indicate it belongs to a specific component. An example docsTags array could look like this: 
```js
[ {  name: 'accordionItemToggled', text: 'Hey there!' } ]
```

Since there is no unique identifier, this could create issues with listeners that have the same names (for example keydown).

andresriveratoro: We could probably improve it by adding the component name in the text for each listener in the JSDoc tag comment, and then perform some logic to remove the component name and display the rest of the text.

andresriveratoro: These decorators get added to the `docsTags` array for each component doc generated by Stencil's `component-docs.json`, but do not extend the `listeners` with this additional information automatically.

k80bowman:
I love this so much, great find!
image

PR #271, opened 2022-01-24
Fix va-alert close event dispatch

andresriveratoro:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/35822

## Testing done
Edge 13, Safari 10, Chrome 45, Firefox 52

Note: Storybook doesn't finish loading in Safari 9 and Chrome 44.

## Screenshots
Screen Shot 2022-01-24 at 15 53 21

## Acceptance criteria
- [x] The va-alert component accepts a function that gets called when the alert is closed
- [x] We are able to see this function work in Storybook
- [x] The user is able to close the Banner component and the Banner onCloseAlert function works.
- [x] This functionality has been tested in all supported browsers (See Testing done)

## Definition of done

### DIFF
--- a/packages/react-components/src/components/Banner/Banner.jsx
+++ b/packages/react-components/src/components/Banner/Banner.jsx
@@ -137,7 +137,7 @@ export class Banner extends Component {
           visible
           full-width
           closeable={showClose}
-          onClose={onCloseAlert}
+          onVaClose={onCloseAlert}
           status={type}
         >
           <h3 slot="headline">{title}</h3>
--- a/packages/react-components/src/components/Banner/Banner.jsx
+++ b/packages/react-components/src/components/Banner/Banner.jsx
@@ -2,7 +2,6 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 // Relative imports.
-
 import { VaAlert } from '@department-of-veterans-affairs/web-components/react-bindings';
 
 const DISMISSED_BANNERS_KEY = 'DISMISSED_BANNERS';
@@ -98,7 +97,7 @@ export class Banner extends Component {
     // Track the dismiss event.
     if (recordEvent) {
       recordEvent({
-        event: 'int-alert-box-close',
+        'event': 'int-alert-box-close',
         'alert-box-headline': title,
       });
     }
@@ -137,7 +136,7 @@ export class Banner extends Component {
           visible
           full-width
           closeable={showClose}
-          onVaClose={onCloseAlert}
+          onCloseEvent={onCloseAlert}
           status={type}
         >
           <h3 slot="headline">{title}</h3>
--- a/packages/storybook/stories/va-alert.stories.jsx
+++ b/packages/storybook/stories/va-alert.stories.jsx
@@ -1,5 +1,6 @@
 /* eslint-disable react/prop-types */
 import React from 'react';
+import { VaAlert } from '@department-of-veterans-affairs/web-components/react-bindings';
 import { getWebComponentDocs, propStructure } from './wc-helpers';
 
 const alertDocs = getWebComponentDocs('va-alert');
@@ -16,7 +17,7 @@ export default {
         disable: true,
       },
     },
-    onclose: {
+    vaClose: {
       description:
         'If closeable is true, this event is triggered when an alert is closed.',
       table: {
@@ -47,7 +48,7 @@ const defaultArgs = {
   'closeable': false,
   'full-width': false,
   'headline': <h3 slot="headline">Alert headline</h3>,
-  'onclose': undefined,
+  'vaClose': undefined,
 };
 
 const Template = ({
@@ -60,8 +61,26 @@ const Template = ({
   closeable,
   'full-width': fullWidth,
   headline,
-  onclose,
+  vaClose,
+  storybookOnlyClose,
 }) => {
+  if (storybookOnlyClose)
+    return (
+      <VaAlert
+        status={status}
+        background-only={backgroundOnly}
+        show-icon={showIcon}
+        disable-analytics={disableAnalytics}
+        visible={visible}
+        close-btn-aria-label={closeBtnAriaLabel}
+        closeable={closeable}
+        full-width={fullWidth}
+        onVaClose={vaClose}
+      >
+        {headline}
+        <div>This is an alert</div>
+      </VaAlert>
+    );
   return (
     <va-alert
       status={status}
@@ -72,7 +91,7 @@ const Template = ({
       close-btn-aria-label={closeBtnAriaLabel}
       closeable={closeable}
       full-width={fullWidth}
-      onclose={onclose}
+      onVaClose={vaClose}
     >
       {headline}
       <div>This is an alert</div>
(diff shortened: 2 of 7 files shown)


pdavies88 (approved):
Thanks for jumping on this fix so quick!!

pdavies88:
@andresriveratoro so since it looks like the bindings need to change to get the `onVaClose` event to fire, do we need a ticket or to put in PR that adjusts current usage on `<va-alert>` to ones that are `<VaAlert>`?

Also I noticed in Storybook it does `<forwardRef>` as the custom element binding which I thought was interesting. I'm assuming it's a Storybook specific thing with bindings
Screen Shot 2022-01-24 at 4 51 39 PM

pdavies88:
Banner React Component Story also fixed:
https://user-images.githubusercontent.com/11822533/150872078-6248d77b-782b-48fa-bcba-d451ddc6caba.mov

andresriveratoro:
> @andresriveratoro so since it looks like the bindings need to change to get the `onVaClose` event to fire, do we need a ticket or to put in PR that adjusts current usage on `<va-alert>` to ones that are `<VaAlert>`?

There are a few `va-alert` (10+) on vets-website that will require updating to the bindings and renaming to the new event name. I'll create a ticket for this.

> Also I noticed in Storybook it does `<forwardRef>` as the custom element binding which I thought was interesting. I'm assuming it's a Storybook specific thing with bindings Screen Shot 2022-01-24 at 4 51 39 PM

I'll get this fixed. Thanks!

packages/storybook/stories/va-alert.stories.jsx
```diff
@@ -60,8 +64,26 @@ const Template = ({
   closeable,
   'full-width': fullWidth,
   headline,
-  onclose,
+  vaClose,
+  storybookOnlyClose,
 }) => {
+  if (storybookOnlyClose)
```
andresriveratoro, line 70: Alternatively, I can create a separate template for a closeable alert.

packages/storybook/stories/va-alert.stories.jsx
```diff
@@ -72,7 +94,7 @@ const Template = ({
       close-btn-aria-label={closeBtnAriaLabel}
       closeable={closeable}
       full-width={fullWidth}
-      onclose={onclose}
+      onVaClose={vaClose}
```
bkjohnson, line 97: Is this doing anything?

packages/web-components/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/web-components",
-  "version": "0.22.0",
+  "version": "1.0.0",
```
bkjohnson, line 3: Major version 0 is a bit special since it indicates that nothing should be considered fully stable. Maybe it's finally time for us to make that bump even without this `close` thing.

packages/storybook/stories/va-alert.stories.jsx
```diff
@@ -60,8 +64,26 @@ const Template = ({
   closeable,
   'full-width': fullWidth,
   headline,
-  onclose,
+  vaClose,
```
bkjohnson, line 67: I think it might be clearer if this was `onVaClose` just to indicate that it is a handler.

packages/storybook/stories/va-alert.stories.jsx
andresriveratoro: Good catch, I'll remove it. 👍

packages/web-components/package.json
andresriveratoro: I didn't know that but thanks! I can revert this if you'd like.

packages/storybook/stories/va-alert.stories.jsx
andresriveratoro: Updated it. 👍

packages/web-components/package.json
k80bowman: I just made a note for us to discuss what kind of guidelines we want to set for versioning this package. Since this would be a breaking change, I am comfortable with bumping the package to 1.0.0 as long as you are OK with it @bkjohnson.

packages/storybook/stories/va-alert.stories.jsx
k80bowman: If we have alerts that aren't closeable, that might be a good idea. Is the default not closeable?

packages/storybook/stories/va-alert.stories.jsx
```diff
@@ -47,7 +51,7 @@ const defaultArgs = {
   'closeable': false,
   'full-width': false,
   'headline': <h3 slot="headline">Alert headline</h3>,
-  'onclose': undefined,
+  'vaClose': undefined,
```
k80bowman, line 54: should this be `onVaClose`?

packages/react-components/src/components/Banner/Banner.jsx
```diff
@@ -137,7 +137,7 @@ export class Banner extends Component {
           visible
           full-width
           closeable={showClose}
-          onClose={onCloseAlert}
+          onVaClose={onCloseAlert}
```
k80bowman, line 140: I'm wondering if we need a more descriptive name here. Would something like `onCloseEvent` be clearer?

packages/storybook/stories/va-alert.stories.jsx
andresriveratoro: I updated it to display bindings if the `onVaClose` prop is present and removed the storybookOnly prop.

packages/storybook/stories/va-alert.stories.jsx
andresriveratoro: I removed this because it was no longer necessary since this prop is being displayed now.

packages/react-components/src/components/Banner/Banner.jsx
andresriveratoro: Most of our custom events have been prefixed with `va`. If we're okay moving away from that, I think that's a much better name. If we want a compromise, how about `onVaCloseEvent`?

packages/web-components/package.json
bkjohnson: I'm fine with it! [Semver guidelines](https://semver.org/#how-do-i-know-when-to-release-100) suggest that we should already be in 1.0.0:

> If your software is being used in production, it should probably already be 1.0.0. If you have a stable API on which users have come to depend, you should be 1.0.0.

Even though we try to hide the `web-components` package and treat `component-library` like the main package, I think it's good for the version to indicate that we're stable.

packages/react-components/src/components/Banner/Banner.jsx
bkjohnson: Either of those is fine - those `va*` event prefixes were originally a workaround to get rid of some stencil build warnings for custom names conflicting with native names. We may want to give those other events some attention at some point :sweat_smile:

k80bowman: yeah, I'm not too keen on the `va*` prefixes, it's not clear on first glance what's going on there.

andresriveratoro: Sounds good to me. I've updated it to `closeEvent` and used as `onCloseEvent`.

k80bowman (approved):

PR #275, opened 2022-01-28
Add va-pagination component

andresriveratoro:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/35098

## Testing done
E2E

## Screenshots
Screen Shot 2022-01-31 at 13 26 45
Screen Shot 2022-01-31 at 13 41 41

## Acceptance criteria
- [x] Build a web-component version of our Pagination React component

## Definition of done

### DIFF
--- a/packages/storybook/stories/va-pagination.stories.jsx
+++ b/packages/storybook/stories/va-pagination.stories.jsx
@@ -21,14 +21,14 @@ const Template = ({
   'show-last-page': showLastPage,
 }) => {
   const [page, setPage] = useState(currentPage);
-  const handlePageSelect = pageNumber => {
-    setPage(pageNumber);
+  const handlePageSelect = event => {
+    setPage(event.detail);
   };
   return (
     <VaPagination
       page={page}
       pages={pages}
-      pageSelect={handlePageSelect}
+      onPageSelect={handlePageSelect}
       showLastPage={showLastPage}
     />
   );
--- a/packages/storybook/stories/va-pagination.stories.jsx
+++ b/packages/storybook/stories/va-pagination.stories.jsx
@@ -8,22 +8,24 @@ export default {
   title: 'Components/va-pagination',
 };
 
-const Template = args => {
-  const [page, setPage] = useState(args.page);
+const Template = ({ page: currentPage, pages, showLastPage }) => {
+  const [page, setPage] = useState(currentPage);
   const handlePageSelect = pageNumber => {
     setPage(pageNumber);
   };
   return (
     <VaPagination
       page={page}
-      pages={args.pages}
+      pages={pages}
       pageSelect={handlePageSelect}
+      showLastPage={showLastPage}
     />
   );
 };
 const defaultArgs = {
   page: 3,
-  pages: 20,
+  pages: 5,
+  showLastPage: true,
 };
 
 export const Default = Template.bind({});
--- a/packages/storybook/stories/va-pagination.stories.jsx
+++ b/packages/storybook/stories/va-pagination.stories.jsx
@@ -10,14 +10,12 @@ export default {
 };
 
 const defaultArgs = {
-  'disable-border': false,
   'page': 3,
   'pages': 5,
   'show-last-page': false,
 };
 
 const Template = ({
-  'disable-border': disableBorder,
   'page': currentPage,
   pages,
   'show-last-page': showLastPage,
@@ -28,7 +26,6 @@ const Template = ({
   };
   return (
     <VaPagination
-      disableBorder={disableBorder}
       page={page}
       pages={pages}
       pageSelect={handlePageSelect}
@@ -42,9 +39,3 @@ Default.args = {
   ...defaultArgs,
 };
 Default.argTypes = propStructure(paginationDocs);
-
-export const Borderless = Template.bind({});
-Borderless.args = {
-  ...defaultArgs,
-  'disable-border': true,
-};
--- a/packages/storybook/stories/va-pagination.stories.jsx
+++ b/packages/storybook/stories/va-pagination.stories.jsx
@@ -2,19 +2,33 @@ import React, { useState } from 'react';
 import { getWebComponentDocs, propStructure } from './wc-helpers';
 import { VaPagination } from '@department-of-veterans-affairs/web-components/react-bindings';
 
+VaPagination.displayName = 'VaPagination';
 const paginationDocs = getWebComponentDocs('va-pagination');
 
 export default {
   title: 'Components/va-pagination',
 };
 
-const Template = ({ page: currentPage, pages, showLastPage }) => {
+const defaultArgs = {
+  'disable-border': false,
+  'page': 3,
+  'pages': 5,
+  'show-last-page': true,
+};
+
+const Template = ({
+  'disable-border': disableBorder,
+  'page': currentPage,
+  pages,
+  'show-last-page': showLastPage,
+}) => {
   const [page, setPage] = useState(currentPage);
   const handlePageSelect = pageNumber => {
     setPage(pageNumber);
   };
   return (
     <VaPagination
+      disableBorder={disableBorder}
       page={page}
       pages={pages}
       pageSelect={handlePageSelect}
@@ -22,14 +36,15 @@ const Template = ({ page: currentPage, pages, showLastPage }) => {
     />
   );
 };
-const defaultArgs = {
-  page: 3,
-  pages: 5,
-  showLastPage: true,
-};
 
 export const Default = Template.bind({});
 Default.args = {
   ...defaultArgs,
 };
 Default.argTypes = propStructure(paginationDocs);
+
+export const Borderless = Template.bind({});
+Borderless.args = {
+  ...defaultArgs,
+  'disable-border': true,
+};
--- a/packages/storybook/stories/va-pagination.stories.jsx
+++ b/packages/storybook/stories/va-pagination.stories.jsx
@@ -22,7 +22,7 @@ const Template = ({
 }) => {
   const [page, setPage] = useState(currentPage);
   const handlePageSelect = event => {
-    setPage(event.detail);
+    setPage(event.detail.page);
   };
   return (
     <VaPagination
--- a/packages/storybook/stories/va-pagination.stories.jsx
+++ b/packages/storybook/stories/va-pagination.stories.jsx
@@ -1,0 +1,33 @@
+import React, { useState } from 'react';
+import { getWebComponentDocs, propStructure } from './wc-helpers';
+import { VaPagination } from '@department-of-veterans-affairs/web-components/react-bindings';
+
+const paginationDocs = getWebComponentDocs('va-pagination');
+
+export default {
+  title: 'Components/va-pagination',
+};
+
+const Template = args => {
+  const [page, setPage] = useState(args.page);
+  const handlePageSelect = pageNumber => {
+    setPage(pageNumber);
+  };
+  return (
+    <VaPagination
+      page={page}
+      pages={args.pages}
+      pageSelect={handlePageSelect}
+    />
+  );
+};
+const defaultArgs = {
+  page: 3,
+  pages: 20,
+};
+
+export const Default = Template.bind({});
+Default.args = {
+  ...defaultArgs,
+};
+Default.argTypes = propStructure(paginationDocs);
--- a/packages/storybook/stories/va-pagination.stories.jsx
+++ b/packages/storybook/stories/va-pagination.stories.jsx
@@ -13,7 +13,7 @@ const defaultArgs = {
   'disable-border': false,
   'page': 3,
   'pages': 5,
-  'show-last-page': true,
+  'show-last-page': false,
 };
 
 const Template = ({
--- a/packages/web-components/src/components/va-pagination/va-pagination.css
+++ b/packages/web-components/src/components/va-pagination/va-pagination.css
@@ -10,6 +10,10 @@ nav {
   width: 100%;
 }
 
+.remove-border {
+  border: 0;
+}
+
 ul {
   display: flex;
   justify-content: center;
--- a/packages/web-components/src/components/va-pagination/va-pagination.css
+++ b/packages/web-components/src/components/va-pagination/va-pagination.css
@@ -1,8 +1,10 @@
+@import '../../mixins/focusable.css';
+
 :host {
-  background-color: #fff;
+  background-color: var(--color-white);
   border-top: 1px solid var(--color-gray-lightest);
   display: flex;
-  font-family: 'Source Sans Pro';
+  font-family: var(--font-source-sans);
   justify-content: center;
   overflow: hidden;
   padding: 2rem 0 1rem;
@@ -28,18 +30,13 @@ button {
   cursor: pointer;
   display: inline-flex;
   font-family: inherit;
-  font-size: 0.9em;
+  font-size: 1.44rem;
   font-weight: 400;
   justify-content: center;
   line-height: 2;
   padding: 0;
   transition: all 0.3s ease-in-out;
   transition-property: color, background-color;
-}
-
-button:focus {
-  outline: 2px solid #f9c642;
-  outline-offset: 2px;
 }
 
 .button-next,
@@ -74,14 +71,10 @@ button:focus {
   width: 3rem;
 }
 
+.button-active,
 .button-inner[aria-current='true'],
 .button-inner:hover,
 .button-inner:focus {
-  background-color: var(--color-primary);
-  color: var(--color-white);
-}
-
-.button-active {
   background-color: var(--color-primary);
   color: var(--color-white);
 }
@@ -111,10 +104,12 @@ button:focus {
 
 @media (min-width: 481px) {
   .pagination-next:not(:empty) .button-next::after {
+    /* right-pointing angle quotation mark › with 2 spaces before it */
     content: '\a0\a0\203a';
   }
 
   .pagination-prev:not(:empty) .button-prev::before {
+    /* left-pointing angle quotation mark ‹ with 2 spaces after it */
     content: '\2039\a0\a0';
   }
 }
--- a/packages/web-components/src/components/va-pagination/va-pagination.css
+++ b/packages/web-components/src/components/va-pagination/va-pagination.css
@@ -1,0 +1,106 @@
+:host {
+  display: block;
+}
+
+nav {
+  border-top: 1px solid var(--color-gray-lightest);
+  display: flex;
+  width: 100%;
+}
+
+ul {
+  display: flex;
+  justify-content: center;
+  list-style-type: none;
+  margin: 0;
+  overflow: hidden;
+  padding: 2rem 0 1rem;
+  position: relative;
+  text-align: center;
+  width: 100%;
+}
+
+li {
+  align-items: center;
+  display: flex;
+}
+
+button {
+  -moz-osx-font-smoothing: auto;
+  -webkit-font-smoothing: auto;
+  background-color: transparent;
+  border: 0;
+  color: var(--color-link-default);
+  cursor: pointer;
+  display: inline-flex;
+  font-size: 0.9em;
+  font-weight: 400;
+  justify-content: center;
+  line-height: 2;
+  padding: 0;
+  transition: all 0.3s ease-in-out;
+  transition-property: color, background-color;
+}
+
+button:focus {
+  outline: 2px solid #f9c642;
+  outline-offset: 2px;
+}
+
+.pagination-prev,
+.pagination-next {
+  display: flex;
+  position: relative;
+  /* border-radius: 0;
+  margin: 0; */
+}
+
+.pagination-prev > button,
+.pagination-next > button {
+  border-radius: 0;
+  margin: 0;
+}
+
+.pagination-prev > button:hover,
+.pagination-next > button:hover {
+  background-color: transparent;
+  color: var(--color-link-default);
+  text-decoration: underline;
+}
+
+.pagination-prev:hover,
+.pagination-next:hover {
+  text-decoration: underline;
+}
+
+.pagination-inner {
+  display: flex;
+  overflow: hidden;
+  white-space: nowrap;
+}
+
+.button-inner {
+  border-radius: 1000px;
+  height: 3rem;
+  margin: 0 0.5rem;
+  width: 3rem;
+}
+
+.button-inner[aria-current='true'],
+.button-inner:hover,
+.button-inner:focus {
+  background-color: var(--color-primary);
+  color: var(--color-white);
+}
+
+.pagination-prev,
+.pagination-next,
+.pagination-inner {
+  padding-bottom: 4px;
+  padding-top: 4px;
+}
+
+.button-active {
+  background-color: var(--color-primary);
+  color: var(--color-white);
+}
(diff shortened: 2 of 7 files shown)


andresriveratoro:
This component has a new prop called `disable-border` because there are instances of Pagination in `vets-website` that have a class provided to just remove the border. `onPageSelect` was renamed to `pageSelect` because Stenciljs complains about prefixing with `on` for non-events. The logic for displaying the buttons is a copy/paste from the React Pagination component. The types for `pageSelect` and `trackEvent` are currently `any` because `() => void` causes the prop name to show up as undefined in Storybook.

packages/web-components/src/components/va-pagination/va-pagination.tsx
```diff
@@ -0,0 +1,245 @@
...
+@Component({
+  tag: 'va-pagination',
+  styleUrl: 'va-pagination.css',
+  shadow: true,
+})
+export class VaPagination {
+  /**
+   * Fires when a page is selected
+   */
+  @Prop() pageSelect: any;
```
bkjohnson, line 18: Can this be an `@Event`? Doing something like this will tell Stencil to have the `on*` handler:

```tsx
  @Event() vaChange: EventEmitter;
  // ...

  this.vaChange.emit()
```

packages/storybook/stories/va-pagination.stories.jsx
```diff
@@ -0,0 +1,50 @@
...
+  const [page, setPage] = useState(currentPage);
+  const handlePageSelect = pageNumber => {
+    setPage(pageNumber);
+  };
+  return (
+    <VaPagination
+      disableBorder={disableBorder}
+      page={page}
+      pages={pages}
+      pageSelect={handlePageSelect}
```
bkjohnson, line 34: Seeing `pageSelect` prop, I'm not expecting that that's an event handler. Stencil [has ways](https://stenciljs.com/docs/events) of indicating that a component will fire a custom event, and that would let us have `onPageSelect` without any build warnings.

packages/storybook/stories/va-pagination.stories.jsx
```diff
@@ -0,0 +1,50 @@
...
+  pages,
+  'show-last-page': showLastPage,
+}) => {
+  const [page, setPage] = useState(currentPage);
+  const handlePageSelect = pageNumber => {
+    setPage(pageNumber);
+  };
+  return (
+    <VaPagination
+      disableBorder={disableBorder}
```
bkjohnson, line 31: Is this supposed to be `removeBorder`?

bkjohnson (commented):
For the `removeBorder`/`disableBorder` thing, I think we may be able to do something that would let people use the [utility classes that they're already using](https://github.com/department-of-veterans-affairs/vets-website/blob/b488760486ab86d8b6a18ee981212ff5a7b897ea/src/applications/disability-benefits/view-payments/components/view-payments-lists/payments/Payments.jsx#L85-L86). We could remove the `<nav>` from the shadow DOM and treat the `<Host>` as a navigation element with `<Host role="navigation">`. That would allow utility classes to be used like we have now:

image

packages/web-components/src/components/va-pagination/va-pagination.tsx
andresriveratoro: I'm not sure if we can do this without updating different handlers in vets-website because they expect a `number` value and don't handle the event themselves.

andresriveratoro: This is the error if I try to prefix a prop with `on` for reference.
```
[ WARN  ]  Build Warn: ./src/components/va-pagination/va-pagination.tsx:18:11
           The @Prop() name "onPageSelect" looks like an
           event. Please use the "@Event()" decorator to
           expose events instead, not properties or
           methods.

     L17:     */
     L18:    @Prop() onPageSelect: any;
```

packages/storybook/stories/va-pagination.stories.jsx
andresriveratoro: I agree that it definitely feels wrong. The real problem is that the handlers for `onPageSelect` in `vets-website` don't handle events like handlers for `VaSelect` as an example.

Here is a handler for va-select used in vets-website. The handler receives the event and extracts what it needs from it.
```jsx
const onValueChange = option => {
    if (Object.values(FACILITY_SORT_METHODS).includes(option.detail.value)) {
      setSelectedSortMethod(option.detail.value);
      dispatch(updateCCProviderSortMethod(option.detail.value));
    } else {
      const selectedFacility = ccEnabledSystems.find(
        facility => facility.id === option.detail.value,
      );
      setSelectedSortMethod(selectedFacility.id);
      dispatch(
        updateCCProviderSortMethod(
          FACILITY_SORT_METHODS.distanceFromFacility,
          selectedFacility,
        ),
      );
    }
  };
```

Here is a handler for Pagination used in vets-website.
```jsx
  const onPageSelect = useCallback(
    selectedPage => {
      setPage(selectedPage); // setPage is a hook with an initial value of 1
      focusElement('#pagination-summary');
    },
    [setPage],
  );
```

Here are some options (not complete)
- We could update Pagination handlers in `vets-website` to handle events and update `onPageSelect` to be an event.
- We could rename Pagination's `onPageSelect` to `pageSelect` (currently in this PR).
- We could use `onPageSelect` as a prop and ignore the build warning.

@k80bowman @bkjohnson @pdavies88 @mendacitees for awareness/suggestions

pdavies88: @andresriveratoro thanks for the thorough comments super helpful for awareness on why. I feel like we are in a catch-22 here and my gut says leave vets-website stuff alone because we could be opening a door here that doesn't need to be opened. Thoughts and feelings on if we remained the prop to pageSelectEvent? That may give someone that little bit more context and avoid the onXYZ error message situation and then add a second line in the comment above the prop explaining why we are doing this.

k80bowman: I like the idea of using pageSelectEvent. I also like the idea of establishing a pattern with the web components for how we handle situations like this. IIRC something similar came up with another web component recently. Should we develop a standard way of handling these types of events?

andresriveratoro: Something else to consider is how it will be displayed in Storybook. I like `pageSelectEvent`, but this will be displayed as a property and not an event in the autogenerated component-docs. I looked through our web components and wasn't able to find an example for function props except events.

packages/web-components/src/components/va-pagination/va-pagination.tsx
bkjohnson: What about something like this?

<details>

```diff
diff --git a/packages/storybook/stories/va-pagination.stories.jsx b/packages/storybook/stories/va-pagination.stories.jsx
index ca42d88..bebff0e 100644
--- a/packages/storybook/stories/va-pagination.stories.jsx
+++ b/packages/storybook/stories/va-pagination.stories.jsx
@@ -21,14 +21,15 @@ const Template = ({
   'show-last-page': showLastPage,
 }) => {
   const [page, setPage] = useState(currentPage);
-  const handlePageSelect = pageNumber => {
-    setPage(pageNumber);
+  const handlePageSelect = pageEvent => {
+    console.log('EVENT', pageEvent);
+    setPage(pageEvent.detail);
   };
   return (
     <VaPagination
       page={page}
       pages={pages}
-      pageSelect={handlePageSelect}
+      onPageSelect={handlePageSelect}
       showLastPage={showLastPage}
     />
   );
diff --git a/packages/web-components/src/components.d.ts b/packages/web-components/src/components.d.ts
index 9b1ec0d..dd5c91b 100644
--- a/packages/web-components/src/components.d.ts
+++ b/packages/web-components/src/components.d.ts
@@ -185,10 +185,6 @@ export namespace Components {
           * The current page number
          */
         "page": number;
-        /**
-          * Fires when a page is selected
-         */
-        "pageSelect": any;
         /**
           * The total number of pages
          */
@@ -724,13 +720,13 @@ declare namespace LocalJSX {
          */
         "maxPageListLength"?: number;
         /**
-          * The current page number
+          * Fires when a page is selected
          */
-        "page"?: number;
+        "onPageSelect"?: (event: CustomEvent<any>) => void;
         /**
-          * Fires when a page is selected
+          * The current page number
          */
-        "pageSelect"?: any;
+        "page"?: number;
         /**
           * The total number of pages
          */
diff --git a/packages/web-components/src/components/va-pagination/va-pagination.tsx b/packages/web-components/src/components/va-pagination/va-pagination.tsx
index 6dac302..d46aa01 100644
--- a/packages/web-components/src/components/va-pagination/va-pagination.tsx
+++ b/packages/web-components/src/components/va-pagination/va-pagination.tsx
@@ -1,4 +1,12 @@
-import { Component, Host, h, Prop, Fragment } from '@stencil/core';
+import {
+  Component,
+  Event,
+  EventEmitter,
+  Host,
+  h,
+  Prop,
+  Fragment,
+} from '@stencil/core';
 import classnames from 'classnames';
 
 declare const window: Window &
@@ -15,7 +23,11 @@ export class VaPagination {
   /**
    * Fires when a page is selected
    */
-  @Prop() pageSelect: any;
+  @Event({
+    composed: true,
+    bubbles: true,
+  })
+  pageSelect: EventEmitter;
 
   /**
    * Fires when a page is selected if enable-analytics is true
@@ -63,7 +75,7 @@ export class VaPagination {
   };
 
   private handlePageSelect = (page, eventID) => {
-    this.pageSelect(page);
+    this.pageSelect.emit({ page });
     if (!this.enableAnalytics) return;
 
     this.handleTrackEvent({
```

</details>

andresriveratoro: We could do this, but we would still have to update current `onPageSelect` handlers in `vets-website` to handle the event.

andresriveratoro: I do feel like this is the best approach even if we have to make changes in `vets-website`.

packages/storybook/stories/va-pagination.stories.jsx
bkjohnson: I [left a comment on the other thread](https://github.com/department-of-veterans-affairs/component-library/pull/275#discussion_r797878103) that's also relevant to this stuff, but I'm more interested in the first option. Based on that diff I put in the comment, that handler for the react Pagination component could be like this for the web component (I removed `useCallback` because I don't know what it's doing):

```js
// selectedPage is the custom event
const onPageSelect = selectedPage => {
    // selectedPage.detail is the data that we give the event - page number in this case
    setPage(selectedPage.detail); // setPage is a hook with an initial value of 1
    focusElement('#pagination-summary');
  },
  [setPage],
);
```

pdavies88: Option 1 is fine with me if we can get it to work. I guess we could just add some Acceptance Criteria that says for the correlating vets-website PR that bumps versions of component-library package. Make sure all of the existing pagination handlers are updated to the correct format as well in that PR.

k80bowman: There are 14 files that import the Pagination component, does this mean we would have to update the pagination handlers in all of those files?

andresriveratoro: @k80bowman That's correct

packages/web-components/src/components/va-pagination/va-pagination.tsx
bkjohnson: > We could do this, but we would still have to update current onPageSelect handlers in vets-website to handle the event.

I don't think that's too much work for teams to do. There might be too much variability there for this to be done automatically with the migration script, but there are only 14 instances where this happens. Manually replacing `myArg` with `myArg.detail` seems simple enough as long as we make sure there's documentation on how to handle the event.

packages/storybook/stories/va-pagination.stories.jsx
bkjohnson: Or teams would be doing it when they do the migration.

packages/web-components/src/components/va-pagination/va-pagination.tsx
k80bowman: I'm Ok with this approach, but we need to add some documentation. I'd say go for it and I'll add some notes to the pagination migration ticket to create some extra documentation on migrating this component. We're going to have to be extra communicative about this one. It might even be good for us to work with a team on migrating their pagination component so we can provide a good example.

packages/web-components/src/components/va-pagination/va-pagination.tsx
```diff
@@ -0,0 +1,229 @@
...
+  @Prop() showLastPage: boolean;
+
+  private handleTrackEvent = (...args) => {
+    if (!this.trackEvent) {
+      if (!window.dataLayer) return;
+      window.dataLayer.push(...args);
+    } else {
+      this.trackEvent(...args);
+    }
+  };
```
bkjohnson, line 63: I don't think we need to work with the datalayer directly in here, we can just fire `component-library-analytics` like we're doing in other places. `vets-website` [listens for that event](https://github.com/department-of-veterans-affairs/vets-website/blob/aa496bd9876a9f204b367b5b4b4234e52a8f7846/src/platform/site-wide/component-library-analytics-setup.js#L104-L107) and then pushes to `window.dataLayer` as appropriate.

packages/web-components/src/components/va-pagination/va-pagination.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+  private handlePageSelect = (page, eventID) => {
+    this.pageSelect(page);
+    if (!this.enableAnalytics) return;
+
+    this.handleTrackEvent({
+      'event': eventID,
+      'paginate-page-number': page,
+    });
```
bkjohnson, line 72: Here's an example of firing the analytics event:


https://github.com/department-of-veterans-affairs/component-library/blob/69655465824e556cc6ea5291be94557e61448986/packages/web-components/src/components/va-progress-bar/va-progress-bar.tsx#L36-L45

packages/web-components/src/components/va-pagination/va-pagination.tsx
andresriveratoro: Thanks for that link. It seems we only have 2 web components registered there. Do we need to create a ticket to add our other web components using analytics?

bkjohnson: Possibly. I've tried to let @bsmartin-ep know when we get a new component but we don't really have a formal process, and it's a bit unclear who is responsible for that bit of code.

k80bowman: I can take a look at our other web components and make a ticket for things we need to update. I also spoke with @bsmartin-ep this morning and put together some [guidelines](https://vfs.atlassian.net/wiki/spaces/DST/pages/2079817745/Component+development+process#Analytics%5BinlineExtension%5D) on how to handle updating/adding the analytics events for our components

packages/storybook/stories/va-pagination.stories.jsx
bkjohnson: I wanted to elaborate a bit on my earlier comment about why I'm in favor of the event handler instead of a callback. A callback would get the job done, but callbacks aren't as robust as the browser's event model. This PR is an example, but I'll break it down a bit further: https://github.com/department-of-veterans-affairs/vets-website/pull/19716

We [originally](https://github.com/department-of-veterans-affairs/vets-website/blob/dcb56da4ff5b2ca649508f39280b7fb8b7a6e50e/src/platform/forms-system/src/js/components/SchemaForm.jsx#L92-L96) had a callback definition like this:

```jsx

onError() {
  const formContext = set('submitted', true, this.state.formContext);
  this.setState({ formContext });
  scrollToFirstError();
}
```

In https://github.com/department-of-veterans-affairs/vets-website/pull/19460 this got changed to parameterize the `submitted` value:

```jsx
onError(hasSubmitted = true) {
  const formContext = set('submitted', hasSubmitted, this.state.formContext);
  this.setState({ formContext });
  scrollToFirstError();
}
```

This _appeared_ fine. However, this callback was [already being called with an argument](https://github.com/department-of-veterans-affairs/react-jsonschema-form/blob/b10769a8c9ce1d69ac51febe8ad07af09cc12de2/src/components/Form.js#L101) that could be an empty array. If this callback had been an event handler this never would have happened. Instead, this caused a bug and led to a support issue.

Once we put a callback out there we become committed to supporting that definition, and if we want to make any big changes we can have a harder time of doing that depending on how the callback is being used by other teams. Relying on the event model decouples things a bit more and gives us more flexibility when it comes to future development. For example, we don't end up calling a callback that might have a signature like `errorCallback(arg1, arg2, somethingElse, data, thatOneFlagSomeoneWanted, options)`.

andresriveratoro: @bkjohnson I appreciate you taking the time and digging through those older PRs to provide this context. Thanks!

packages/web-components/src/components/va-pagination/va-pagination.tsx
```diff
@@ -0,0 +1,232 @@
...
+              this.handlePageSelect(pageNumber, 'nav-paginate-number')
+            }
+            onKeyDown={e => this.handleKeyDown(e, pageNumber)}
+            type="button"
+          >
+            {pageNumber}
+          </button>
+        </li>
+      );
+    });
```
bkjohnson, line 153: Is there a reason that the markup and possibly styles are changing?

https://github.com/department-of-veterans-affairs/component-library/blob/b346faba577fb90c0efc825e7c4268e5893f800f/packages/react-components/src/components/Pagination/Pagination.jsx#L187-L205

packages/web-components/src/components/va-pagination/va-pagination.tsx
```diff
@@ -0,0 +1,232 @@
...
+                }
+                onKeyDown={e => this.handleKeyDown(e, this.page - 1)}
+                type="button"
+              >
+                Prev
+              </button>
+            </li>
+          )}
+          {/* END PREV BUTTON */}
+        </ul>
```
bkjohnson, line 175: Same question here. In the React component this is rendered as:

```jsx
    <span className="va-pagination-prev">
      <a>Prev</a>
    </span>
```

bkjohnson (commented):
I'm not totally sure why we're going from anchors to buttons and including unordered lists, or what the implications of those changes are. I understand that list semantics exist but I'm not sure if this is solving a problem or not. I'm not against it, I'm just trying to understand the possible enhancement/fix.

Additionally, what do we gain by having `<ul class="pagination-prev">` and `<ul class="pagination-next">`, each with only one `<li>`? If we go with the `<ul>` approach, would it make sense for everything to be in the same list and use selectors like `:first-child` and `:last-child` to apply special styling?

packages/web-components/src/components/va-pagination/va-pagination.tsx
```diff
@@ -0,0 +1,232 @@
...
+    } else {
+      start = 1;
+      end = totalPages + 1;
+    }
+    return Array.from({ length: end - start }, (_, i) => i + start);
+  };
+
+  private handleKeyDown = (e, pageNumber) => {
+    const keyCode = e.key;
+    if (keyCode === 'Enter' || keyCode === ' ') {
```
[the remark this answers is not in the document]
andresriveratoro: The main motivation for the markup changes is because we have the opportunity to use the correct elements without having to resort to using roles and workarounds such as `tabIndex` to anchors without `href`. This is 100% for assistive technologies. 

Using NVDA, the React component reads out as "Link previous page, link page 1, link page 2, ...link next page". This web component with the markup changes reads out as "Navigation landmark list with 1 item: button previous page, out of list... list with 5 items: button page 1, button page 2, ...out of list, list with 1 items: button next page". I think this adds a lot of value to screen readers.

I wanted to wrap the entire thing in a `ul`, but our pagination component design has a center aligned container that contains the pages. I had to use an `ul` here because using a `div` or `span` inside `ul` fails the axe check.

packages/web-components/src/components/va-pagination/va-pagination.tsx
andresriveratoro: See my previous comment above.

packages/web-components/src/components/va-pagination/va-pagination.tsx
andresriveratoro: Here are the only visual differences I've found:

The outline is clipped on the right side of the `50` in the React component. This is fixed in the web component.
Screen Shot 2022-02-08 at 20 31 47
Screen Shot 2022-02-08 at 20 32 16

The outline is smaller for the previous and next page anchors in the React component. They are bigger and consistent with the rest of the buttons in the web component.
Screen Shot 2022-02-08 at 20 31 58
Screen Shot 2022-02-08 at 20 32 23

packages/web-components/src/components/va-pagination/va-pagination.css
```diff
@@ -0,0 +1,140 @@
...
+  padding-bottom: 4px;
+  padding-top: 4px;
+}
+
+@media (max-width: 768px) {
+  .pagination-prev {
+    margin-left: -3rem;
+    padding-right: 1rem;
+  }
+}
```
bkjohnson, line 101: Can we combine all of the `@media (max-width: 768px) {` groups?

packages/web-components/src/components/va-pagination/va-pagination.css
```diff
@@ -0,0 +1,140 @@
...
+
+@media (min-width: 481px) {
+  .pagination-prev:not(:empty) .button-prev::before {
+    content: '\2039\a0\a0';
+  }
+}
```
bkjohnson, line 107: Same for `min-width: 481 px`.

packages/web-components/src/components/va-pagination/va-pagination.tsx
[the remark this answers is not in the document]
bkjohnson: Fantastic, thank you for the explanation.

bkjohnson (commented):
Just some minor CSS comments. This looks good overall, we could probably schedule a staging review.

packages/web-components/src/components/va-pagination/va-pagination.tsx
Mottie: Is this an MVP for this web component? I really dislike the current pager because as you can see in the screenshot, there is no easy way to get to page 1 &ndash; imagine being on page 25. You _have_ to click the "Prev" button until you get to the first page.

k80bowman: Yes, for right now we are just focusing on replicating the React component as a web component. I agree that sounds frustrating. @rtwell would we need to go through the DSC to fix this design issue?

pdavies88 (approved):

packages/web-components/src/components/va-pagination/va-pagination.css
```diff
@@ -0,0 +1,134 @@
...
+
+@media (min-width: 481px) {
+  .pagination-next:not(:empty) .button-next::after {
+    content: '\a0\a0\203a';
+  }
+
+  .pagination-prev:not(:empty) .button-prev::before {
+    content: '\2039\a0\a0';
+  }
```
bkjohnson, line 119: Could you add a comment to these describing what those characters are?

bkjohnson (approved):
:tada:

PR #278, opened 2022-02-01
36046 va-accordion IE11 bugs

pdavies88:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/36046

## Testing done
Screenshot (23)

## Screenshots
**Initial Bug**
Screenshot (22)
**Cause of the Issue**
Screenshot (21)
**Resolution**
Screenshot (20)
Screenshot (19)

## Acceptance criteria
- [X] va-accordion should open/close on all supported browsers

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-accordion/va-accordion-item.tsx
+++ b/packages/web-components/src/components/va-accordion/va-accordion-item.tsx
@@ -87,14 +87,8 @@ export class VaAccordionItem {
     // IE11 is unable to run onSlotChange
     // Check needed to populate Accordion Item Header information
     // When using slot="headline" to set the information
-    const ieSlotCheckTag =
-      this.el.querySelector('[slot="headline"]') !== null
-        ? this.el.querySelector('[slot="headline"]').tagName.toLowerCase()
-        : false;
-    const ieSlotCheckHeader =
-      this.el.querySelector('[slot="headline"]') !== null
-        ? this.el.querySelector('[slot="headline"]').innerHTML
-        : false;
+    const ieSlotCheckTag = this.el.querySelector('[slot="headline"]')?.tagName.toLowerCase()
+    const ieSlotCheckHeader = this.el.querySelector('[slot="headline"]')?.innerHTML
 
     const Header = () =>
       h(
--- a/packages/web-components/src/components/va-accordion/va-accordion-item.tsx
+++ b/packages/web-components/src/components/va-accordion/va-accordion-item.tsx
@@ -84,9 +84,21 @@ export class VaAccordionItem {
   }
 
   render() {
+    // IE11 is unable to run onSlotChange
+    // Check needed to populate Accordion Item Header information
+    // When using slot="headline" to set the information
+    const ieSlotCheckTag =
+      this.el.querySelector('[slot="headline"]') !== null
+        ? this.el.querySelector('[slot="headline"]').tagName.toLowerCase()
+        : false;
+    const ieSlotCheckHeader =
+      this.el.querySelector('[slot="headline"]') !== null
+        ? this.el.querySelector('[slot="headline"]').innerHTML
+        : false;
+
     const Header = () =>
       h(
-        this.slotTag || `h${this.level}`,
+        this.slotTag || ieSlotCheckTag || `h${this.level}`,
         null,
         <button
           ref={el => { this.expandButton = el }}
@@ -94,11 +106,10 @@ export class VaAccordionItem {
           aria-expanded={this.open ? 'true' : 'false'}
           aria-controls="content"
         >
-          {this.slotHeader || this.header}
+          {this.slotHeader || ieSlotCheckHeader || this.header}
           {this.subheader ? (<p>{this.subheader}</p>) : false}
         </button>,
       );
-
     return (
       <Host>
         <Header />
--- a/packages/web-components/src/components/va-accordion/va-accordion.tsx
+++ b/packages/web-components/src/components/va-accordion/va-accordion.tsx
@@ -58,8 +58,8 @@ export class VaAccordion {
     let headerLevel
     getSlottedNodes(clickedItem, null).map(
       (node: HTMLSlotElement) => {
-        headerText = node.innerHTML !== undefined ? node.innerHTML : clickedItem.header
-        headerLevel = node.tagName !== undefined ? parseInt(node.tagName.toLowerCase().split('')[1]) : clickedItem.level
+        headerText = node?.innerHTML
+        headerLevel = node?.tagName?.toLowerCase().split('')[1]
     })
 
     if (this.openSingle) {
@@ -77,7 +77,7 @@ export class VaAccordion {
         details: {
           header: headerText || clickedItem.header,
           subheader: clickedItem.subheader,
-          level: headerLevel || clickedItem.level,
+          level: headerLevel !== undefined ? parseInt(headerLevel) : headerLevel || clickedItem.level,
           sectionHeading: this.sectionHeading,
         },
       };
--- a/packages/web-components/src/components/va-accordion/va-accordion.tsx
+++ b/packages/web-components/src/components/va-accordion/va-accordion.tsx
@@ -27,6 +27,10 @@ import { getSlottedNodes } from '../../utils/utils';
  * - Allow users to click anywhere in the header area to expand or collapse the content; a larger target is easier to manipulate.
  * - Make sure interactive elements within the collapsible region are far enough from the headers that users don’t accidentally trigger a collapse. (The exact distance depends on the device.
  */
+
+/**
+ * @accordionItemToggled This event is fired when an accordion item is opened or closed
+ */
 @Component({
   tag: 'va-accordion',
   styleUrl: 'va-accordion.css',
@@ -35,6 +39,10 @@ import { getSlottedNodes } from '../../utils/utils';
 export class VaAccordion {
   @Element() el!: any;
 
+  /**
+   * The event used to track usage of the component. This is emitted when an
+   * accordion item is toggled and disableAnalytics is not true.
+   */
   @Event({
     eventName: 'component-library-analytics',
     composed: true,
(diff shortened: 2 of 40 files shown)


packages/web-components/src/components/va-accordion/va-accordion.tsx
```diff
@@ -58,8 +58,8 @@ export class VaAccordion {
     let headerLevel
     getSlottedNodes(clickedItem, null).map(
       (node: HTMLSlotElement) => {
-        headerText = node.innerHTML
-        headerLevel = parseInt(node.tagName.toLowerCase().split('')[1])
+        headerText = node.innerHTML !== undefined ? node.innerHTML : clickedItem.header
+        headerLevel = node.tagName !== undefined ? parseInt(node.tagName.toLowerCase().split('')[1]) : clickedItem.level
```
bkjohnson, line 62: If we had something like this in IE, what would happen? There isn't a `header` or `level` prop, so how would things work?

```html
<va-accordion>
  <va-accordion-item>
    <h3 slot="headline">How do I X?</h3>
    <div>Do Y then Z</div>
  </va-accordion-item>
</va-accordion>
```

pdavies88: @bkjohnson so that issue you can see in the "Initial Bug" Screenshot I posted above. The first accordion item has an empty header. Which is why I had to make adjustments to the `va-accordion-item.tsx` in order to make sure if you are on IE11 the accordion item header gets included. So the updates in that file are to solve this issue.

packages/web-components/src/components/va-accordion/va-accordion-item.tsx
```diff
@@ -84,21 +84,32 @@ export class VaAccordionItem {
...
+    // When using slot="headline" to set the information
+    const ieSlotCheckTag =
+      this.el.querySelector('[slot="headline"]') !== null
+        ? this.el.querySelector('[slot="headline"]').tagName.toLowerCase()
+        : false;
+    const ieSlotCheckHeader =
+      this.el.querySelector('[slot="headline"]') !== null
+        ? this.el.querySelector('[slot="headline"]').innerHTML
+        : false;
+
```
bkjohnson, line 98: Ok, I think I'm understanding this now based on the previous comment. We've got these because `this.slotTag` and `this.slotHeader` aren't working in IE since the slotchange handler isn't running, right? And it also looks like IE is able to query by `[slot="headline"]`.

I know optional chaining doesn't work on IE, but could you try it here to see if our build handles it appropriately since its configured for `es5`? Something like:

```js
 const ieSlotHeader = this.el.querySelector('[slot="headline"]')?.innerHTML;
```

packages/web-components/src/components/va-accordion/va-accordion.tsx
bkjohnson: We're just setting these values for the analytics event, but we already have a fallback for getting the values from `clickedItem`:

https://github.com/department-of-veterans-affairs/component-library/blob/69655465824e556cc6ea5291be94557e61448986/packages/web-components/src/components/va-accordion/va-accordion.tsx#L84-L87

Could we just leave `headerText` alone here and let that `undefined || clickedItem.header` expression do the work for us?

We might get an error on `.toLowerCase()` if `node.tagName` is undefined, so if our build can compile optional chaining for `es5` could we try that here as well?

packages/web-components/src/components/va-accordion/va-accordion.tsx
```diff
@@ -77,7 +77,7 @@ export class VaAccordion {
         details: {
           header: headerText || clickedItem.header,
           subheader: clickedItem.subheader,
-          level: headerLevel || clickedItem.level,
+          level: headerLevel !== undefined ? parseInt(headerLevel) : headerLevel || clickedItem.level,
```
pdavies88, line 80: Comes back as text when using getSlottedNodes and pulling from the tagName so still need to convert to Int to match format of level props which is Int

packages/web-components/src/components/va-accordion/va-accordion-item.tsx
pdavies88: @bkjohnson optional chaining works. I updated the PR and screenshots are here:
Screenshot (24)
Screenshot (26)
Screenshot (25)

packages/web-components/src/components/va-accordion/va-accordion.tsx
```diff
@@ -65,8 +65,8 @@ export class VaAccordion {
     let headerText;
     let headerLevel;
     getSlottedNodes(clickedItem, null).map((node: HTMLSlotElement) => {
-      headerText = node.innerHTML;
-      headerLevel = parseInt(node.tagName.toLowerCase().split('')[1]);
+      headerText = node?.innerHTML
+      headerLevel = node?.tagName?.toLowerCase().split('')[1]
```
k80bowman, line 69: Do we have tests around what happens when these conditionals aren't met?

packages/web-components/src/components/va-accordion/va-accordion-item.tsx
```diff
@@ -84,21 +84,26 @@ export class VaAccordionItem {
   }
 
   render() {
+    // IE11 is unable to run onSlotChange
+    // Check needed to populate Accordion Item Header information
+    // When using slot="headline" to set the information
+    const ieSlotCheckTag = this.el.querySelector('[slot="headline"]')?.tagName.toLowerCase()
+    const ieSlotCheckHeader = this.el.querySelector('[slot="headline"]')?.innerHTML
```
k80bowman, line 91: Is it possible to add tests for this additional logic or are we going to run into the the difficulties we've had before with writing tests for IE11 stuff?

packages/web-components/src/components/va-accordion/va-accordion.tsx
pdavies88: @k80bowman that's correct. I really just moved the parseInt() from getSlottedNodes into the analytics event because that is where the data is used, and `clickedItem.level` gives us an Integer so I'm trying to keep the types given to the Analytics event the same. When we pull from getSlottedNodes we get the DOM Node back and then I'm parsing through the tagName to get us the heading level which would be 1-6 but doing that gives us that number as a string instead of an integer and works in all other browsers but IE11. In IE11 that parsing for the tagName via getSlottedNodes returns an `EmptyTextNode` so if we are in IE11 we are now doing an optional conditional which would return `undefined` and then `clickedItem.level` instead.
So it boils down to this:
* Modern browser where someone uses `slot="headline"` we need to change the values type from a string to an integer
* IE11 where someone uses `slot="headline"` tagName property doesn't exist therefore return `undefined` which would then fall to the second value which is `clickedItem.level`
* All Browsers (Modern and IE11) someone sets the level directly like: `<va-accordion-item header="Second Amendment" level={5} id="second">` then `clickedItem.level` will be used

packages/web-components/src/components/va-accordion/va-accordion.tsx
pdavies88: They are optional conditionals so they would just return `undefined` when not met. 
By getting back `undefined` which is falsy that helps determine which value should be used in the analytics event on lines 78 and 80

packages/web-components/src/components/va-accordion/va-accordion-item.tsx
pdavies88: @k80bowman so the e2e testing suite that Stencil uses is powered by Puppeteer which runs their tests via headless Chrome so trying to run a specific suite through IE11 I don't believe is possible.

Overall everything being done is to bulletproof the code further. Anything that fails via the optional chaining it will return `undefined` and thus fall back to the next value in the OR statement which in this case would be the prop of `this.level` and the prop of `this.header` which is shown on lines 95 and 103

k80bowman: That's fine, I knew that we might have problems with Puppeteer. I just wanted to check if adding tests would be possible. Especially with bug fixes I like to make sure we have tests to prevent any future regressions.

packages/web-components/src/components/va-accordion/va-accordion.tsx
k80bowman: Do we have tests for what happens when we get a falsy value? I just want to make sure that we have tests to prevent a regression in the future, especially since this is a bug fix.

bkjohnson: Is the `node` going to be undefined here, or is it `innerHTML` and `tagName` that might be undefined?

packages/web-components/src/components/va-accordion/va-accordion.tsx
bkjohnson: I'm not sure this is necessary. If we call `parseInt` on an undefined value we just get `NaN`, it doesn't throw an error:

image

Is something like this possible?

```js
      headerLevel = parseInt(node.tagName?.toLowerCase().split('')[1]);
      
      // ...
                level: headerLevel || clickedItem.level,
```

The expression `NaN || 3` should evaluate to 3 so I think we would be covered.

packages/web-components/src/components/va-accordion/va-accordion.tsx
pdavies88: @bkjohnson node in IE11 returns a string of EmptyTextNode as shown in the screenshots and thus has no properties available on it

packages/web-components/src/components/va-accordion/va-accordion.tsx
pdavies88: @bkjohnson I reverted the parseInt() location back as suggested and pushed up those changes

packages/web-components/src/components/va-accordion/va-accordion.tsx
pdavies88: Test code that I used in IE11 which included both the `slot=headline` method and the `header` props method
Screenshot (29)

The `header` props method returns back `EmptyTextNode` for `node` and the `slot=headline` method returns the `<h6>` element from the sample code which is why we need the optional chaining checks
Screenshot (28)

packages/web-components/src/components/va-accordion/va-accordion.tsx
```diff
@@ -65,8 +65,8 @@ export class VaAccordion {
     let headerText;
     let headerLevel;
     getSlottedNodes(clickedItem, null).map((node: HTMLSlotElement) => {
-      headerText = node.innerHTML;
-      headerLevel = parseInt(node.tagName.toLowerCase().split('')[1]);
+      headerText = node?.innerHTML;
+      headerLevel = parseInt(node?.tagName?.toLowerCase().split('')[1]);
```
bkjohnson, line 69: This looks good - I like that we were able to minimize the IE-specific changes.

bkjohnson (commented):
Do things still work on IE even with the most recent changes?

k80bowman (approved):

PR #280, opened 2022-02-02
Fix va-alert heading to not announce as clickable when using NVDA

andresriveratoro:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/34161

## Testing done
Tested with NVDA

## Screenshots
N/A

## Acceptance criteria
- [x] va-alert heading should not read as clickable using NVDA

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-alert/test/va-alert.e2e.ts
+++ b/packages/web-components/src/components/va-alert/test/va-alert.e2e.ts
@@ -9,7 +9,7 @@ describe('va-alert', () => {
     const element = await page.find('va-alert');
 
     expect(element).toEqualHtml(`
-      <va-alert class="hydrated" role="presentation">
+      <va-alert class="hydrated">
         <mock:shadow-root>
           <div class="alert info">
             <i aria-hidden="true" role="img"></i>
--- a/packages/web-components/src/components/va-alert/test/va-alert.e2e.ts
+++ b/packages/web-components/src/components/va-alert/test/va-alert.e2e.ts
@@ -15,11 +15,7 @@ describe('va-alert', () => {
             <i aria-hidden="true" role="img"></i>
             <div class="body">
               <slot name="headline"></slot>
-              <div id="body-click" role="presentation">
-                <div>
-                  <slot></slot>
-                </div>
-              </div>
+              <slot></slot>
             </div>
           </div>
         </mock:shadow-root>
--- a/packages/web-components/src/components/va-alert/test/va-alert.e2e.ts
+++ b/packages/web-components/src/components/va-alert/test/va-alert.e2e.ts
@@ -15,7 +15,7 @@ describe('va-alert', () => {
             <i aria-hidden="true" role="img"></i>
             <div class="body">
               <slot name="headline"></slot>
-              <div role="presentation">
+              <div id="body-click" role="presentation">
                 <div>
                   <slot></slot>
                 </div>
--- a/packages/web-components/src/components/va-alert/test/va-alert.e2e.ts
+++ b/packages/web-components/src/components/va-alert/test/va-alert.e2e.ts
@@ -9,7 +9,7 @@ describe('va-alert', () => {
     const element = await page.find('va-alert');
 
     expect(element).toEqualHtml(`
-      <va-alert class="hydrated">
+      <va-alert class="hydrated" role="presentation">
         <mock:shadow-root>
           <div class="alert info">
             <i aria-hidden="true" role="img"></i>
--- a/packages/web-components/src/components/va-alert/test/va-alert.e2e.ts
+++ b/packages/web-components/src/components/va-alert/test/va-alert.e2e.ts
@@ -13,11 +13,9 @@ describe('va-alert', () => {
         <mock:shadow-root>
           <div class="alert info">
             <i aria-hidden="true" role="img"></i>
-            <div class="body">
+            <div class="body" role="presentation">
               <slot name="headline"></slot>
-              <div role="presentation">
-                <slot></slot>
-              </div>
+              <slot></slot>
             </div>
           </div>
         </mock:shadow-root>
--- a/packages/web-components/src/components/va-alert/test/va-alert.e2e.ts
+++ b/packages/web-components/src/components/va-alert/test/va-alert.e2e.ts
@@ -15,7 +15,9 @@ describe('va-alert', () => {
             <i aria-hidden="true" role="img"></i>
             <div class="body">
               <slot name="headline"></slot>
-              <slot></slot>
+              <div role="presentation">
+                <slot></slot>
+              </div>
             </div>
           </div>
         </mock:shadow-root>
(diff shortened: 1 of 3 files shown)


k80bowman (approved):

andresriveratoro:
https://github.com/nvaccess/nvda/issues/13262#issuecomment-1020835250 for additional info

packages/web-components/src/components/va-alert/va-alert.tsx
```diff
@@ -153,12 +157,16 @@ export class VaAlert {
...
-      <Host>
+      <Host role="presentation">
         <div role={role} aria-live={ariaLive} class={classes}>
           <i aria-hidden="true" role="img"></i>
-          <div class="body" onClick={this.handleAlertBodyClick.bind(this)}>
+          <div class="body">
             {!backgroundOnly && <slot name="headline"></slot>}
-            <slot></slot>
+            <div id="body-click" role="presentation">
+              <div>
```
bkjohnson, line 166: In the NVDA link, there is this snippet:

```html
<div> <!-- for display:block to create consistency of reading in NVDA -->
```

If the extra div is just there for setting the display, could we have css that's like:

```css

/** For NVDA reading **/
#body-click {
  display: block;
}
```

And not need the `<div>` on 166?

This isn't a huge deal, but I can see this kind of thing easily getting removed in the future if someone comes in here and sees a cascade of divs without any particular styling applied to them. They might remove it as a "refactoring cleanup" and re-introduce the problem. Including CSS in the solution might make things more explicit.

bkjohnson (commented):
That NVDA link also had:

> Typically, the nature of interaction with an element should be conveyed via it's role. NVDA shouldn't have to say clickable. If NVDA reports clickable AND the name of a role that already indicates interactivity (eg 'link'), then it is an NVDA bug.

Does this mean that we might have an a11y issue elsewhere since NVDA isn't able to recognize the role, or is this happening because we're specifically attaching an `onClick` to a div? I wonder if we could use [`@Listen('click')`](https://stenciljs.com/docs/events#listen-decorator) instead to still handle the analytics functionality while not giving NVDA the wrong idea about interactivity.

packages/web-components/src/components/va-alert/va-alert.tsx
andresriveratoro: This isn't working right now. It still says clickable no matter if I use addEventListener, onClick, and role. This is looking like a [React 17 issue](https://github.com/facebook/react/issues/20895) with screen readers. This "clickable" problem is happening on our other components including React components in Storybook.

I'll try the @Listen.

andresriveratoro: Without `role="presentation"` on the React root it doesn't matter what I tried: `onClick` on div with class of `body` or separate div, `addEventListener` and `@Listen` didn't prevent NVDA from announcing va-alert as clickable.

With `role="presentation"` on the React root AND `role="presentation"` on the div with the onClick, va-alert is no longer announced as clickable.

packages/web-components/src/components/va-alert/test/va-alert.e2e.ts
```diff
@@ -13,7 +13,7 @@ describe('va-alert', () => {
         <mock:shadow-root>
           <div class="alert info">
             <i aria-hidden="true" role="img"></i>
-            <div class="body">
+            <div class="body" role="presentation">
```
andresriveratoro, line 16: This fixes the issue from the component-library side of things. Any React applications that consume this component will need to be updated with role="presentation" to prevent NVDA from announcing the element as clickable.

packages/web-components/src/components/va-alert/va-alert.tsx
```diff
@@ -156,7 +156,11 @@ export class VaAlert {
       <Host>
         <div role={role} aria-live={ariaLive} class={classes}>
           <i aria-hidden="true" role="img"></i>
-          <div class="body" onClick={this.handleAlertBodyClick.bind(this)}>
+          <div
+            class="body"
+            onClick={this.handleAlertBodyClick.bind(this)}
+            role="presentation"
```
andresriveratoro, line 162: This fixes the issue from the component-library side of things. Any React applications that consume this component will need to be updated with role="presentation" to prevent NVDA from announcing the element as clickable.

pdavies88 (approved):
Thanks for digging in on alternatives glad role presentation allows it to work correctly

packages/web-components/src/components/va-alert/va-alert.tsx
bkjohnson: What do you mean by consuming React apps will need to be updated? If an app has this:

```jsx
return (
  <div>
    <va-alert>
      Alert content
    </va-alert>
  <div>
  );
```

Will it need to be updated to this?

```jsx
return (
  <div>
    <va-alert role="presentation">
      Alert content
    </va-alert>
  <div>
  );
```

Or do they just need to be updated to the published version with this fix?

andresriveratoro: I was incorrect in my previous comment. **This issue is now fixed completely at the component level.** NVDA won't read this element as "clickable, heading level 3, ..." with `role="presentation` on the div with the onClick.

There is another bug related to React and NVDA where NVDA reads the React parent container as "clickable" even if there is no interactive content. This is easily testable in Storybook or [this example codesandbox](https://codesandbox.io/s/epic-babbage-qc9xj). The fix for that is to add `role="presentation"` to the root element where React mounts.

---

NVDA reads this as "clickable, heading level 2".
```jsx
const rootElement = document.getElementById("root");
ReactDOM.render(
  <StrictMode>
    <App />
  </StrictMode>,
  rootElement
);

<div className="App">
      <h2>This is a heading within a React application.</h2>
</div>
```

If `#root` has `role="presentation"` set, NVDA correctly reads this as "heading level 2".

---

I was able to fix this issue in Storybook by adding this code snippet below in `.storybook/preview.js`. I've removed it from this PR because this change only affects stories, but I can add it back in until it's fixed by React/NVDA. 

```jsx
document.body.onload = function () {
  document.querySelector('#root').setAttribute('role', 'presentation');
};
```

The work for https://github.com/department-of-veterans-affairs/va.gov-team/issues/34161 is completed with the changes in this PR. The remaining React/NVDA issue should not be part of this PR. We should create a new issue for that.

andresriveratoro: Should I re-add the Storybook fix for this issue? It would make it easier to test these types of issues while we wait for the official React/NVDA fix.

bkjohnson: I think making sure storybook reads correctly is a good idea. We could have that `onload` handler in the `preview.js`, but storybook also gives us a way to define our own custom root and put whatever tags we want on it:

https://storybook.js.org/docs/react/configure/story-rendering#adding-to-body

bkjohnson (approved):
Looks good, just be sure to add the fix for storybook in before merge. I don't have strong opinions on the `preview.js` or `preview-body.html`

andresriveratoro:
I tried adding an element to `preview-body.html` but that didn't change what React mounted to. We have the option of using a global decorator and modifying the template, but I wasn't able to get it working. https://storybook.js.org/docs/react/writing-stories/decorators#global-decorators

bkjohnson:
In that case let's just go with what we know works :shrug:

PR #291, opened 2022-02-14
Add responsive table web component

bkjohnson:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/35107

This adds `<va-table>` and `<va-table-row>`. Teams will be expected to use `<span>` elements as children of the `<va-table-row>`. This is to prevent table elements like `<td>` and `<th>` from being removed in static HTML builds. Table semantics will be added as appropriate.

Most of the CSS comes from the [existing sass file in `formation`](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/c4cd826702d9fa5e1a84a8599787d437722d3922/packages/formation/sass/modules/_m-responsive-tables.scss).

## Testing done

- Storybook :eyes: 
- E2E tests
- Unit tests for helper

## Screenshots

Table web component in storybook

And in mobile:

Responsive list table

## Acceptance criteria
- [ ]

## Definition of done


packages/web-components/src/components/va-table/va-table.tsx
```diff
@@ -0,0 +1,69 @@
...
+        if (index === 0) {
+          // Right align columns with numeric data
+          if (isNumeric(cell.textContent)) {
+            alignment[colNum] = 'medium-screen:vads-u-text-align--right';
+            headers[colNum].classList.add(alignment[colNum]);
+          }
+        }
+        if (alignment[colNum]) {
+          cell.classList.add(alignment[colNum]);
+        }
```
bkjohnson: This seems better than setting a `data-*` attribute in JS just to use it as a selector in the CSS.

packages/web-components/src/components/va-table/test/va-table.e2e.ts
```diff
@@ -0,0 +1,116 @@
...
+          <span>Dog</span>
+          <span>Cat</span>
+          <span>Mouse</span>
+        </va-table-row>
+      </va-table>
+    `);
+
+    const element = await page.find('va-table > va-table-row');
+    const cols = await element.findAll('span');
+    expect(element.getAttribute('role')).toEqual('row');
```
bkjohnson, line 33: This is kind of snuck in here. I didn't create a separate e2e test file just for `<va-table-row>` - let me know what your opinions are. Should this be its own test in this file, should it get a new file, or should I leave it as is?

packages/web-components/src/components/va-table/va-table.tsx
```diff
@@ -0,0 +1,75 @@
...
+      Array.from(cells).forEach((cell: HTMLSpanElement, colNum) => {
+        // Look at the first row of data to determine type of data in column
+        // Right align columns with numeric data
+        if (index === 0 && isNumeric(cell.textContent)) {
+          alignment[colNum] = 'medium-screen:vads-u-text-align--right';
+          headers[colNum].classList.add(alignment[colNum]);
+        }
+        if (alignment[colNum]) {
+          cell.classList.add(alignment[colNum]);
+        }
```
bkjohnson, line 49: This seems better than setting a `data-*` attribute in JS just to use it as a selector in the CSS.

packages/storybook/stories/va-table.stories.jsx
```diff
@@ -0,0 +1,70 @@
...
+          ))}
+        </va-table-row>
+
+        {data.map(row => (
+          <va-table-row>
+            {row.map(item => (
+              <span>{item}</span>
+            ))}
+          </va-table-row>
+        ))}
```
bkjohnson, line 56: I'm going to see if I can write a function that will be exported along with the rest of `component-library` that will accept a `fields` and `data` arg structured the same way as the current `<Table>` props. This should help with migration since we should be able to take this:

```jsx
import Table from '@department-of-veterans-affairs/component-library/Table';

// ...

const content = <Table data={data} fields={fields} />;
```
and have our ESLint rule suggest this:


```jsx
import { tableChildren } from '@department-of-veterans-affairs/component-library';

// ...

const content = <va-table>{tableChildren(fields, data)}</va-table>;
```

If it works I'll use it here in the story.

bkjohnson (commented):
I may be making some minor changes here and there but this feels like it's ready for a review.

packages/web-components/src/components/va-table/test/va-table.e2e.ts
andresriveratoro: I think it's fine as is since this logic is handled in `va-table` and `va-table-row`'s only responsibility is to accept children.

andresriveratoro (commented):
I think you covered this on Monday, but I don't remember the answer. `This is to prevent table elements like <td> and <th> from being removed in static HTML builds.` Why does this happen in static HTML builds?

LGTM overall

bkjohnson:
I'm still not totally sure where it's happening, but I suspect it's because of semantics. Whatever is doing the build will see something like this:

```html
<va-table-row>
  <td>Foo</td>
  <td>Bar</td>
</va-table-row>
```

It knows the rules for `<td>` elements like their default role and allowed parents/children, but it doesn't know anything about `<va-table-row>`. Since that element doesn't have table semantics (`role="row"` will get added later at runtime) it decides that the `<td>` is where it doesn't belong, so the element is removed.

We might be able to avoid this if we were able to define our custom elements in the build process like we're doing in the browser, but I didn't dig that deep to find out if that's possible or where I would even do it.

pdavies88 (approved):

pdavies88:
Do we want to add examples for Missing Data and Custom Components like in the React Component story?

bkjohnson:
Good call, I'll work on adding more stories.

packages/web-components/src/components/va-table/va-table.tsx
```diff
@@ -0,0 +1,68 @@
...
+  styleUrl: 'va-table.css',
+  shadow: true,
+})
+export class VaTable {
+  @Element() el: HTMLElement;
+
+  /**
+   * The title of the table
+   */
+  @Prop() caption: string;
```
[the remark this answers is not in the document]
k80bowman: could we use something like tableTitle instead? Something that makes it clear that this prop provides the text for the title.

bkjohnson:
I put this scary message on the story just in case the published version makes its way into `vets-website` before we want it to:

image

pdavies88:
Rechecked with the new stories and LGTM

k80bowman (approved):

PR #294, opened 2022-02-17
Fix va-breadcrumbs page reload when using Link elements

andresriveratoro:
## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/635

## Testing done
vets-website (localhost)

## Acceptance criteria
- [x] va-breadcrumbs routes without a page reload for `<Link>` elements 

## Definition of done


packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.css
```diff
@@ -24,67 +20,40 @@ ul {
   width: 100%;
 }
 
-li {
+::slotted(li) {
   display: inline-block;
-  line-height: 1.5;
-  margin: 0.4rem 0;
+  line-height: 1.5 !important;
+  margin: 0.4rem 0 !important;
```
andresriveratoro, line 26: These are here because external styles were affecting `li` somehow.

packages/web-components/src/global/main.css
```diff
@@ -5,3 +5,30 @@ va-process-list li :is(h2, h3, h4, h5) {
...
+  outline-offset: 0.2rem;
+}
+
+va-breadcrumbs li > a[aria-current='page'] {
+  color: var(--color-gray-dark);
+  cursor: default;
+  font-weight: bold;
+  pointer-events: none;
+  text-decoration: none;
+}
```
andresriveratoro, line 34: We might have to come up with a cleaner solution if we continue to have web components with nested slot content. Maybe an additional css file in the component directory for these nested rules and import it here?

k80bowman: I do like the idea of keeping the CSS intended for a component with that component. Also, you're right, if we have more of these types of components, I could see this file getting huge and unorganized.

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.css
andresriveratoro: This is coming from Formation.

<details><summary>vets-website</summary>
Screen Shot 2022-02-17 at 10 15 32
</details>

<details><summary>Storybook</summary>
Screen Shot 2022-02-17 at 10 20 24
</details>

andresriveratoro:
Clicking on a <Link> element on `vets-website`:
https://user-images.githubusercontent.com/36863582/154514385-307fdfe0-8bcc-444b-8928-af87b0874f25.mov

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.css
k80bowman: I'm not keen on using `!important` here. And I'm wondering why Formation styles are overriding component styles in the cascade, because ideally it would be the other way around.

packages/web-components/src/global/main.css
bkjohnson: shame file 2.0 :sweat_smile:

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.css
bkjohnson: Slotted content is technically in the light DOM, so global styles will still affect it.

bkjohnson (commented):
Did you test to see if the [DOM manipulation](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/635#issuecomment-1040879847) would break a `<Link>`? I'd like to see what our other options are before we change the api so that developers have to pass in a series of `<li><a /></li>` instead of just `<a />`.

bkjohnson:
Another option would be to have our consumers make some slight changes to use `useNavigate` in a handler instead of use `<Link>` children. [API docs](https://reactrouter.com/docs/en/v6/getting-started/overview#navigation):

> Use `Link` to let the user change the URL or `useNavigate` to do it yourself (like after form submissions)

This might look like:

```jsx
import { useNavigate } from "react-router-dom";

const navigate = useNavigate();

const handleClick = e => {
  navigate(e.target.href);  
}

return (
  <va-breadcrumbs onClick={handleClick}>
    <a href="my/page">My Page</a>
  </va-breadcrumbs>
);
```

andresriveratoro:
> Did you test to see if the [DOM manipulation](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/635#issuecomment-1040879847) would break a `<Link>`? I'd like to see what our other options are before we change the api so that developers have to pass in a series of `<li><a /></li>` instead of just `<a />`.

It's working as expected with `<a>` and `<Link>` without requiring a dev to use their own `li` wrappers. I moved away from the slot ref way because it wasn't required (to limit changes) although it was working that way as well. I think the key is `parentNode.replaceChild(li, crumb)` - returning the node with a wrapper instead of trying to construct a new anchor.

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.css
andresriveratoro: The latest changes on `vets-website`:

Screen Shot 2022-02-17 at 13 26 21

k80bowman: Perfect, thank you!

bkjohnson: > shouldn't component styles be loaded after global styles in the cascade, thereby overriding them (unless there is a specificity issue)?

It's a specificity thing, unfortunately. [This answer](https://answerbun.com/stack-overflow/slotted-css-selector-for-nested-children-in-shadowdom-slot/#w) has some good explanations, specifically (no pun intended) from the section on the workaround for styling the light DOM:

> **Note: `::slotted([Simple Selector])`** confirms to Specificity rules,
but (being simple) does not add weight to lightDOM skin selectors, so never gets higher Specificity.
You might need `!important` in some (rare) use cases.

And the reason we can only use simple selectors inside `::slotted()` is for performance reasons.

packages/web-components/src/global/main.css
```diff
@@ -1,4 +1,5 @@
 @import 'variables.css';
+@import '../components/va-breadcrumbs/va-breadcrumbs-slot.css';
```
bkjohnson, line 2: I like this more than what I did with just putting the `va-process-list` selector directly in here, but we may want to think about what we can do to limit this kind of thing from becoming a burden to maintain.

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
```diff
@@ -68,26 +65,16 @@ export class VaBreadcrumbs {
...
-    );
+    anchorNodes.forEach((crumb: HTMLAnchorElement, index: number) => {
+      const li = document.createElement('li');
+      li.classList.add('va-breadcrumbs-li');
+      if (index === anchorNodes.length - 1) {
+        crumb.setAttribute('aria-current', 'page');
+      }
+      crumb.parentNode.replaceChild(li, crumb);
+      li.appendChild(crumb);
+    });
```
bkjohnson, line 77: This is cool, I'm glad this wrapping worked. Could you add a comment to the code that explains this a bit?

I also like the `.classList.add()` and think that we may be able to use that to help us manage the global styles in our `main.css`.

For example, we could do something like:

```js
      li.classList.add('vads-u-display--none medium-screen:vads-u-display--block');
```

I'm not saying that we should do that now, it's just food for thought. I think in order for something like that to work with all of the styling we're using here we'd need to spend some time on further developing our utility classes.

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs-slot.css
```diff
@@ -0,0 +1,66 @@
...
+}
+
+va-breadcrumbs li > a:hover {
+  background: rgba(0, 0, 0, 0.05);
+}
+
+va-breadcrumbs li > a:focus {
+  outline: 2px solid var(--color-gold-light);
+  outline-offset: 0.2rem;
+}
```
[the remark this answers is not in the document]
andresriveratoro: This screenshot is from a page in `vets-website` using this web component. Most of the styles are the same with some exceptions.

Screen Shot 2022-02-17 at 19 22 17

I left them in because I don't like depending on external styles. That could lead to our components not displaying properly if styling changes are made elsewhere. Although it may lead to more chore work, I like it because it's more predictable and controllable. It makes our component-library components more self-contained. If styles need to be updated, there will be tickets created for that and it won't catch us by surprise. I would rather work on a ticket that is about updating styling and not about fixing a bug in our components because of external styling changes.

Should our components inherit external styles or should they contain all the styling? @bkjohnson @k80bowman @pdavies88 @mendacitees

bkjohnson (commented):
Now that we are displaying the slotted content instead of grabbing it and copying it into the shadow DOM, is there an opportunity to use the `::slotted()` selector to keep some of the styling internal?

andresriveratoro:
> Now that we are displaying the slotted content instead of grabbing it and copying it into the shadow DOM, is there an opportunity to use the `::slotted()` selector to keep some of the styling internal?

In `packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.css` `::slotted(li)` works, but `a` and `::slotted(a)` do not.

andresriveratoro:
I tried moving the li styles out of the global styles, but I ran into specificity issues even using a class.

Screen Shot 2022-02-18 at 16 00 32

bkjohnson (approved):
This seems to solve the `<Link>` problem which is the important part, and I don't think that will be impacted if we end up making any future CSS changes.

PR #297, opened 2022-02-21
35152 Promo Banner Web Component

pdavies88:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/35152

## Testing done
Screen Shot 2022-02-21 at 6 54 38 PM

## Screenshots
Screen Shot 2022-02-21 at 6 56 21 PM
Screen Shot 2022-02-21 at 6 56 29 PM
Screen Shot 2022-02-21 at 6 57 56 PM

## Acceptance criteria
- [x] va-promo-banner is written
- [x] E2E tests are written
- [ ] Add Chromatic link to [#36836](https://github.com/department-of-veterans-affairs/va.gov-team/issues/36836), assign ticket to Courtney, and move ticket to Current Sprint column
- [ ] Merge component

## Definition of done


packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
```diff
@@ -0,0 +1,170 @@
...
+export class VaPromoBanner {
+  @Element() el: HTMLElement;
+
+  /**
+   * The rendering custom markup via the slot instead of the `<a>` with `text` in it
+   * */
+  @Prop() renderCustom: boolean = false;
+
+  /**
+   * `href` attribute for the `<a>` tag. Only gets used if `render` is _not_ used
```
andresriveratoro, line 27: Should this be `renderCustom`?

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  /**
+   * `target` attribute for the `<a>` tag. Only gets used if `render` is _not_ used
+   */
+  @Prop() target: string;
+
+  /**
+   * Controls which icon gets used. Options are 'announcement', 'news', or 'email-signup'.
+   * */
+  @Prop() type: string = 'announcement';
```
andresriveratoro, line 39: We could use `'announcement' | 'news' | 'email-signup'` instead of `string` to better represent `type`'s type.

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
```diff
@@ -0,0 +1,170 @@
...
+            {this.renderCustom ? (
+              <slot></slot>
+            ) : (
+              <a
+                class="va-banner-content-link"
+                href={this.href}
+                target={this.target}
+                onClick={e => this.handleLinkClick(e)}
+              >
+                <slot></slot> <i aria-hidden="true" role="presentation" />
```
andresriveratoro, line 153: Is passing `<Link>` a possible use case here? Will it cause a page reload?

andresriveratoro (commented):
LGTM overall

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
pdavies88: So it was originally `render` but I thought that was a confusing prop name since theres a render function, but this is supposed to allow for a prop that indicates custom html that isn't inside of an anchor tag. Def open to suggestions on renaming it

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
pdavies88: Do we have an example of using the `<Link>` tag in another Web Component? Thought that was a React Router thing

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
andresriveratoro: Oh, I meant renaming `render` to `renderCustom` in the comment to reflect the latest prop name.

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
andresriveratoro: After a second look at this, devs can use `renderCustom` if they need to use `<Link>`.

Here's an example for my use case, but I don't think it's useful here. I needed to wrap children with a `li` element and tried this approach, but it caused page reloads when using React Router's `<Link>`.
https://github.com/department-of-veterans-affairs/component-library/blob/69655465824e556cc6ea5291be94557e61448986/packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx#L70-L91

This was the solution and accomplished the same thing without transforming the node.
https://github.com/department-of-veterans-affairs/component-library/blob/c145ac3802840c39f0b6e94b14c94146829a3a90/packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx#L67-L78

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
```diff
@@ -0,0 +1,170 @@
...
+  styleUrl: 'va-promo-banner.css',
+  shadow: true,
+})
+export class VaPromoBanner {
+  @Element() el: HTMLElement;
+
+  /**
+   * The rendering custom markup via the slot instead of the `<a>` with `text` in it
+   * */
+  @Prop() renderCustom: boolean = false;
```
bkjohnson, line 24: Could we just determine this automatically based on the slotted content? We may be able to see if it is an actual `Node` or just text content and then render with the `<a>` wrapping the `<slot>` or not.

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  /**
+   * Analytics tracking function(s) will not be called
+   */
+  @Prop() disableAnalytics: boolean;
+
+  /**
+   * Id of Promo Banner Instance for tracking in Storage
+   */
+  @Prop() bannerId: string;
```
bkjohnson, line 49: What might this be, and do we need a separate prop for it or could we just use the element's `id`, or some combination [like we're doing in `<va-banner>`](https://github.com/department-of-veterans-affairs/component-library/blob/3775aeb5fb3ea3f8d3ea26dd57afc5fb268d096b/packages/web-components/src/components/va-banner/va-banner.tsx#L71-L72)?

packages/web-components/src/components/va-promo-banner/va-promo-banner.css
```diff
@@ -0,0 +1,120 @@
...
+.va-banner-close button i::before {
+  content: '\F057'; /* fa-times-circle */
+}
+
+i {
+  font-family: 'Font Awesome 5 Free';
+  font-style: normal;
+  font-weight: 900;
+  -moz-osx-font-smoothing: grayscale;
+  -webkit-font-smoothing: antialiased;
```
bkjohnson, line 103: I don't think we need to add vendor-specific stuff like this, I've noticed `postcss` handling that for us before when necessary.

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
pdavies88: Yeah if they are going to do something atypical like that render-custom should be the prop to use

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
pdavies88: Updates have been made

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
pdavies88: Updates have been made

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
pdavies88: So its coming from here: https://github.com/department-of-veterans-affairs/content-build/blob/master/src/site/components/banners.drupal.liquid#L33
So we are actually getting an ID from Drupal in this case where as in the Banner we are making it up on the fly in the component.
For that I think we would need to keep it because we would need to map to it but correct my if you have other thoughts happy to update

packages/web-components/src/components/va-promo-banner/va-promo-banner.css
pdavies88: Beautiful going to remove

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
andresriveratoro: Awesome, thank you for that information.

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
pdavies88: update to this.el.id to do the matching on the banner for storage

pdavies88:
Per conversation with @k80bowman and @bkjohnson: https://dsva.slack.com/archives/G01BJ3ESXL4/p1645808762289199

We are going to remove the `render-custom` ability from this component since it is not used nor has been used in the past. If a team requests this feature in the future we can reimplement it at that time.

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
pdavies88: @bkjohnson I updated it to use the native `id` attribute and remove `render-custom`

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
pdavies88: Removed per: https://github.com/department-of-veterans-affairs/component-library/pull/297#issuecomment-1051216595

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
```diff
@@ -0,0 +1,156 @@
...
+    const isBannerDismissed = this.dismissedBanners?.includes(this.el.id);
+
+    // Do not render if the promo banner is dismissed.
+    if (isBannerDismissed) {
+      return null;
+    }
+
+    return (
+      <Host>
+        <div class="va-banner-body">
```
bkjohnson, line 127: Would it be possible to use the `<Host>` for this purpose or do we need the extra nesting for position/display reasons?

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
```diff
@@ -0,0 +1,156 @@
...
+      DISMISSED_PROMO_BANNERS_LOCAL_STORAGE_KEY,
+    );
+    this.dismissedBanners = dismissedBannersString
+      ? JSON.parse(dismissedBannersString)
+      : [];
+  }
+
+  render() {
+    // Derive if the banner is dismissed based on the id attribute set on the component.
+    const isBannerDismissed = this.dismissedBanners?.includes(this.el.id);
```
bkjohnson, line 118: If `id` is necessary in order for dismiss functionality to work it may be good to have a test to show that dismiss _won't_ work without an `id`. In the React version we left those implementation details up to the consumer, so now that we're internalizing it having a test case describing the intentional behavior could act as a form of documentation.

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
```diff
@@ -0,0 +1,156 @@
...
+        (targetEl?.tagName === 'SLOT' &&
+          targetEl?.parentElement.tagName === 'A');
+      // If it's a link being clicked, dispatch an analytics event
+      if (targetAnchor) {
+        const detail = {
+          componentName: 'va-promo-banner',
+          action: 'linkClick',
+          details: {
+            text:
+              targetEl.tagName === 'A' ? targetEl.innerText : this.el.innerText,
```
bkjohnson, line 95: Is this leftover from the work on `customRender`? Would the click target here ever not be an `<a>`?

packages/web-components/src/components/va-promo-banner/va-promo-banner.css
```diff
@@ -0,0 +1,122 @@
...
+.va-banner-content {
+  padding: 0.8rem;
+  line-height: 2rem;
+}
+
+.va-banner-content-link {
+  display: inline-block;
+  text-decoration: none;
+  color: var(--color-link-default);
+}
```
bkjohnson, line 71: The React implementation has `text-decoration: underline` on `:hover`. We probably want to add that.

bkjohnson (commented):
I'm not big on the layout table styling, but I know that's just what the previous React implementation used. At some point we may want to see if something like flexbox would be more appropriate but I don't think we necessarily need to do that kind of work right now.

pdavies88:
> I'm not big on the layout table styling, but I know that's just what the previous React implementation used. At some point we may want to see if something like flexbox would be more appropriate but I don't think we necessarily need to do that kind of work right now.

Per team convo we are going to change to flexbox for the layout

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
pdavies88: Needed for the positioning and display but I should be able to tighten this up with flexbox

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
pdavies88: Yes just removed it

packages/web-components/src/components/va-promo-banner/va-promo-banner.css
pdavies88: Added good catch

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
pdavies88: @bkjohnson added a test for this and also converted the css to flex instead of table

k80bowman (approved):

packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx
```diff
@@ -0,0 +1,142 @@
...
+          <div class="va-banner-close">
+            <button
+              type="button"
+              aria-label="Dismiss this promo banner"
+              onClick={() => this.closeHandler()}
+            >
+              <i aria-hidden="true" role="presentation" />
+            </button>
+          </div>
+        </div>
```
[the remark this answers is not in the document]
pdavies88: @bkjohnson thanks for the code I have made those updates and everything is looking good

bkjohnson (approved):

pdavies88:
Passing per: https://github.com/department-of-veterans-affairs/va.gov-team/issues/36836

PR #298, opened 2022-02-22
35255 va modal dev

andresriveratoro:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/35255

## Testing done
E2E, Storybook

## Screenshots
Screen Shot 2022-02-28 at 11 12 02

## Acceptance criteria
- [x] `va-modal` is written
- [x] E2E tests are written
- [x] Add Chromatic link to #36837, assign ticket to Courtney, and move ticket to Current Sprint column
- [ ] Merge component

## Definition of done


andresriveratoro:
## Differences from React component
- Replaced `contents` with `slot`
- Renamed `close` prop to `closeEvent`
- Renamed `title` prop to `modalTitle`
- Removed `focusSelector` prop because it wasn’t being used anywhere (not even in the React component)
- Replaced `aria-labelledby` with `aria-label` would fail an accessibility check when `modalTitle` wasn't provided (`modalTitle` is optional)
- Set `aria-label` default value to `‘Modal’`
- Replaced `react-focus-on` package with `aria-hidden`, `body-scroll-lock` `focus-trap` packages

packages/web-components/src/components/va-modal/va-modal.tsx
```diff
@@ -0,0 +1,304 @@
...
+    this.isVisibleDirty = true;
+  }
+
+  componentDidLoad() {
+    if (this.visible) this.setupModal();
+  }
+
+  componentDidUpdate() {
+    // This dirty check prevents the component from updating endlessly.
+    if (!this.isVisibleDirty) return;
```
andresriveratoro, line 138: This is needed to provide the same functionality as React's componentDidUpdate. Stencil's `componentDidUpdate` doesn't provide previous prop values that we can use to decide if `setupModal` or `teardownModal` should be invoked. We can use `@Watch` to know when `visible` changes and perform the check in `componentDidUpdate` to prevent endless updating.

packages/web-components/src/components/va-modal/va-modal-slot.css
```diff
@@ -0,0 +1,3 @@
+va-modal p:first-child {
+  margin-top: 0;
+}
```
andresriveratoro, line 3: I could have used `::slotted(p:first-child)` with an `!important`, but I'm not sure if we want that.

packages/web-components/src/components.d.ts
```diff
@@ -166,6 +166,41 @@ export namespace Components {
...
+        "modalTitle"?: string;
+        /**
+          * Primary button text and action
+         */
+        "primaryButton"?: { text: string; action: () => void };
+        /**
+          * Secondary button text and action
+         */
+        "secondaryButton"?: { text: string; action: () => void };
+        "status"?: 'continue' | 'error' | 'info' | 'success' | 'warning';
```
pdavies88, line 198: @andresriveratoro missing the comment here and then looking at storybook we are getting an `undefined` prop being set below `visible`: https://60f9b557105290003b387cd5-dighdkhdhk.chromatic.com/?path=/docs/components-va-modal--default

packages/web-components/src/components/va-modal/va-modal.css
```diff
@@ -0,0 +1,201 @@
+@import '../../mixins/buttons.css';
+@import '../../mixins/focusable.css';
```
pdavies88, line 2: sorry minor - prob don't need the focusable import since buttons.css is already importing it

packages/web-components/src/components/va-modal/va-modal.css
```diff
@@ -0,0 +1,201 @@
...
+.va-modal-body {
+  margin-right: 2rem;
+  overflow-wrap: break-word;
+  padding: 2rem;
+  word-break: break-word;
+  word-wrap: break-word;
+}
+
+h1 {
+  font-family: Bitter, Georgia, Cambria, 'Times New Roman', Times, serif;
```
pdavies88, line 55: sorry minor - we can use --font-serif for that https://github.com/department-of-veterans-affairs/component-library/blob/e157e3c011afb64c8431cc7c7b266ab4c786ac03/packages/web-components/src/global/variables.css#L31

packages/web-components/src/components/va-modal/va-modal.css
```diff
@@ -0,0 +1,201 @@
...
+  outline: 2px solid var(--color-gold-light);
+  outline-offset: 2px;
+}
+
+.va-modal-title {
+  margin: 0 0 1.6rem 0;
+}
+
+i::before {
+  -webkit-font-smoothing: antialiased;
```
pdavies88, line 72: sorry minor - we can remove the browser prefixes since postcss will handle that for us

packages/web-components/src/components/va-modal/va-modal.css
```diff
@@ -0,0 +1,201 @@
...
+  font-family: 'Font Awesome 5 Free';
+  font-style: normal;
+  font-variant: normal;
+  font-weight: 900;
+  line-height: 1;
+  text-rendering: auto;
+}
+
+.alert-actions button {
+  -moz-osx-font-smoothing: grayscale;
```
pdavies88, line 84: same thing on the browser prefixes

packages/web-components/src/components/va-modal/va-modal.css
```diff
@@ -0,0 +1,201 @@
...
+  color: var(--color-primary);
+}
+
+.alert-actions button:hover,
+.alert-actions button:focus {
+  background-color: var(--color-primary-darker);
+}
+.alert-actions .button-secondary:hover,
+.alert-actions .button-secondary:focus {
+  background-color: #dce4ef;
```
pdavies88, line 105: Weird that we don't have this in the variables.css. @bkjohnson I know you ran into something similar on process list with a gray that didn't 100% match wanted to get your thoughts on this

packages/web-components/src/components/va-modal/va-modal.css
```diff
@@ -0,0 +1,201 @@
...
+  padding: 0;
+}
+
+.va-modal-alert-continue {
+  border-left-color: var(--color-green);
+}
+
+.va-modal-alert-continue::before {
+  color: var(--color-green);
+  content: '\F023';
```
pdavies88, line 149: sorry minor - can you add the fa icon name here as a comment and for the ones below like you had done with  /* fa-times-circle*/

packages/web-components/package.json
```diff
@@ -27,8 +27,13 @@
   },
   "dependencies": {
     "@stencil/core": "^2.10.0",
+    "@stencil/postcss": "^2.0.0",
```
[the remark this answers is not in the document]
andresriveratoro: I'm not sure how this happened, but great catch!

packages/web-components/src/components.d.ts
andresriveratoro: I can fix this by updating the type to `any` . I haven't been able to figure out why this happens in Storybook when specifying object or function types.

pdavies88: Yeah its weird because the whole "Style of modal alert - info, error, success, warning" seems to get eaten during compilation

packages/web-components/src/components/va-modal/va-modal.css
andresriveratoro: If I remove these styles, it'll look slightly different in Storybook when compared to the React component.

andresriveratoro: Are we okay with that?

pdavies88: Interesting @bkjohnson I was under the impression that postcss would handle all the browser specific prefixes. Am I wrong on that? Cause if so we should ignore my comments on this

andresriveratoro: The visual difference is minor (it looks slightly more bold without it). Is PostCSS supposed to be handling this in Storybook?

With it:
Screen Shot 2022-03-01 at 12 53 18

Without it:
Screen Shot 2022-03-01 at 12 52 53

packages/web-components/package.json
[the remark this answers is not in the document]
bkjohnson: Branch merge conflicts are a pain :upside_down_face:

packages/web-components/src/components/va-modal/va-modal.css
bkjohnson: I had seen postcss do similar things but I may have been wrong that it was doing this for us.

Anyway, if we want this kind of font smoothing we may want to put it in that `buttons.css` file that gets imported. Something to keep in mind though, from [MDN's documentation regarding `font-smooth`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-smooth):

>  This feature is non-standard and is not on a standards track.

I'd also like to write components that don't target or favor specific browsers or OSes. If the build process adds things in for us that's one thing but adding support in the source gets messy (IE is an example of this).

andresriveratoro: I'm okay with removing it since the difference is minor. Should we do that?

packages/web-components/src/components/va-modal/va-modal.css
bkjohnson: It's already part of the [design system color palette](https://design.va.gov/design/color-palette), we just haven't used it in web components before. We can just add it as a new variable.

image

packages/web-components/src/components/va-modal/va-modal.tsx
```diff
@@ -0,0 +1,310 @@
...
+    const ariaLabel = `${modalTitle} modal` || 'Modal';
+    const wrapperClass = classnames('va-modal-inner', {
+      'va-modal-alert': status,
+      [`va-modal-alert-${status}`]: status,
+    });
+    const bodyClass = status ? 'va-modal-alert-body' : 'va-modal-body';
+    const titleClass = classnames(
+      status ? 'va-modal-alert-title' : 'va-modal-title',
+    );
+    const contentClass = classnames({ 'va-modal-alert-text': status });
```
bkjohnson, line 239: Could we could simplify this `status`-based class logic with something like this in the CSS?

```css


.va-modal-body {
  margin-right: 2rem;
  overflow-wrap: break-word;
  padding: 2rem;
  word-break: break-word;
  word-wrap: break-word;
}

/* the va-modal-alert-body*/
:host([status]) .va-modal-body {
  display: table-cell;
  padding-left: 0;
  padding-right: 0;
  width: 100%;
  vertical-align: middle;
}

```

packages/web-components/src/components/va-modal/va-modal.tsx
```diff
@@ -0,0 +1,310 @@
...
+  /**
+   * This method traps the focus inside our web component, prevents scrolling outside
+   * the modal, and adds aria-hidden="true" to all elements outside the web component.
+   * Fires analytics event unless disableAnalytics is true.
+   */
+  private setupModal() {
+    const initialFocus = (this.el.querySelector(this.initialFocusSelector) ||
+      this.el.shadowRoot.querySelector(
+        this.initialFocusSelector,
+      )) as HTMLElement;
```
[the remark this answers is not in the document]
andresriveratoro: I think we would still need `isVisibleDirty` because we can't call `setupModal` or `teardownModal` until after `render`. We could do the `isVisibleDirty` check inside componentShouldUpdate, but Stencil docs state `...it's not good to rely on it to watch for prop changes, instead use the @Watch decorator for that.`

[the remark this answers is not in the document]
andresriveratoro: The query selector is to allow developers to be able to focus elements in either light or shadow DOM. I can remove this functionality, but I think it adds a lot of value. As a developer using this component, I expect to focus the element I want when providing a value for `initialFocusSelector`.

packages/web-components/src/components/va-modal/va-modal.tsx
```diff
@@ -0,0 +1,310 @@
...
+
+  /**
+   * Primary button text and action
+   */
+  @Prop() primaryButton?: any;
+
+  /**
+   * Secondary button text and action
+   */
+  @Prop() secondaryButton?: any;
```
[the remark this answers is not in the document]
andresriveratoro: I like that suggestion. I'll update it to that.

k80bowman (approved):

pdavies88 (approved):
Thanks for making the updates

packages/web-components/src/components/va-modal/va-modal.tsx
```diff
@@ -0,0 +1,336 @@
...
+  @Listen('click')
+  handleClick(e: MouseEvent) {
+    if (!this.clickToClose) return;
+
+    // event.target is always the shadow host
+    // event.composedPath()[0] returns the node clicked when shadow root is open
+    if (this.visible && e.composedPath()[0] === this.el) {
+      this.handleClose(e);
+    }
+  }
```
bkjohnson, line 129: Can you add a comment explaining that this will close the modal if the gray overlay is clicked?

packages/web-components/src/components/va-modal/va-modal.tsx
```diff
@@ -0,0 +1,336 @@
...
+    return (
+      <Host
+        aria-label={ariaLabel}
+        aria-modal="true"
+        class="va-modal"
+        role={ariaRole(status)}
+      >
+        <div class={wrapperClass} tabIndex={-1}>
+          {!hideCloseButton && (
+            <div ref={el => (this.closeButtonContainer = el as HTMLDivElement)}>
```
bkjohnson, line 283: Do we need this container? I'm not sure what it's doing, and it's not interactive so I don't see us focusing it.

packages/web-components/src/components/va-modal/va-modal.tsx
```diff
@@ -0,0 +1,336 @@
...
+          <div class={bodyClass}>
+            <div role="document">
+              {modalTitle && (
+                <h1 class={titleClass} tabIndex={-1}>
+                  {modalTitle}
+                </h1>
+              )}
+              <div class={contentClass}>
+                <slot></slot>
+              </div>
```
bkjohnson, line 304: The `contentClass` could be:

```css
.va-modal-alert-text {
  margin-top: 2rem;
  margin-bottom: 0;
}
```
I toggled the `margin-bottom` and didn't notice any difference - do we need that?

Also, could we get away with `margin-bottom: 2rem` on the `h1` just above this and not need the extra class and wrapping element?

bkjohnson (commented):
Small things. I'm mainly looking to see if we can reduce some of the nesting.

Also, here's a diff of what I did to test out the focus trap changes. The behavior seemed fine to me, but I'll let you be the judge:

<details>

<summary>Modal diff</summary>

```diff
diff --git a/packages/web-components/src/components/va-modal/va-modal.tsx b/packages/web-components/src/components/va-modal/va-modal.tsx
index 77af499..3073360 100644
--- a/packages/web-components/src/components/va-modal/va-modal.tsx
+++ b/packages/web-components/src/components/va-modal/va-modal.tsx
@@ -23,9 +23,8 @@ import { hideOthers, Undo } from 'aria-hidden';
   shadow: true,
 })
 export class VaModal {
-  alertActions: HTMLDivElement;
   closeButton: HTMLButtonElement;
-  closeButtonContainer: HTMLDivElement;
+  wrapper: HTMLDivElement;
   isVisibleDirty: boolean; // used to determine if visible has changed (workaround)
   focusTrap: FocusTrap;
   undoAriaHidden: Undo; // used to undo aria hidden on close
@@ -194,15 +193,12 @@ export class VaModal {
       this.el.shadowRoot.querySelector(
         this.initialFocusSelector,
       )) as HTMLElement;
-    this.focusTrap = createFocusTrap(
-      [this.el, this.alertActions, this.closeButtonContainer],
-      {
-        // trap is removed in teardownModal - disable escape key deactivating the focus trap
-        escapeDeactivates: false,
-        // the element we want to target immediately after opening modal
-        initialFocus: initialFocus || this.closeButton,
-      },
-    );
+    this.focusTrap = createFocusTrap([this.el, this.wrapper], {
+      // trap is removed in teardownModal - disable escape key deactivating the focus trap
+      escapeDeactivates: false,
+      // the element we want to target immediately after opening modal
+      initialFocus: initialFocus || this.closeButton,
+    });
     this.focusTrap.activate();
     disableBodyScroll(this.el);
     this.undoAriaHidden = hideOthers(this.el);
@@ -278,19 +274,21 @@ export class VaModal {
         class="va-modal"
         role={ariaRole(status)}
       >
-        <div class={wrapperClass} tabIndex={-1}>
+        <div
+          class={wrapperClass}
+          tabIndex={-1}
+          ref={(el: HTMLDivElement) => (this.wrapper = el)}
+        >
           {!hideCloseButton && (
-            <div ref={el => (this.closeButtonContainer = el as HTMLDivElement)}>
-              <button
-                aria-label={btnAriaLabel}
-                class="va-modal-close"
-                onClick={e => this.handleClose(e)}
-                ref={el => (this.closeButton = el as HTMLButtonElement)}
-                type="button"
-              >
-                <i aria-hidden="true" />
-              </button>
-            </div>
+            <button
+              aria-label={btnAriaLabel}
+              class="va-modal-close"
+              onClick={e => this.handleClose(e)}
+              ref={el => (this.closeButton = el as HTMLButtonElement)}
+              type="button"
+            >
+              <i aria-hidden="true" />
+            </button>
           )}
           <div class={bodyClass}>
             <div role="document">
@@ -305,10 +303,7 @@ export class VaModal {
             </div>
             {((primaryButtonClick && primaryButtonText) ||
               (secondaryButtonClick && secondaryButtonText)) && (
-              <div
-                class="alert-actions"
-                ref={el => (this.alertActions = el as HTMLDivElement)}
-              >
+              <div class="alert-actions">
                 {primaryButtonClick && primaryButtonText && (
                   <button
                     onClick={e => this.handlePrimaryButtonClick(e)}
```

</details>

We may also want to leave a comment saying that one focus trap is for the slotted children in the light dom, and the other(s) are for elements inside the shadow DOM.

andresriveratoro:
> Small things. I'm mainly looking to see if we can reduce some of the nesting.
> 
> Also, here's a diff of what I did to test out the focus trap changes. The behavior seemed fine to me, but I'll let you be the judge:
> 
> Modal diff
> We may also want to leave a comment saying that one focus trap is for the slotted children in the light dom, and the other(s) are for elements inside the shadow DOM.

I had to add these containers to be able to get the tabbing and alt-tabbing to work properly for all possible cases (hidden close button, focusable elements in the light DOM AND primary/secondary buttons, focusable elements in the light DOM AND primary/secondary buttons AND close button, etc). I ran into different issues where `tabbing` would work properly but `alt-tabbing` didn't work as expected. 

Here's a video using `this.el` and `this.wrapper` as the focus trap when we are showing the close button, passing in a focusable element (anchor) and also showing the primary and secondary buttons. If a focusable element is passed, I am unable to continue tabbing through the elements. I then try alt-tabbing and it gets stuck in the bottom elements. The order of the elements passed in the focus trap array matters, and I believe that's the reason for this weird behavior. My assumption  for this behavior when using `this.el` and `this.wrapper` is that once focus moves to the light DOM it tries to continue to the first focusable element (close button) in the next container element. Using alt-tab once focus is in the light DOM moves on to the last element (secondary button) of the previous container in the focus trap.


https://user-images.githubusercontent.com/36863582/157299873-ee69b5a7-2777-40ba-a862-923e18cdcea6.mov

packages/web-components/src/components/va-modal/va-modal.tsx
andresriveratoro: Good catch. I'll remove this.

andresriveratoro: We need this container for alt-tabbing to work properly. I tried setting the focus trap to `[this.el, this.alertActions, this.wrapper]` and that only works for tabbing. I'm assuming that when we try to alt-tab out of the light DOM it focuses the last focusable element of `this.wrapper` instead of the behavior that we want which is to focus the close button and complete the loop.

bkjohnson (approved):
Gotcha. Thanks for checking that out.

andresriveratoro:
Tested at 400% zoom:

https://user-images.githubusercontent.com/36863582/157743084-c78e7384-5d31-4778-a069-a14a77de950d.mov

Full width buttons @ max-width: 320px:
Screen Shot 2022-03-10 at 14 44 57

On hover color change:
Screen Shot 2022-03-10 at 14 50 20

k80bowman (approved):

PR #310, opened 2022-03-02
36950 Storybook Upgrade Webpack 5

pdavies88:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/36950

Guidance provided by https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#webpack-5
Resolution of build issue: https://github.com/storybookjs/storybook/issues/15336

## Testing done
Tested Storybook locally and running properly

## Screenshots
Screen Shot 2022-03-02 at 2 56 55 PM
Screen Shot 2022-03-02 at 2 57 14 PM

## Acceptance criteria
- [X]  Storybook and the dependencies mentioned are upgraded and verified as working properly

## Definition of done

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,5 +18,8 @@
     "node-fetch": "^2.6.7",
     "trim": "^0.0.3",
     "prism": "^1.27.0"
+  },
+  "dependencies": {
+    "webpack": "5"
   }
 }
--- a/packages/react-components/package.json
+++ b/packages/react-components/package.json
@@ -79,7 +79,7 @@
     "karma": "^6.3.14",
     "karma-chrome-launcher": "^3.1.0",
     "karma-mocha": "^2.0.1",
-    "karma-webpack": "^4.0.2",
+    "karma-webpack": "^5.0.0",
     "lint-staged": "^10.5.4",
     "mocha": "^4.0.1",
     "pinst": "^2.1.4",
--- a/packages/react-components/testing/karma.conf.js
+++ b/packages/react-components/testing/karma.conf.js
@@ -38,7 +38,7 @@ module.exports = function karmaConfig(config) {
         },
       },
       // target: 'node',
-      node: { global: true, fs: 'empty' },
+      node: { global: true },
     },
 
     // list of files / patterns to load in the browser
--- a/packages/storybook/package.json
+++ b/packages/storybook/package.json
@@ -19,12 +19,14 @@
     "@storybook/addon-docs": "^6.4.19",
     "@storybook/addon-essentials": "^6.4.19",
     "@storybook/addon-links": "^6.4.19",
+    "@storybook/builder-webpack5": "^6.4.19",
+    "@storybook/manager-webpack5": "^6.4.19",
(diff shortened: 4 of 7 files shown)


package.json
```diff
@@ -18,5 +18,8 @@
     "node-fetch": "^2.6.7",
     "trim": "^0.0.3",
     "prism": "^1.27.0"
+  },
```
pdavies88, line 21: Following Monorepo approach: https://github.com/storybookjs/storybook/issues/15336#issuecomment-869472304
Other method of `@storybook/react-docgen-typescript-plugin` disabling `typescript: { reactDocgen: false },` is not ideal as this was causing issues with our React Docs because it disables Set String and converts everything to Set Objects which will cause the docs to fail in certain instances

packages/react-components/testing/karma.conf.js
```diff
@@ -38,7 +38,7 @@ module.exports = function karmaConfig(config) {
         },
       },
       // target: 'node',
-      node: { global: true, fs: 'empty' },
```
pdavies88, line 41: Legacy webpack setting that is no longer used on node

packages/storybook/package.json
```diff
@@ -19,12 +19,14 @@
...
+    "@storybook/manager-webpack5": "^6.4.19",
     "@storybook/react": "^6.4.19",
     "@whitespace/storybook-addon-html": "^5.0.0",
     "babel-loader": "^8.2.2",
     "css-loader": "^5.2.4",
-    "node-sass": "^6.0.0",
+    "node-sass": "^7.0.0",
     "react-dom": "^17.0.2",
-    "sass-loader": "^10.2.1"
+    "sass-loader": "^12.4.0"
```
bkjohnson, line 30: I don't quite remember all of the details around the sass dependencies. Did we do some earlier work because we couldn't get to `sass-loader` version 12 without other issues?

packages/react-components/package.json
```diff
@@ -79,7 +79,7 @@
     "karma": "^6.3.14",
     "karma-chrome-launcher": "^3.1.0",
     "karma-mocha": "^2.0.1",
-    "karma-webpack": "^4.0.2",
+    "karma-webpack": "^5.0.0",
```
bkjohnson, line 82: Did the `dependencies` addition in the root `package.json` cause this to error?

packages/storybook/package.json
pdavies88: @bkjohnson yes so in order for us to get node-sass to v7.0.0, which is really what started this all because of the security issue, we need to use sass-loader v12.4.0 and to use sass-loader v12.4.0 we have to use webpack 5 in storybook

packages/react-components/package.json
pdavies88: That's correct upgrading to webpack 5 in the root caused karma to fail thus causing me to upgrade to karma 5 which supports webpack 5

k80bowman:
Is this change going to affect our published package at all?

k80bowman (approved):

PR #313, opened 2022-03-04
36964 Va Telephone Vanity Numbers

pdavies88:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/36964

## Testing done
Screen Shot 2022-03-15 at 11 35 55 AM

## Screenshots
Screen Shot 2022-03-15 at 11 37 55 AM

## Acceptance criteria
- [X] va-telephone accepts and displays vanity phone numbers

## Definition of done


packages/storybook/stories/va-telephone.stories.jsx
```diff
@@ -9,32 +9,70 @@ import {
...
 const Contacts = () => (
-  <Table
-    fields={fields}
-    data={Object.entries(contactsMap).map(c => ({
-      key: c[0],
-      value: <va-telephone contact={c[1].phoneNumber} />,
-      description: <div style={{ maxWidth: '30em' }}>{c[1].description}</div>,
-    }))}
-  />
+  <va-table>
```
pdavies88, line 58: Converted to use `va-table` instead of the other React Component

k80bowman: I know this is just in our storybook, but should we be using this before work is done? @bkjohnson

bkjohnson: I'd like to have this work be separate. I had planned to write a ticket [for this](https://dsva.slack.com/archives/CBU0KDSB1/p1645220464229119?thread_ts=1645202969.999719&cid=CBU0KDSB1) which would hopefully result in the `CONTACTS` object being exported from the `web-components` sub-package, so I'd also like to not redefine all of that in here just for the table.

pdavies88: @bkjohnson @k80bowman sounds good I just reverted. For future reference when that update is made here is the commit: https://github.com/department-of-veterans-affairs/component-library/pull/313/commits/06c4ca348295d8689d1b69c8ddb738ada132b8e6

packages/web-components/src/components/va-telephone/va-telephone.tsx
```diff
@@ -117,7 +192,7 @@ export class VaTelephone {
       </Fragment>
     ) : (
       <a
-        href={VaTelephone.createHref(contact, extension)}
+        href={VaTelephone.createHref(contact, extension, this.alphaPhoneNumber)}
```
bkjohnson, line 195: The test shows that `VETS` isn't included in the href, but it looks like we may be putting it in anyway. This makes me think that the browser is stripping out invalid characters from a `tel` link which is great, but let's not trust all browsers to do that. We should probably leave the `href` alone.

packages/web-components/src/components/va-telephone/test/va-telephone.e2e.ts
```diff
@@ -132,6 +132,18 @@ describe('va-telephone', () => {
...
   });
 
+  it('handles a vanity number as a contact prop', async () => {
+    const page = await newE2EPage();
+    await page.setContent('<va-telephone contact="877222VETS"></va-telephone>');
+
+    const link = await page.find('va-telephone >>> a');
+    expect(link.getAttribute('aria-label')).toEqual(
+      '1. 8 7 7. 2 2 2. V E T S. 8 3 8 7.',
+    );
```
bkjohnson, line 142: I don't think the current implementation puts the vanity letters into the aria-label. Is this something we want? cc @CourtneyChambers77 

https://github.com/department-of-veterans-affairs/component-library/blob/41015c4b2de355949e03ab0c450fc8e7535804b4/packages/react-components/src/components/Telephone/Telephone.jsx#L135-L139

https://github.com/department-of-veterans-affairs/component-library/blob/41015c4b2de355949e03ab0c450fc8e7535804b4/packages/react-components/src/components/Telephone/Telephone.jsx#L53-L65

image

packages/web-components/src/components/va-telephone/va-telephone.tsx
```diff
@@ -53,14 +101,23 @@ export class VaTelephone {
     num: string,
     extension: number,
     international: boolean = false,
+    alphaPhoneNumber: any,
```
bkjohnson, line 104: Let's try to use a more specific type where we can, or a more descriptive name. This name alone isn't enough to tell me if this param is a boolean, string, or maybe even a function.

bkjohnson (changes requested):

packages/web-components/src/components/va-telephone/va-telephone.tsx
pdavies88: @bkjohnson the href if left alone becomes href=tel:+1877222VETS which isn't a valid phone number which is why we also need to convert it in the createHref as well from letters to numbers

packages/web-components/src/components/va-telephone/va-telephone.tsx
pdavies88: I have updated to add the specific type of function

packages/web-components/src/components/va-telephone/test/va-telephone.e2e.ts
pdavies88: I have updated the PR so that the aria-label matches the image above and removes the extra "VETS"

packages/web-components/src/components/va-telephone/va-telephone.tsx
```diff
@@ -53,14 +101,23 @@ export class VaTelephone {
...
+      const vanity = /^[a-zA-Z]+$/.test(last4);
       formattedNum = `${area}-${local}-${last4}`;
       if (international) formattedNum = `+1-${formattedNum}`;
       if (extension) formattedNum = `${formattedNum}, ext. ${extension}`;
+      if (vanity) {
+        const alphaConversion = last4
+          .split('')
+          .map(digit => alphaPhoneNumber(digit))
+          .join('');
+        formattedNum = `+1-${formattedNum} (${alphaConversion})`;
```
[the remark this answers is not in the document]
pdavies88: Good catch - I have removed the extra +1

k80bowman (approved):

packages/web-components/src/components/va-telephone/va-telephone.tsx
```diff
@@ -71,20 +128,45 @@ export class VaTelephone {
...
+          .filter(n => n)
+          .map(chunk => chunk.split('').join(' '))
+          .join('. ')
+      : number
+          .split(/[^\d\w]+/)
+          .filter(n => n)
+          .map(chunk =>
+            chunk === 'ext' ? 'extension' : chunk.split('').join(' '),
+          )
+          .join('. ');
```
bkjohnson, line 144: Could we shift the conditional logic into defining the regex that gets used instead?

packages/web-components/src/components/va-telephone/va-telephone.tsx
```diff
@@ -71,20 +128,45 @@ export class VaTelephone {
...
+      vanityNumber =
+        contact.substring(0, contact.length - 4) +
+        last4
+          .split('')
+          .map(digit => alphaPhoneNumber(digit))
+          .join('');
+    }
+    const href = `tel:${
+      isN11 ? contact : vanity ? `+1${vanityNumber}` : `+1${contact}`
+    }`;
```
bkjohnson, line 169: This is a little hard for me to understand, and I may ask for some simplifications. But regardless, can we get [some unit tests added here](https://github.com/department-of-veterans-affairs/component-library/blob/3d956e06839b6bfddc1d00a64a222f3829209297/packages/web-components/src/components/va-telephone/test/va-telephone.spec.tsx) so that even if the function is tricky to understand, the tests describe what it can do, what output to expect from the input, etc.

packages/web-components/src/components/va-telephone/va-telephone.tsx
```diff
@@ -61,19 +68,24 @@ export class VaTelephone {
       formattedNum = `${area}-${local}-${last4}`;
       if (international) formattedNum = `+1-${formattedNum}`;
       if (extension) formattedNum = `${formattedNum}, ext. ${extension}`;
+      if (vanity) {
+        formattedNum = `${area}-${local}-${vanity} (${last4})`;
+      }
```
bkjohnson, line 73: This seems much more straightforward :+1:

packages/web-components/src/components/va-telephone/va-telephone.tsx
```diff
@@ -61,19 +68,24 @@ export class VaTelephone {
...
    * periods, e.g. "800-555-1212" becomes "8 0 0. 5 5 5. 1 2 1 2"
    */
-  static formatTelLabel(number: string): string {
-    return number
-      .split(/[^\d\w]+/)
+  static formatTelLabel(number: string, vanity: string): string {
+    const numberType = vanity
+      ? number.split(/[^\d]+/)
+      : number.split(/[^\d\w]+/);
+    return numberType
```
bkjohnson, line 88: Can we add the `vanity` param to the JSDoc above? It might be nice to have a short explanation why we're taking a string param but treating it like a boolean. I was a bit confused when I saw the string param only be used in the conditional part of a ternary, but then I realized that it's because we're passing in the string prop as the arg.

bkjohnson (commented):
Can we get unit tests on those static formatting functions?

pdavies88:
> Can we get unit tests to cover the changes in those static formatting functions?
@bkjohnson just updated with additional unit tests
Screen Shot 2022-03-15 at 12 46 32 PM

packages/web-components/src/components/va-telephone/va-telephone.tsx
pdavies88: @bkjohnson updated and added context into why/how we are using vanity here

k80bowman (approved):
The changes all look good to me. Great job!

packages/web-components/src/components/va-telephone/test/va-telephone.spec.tsx
```diff
@@ -4,48 +4,63 @@ describe('formatPhoneNumber', () => {
   const contact = '8885551234';
   const N11 = '911';
   const extension = 123;
+  const vanity = 'ABCD';
   it('formats a contact number with no extension', () => {
-    expect(VaTelephone.formatPhoneNumber(contact, null)).toBe('888-555-1234');
+    expect(VaTelephone.formatPhoneNumber(contact, null, null, null)).toBe('888-555-1234');
```
bkjohnson, line 9: It might be nice if the `vanity` param in `formatPhoneNumber` defaulted to `null` so we didn't have to do this, but that's nitpicking.

bkjohnson (approved):

PR #314, opened 2022-03-07
Replace breadcrumb item break with font awesome icons

longmd:
## Description
This PR address an inconsistency in breadcrumb break rendering. Occasionally when refreshing page or switch to new page, the breadcrumb renders a random set of characters instead of the chevron icon assigned in the stylesheet. This PR replaces that text chevron with a Font Awesome 5 chevron to provide consistent rendering.

## Original Issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/34502

## Screenshots
Screen Shot 2022-02-11 at 10 12 39 AM

Screen Shot 2022-03-07 at 10 57 59 AM

## Acceptance criteria
- [x] Breadcrumbs render consistently on every page load

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs-slot.css
+++ b/packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs-slot.css
@@ -7,9 +7,12 @@ va-breadcrumbs .va-breadcrumbs-li {
 }
 
 va-breadcrumbs .va-breadcrumbs-li::after {
-  content: ' › ';
+  content: '\f054';
   display: inline-block;
+  font: 900 .75rem 'Font Awesome 5 Free';
+  line-height: 1;
   padding: 0 0.56rem;
+  vertical-align: middle;
 }
 
 va-breadcrumbs .va-breadcrumbs-li:last-child::after {


k80bowman:
@longmd can you give us more details about when the characters appear? We need to be able to replicate this. For some of these things, it might be more helpful if you filled out a [bug ticket](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/new/choose) for us. That has a template that should prompt you for the information we need.

longmd:
@k80bowman , this one is completely random. They basically appear and disappear when they want to on either a page refresh or when navigating between pages. I haven't been able to come up with specific behavior to trigger the characters. You could refresh 80x in one sitting and never get them to appear, and then in another sitting you could just refresh once and they are there. 

I picked this ticket up from an engineer who is no longer with the team, but this was the desired solution that was discussed in the original issue noted above. I'm happy to submit a bug ticket if you feel that would be more helpful, but I don't have specific steps for you to replace because of the randomness of this.

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs-slot.css
```diff
@@ -7,9 +7,12 @@ va-breadcrumbs .va-breadcrumbs-li {
 }
 
 va-breadcrumbs .va-breadcrumbs-li::after {
-  content: ' › ';
+  content: '\f054';
   display: inline-block;
+  font: 900 .75rem 'Font Awesome 5 Free';
+  line-height: 1;
   padding: 0 0.56rem;
+  vertical-align: middle;
```
bkjohnson, line 15: I think using fontawesome may be unnecessary for this. Did you look at using the [encoded version of the unicode character](https://unicode-table.com/en/203A/) to see if that would resolve the problem?

image

longmd: @bkjohnson , I had not. Just ran with the solution that was proposed in the issue ticket by the previous lead engineer on this project. If you'd prefer we just use the unicode, I'm good with that and will make the change.

k80bowman (approved):

PR #318, opened 2022-03-15
Add sorting functionality to va-table component

bkjohnson:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/36938

Adds sorting functionality to `<va-table>` (originally added in https://github.com/department-of-veterans-affairs/component-library/pull/291). Due to the aria-label on the sort button this implementation of a sortable table has no initial "unsorted" state - it is either sorted ascending or descending.

I used these quicksort resources while working on this implementation:

- https://en.wikipedia.org/wiki/Quicksort#Hoare_partition_scheme
- https://www.guru99.com/quicksort-in-javascript.html

## Testing done

- Storybook :eyes: 
- Unit tests on `quicksort` function
- E2E tests on some of the sorting features (adding the button)

## Screenshots

Rows are sorted initially despite their order as original children:

Peek 2022-03-18 15-32

## Acceptance criteria
- [ ]

## Definition of done


packages/web-components/src/components/va-table/va-table-row.css
```diff
@@ -23,6 +23,10 @@
   padding: 1rem 1.5rem;
 }
 
+::slotted(*:empty):before {
+  content: '---';
+}
```
bkjohnson, line 28: This isn't related to sorting at all - it's just a small improvement.

packages/web-components/src/components/va-table/va-table.tsx
```diff
@@ -53,11 +78,6 @@ export class VaTable {
           cell.classList.add(alignment[colNum]);
         }
 
-        // Indicate missing cell content with "---"
-        if (cell.innerText === '') {
-          cell.innerText = '---';
-        }
```
bkjohnson, line 59: So we don't have to do the same thing in JS.

packages/web-components/src/global/main.css
```diff
@@ -6,3 +6,11 @@ va-process-list li :is(h2, h3, h4, h5) {
...
   padding-top: 0.3rem;
 }
+
+va-table > va-table-row[slot] button {
+  display: inline-flex;
+  align-items: center;
+}
+va-table > va-table-row[slot] button svg {
+  width: 2.4rem;
+}
```
k80bowman, line 16: Could we put this in a va-table-slot.css file and import it the way we do with the breadcrumbs slot CSS above?

packages/web-components/src/utils/dom.spec.ts
```diff
@@ -0,0 +1,47 @@
+import { quicksort } from './dom';
+
+describe('quicksort', () => {
```
k80bowman, line 3: I'd like to see some more tests here. What happens when the text inside the cells is a number? What if you have a mix of cells with letters and numbers? What happens if the text includes a symbol or starts with a symbol?

packages/web-components/src/global/main.css
bkjohnson: Yeah, that's on my todo list.

packages/web-components/src/components/va-table/va-table.tsx
```diff
@@ -1,5 +1,8 @@
-import { Component, Element, Host, Prop, h } from '@stencil/core';
+import { Component, Element, Host, Prop, State, h } from '@stencil/core';
 import { isNumeric } from '../../utils/utils';
+import ascendingIcon from '../../assets/sort-arrow-up.svg?format=text';
+import descendingIcon from '../../assets/sort-arrow-down.svg?format=text';
```
bkjohnson, line 4: I'd like to explore moving this kind of thing into a `<va-icon>` component in the future, where a prop can be used to identify the icon name for a dynamic import.

Also, the `?format=text` was the only way I was able to import the raw svg instead of the encoded string.

packages/web-components/src/components/va-table/va-table.tsx
```diff
@@ -37,6 +55,30 @@ export class VaTable {
...
+      button.classList.add(
+        'vads-u-padding--0',
+        'vads-u-margin--0',
+        'vads-u-text-decoration--none',
+        'vads-u-font-weight--bold',
+        'vads-u-background-color--gray-lightest',
+        'vads-u-color--base',
+      );
+      button.innerHTML = `${columns[this.sortColumn]}${icon}`;
+      button.onclick = this.handleSort.bind(this);
```
bkjohnson, line 74: I'm doing all of this stuff the old way instead of with JSX because this button will exist in the light DOM.

packages/web-components/src/components/va-table/va-table.tsx
```diff
@@ -66,6 +103,29 @@ export class VaTable {
...
 
+  componentDidUpdate() {
+    // Update the sort icon
+    if (this.sortColumn >= 0) {
+      const icon = this.sortAscending ? descendingIcon : ascendingIcon;
+      const button = this.el.querySelector('va-table-row[slot] button');
+      const node = button.children[0];
+      node.outerHTML = icon;
+    }
+  }
```
bkjohnson, line 114: The `sortAscending` state variable gets updated on sort, so this function will run after a sort.

packages/web-components/src/components/va-table/va-table.tsx
```diff
@@ -66,6 +103,29 @@ export class VaTable {
...
+  private handleSort(): void {
+    const cellSelector = row => row.children[this.sortColumn].textContent;
+    // Starting at index 1 to skip the header row
+    const sortingFunc = this.sortAscending ? quicksort : reverseQuicksort;
+    sortingFunc(
+      this.el.childNodes,
+      1,
+      this.el.childNodes.length - 1,
+      cellSelector,
+    );
```
bkjohnson, line 125: Here we call quicksort with a live `NodeList`. It happens in place because I didn't want to have to develop our own virtual DOM or other DOM abstraction.

bkjohnson (commented):
I wasn't able to add all the tests as I wanted, but I think all the functionality is here.

pdavies88 (approved):

packages/web-components/src/utils/dom.spec.ts
k80bowman: These are great, thanks for adding all the tests!

packages/web-components/src/utils/dom.ts
```diff
@@ -0,0 +1,78 @@
...
+function swapChildren(rows: NodeList, i: number, j: number): void {
+  const parent = rows[0].parentNode;
+  const one = rows[j];
+  const two = rows[i];
+  const temp = one.nextSibling;
+  parent.insertBefore(one, two);
+  parent.insertBefore(two, temp);
+}
+
+function partition(
```
k80bowman, line 13: Would you mind adding a comment explaining this function? It's the only one in the file without a description.

packages/web-components/src/utils/dom.ts
```diff
@@ -0,0 +1,78 @@
+/**
+ * Swap the DOM elements at each index
+ */
+function swapChildren(rows: NodeList, i: number, j: number): void {
```
k80bowman, line 4: Could you use some more descriptive param names? It's not clear what `i` and `j` are.

packages/web-components/src/utils/dom.ts
```diff
@@ -0,0 +1,78 @@
...
+  rows: NodeList,
+  lo: number,
+  hi: number,
+  selector: Function,
+  comparator: Function,
+): number {
+  const pivot = selector(rows[Math.floor((lo + hi) / 2)]);
+
+  let i = lo;
+  let j = hi;
```
k80bowman, line 23: Could these variable names be more descriptive as well, please?

packages/web-components/src/utils/dom.ts
```diff
@@ -0,0 +1,78 @@
...
+export function quicksort(
+  rows: NodeList,
+  lo: number,
+  hi: number,
+  selector = row => row.textContent,
+  comparator = (a, b) => a < b,
+): void {
+  if (lo >= hi || lo < 0) return;
+
+  const p = partition(rows, lo, hi, selector, comparator);
```
k80bowman, line 61: Once again, could you please use a more descriptive variable name? Sorry, I'm not a fan of the single-letter variable and param names. They make it harder for someone else to come in later and figure out what's going on.

k80bowman (changes requested):
Just a few requests, but otherwise this looks great. Thanks!

packages/web-components/src/utils/dom.ts
```diff
@@ -0,0 +1,78 @@
+/**
```
k80bowman, line 1: I wonder if we should name this file something more descriptive. I'm assuming you named it `dom.ts` because it has to do with sorting dom nodes, but maybe it should be `sort` or something similar. I'd rather have more files with specific names/purposes than fewer files with generic names/purposes.

bkjohnson: Yeah, I named it that way because of the DOM manipulations and to distinguish it from other less-specific sorting functions. That's not something we need to worry about for now though, so I'll rename it to `dom-sort`.

k80bowman (approved):

bkjohnson:
The sorting still doesn't work on IE11. Based on the discovery I've done so far, it would take a significant amount of work to resolve that due to the way the polyfills work and the difference between `.children` and `.childNodes`.

PR #322, opened 2022-03-16
Add slot to va-text-input

andresriveratoro:
## Chromatic

https://35734-va-text-input-label--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/35734

## Testing done
E2E and Storybook

## Screenshots
Screen Shot 2022-03-31 at 18 08 20

## Acceptance criteria
- [x] Add ability to pass in HTMLElements to be used for label

## Definition of done


packages/storybook/stories/va-text-input.stories.jsx
```diff
@@ -22,9 +22,9 @@ const defaultArgs = {
   'autocomplete': false,
   'enable-analytics': false,
   'required': false,
-  'error': null,
-  'maxlength': null,
-  'value': null,
+  'error': undefined,
+  'maxlength': undefined,
+  'value': undefined,
```
andresriveratoro, line 27: These were updated to `undefined` to prevent them from showing up in Storybook's `Show Code` as `error={null}`.

packages/web-components/src/components/va-text-input/va-text-input.tsx
```diff
@@ -40,12 +41,20 @@ export class VaTextInput {
...
    */
-  @Prop() inputmode?: string = '';
+  @Prop() inputmode?:
+    | 'decimal'
+    | 'email'
+    | 'numeric'
+    | 'search'
+    | 'tel'
+    | 'text'
+    | 'url';
```
andresriveratoro, line 51: The type was updated to have better documentation in Storybook.

packages/web-components/src/components/va-text-input/va-text-input.tsx
```diff
@@ -40,12 +41,20 @@ export class VaTextInput {
...
 
   /**
    * The type attribute.
    */
-  @Prop() type?: string = 'text';
+  @Prop() type?: 'email' | 'number' | 'search' | 'tel' | 'text' | 'url' =
+    'text';
```
andresriveratoro, line 57: See comment for `inputmode`.

packages/web-components/src/components/va-text-input/va-text-input.tsx
```diff
@@ -140,20 +151,20 @@ export class VaTextInput {
...
-        )}
+        <label htmlFor="inputField">
+          {this.label && (
+            <Fragment>
+              {this.label}{' '}
+              {this.required && <span class="required">(*Required)</span>}
+            </Fragment>
+          )}
+          <slot></slot>
+        </label>
```
andresriveratoro, line 167: This was done so engineers have the option to use a label and insert content between the label and input field. Engineers can also bypass the label property to have finer control over what is rendered above the input field.

pdavies88:
> I can remove the Experimental story if we don't want that example here.

I mean I think we should have a story I don't love the experimental name was thinking more along the lines of "HTML within Text Input" just like what it is. I mean at the core of this ticket I would much rather have developers put this HTML before the component than inside it but thats not part of the request lol.

pdavies88 (approved):

k80bowman:
Just to make sure I understand what you've done here, the implementation on this is changing. The label prop now only accepts text, but engineers can also skip use of the label prop and insert an HTML element instead. Or they could have a label and an HTML element. Is this correct?

If so, how will this affect implementations of this component that are already in place?

pdavies88:
Ah gotcha yeah maybe we should put a link to that in the notes at the top or change the name. Cause without that context I was kinda like this name is odd cause we are recommending this in our official storybook so it makes it like not experimental at that point.

pdavies88:
@k80bowman @andresriveratoro side note if we implement this should we remove this page from the docs: https://design.va.gov/experimental-design/hint-text#within-an-additional-info-component now that implementation would be part of the official code base?

andresriveratoro:
I only included the story because without it engineers won't have an example using the slot.

Here's a `TextInput` label used in vets-website:
```jsx
  <>
    <strong>Name of issue</strong>
    <span className="form-required-span">(*Required)</span>
    <p className="vads-u-font-weight--normal label-description">
      You can only add an issue that you’ve already received a VA decision
      notice for.
    </p>
  </>
```

Using `va-text-input`:
```jsx
<va-text-input label="Name of issue" {...otherProps}>
  <p className="vads-u-font-weight--normal label-description">
    You can only add an issue that you’ve already received a VA decision notice for.
  </p>
</va-text-input>
```

Should I update the story to this and rename it?

pdavies88:
IMO yes we should rename and provide a concrete example that a developer on another team could run with and allow them to move away from the react component cause that's the ultimate goal.

k80bowman:
I'm a little nervous about implementing an experimental design in the design system without consulting the DSC. Let's hold off on merging this until I can talk with Carol and make sure this is OK. We may need to bring it up with the DSC before merging.

I also think we should rename the story and provide some more guidance on when and how this feature should be used.

andresriveratoro:
I just realized that last example is also under our Experimental Design, but it is being used in `vets-website`.

packages/web-components/src/components/va-text-input/va-text-input.tsx
bkjohnson: This seems fine, holding off on approval for an a11y :+1:

packages/web-components/src/components/va-text-input/test/va-text-input.e2e.ts
```diff
@@ -11,6 +11,9 @@ describe('va-text-input', () => {
     expect(element).toEqualHtml(`
       <va-text-input class="hydrated">
         <mock:shadow-root>
+          <label for="inputField">
+            <slot></slot>
+          </label>
```
bkjohnson, line 16: This is in the `renders without a label` test - should this be happening?

andresriveratoro: I've updated this test to use a `label` prop and added a separate test when the slot is used as a label.

andresriveratoro: I updated the test because a developer would be providing either a `label` prop string or children to be used inside the label tag. The original render test was updated to cover the case where a label prop is used.

bkjohnson (commented):
We should probably have @CourtneyChambers77 take a look at this.

andresriveratoro:
I updated the default story to make it easier for @CourtneyChambers77 to do an accessibility test using slot content in the label. I'll revert this before merging this PR.

andresriveratoro:
The story changes have been reverted. Having the slot inside the label is not a problem as long as a few guidelines are followed. According to the [spec](https://html.spec.whatwg.org/multipage/forms.html#the-label-element), acceptable children are "[Phrasing content](https://html.spec.whatwg.org/multipage/dom.html#phrasing-content-2), but with no descendant [labelable elements](https://html.spec.whatwg.org/multipage/forms.html#category-label) unless it is the element's [labeled control](https://html.spec.whatwg.org/multipage/forms.html#labeled-control), and no descendant [label](https://html.spec.whatwg.org/multipage/forms.html#the-label-element) elements." 

Using the experimental design example in this way would be going against the spec because `va-additional-info` is not a form control and it contains a `div`.

The goal of this issue was to give developers more control over what is displayed and how it's displayed before the input field. If we move the `slot` outside of the label **element** and engineers don't provide a `label` **prop**, the responsibility of adding the label element and any other markup falls to them but they would have all the flexibility they would need. If we leave the slot inside the label, engineers could potentially insert elements that go against the spec. I prefer the first option, but would love to know your thoughts @bkjohnson @pdavies88 @k80bowman @mendacitees.

pdavies88:
@andresriveratoro I would agree with you to err on the side of caution and rather have them do the html prior instead of going against the spec. However I do think that would invalidate the ticket and the experimental design so there's that to think about as well. Maybe we want to add it to the Monday meeting

andresriveratoro:
It looks like we have a critical accessibility error when passing in `<label htmlFor="inputField">` from the light DOM - the input is never linked to this label.  

I modified the code to this:
```jsx
{this.label && (
  <label htmlFor="inputField">
    <Fragment>
      {this.label}{' '}
      {this.required && <span class="required">(*Required)</span>}
    </Fragment>
  </label>
)}
<slot></slot>
```

The experimental design would work with this change when a label is provided. 
Screen Shot 2022-03-31 at 10 46 45

packages/web-components/src/components/va-text-input/va-text-input.tsx
```diff
@@ -140,20 +151,20 @@ export class VaTextInput {
...
+    const inputMode = this.inputmode ? this.inputmode : null; // Null so we don't add the attribute if we have an empty string
     const type = this.getInputType();
 
     return (
       <Host>
         {this.label && (
           <label htmlFor="inputField">
-            {this.label}{' '}
-            {this.required && <span class="required">(*Required)</span>}
+            <Fragment>
```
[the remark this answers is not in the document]
andresriveratoro: Removed

andresriveratoro:
@pdavies88 I've updated the story name to 'With Hint Text'.

bkjohnson (approved):
:shipit:

PR #325, opened 2022-03-17
Add va-search web component

andresriveratoro:
## Chromatic

https://37972-va-search--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/37972

## Acceptance criteria
- [x] Search input and button are implemented and in Storybook
    - [x] Indicate that component is not ready for use
- [x] E2E/unit tests are written
- [x] Add Chromatic link to [#37973](https://github.com/department-of-veterans-affairs/va.gov-team/issues/37973), assign ticket to Courtney, and move ticket to Current Sprint column

## Definition of done


packages/web-components/src/components/va-search/va-search.tsx
```diff
@@ -0,0 +1,43 @@
+import { Component, Host, h, Prop, Event, EventEmitter } from '@stencil/core';
+
+@Component({
+  tag: 'va-search',
+  styleUrl: 'va-search.css',
+  shadow: true,
+})
+export class VaSearch {
+  @Prop() action: string;
```
pdavies88, line 9: Missing the jsdoc comments could prop just copy the definitions: https://www.w3schools.com/tags/tag_form.asp

packages/storybook/stories/va-search.stories.jsx
```diff
@@ -0,0 +1,41 @@
...
+    label={label}
+    method={method}
+    onSubmitEvent={submitEvent}
+  />
+);
+
+const defaultArgs = {
+  action: undefined,
+  label: 'Search',
+  method: undefined,
```
pdavies88, line 33: Thoughts on doing method as a dropdown option since it'd only be get or post

packages/web-components/src/components/va-search/va-search.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <Host>
+        <form
+          id="va-search-form"
+          action={action}
+          method={method}
+          onSubmit={(event: KeyboardEvent | MouseEvent) =>
+            this.handleSubmitEvent(event)
+          }
+        >
+          <input id="va-search-input" aria-label={label} type="text"></input>
```
pdavies88, line 35: Should we use the `va-text-input` component

pdavies88:
@andresriveratoro general note not sure if its happening for you but when I type into the input field and use either the F or S key the zoom full screen option in storybook toggles for me on the search component

packages/web-components/src/components/va-search/va-search.tsx
andresriveratoro: I'm against using `va-text-input` here because the native HTML element does everything we need for this search component. `va-text-input` also uses `aria-describedby` (I think we would have to create a visually hidden element with text to work around this), and we wouldn't use the error or label. 

We could make `va-text-input` more flexible and revisit this later.

andresriveratoro:
> @andresriveratoro general note not sure if its happening for you but when I type into the input field and use either the F or S key the zoom full screen option in storybook toggles for me on the search component

This happens to me too. We could disable Storybook shortcuts to fix this.

packages/web-components/src/components/va-search/va-search.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  @Prop() method?: 'GET' | 'POST' | 'dialog';
+
+  @Event({ bubbles: true, composed: true })
+  submitEvent: EventEmitter;
+
+  private handleSubmitEvent(event: KeyboardEvent | MouseEvent) {
+    event.preventDefault();
+    this.submitEvent.emit(event);
+  }
```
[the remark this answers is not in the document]
andresriveratoro: I'm trying this in Storybook, but the function passed in isn't being fired.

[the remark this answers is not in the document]
bkjohnson: That's my bad - I got `bubbles` and `composed` confused again. According to [some documentation](https://javascript.info/shadow-dom-events#event-composed):

> Most events successfully bubble through a shadow DOM boundary. There are few events that do not.
>
> This is governed by the `composed` event object property.

Since `composed` isn't true for the `submit` event that the DOM fires, we may have to [create and fire](https://developer.mozilla.org/en-US/docs/Web/Events/Creating_and_triggering_events) a synthetic [`SubmitEvent`](https://developer.mozilla.org/en-US/docs/Web/API/SubmitEvent/SubmitEvent) if we want to rely on the native `onSubmit` handler.

packages/web-components/src/components/va-search/va-search.css
```diff
@@ -0,0 +1,53 @@
...
+  max-width: none;
+  min-width: 24.8rem;
+  padding: 1.2rem 0.8rem;
+}
+
+#va-search-input:focus {
+  outline: 2px solid var(--color-gold-light);
+  outline-offset: 2px;
+  z-index: 1;
+}
```
bkjohnson, line 27: Could we add the `input` selector to `focusable.css` instead?

https://github.com/department-of-veterans-affairs/component-library/blob/3c44d4a5e3b134c8d8be692fde19caa9e708cdda/packages/web-components/src/mixins/focusable.css#L3-L9

packages/web-components/src/components/va-search/va-search.tsx
```diff
@@ -0,0 +1,84 @@
...
+          aria-controls={ariaControls}
+          aria-label={label}
+          type="text"
+          onChange={handleOnChange}
+        />
+        <button id="va-search-button" type="submit" aria-label={label}>
+          <i aria-hidden="true" class="fa fa-search" />
+          {buttonText && <span id="va-search-button-text">{buttonText}</span>}
+        </button>
+        {/* </form> */}
```
bkjohnson, line 80: Here's an example of what I'm talking about:

```jsx
        <form
          id="va-search-form"
          action={action}
          method={method}
          onSubmit={e => {
            /***********/
            // We prevent the postback and create our own submit event 
            // with `composed` and `bubbles` set to true
            console.log('Inner submit handler');
            e.preventDefault();
            const se = new SubmitEvent('submit', {
              bubbles: true,
              composed: true,
            });
            this.el.dispatchEvent(se);
            console.log('Submit event', se);
            /***********/
          }}
        >
          <input id="va-search-input" aria-label={label} type="text"></input>
          <button id="va-search-button" type="submit" aria-label={label}>
            <i aria-hidden="true" class="fa fa-search" />
          </button>
        </form>
      </Host>
    );
  }
}
```

packages/storybook/stories/va-search.stories.jsx
```diff
@@ -0,0 +1,49 @@
...
+  <va-search
+    action={action}
+    label={label}
+    // method={method}
+    onBlur={() => console.log('BLUR')}
+    onChange={() => console.log('CHANGE')}
+    onClick={() => console.log('CLICK')}
+    onKeyDown={() => console.log('KEYDOWN')}
+  />
+);
```
bkjohnson, line 36: And in storybook we can add a handler for the default event like this:

```jsx
const Template = ({ action, label, method, submit }) => (
  <va-search
    action={action}
    label={label}
    method={method}
    onSubmit={e => {
      console.log('Outer submit handler');
      console.log(e);
    }}
  />
);
```

bkjohnson (commented):
Adding some examples for the native event stuff.

packages/storybook/stories/va-search.stories.jsx
andresriveratoro: This isn't working for `onChange`.

packages/web-components/src/components/va-search/va-search.tsx
```diff
@@ -0,0 +1,216 @@
...
+          aria-expanded={inputAriaExpanded}
+          aria-haspopup="listbox"
+          aria-label={label}
+          autocomplete="off"
+          onBlur={handleInputBlurEvent}
+          onChange={handleInputChangeEvent}
+          onClick={handleInputClickEvent}
+          onFocus={handleInputFocusEvent}
+          onKeyDown={handleInputKeyDownEvent}
+          role="combobox"
```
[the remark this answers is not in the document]
andresriveratoro: Updated to retrieve the attribute value from `va-search`

packages/web-components/src/components/va-search/va-search.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+  /**
+   * Identifies the element (or elements) whose contents or presence are controlled by the element on which this attribute is set
+   */
+  @Prop() inputAriaControls?: string;
+
+  /**
+   * Adds an aria-label attribute to the input and button
+   */
+  @Prop() inputAriaExpanded: string;
```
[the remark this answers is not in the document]
andresriveratoro: Updated to retrieve attribute values instead of using props

packages/web-components/src/components/va-search/va-search.tsx
```diff
@@ -0,0 +1,216 @@
...
+          aria-activedescendant={inputAriaActiveDescendant}
+          aria-autocomplete="none"
+          aria-controls={inputAriaControls}
+          aria-expanded={inputAriaExpanded}
+          aria-haspopup="listbox"
+          aria-label={label}
+          autocomplete="off"
+          onBlur={handleInputBlurEvent}
+          onChange={handleInputChangeEvent}
+          onClick={handleInputClickEvent}
```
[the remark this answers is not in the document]
andresriveratoro: Good catch. I've removed the input onClick custom event.

packages/storybook/stories/va-search.stories.jsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { VaSearch } from '@department-of-veterans-affairs/web-components/react-bindings';
+import { getWebComponentDocs, propStructure } from './wc-helpers';
+import { generateEventsDescription } from './events';
+
```
pdavies88, line 5: @andresriveratoro Looks like we are missing: `VaSearch.displayName = 'VaSearch';`

packages/storybook/stories/va-search.stories.jsx
```diff
@@ -0,0 +1,40 @@
...
+    onInputChangeEvent={e => console.log(e, 'INPUT CHANGE FIRED')}
+    onInputClickEvent={e => console.log(e, 'INPUT CLICK FIRED')}
+    onInputKeyDownEvent={e => console.log(e, 'INPUT KEYDOWN FIRED')}
+    onButtonClickEvent={e => console.log(e, 'BUTTON CLICK FIRED')}
+    onButtonFocusEvent={e => console.log(e, 'BUTTON FOCUS FIRED')}
+    onButtonKeyDownEvent={e => console.log(e, 'BUTTON KEYDOWN FIRED')}
+  />
+);
+
+const defaultArgs = {
```
pdavies88, line 32: @andresriveratoro do we want to tie in: button-text, hide-button, and input-value props to this story

packages/web-components/src/components/va-search/va-search.tsx
```diff
@@ -0,0 +1,183 @@
...
+      hideButton,
+      inputValue,
+      label,
+    } = this;
+
+    return (
+      <Host>
+        <input
+          id="va-search-input"
+          aria-activedescendant={this.el.getAttribute('aria-activedescendant')}
```
pdavies88, line 152: @andresriveratoro correct me if I am wrong here but I am assuming these are being added so if someone wants to do like:
```
<va-search aria-activedescendant="XYZ" />
```
It would get added to the input element in the light Dom because otherwise the aria element wouldn't be hit since it would exist in the shadow Dom. Is that correct?

pdavies88:
@andresriveratoro so do we want to do something about the 'S' and 'F' keys triggering the window sizing in SB. I feel like if we go live with this and someone happens upon it we are going to get a ticket to fix it.

packages/storybook/stories/va-search.stories.jsx
```diff
@@ -0,0 +1,40 @@
...
+    },
+  },
+};
+
+const Template = ({ label }) => (
+  <VaSearch
+    label={label}
+    onInputBlurEvent={e => console.log(e, 'INPUT BLUR FIRED')}
+    onInputChangeEvent={e => console.log(e, 'INPUT CHANGE FIRED')}
+    onInputClickEvent={e => console.log(e, 'INPUT CLICK FIRED')}
```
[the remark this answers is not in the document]
andresriveratoro: Good catch. This event was removed in the web component. I've updated the story to remove it as well.

packages/web-components/src/components/va-search/va-search.tsx
andresriveratoro: Your example of how to use it is correct. Previously, I had created props for this functionality, but it wasn't required because we can grab the value from the wrapper `<va-search>` and use it on the `input`.

This input element is in the shadow DOM. Are you asking if `aria-activedescendant` will work if the element specified exists in the light DOM?

packages/storybook/stories/va-search.stories.jsx
andresriveratoro: Thanks, I forgot to add these in.

packages/storybook/stories/va-search.stories.jsx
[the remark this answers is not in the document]
andresriveratoro: Storybook shortcuts have been disabled to allow users to input text without triggering shortcuts that may frustrate users.

packages/web-components/src/components/va-search/va-search.tsx
andresriveratoro: That's correct. In order to get around that we have to grab the value from the host and explicitly set it on whatever element requires it.

packages/web-components/src/components/va-search/va-search.tsx
```diff
@@ -0,0 +1,183 @@
...
+          aria-autocomplete="none"
+          aria-controls={this.el.getAttribute('aria-controls')}
+          aria-expanded={this.el.getAttribute('aria-expanded')}
+          aria-haspopup={this.el.getAttribute('aria-haspopup')}
+          aria-label={label}
+          autocomplete="off"
+          onBlur={handleInputBlurEvent}
+          onChange={handleInputChangeEvent}
+          onFocus={handleInputFocusEvent}
```
pdavies88, line 161: @andresriveratoro should we remove handleInputFocusEvent since we aren't using it in Storybook

pdavies88 (approved):
Everything looks good thanks for the updates. Only last thing was my comment about: `handleInputFocusEvent` wanted to see if we should have it in SB or if it should be removed from the web component since it seems like `blur` event handles when we tab into the input element.

packages/web-components/src/components/va-search/va-search.tsx
```diff
@@ -0,0 +1,183 @@
...
+   */
+  @Prop() label: string = 'Search';
+
+  // Input Event Handlers
+  private handleInputBlurEvent = (event: FocusEvent) => {
+    this.inputBlurEvent.emit(event);
+  };
+
+  private handleInputChangeEvent = (event: Event) => {
+    // const target = event.target as HTMLInputElement;
```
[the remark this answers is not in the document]
andresriveratoro: I was going to pass in `event.target.value` to this custom event, but opted to pass in the original event in `detail`. Handlers will have to be updated to point to `detail.path[0].value` to grab the input's value.

bkjohnson (approved):
This looks good :+1:

PR #331, opened 2022-03-21
Additional Info Disable Blue Line Prop

pdavies88:
## Chromatic

https://additional-info-blue-line-prop--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/37751

## Testing done / Screenshots
Screen Shot 2022-03-21 at 9 17 52 AM

## Acceptance criteria
- [X] Adding the prop and updating the documentation to include information about the prop in Storybook

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-additional-info/va-additional-info.css
+++ b/packages/web-components/src/components/va-additional-info/va-additional-info.css
@@ -3,14 +3,9 @@
   display: block;
 }
 
-:host([disable-border='true']) a[aria-expanded='true'] ~ #info,
-:host([disable-border='']) a[aria-expanded='true'] ~ #info {
-  padding-left: 0;
-  border-left: 0;
-  border-left-color: transparent;
-}
-
-a[aria-expanded='true'] ~ #info {
+/* Add left side blue border if disable border prop disabled or not present */
+:host(:not([disable-border])) a[aria-expanded='true'] ~ #info,
+:host([disable-border='false']) a[aria-expanded='true'] ~ #info {
   padding-left: calc(2rem - 7px);
   border-left: 7px solid transparent;
   border-left-color: var(--color-primary-alt-dark);


packages/web-components/src/components/va-additional-info/va-additional-info.tsx
```diff
@@ -99,8 +104,10 @@ export class VaAdditionalInfo {
             <i class="fa-angle-down" role="presentation" />
           </span>
         </a>
-
-        <div id="info" class={this.open ? 'open' : 'closed'}>
+        <div
+          id={this.disableBorder ? 'info-no-border' : 'info'}
```
bkjohnson, line 108: What does changing the `id` do for us?

bkjohnson (commented):
It's unclear how this change solves the problem, could you provide some explanation?

packages/web-components/src/components/va-additional-info/va-additional-info.tsx
pdavies88: @bkjohnson it makes it so the extra CSS that creates the padding left and blue border doesn't get added in: https://github.com/department-of-veterans-affairs/component-library/blob/f13301ccda884d93e8fcd68ee36c24bf4a8b1c70/packages/web-components/src/components/va-additional-info/va-additional-info.css#L5-L9

packages/core/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@department-of-veterans-affairs/component-library",
   "description": "VA.gov component library. Includes React and web components.",
-  "version": "8.0.0",
+  "version": "8.0.1",
```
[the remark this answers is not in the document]
pdavies88: I updated it to minor instead of patch

pdavies88:
@bkjohnson without the CSS here: https://github.com/department-of-veterans-affairs/component-library/blob/f13301ccda884d93e8fcd68ee36c24bf4a8b1c70/packages/web-components/src/components/va-additional-info/va-additional-info.css#L5-L9 the blue border and padding will not be added in. So adding the prop and enabling it to true changes the name of the id so the CSS rule that adds in the border and padding will not match therefore the border and padding will not be displayed giving us the correct look on the component

packages/web-components/src/components/va-additional-info/va-additional-info.tsx
bkjohnson: I feel like changing the `id` isn't super obvious. Are we able to accomplish this with an attribute selector instead?

pdavies88: @bkjohnson I have reworked to use the :host([disable-border='false']) selector to differentiate when the css should be added in

packages/web-components/src/components/va-additional-info/va-additional-info.css
```diff
@@ -2,7 +2,8 @@
 :host {
   display: block;
 }
-a[aria-expanded='true'] ~ #info {
+
+:host([disable-border='false']) a[aria-expanded='true'] ~ #info {
```
bkjohnson, line 6: This is closer, but what about when `disable-border` isn't passed at all?

image

pdavies88: @bkjohnson updated so it will only change the css when the prop is true

packages/web-components/src/components/va-additional-info/va-additional-info.css
```diff
@@ -2,6 +2,13 @@
 :host {
   display: block;
 }
+
+:host([disable-border='true']) a[aria-expanded='true'] ~ #info {
```
bkjohnson, line 6: What about situations when `true` isn't explicitly passed, but the prop is just truthy? That's [a valid boolean attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#boolean_attributes):


> HTML5 defines restrictions on the allowed values of boolean attributes: If the attribute is present, **its value must either be the empty string** (equivalently, the attribute may have an unassigned value), or a value that is an ASCII case-insensitive match for the attribute's canonical name

Here's [an example](https://staging.va.gov/kansas-city-health-care/locations/shawnee-va-clinic/index.html):

image

pdavies88: @bkjohnson good call I added `:host([disable-border='']) a[aria-expanded='true'] ~ #info` to account for that so we are now covered for that use case

packages/web-components/src/components/va-additional-info/va-additional-info.css
```diff
@@ -2,6 +2,14 @@
 :host {
   display: block;
 }
+
+:host([disable-border='true']) a[aria-expanded='true'] ~ #info,
+:host([disable-border='']) a[aria-expanded='true'] ~ #info {
+  padding-left: 0;
+  border-left: 0;
+  border-left-color: transparent;
+}
```
bkjohnson, line 11: Is this the most direct approach we could take?

pdavies88: @bkjohnson I updated the css to the following to add in the line if the prop is not present or false:
https://github.com/department-of-veterans-affairs/component-library/blob/049b9c4d23b3fef4be19bd0eda58f31e0171709d/packages/web-components/src/components/va-additional-info/va-additional-info.css#L6-L12

I am assuming that is what you were trying to point out via the most recent comment.

If I am incorrect and additional work needs to be done to complete this ticket please let me know what needs to occur and I will gladly adjust. Thanks!

packages/web-components/src/components/va-additional-info/va-additional-info.css
```diff
@@ -2,7 +2,10 @@
 :host {
   display: block;
 }
-a[aria-expanded='true'] ~ #info {
+
+/* Add left side blue border if disable border prop disabled or not present */
+:host(:not([disable-border])) a[aria-expanded='true'] ~ #info,
+:host([disable-border='false']) a[aria-expanded='true'] ~ #info {
```
bkjohnson, line 8: I think this is good. Using `:host(:not([disable-border]))` seems like it will keep things easier to read and maintain in the future :+1: 

_(The rest of this is more of a public reminder for myself)_

At some point in the future I'd like to see if we can revisit things like that second line (to remove it), along with other examples like this:

https://github.com/department-of-veterans-affairs/component-library/blob/ba2bca2c9c1251d4960789addba0ad7659ae93f3/packages/web-components/src/components/va-accordion/va-accordion-item.css#L16

The reason is because I don't think that engineers will be using the component like:

```html
<va-additional-info disable-border="false" trigger="Whatever">
  Foo
</va-additional-info>
```

Instead, I think it's more likely that they will just not set `disable-border` if they want to keep the default behavior:

```html
<va-additional-info trigger="Whatever">
  Foo
</va-additional-info>
```

The reason we have our CSS like that is a result of how we are using storybook, so that can be for another ticket.

bkjohnson (approved):

PR #332, opened 2022-03-21
Add helper function for va-table migration

bkjohnson:
## Chromatic

https://table-migration-helper--60f9b557105290003b387cd5.chromatic.com

## Description
Part of https://github.com/department-of-veterans-affairs/va.gov-team/issues/35133

Based on an earlier comment: https://github.com/department-of-veterans-affairs/component-library/pull/291#discussion_r812494739

This should help make it easier to write an ESLint rule to migrate the React `<Table>` component to `<va-table>`. The helper function will take two arguments - `data` and `fields`, both structured the same way as the React `<Table>` props of the same name. It will then generate the appropriate `<va-table-row>` children.

---

I'm not committing the changes for storybook testing because part of them involve removing this alias:

https://github.com/department-of-veterans-affairs/component-library/blob/ba2bca2c9c1251d4960789addba0ad7659ae93f3/packages/storybook/.storybook/main.js#L27-L31

So that this import can work:

```js
import { generateTableChildren } from '@department-of-veterans-affairs/component-library';
```

## Testing done

- Local storybook :eyes: 
- Publishing a prerelease to use in `vets-website` locally
    - The data is structured to match [actual `<Table>` props](https://github.com/department-of-veterans-affairs/vets-website/blob/2e354510fc8a704fbd1bb7c84e3dad182f4eee94/src/applications/medical-copays/components/PaymentHistoryTable.jsx#L8-L35)

## Screenshots

Testing changes in `vets-website`:

Local changes in vets-website

table rows working in browser

## Acceptance criteria
- [ ]

## Definition of done


packages/web-components/src/utils/table.tsx
```diff
@@ -0,0 +1,41 @@
+/** @jsx React.createElement */
+import React from 'react';
```
bkjohnson, line 2: The `@jsx` comment is [changing typescript's `jsxFactory`](https://www.typescriptlang.org/tsconfig#jsxFactory) for this specific file, and the React import is so I can use that factory instead of `import { h } from `@stencil/core;`

packages/web-components/src/utils/table.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Useful for generating child content for a `<va-table>`
+ *
+ * @param {array} data - An array of arrays, where the inner arrays represent row data
+ * @param {array} fields - The field/column names
+ */
+export function generateTableChildren(data: any[], fields: FieldType[]): any[] {
+  return [
+    /** @ts-ignore */
```
bkjohnson, line 21: The `@ts-ignore` comments are because without the `h` JSX factory import from `components.d.ts` which also includes `<va-table-row>` as a [JSX "Intrinsic Element"](https://github.com/department-of-veterans-affairs/component-library/blob/ba2bca2c9c1251d4960789addba0ad7659ae93f3/packages/web-components/src/components.d.ts#L1240), we get errors. I tried some different imports and couldn't get it to work, and I didn't want to duplicate code, so I just ignored the warning since it shouldn't be an issue when this function gets used.

bkjohnson (commented):
Let me know if you think we should be more typescript-y with this.

packages/web-components/src/utils/table.spec.tsx
```diff
@@ -0,0 +1,34 @@
+/** @jsx React.createElement */
```
k80bowman, line 1: similar to the comment I left on the table sort PR, I wonder if this file should have a more specific name. Yes, these functions are specific to the table component, but we could be more specific here and call it something like `table-migration` maybe.

packages/web-components/src/utils/table.tsx
```diff
@@ -0,0 +1,46 @@
+/** @jsx React.createElement */
```
k80bowman, line 1: Similar to the comment I left on the table sort PR, I'd suggest that we name these files something more specific. Yes, these are functions that have to do with the table component, but we could be more specific and call it something like `table-migration`, maybe.

pdavies88 (approved):
LGTM

packages/web-components/src/utils/table.tsx
bkjohnson: I renamed the file and left a comment (https://github.com/department-of-veterans-affairs/component-library/pull/332/commits/714bc6d2d533b6cb8d0c1ad9ab814b4cc90e280f) to remove this in the future once teams are no longer using the React `<Table>`

PR #337, opened 2022-03-29
Added a prop for the required label to the the va-text-input...

brianseek:
…component moving the hard coded label to the default along with tests for default and passed prop.

## Chromatic

## Description
This is a PR to add a new optional prop for the (*Required) label text. Our use case for needing this change is so that we can fully implement Spanish translation in the Check In Experience app. This would be a nice to have pattern for all form field components, but this PR only covers text input. The current `(*Required)` text is set to the default when no value is present.

I made the entire previously hardcoded string overridable from the opening parenthesis to the closing parenthesis as well as the `*`, I figured that this would be better than just the text portion incase a future need to override the whole string comes along. I could see this happening with a custom icon or something similar that another application may need.

Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/39012
Zen hub: https://app.zenhub.com/workspaces/check-in-experience-61fc23a2cb8a14001132e102/issues/department-of-veterans-affairs/va.gov-team/39012

## Testing done
Updated va-text-input.e2e test to include default required label and passed in prop label.

## Screenshots
Screen Shot 2022-03-29 at 2 00 53 PM

## Definition of done

### DIFF
--- a/packages/storybook/stories/va-text-input.stories.jsx
+++ b/packages/storybook/stories/va-text-input.stories.jsx
@@ -22,6 +22,7 @@ const defaultArgs = {
   'autocomplete': false,
   'enable-analytics': false,
   'required': false,
+  'requiredLabel': null,
   'error': null,
   'maxlength': null,
   'value': null,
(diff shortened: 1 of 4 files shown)


pdavies88:
Closing in favor of internationalization PRs: #361 #360 #362 #358

Issue #338, opened 2022-03-29
Consider internationalization patterns


bkjohnson:
## Description

Some of our components have hard-coded English strings, like this:

https://github.com/department-of-veterans-affairs/component-library/blob/3c44d4a5e3b134c8d8be692fde19caa9e708cdda/packages/web-components/src/components/va-text-input/va-text-input.tsx#L154

It would be good if our components were language-aware and could present the right text for non-prop or children values, without engineers using the library having to provide translations. Having a system approach to internationalization would help ensure that users have a consistent experience, and application teams wouldn't need to take on this work themselves.

### Related tickets

- https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/667

## Details

The team building the check-in app is [setting the `lang` attribute of the element with `id="content"`](https://github.com/department-of-veterans-affairs/vets-website/blob/91682cf9b878818a704f9023bd9f1c29f0dd8faf/src/applications/check-in/utils/i18n/i18n.js#L74-L76) to reflect the content being in Spanish or English. They are also using the following dependencies for their PoC:

- [`i18next`](https://www.npmjs.com/package/i18next)
- [`i18next-browser-languagedetector`](https://www.npmjs.com/package/i18next-browser-languagedetector)
- [`react-i18next`](https://www.npmjs.com/package/react-i18next)

### i18next

The i18next family of libraries seems like a well supported and widely used collection, so we may want to follow that pattern. The `lang` attribute of the `<html>` tag seems like the best indicator for us to use, but since the VA isn't there yet, we can [add our own language detection](https://github.com/i18next/i18next-browser-languageDetector#adding-own-detection-functionality) which looks at the `lang` attribute of the tag with `id="content"`.

Our custom lookup could [use this file as a reference](https://github.com/i18next/i18next-browser-languageDetector/blob/34e9180e753ea3343b0ec5d079dd36fe46851566/src/browserLookups/htmlTag.js#L6) and replace the highlighted line with something like:

```js
    const contentContainer = document.getElementById('content');

    return contentContainer?.getAttribute('lang');
```

We could configure the language detection to use our custom look first, and then fallback to the `<html>` element if no element or language was found.

### Build functionality ourselves

[...]

PR #346, opened 2022-04-01
Add infrastructure for internationalization

bkjohnson:
## Chromatic

https://338-i18next-demo--60f9b557105290003b387cd5.chromatic.com

## Description
Part of https://github.com/department-of-veterans-affairs/component-library/issues/338

This is not a final version of an internationalization strategy for the Design System or the component library. This sets up the necessary dependencies for individual components to use `i18next` and register a `languageChanged` handler which forces a re-render.

If you want to check out the changes on storybook, this [earlier chromatic build](https://60f9b557105290003b387cd5-tvopxmylsb.chromatic.com/) will be useful.

## Testing done

- Storybook :eyes: 
- Testing the [pre-release](https://github.com/department-of-veterans-affairs/component-library/releases/tag/v8.4.0-0) in vets-website

## Screenshots

Dynamic translation for `<va-text-input>`

Rendered text changing to reflect language of the `<main>` tag

Dynamic translation for `<Date>`

Rendered text changing to reflect language of the `<main>` tag

Testing in `vets-website`:

pre-release version in vets-website translating dynamically

## Acceptance criteria
- [ ]

## Definition of done


packages/core/src/i18n/i18n-setup.js
```diff
@@ -0,0 +1,41 @@
...
+i18next.use(languageDetector).init({
+  fallbackLng: 'en',
+  detection: {
+    order: ['mainTag', 'htmlTag'],
+  },
+  resources: {
+    en: { translation: enTranslation },
+    es: { translation: esTranslation },
+  },
+});
```
bkjohnson, line 19: This initialization will detect the language on a page load - language detection doesn't automatically happen when things _in_ the DOM change.

packages/core/src/i18n/i18n-setup.js
```diff
@@ -0,0 +1,39 @@
...
+      if (mutation.type === 'attributes' && mutation.attributeName === 'lang') {
+        i18next.changeLanguage(element.getAttribute('lang'));
+      }
+    });
+  });
+
+  observer.observe(element, {
+    attributes: true,
+  });
+});
```
bkjohnson, line 39: This sets up a MutationObserver to change the language in the `i18next` dependency anytime the lang in `<main lang="">` changes. This is what supports the dynamic translation.

packages/core/src/i18n/language-detector.js
```diff
@@ -0,0 +1,9 @@
+export default {
+  name: 'mainTag',
+
+  lookup(options) {
+    const mainTag = document.querySelector('main');
+
+    return mainTag?.getAttribute('lang');
+  },
+};
```
bkjohnson, line 9: This, along with the `languageDetector.addDetector(mainTag);` in the setup file, allows the language detection to look at the `lang` attribute of the `<main>` tag. It's modeled after the dependency's [built-in `htmlTag` detector](https://github.com/i18next/i18next-browser-languageDetector/blob/34e9180e753ea3343b0ec5d079dd36fe46851566/src/browserLookups/htmlTag.js#L1-L14).

packages/core/src/i18n/translations/en.js
```diff
@@ -0,0 +1,4 @@
+export default {
+  'required': '(*Required)',
+  'max-chars': '(Max. {{length}} characters)',
+};
```
bkjohnson, line 4: In a fully internationalized component library, these translations will likely live somewhere else as part of the platform's larger internationalization strategy.

packages/react-components/scripts/build.js
```diff
@@ -52,6 +52,7 @@ const fileNames = [].concat.apply(
       ignore: ['./**/*.unit.spec.@(js|jsx)', './**/*.stories.@(js|jsx)'],
     }),
     glob.sync('./src/helpers/*.js'),
+    glob.sync('../core/src/i18n/**/*.js'),
```
bkjohnson, line 55: The build for `react-components` doesn't give us a proper bundle - it's just a bunch of JS files at the root level with all of their imports flattened. This makes sure that the "built" files will have access to the shared i18n setup.

packages/react-components/index.js
```diff
@@ -38,6 +38,8 @@ import Telephone, {
 import TextArea from './TextArea';
 import TextInput from './TextInput';
 
+import './i18n-setup';
```
bkjohnson, line 41: And this import initializes the `i18next` dependency with our language detection and translation files, so when components import the dependency it's ready to use.

packages/react-components/src/components/Date/SimpleDate.jsx
```diff
@@ -41,6 +42,15 @@ export const SimpleDate = ({
...
   const { day, month, year } = date;
 
+  const [lang, setLang] = useState();
+
+  useEffect(() => {
+    i18next.on('languageChanged', lang => {
+      console.log('language changed SimpleDate');
+      setLang(lang);
+    });
+  }, []);
```
bkjohnson, line 52: This effect hook is [basically like the `componentDidMount()` lifecycle event](https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect):

>  While passing `[]` as the second argument is closer to the familiar `componentDidMount` and `componentWillUnmount` mental model ...

We're doing this because `i18next` changing the language it's using internally as a result of the mutation observer doesn't trigger a re-render. Teams who use the component library _could_ force a re-render themselves and the output might be what they expect, but in the testing I did in storybook it's best if we have our translation-enhanced components force themselves to re-render when a language change is detected. With this approach the translation happens automatically _as long as_ the dynamic language switching also updates the `lang` attribute in `<main>`.

The state variable isn't really used other than to trigger the re-render so there's probably a better way of triggering a re-render, even if it's just naming it `_lang`.

packages/react-components/src/components/Date/SimpleDate.jsx
```diff
@@ -81,7 +91,9 @@ export const SimpleDate = ({
     <fieldset className={fieldsetClass} id={inputId.current}>
       <legend className="vads-u-font-size--base vads-u-font-weight--normal">
         {label || 'Date of birth'}
-        {required && <span className="form-required-span">(*Required)</span>}
+        {required && (
+          <span className="form-required-span">{i18next.t('required')}</span>
```
bkjohnson, line 95: This is the basic use of the `i18next` library - pass the `t()` function a key and it will return a value based on the language that is currently configured. Here's the [dependency documentation](https://www.i18next.com/overview/api#t).

packages/web-components/src/components/va-text-input/va-text-input.tsx
```diff
@@ -6,8 +6,20 @@ import {
...
+import i18next from 'i18next';
+// import { i18next } from '../../../../core/src/i18n/i18n-setup';
 import { consoleDevError } from '../../utils/utils';
+import { Build } from '@stencil/core';
+
+/** The .t() function will return the key instead of the value
+ */
+if (Build.isTesting) {
+  i18next.init({ lng: 'cimode' });
+}
```
bkjohnson, line 20: I don't like having this in the component's definition file. I originally tried having the `cimode` get set in the test file, but I think the instance of `i18next` in the compiled bundle is different than the one that's getting imported in the test file, so I wasn't initializing the right thing. :disappointed:

packages/web-components/src/components/va-text-input/va-text-input.tsx
```diff
@@ -146,6 +158,16 @@ export class VaTextInput {
     }
   };
 
+  connectedCallback() {
+    i18next.on('languageChanged', () => {
+      console.log('language changed text-input');
+      forceUpdate(this.el);
+    });
+  }
```
bkjohnson, line 166: This is the same pattern that we're using with the `useEffect` hook in the `<Date>` component. The component forces a re-render when it detects a language change, and then the next calls to `i18next.t()` will give output using the newly set language.

packages/web-components/src/components/va-text-input/va-text-input.tsx
```diff
@@ -173,8 +197,10 @@ export class VaTextInput {
           inputmode={inputMode}
           maxlength={this.maxlength}
         />
-        {this.maxlength && this.value.length >= this.maxlength && (
-          <small aria-live="polite">(Max. {this.maxlength} characters)</small>
+        {this.maxlength && this.value?.length >= this.maxlength && (
+          <small aria-live="polite">
+            {i18next.t('max-chars', { length: this.maxlength })}
+          </small>
```
bkjohnson, line 203: The optional chaining here is to fix a bug I noticed in storybook where the component doesn't render due to the `value` prop not being set:

image

---

The `i18next` usage is just an example of how we can pass values into the translation string.

packages/web-components/src/index.ts
```diff
@@ -1,3 +1,5 @@
+import '../../core/src/i18n/i18n-setup';
```
bkjohnson, line 1: This is similar to the import in the `react-components` index file, except typescript is able to use the relative import successfully.

packages/storybook/.storybook/main.js
```diff
@@ -24,11 +24,11 @@ module.exports = {
...
-    // This is so the docs page for each component can populate the args table.
-    // The propTypes are stripped out in the built version, so we need to use
-    // the source files
-    config.resolve.alias['@department-of-veterans-affairs/component-library$'] =
-      path.resolve(__dirname, '../../react-components/src');
+    // // This is so the docs page for each component can populate the args table.
+    // // The propTypes are stripped out in the built version, so we need to use
+    // // the source files
+    // config.resolve.alias['@department-of-veterans-affairs/component-library$'] =
+    //   path.resolve(__dirname, '../../react-components/src');
```
bkjohnson, line 31: I commented this out so that I could test in storybook more confidently by knowing that I'm importing things in a way that the would be imported in `vets-website`. It will be reverted before merge.

packages/storybook/.storybook/preview.js
```diff
@@ -6,7 +6,7 @@ import '@department-of-veterans-affairs/component-library/dist/main.css';
 import {
   applyPolyfills,
   defineCustomElements,
-} from '@department-of-veterans-affairs/web-components/loader';
+} from '@department-of-veterans-affairs/component-library';
```
bkjohnson, line 9: This was also part of testing and will be reverted before merge.

bkjohnson (commented):
I originally tried running the i18n setup in `core/src/main.js`, but then the imports in `react-components` and `web-components` weren't initialized and I ended up with no output from the `i18next.t()` function. Some of the awkwardness here is a result of our ongoing tech debt.

Also, the react component tests are failing in a very annoying way with no error messages:

image

packages/web-components/src/components/va-text-input/va-text-input.tsx
```diff
@@ -173,7 +173,7 @@ export class VaTextInput {
           inputmode={inputMode}
           maxlength={this.maxlength}
         />
-        {this.maxlength && this.value.length >= this.maxlength && (
+        {this.maxlength && this.value?.length >= this.maxlength && (
```
bkjohnson, line 176: This is a small bugfix that might only impact storybook.

bkjohnson (commented):
I've rolled back all of the testing changes to `<va-text-input>` and `<SimpleDate>`. What's left should be enough so that future PRs can be targeted to specific components and be about:

- Adding translations as needed
- Importing `i18next` and calling the `t()` function
- Updating tests as needed
- Adding stories to demo the translation capability

Before we merge I'll update the minor versions of the three packages for publishing.

pdavies88 (approved):

PR #349, opened 2022-04-06
39550 va-number-input Web Component

pdavies88:
## Chromatic

https://39550-number-input-wc--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/39550

## Testing done
Screen Shot 2022-04-06 at 11 52 18 AM

## Screenshots
Screen Shot 2022-04-06 at 11 54 05 AM

## Acceptance Criteria
- [X] `va-number-input` is written and added to Storybook
- [X] E2E tests are written
- [ ] Add Chromatic link to https://github.com/department-of-veterans-affairs/va.gov-team/issues/39674, assign ticket to Courtney, and move ticket to Current Sprint column

## Definition of done


packages/web-components/src/components/va-number-input/va-number-input.tsx
```diff
@@ -0,0 +1,150 @@
...
+  componentLibraryAnalytics: EventEmitter;
+
+  private handleChange = (e: Event) => {
+    const target = e.target as HTMLInputElement;
+    this.value = target.value;
+    this.vaChange.emit({ value: this.value });
+  };
+
+  private handleBlur = () => {
+    this.vaBlur.emit();
```
bkjohnson, line 107: Would it be possible to use a native event instead of a custom one? The [`blur` event is composed but it doesn't bubble](https://developer.mozilla.org/en-US/docs/Web/API/Element/blur_event), but I'm not sure if the no bubbling would matter here. For example:

```jsx
<va-number-input onBlur={(e) => console.log('blur event', e)} />
```

The `onBlur` handler should still fire, but when the web component is blurred instead of the inner `<input>`. The effect might still be the same, but we'd need to test to know for sure.

packages/web-components/src/components/va-number-input/va-number-input.tsx
```diff
@@ -0,0 +1,150 @@
...
+        {this.error && <span id="error-message">{this.error}</span>}
+        <input
+          id="inputField"
+          type="number"
+          aria-describedby={describedBy}
+          inputmode={inputMode}
+          max={this.max}
+          min={this.min}
+          value={this.value}
+          onInput={this.handleChange}
```
bkjohnson, line 144: The React version is using the `change` native event instead of `input`:

https://github.com/department-of-veterans-affairs/component-library/blob/77358b99680ae3ffca2134ce97e4c46b06a7a0d2/packages/react-components/src/components/NumberInput/NumberInput.jsx#L78

It's probably my fault that `<va-text-input>` has switched things around, but if we stick with `onInput` I think we might not need the custom events (and the react bindings :crossed_fingers: ). Here's a [demo](https://github.com/department-of-veterans-affairs/component-library/compare/native-event-demo) I put together a few weeks ago.

And when I run that locally and type "hello" into the input, I get:

logging output from `input` handler for each character typed

packages/web-components/src/components/va-number-input/va-number-input.tsx
```diff
@@ -0,0 +1,150 @@
...
+  Prop,
+  Host,
+  h,
+  Event,
+  EventEmitter
+} from '@stencil/core';
+
+@Component({
+  tag: 'va-number-input',
+  styleUrl: 'va-number-input.css',
```
bkjohnson, line 13: Could we point to the CSS file for the text input? Alternatively, could we `@import` that file in `va-number-input.css` so we don't have to redefine a lot of the same things?

bkjohnson (commented):
The tests look good - I just have some comments on CSS reuse and possibly relying on native events so that react bindings aren't needed.

packages/web-components/src/components/va-number-input/va-number-input.tsx
pdavies88: I tried to point it to the va-text-input.css but that gave a Rollup error so I chose to do an import in css file instead which worked correctly

pdavies88:
@andresriveratoro @bkjohnson @k80bowman I removed the custom events and event emitters and switched to (onBlur and onInput) and the code works in Storybook.
Screen Shot 2022-04-12 at 9 33 05 AM

But this does mean that I had to comment out our tests for blur and input on this component since we no longer have events to spy on.

In Andres `va-search` PR we leave the custom events and event emitters in so that we can test them in the testing suite:
https://github.com/department-of-veterans-affairs/component-library/pull/325/files#diff-d61a0a911521c70bf990c7e1fdce88087008ff6030c88632d7b30fe512193156R103
https://github.com/department-of-veterans-affairs/component-library/pull/325/files#diff-d61a0a911521c70bf990c7e1fdce88087008ff6030c88632d7b30fe512193156R22

Should we leave the custom events and event emitters in so that we can test them?

bkjohnson:
Let's test for the native events instead.

k80bowman:
I agree with Brooks. I'd rather not leave code it only for testing purposes.

pdavies88:
@bkjohnson @k80bowman @andresriveratoro I updated the testing suite so that we are now testing the native events.

packages/web-components/src/components/va-number-input/va-number-input.tsx
```diff
@@ -0,0 +1,130 @@
...
+          type="number"
+          aria-describedby={describedBy}
+          inputmode={inputMode}
+          name={this.name}
+          max={this.max}
+          min={this.min}
+          value={this.value}
+          onInput={this.handleInput}
+          onBlur={this.handleBlur}
+        />
```
bkjohnson, line 126: Oh, if we have `required={this.required}` in here [the `:inactive` thing should work](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required#constraint_validation).

> If the element is required and the element's value is the empty string, then the element is suffering from [valueMissing]() and the element will match the [:invalid](https://developer.mozilla.org/en-US/docs/Web/CSS/:invalid) pseudo class.

I noticed that we aren't doing that in `<va-text-input>` and I'm guessing that's just an oversight rather than an intentional decision. We can fix that in another PR.

packages/web-components/src/components/va-number-input/va-number-input.tsx
```diff
@@ -0,0 +1,130 @@
...
+          label: this.label,
+          value: this.value,
+        },
+      });
+    }
+  };
+
+  render() {
+    const describedBy =
+      `${this.ariaDescribedby} ${this.error ? 'error-message' : ''}`.trim() ||
```
bkjohnson, line 102: I know this is another thing that's copied over from `<va-text-input>`, but I'm curious how JAWS/VoiceOver/NVDA would handle this and I don't think we tested it back then. `error-message` is an id inside the shadow DOM, and a value coming in from the prop would be for something on the outside. @CourtneyChambers77, can you check the [story with an error message and `aria-describedby`](https://39550-number-input-wc--60f9b557105290003b387cd5.chromatic.com/?path=/story/components-va-number-input--aria-describedby) to see if the screen reader finds what it needs to?

packages/storybook/stories/va-number-input.stories.jsx
```diff
@@ -0,0 +1,94 @@
...
+  min: 0,
+  max: 4,
+};
+
+export const AriaDescribedby = Template.bind({});
+AriaDescribedby.args = {
+  ...defaultArgs,
+  'error': 'This is also an error',
+  'aria-describedby': 'testing-id',
+};
```
bkjohnson, line 94: In order to test this with screen readers we may want to make sure the template renders with a `<div id="testing-id">This explains the error in detail</div>` or something like that.

bkjohnson (commented):
That `:invalid` pseudoselector is pretty cool - I'm wondering if we can rely on that a bit more for "free" html validation. Now that we've got an accessibility expert on the team I'm also wondering if the `aria-describedby` approach that we used in the text input web component is working like we want it to.

packages/web-components/src/components/va-number-input/va-number-input.css
```diff
@@ -0,0 +1,5 @@
+@import '../va-text-input/va-text-input.css';
+
+input:invalid {
```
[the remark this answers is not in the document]
pdavies88: @bkjohnson what I had found was these are the parameters that work with it: required, minlength, maxlength, min, max, type and pattern.
Since min and max is unique to the `va-number-input` I added it.
Here's an article going into more depth: https://funwithforms.com/posts/valid-invalid-pseudo-classes/

packages/web-components/src/components/va-number-input/va-number-input.tsx
pdavies88: @bkjohnson you are correct I had mimicked what we had done in `va-text-input`. Let me know if you want me to write up a ticket to add in the required attribute to both the `va-text-input` and `va-number-input` components

packages/storybook/stories/va-number-input.stories.jsx
pdavies88: @bkjohnson so right now it outputs like:
```
<input id="inputField" type="number" aria-describedby="testing-id error-message" inputmode="numeric" name="my-input">
```
and that exists within the shadow DOM. Are you thinking that moving aria-describedby out into its own `div` with the prop setting the id and the error message setting the content above the input would be a better move? Should we then visually hide that element?

packages/web-components/src/components/va-number-input/va-number-input.tsx
pdavies88: PR for suggested work: https://github.com/department-of-veterans-affairs/va.gov-team/issues/39980

packages/storybook/stories/va-number-input.stories.jsx
bkjohnson: I was imagining something like this for the story:

```jsx

return (
  <>
    <div id="testing-id">Some more detail on the error</div>
    <va-number-input error="You had an error" aria-describedby="testing-id" />
  </>
);

```

Because that's the way I understand `aria-describedby` to work. But it sounds like we'll be doing that with another ticket.

bkjohnson (approved):

PR #354, opened 2022-04-07
Add typeahead functionality to va-search-input

andresriveratoro:
## Chromatic
DIRECT LINK TO ISOLATED COMPONENT:
https://37974-typeahead--60f9b557105290003b387cd5.chromatic.com/iframe.html?id=components-va-search-input--with-typeahead&viewMode=story

https://37974-typeahead--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/37974

## Testing done
E2E, Storybook

## Screenshots
Screen Shot 2022-04-21 at 12 29 31

Screen Shot 2022-04-21 at 12 29 49

## Acceptance criteria
- [x] Typeahead dropdown is implemented and in Storybook (large and small versions)
    - [x] Remove indicator that va-search-input component is not ready for use
- [x] E2E/unit tests are written
- [x] Add Chromatic link to #37975 , assign ticket to Courtney, and move ticket to Current Sprint column
- [ ] Once accessibility review is complete, alert Carol that the full va-search-input component is ready for staging review

## Definition of done


packages/web-components/src/components/va-search/va-search.tsx
```diff
@@ -168,6 +418,35 @@ export class VaSearch {
...
+            aria-label="Search Suggestions"
+            role="listbox"
+          >
+            {formattedSuggestions.map(
+              (suggestion: string | HTMLElement, index: number) => {
+                return (
+                  <li
+                    id={`listbox-option-${index}`}
+                    aria-selected={activeIndex === index ? 'true' : undefined}
+                    aria-hidden="true"
```
bkjohnson, line 433: Do we want to hide the suggestions from screen readers? I'm wondering if we could have [`aria-live="polite"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-live) somewhere to let the user know that there are updates to the document.

packages/web-components/src/components/va-search/va-search.tsx
```diff
@@ -79,6 +84,61 @@ export class VaSearch {
...
   buttonKeyDownEvent: EventEmitter;
 
+  /**
+   * Fires when a suggestion is clicked
+   */
+  @Event({
+    composed: true,
+    bubbles: true,
+  })
+  suggestionClickEvent: EventEmitter;
```
bkjohnson, line 94: I think we could drop the `Event` part of these event names.

packages/web-components/src/components/va-search/va-search.tsx
```diff
@@ -79,6 +84,61 @@ export class VaSearch {
...
+
+  /**
+   * An array of formatted suggestions
+   */
+  @State() formattedSuggestions: string[] = [];
+
+  /**
+   * Index of selected suggestion
+   */
+  @Prop() activeIndex?: number;
```
bkjohnson, line 140: Are we expecting engineers to be setting `activeIndex` when they use the component?

packages/web-components/src/components/va-search/va-search.tsx
```diff
@@ -94,6 +154,31 @@ export class VaSearch {
...
 
+  /**
+   * A boolean that controls whether suggestions are visible
+   */
+  @Prop() showSuggestions?: boolean;
+
+  /**
+   * An array of strings containing suggestions
+   */
+  @Prop() suggestions: any;
```
bkjohnson, line 165: We should try to avoid array props if we can since they won't be easy to use in the liquid templates.

packages/web-components/src/components/va-search/va-search.tsx
```diff
@@ -124,37 +229,182 @@ export class VaSearch {
...
+    }
+  };
+
+  private handleSuggestionMouseDownEvent = (event: MouseEvent) => {
+    this.suggestionMouseDownEvent.emit(event);
+  };
+
+  private handleSuggestionMouseOverEvent = (event: MouseEvent) => {
+    this.suggestionMouseOverEvent.emit(event);
+  };
```
bkjohnson, line 320: What's happening on mouseover that we need to support?

andresriveratoro: [This](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/applications/search/components/SearchDropdown/SearchDropdownComponent.js#L694-L696) is how it's being used in the SearchDropdownComponent in `vets-website`. I think it's safe to remove.

andresriveratoro:
In order to make this component easier to use, the state of the listbox was internalized. This allowed me to remove all custom events and replace them with just `input` and `submit`.

The new API has been updated to:
```jsx
  <VaSearch
    buttonText={buttonText}
    value={value}
    label={label}
    onInput={e => console.log(e)}
    onSubmit={e => console.log(e)}
    suggestions={suggestions}
  />
```

packages/web-components/src/components/va-search/va-search.tsx
```diff
@@ -15,159 +16,348 @@ import {
...
+  private handleListboxKeyDown = (event: KeyboardEvent, index: number) => {
+    const options = this.el.shadowRoot.querySelectorAll(
+      '[role="option"]',
+    ) as NodeListOf<HTMLLIElement>;
+
+    switch (event.key) {
+      case 'ArrowUp':
+        const lastOption = options[options.length - 1];
+        if (index === 0) {
+          if (!lastOption) return;
```
bkjohnson, line 189: I'm trying to visualize what "no last option" would look like and I'm not sure I understand. What is this doing for us?

packages/web-components/src/components/va-search/va-search.tsx
```diff
@@ -15,159 +16,348 @@ import {
...
+          this.selectSuggestion(lastOption);
+        } else {
+          if (!options[index - 1]) return;
```
bkjohnson, line 192: Similar question here. We know that `index` will be at least 1, so what kind of value in the `NodeList` would be falsey?

packages/web-components/src/components/va-search/va-search.tsx
```diff
@@ -15,159 +16,348 @@ import {
...
+    const inputValue = this.inputRef.value.toLowerCase();
+    if (lowercaseSuggestion.includes(inputValue)) {
+      return (
+        <Fragment>
+          {inputValue}
+          <strong>{lowercaseSuggestion.replace(inputValue, '')}</strong>
+        </Fragment>
+      );
+    }
+    return <strong>{suggestion}</strong>;
```
bkjohnson, line 258: I want to double check my understanding. We're formatting the suggestions based on the `suggestions` change instead of the input change, but consumers of the component _should_ change  `suggestions` based on the input change, is that right? I ask because I expected the bolding to continue creeping into the tail end of "benefits" as I deleted characters and it didn't happen:

bolded suggestions after "benefit"

I think that's fine, but we should make sure we document when we apply the formatting so that teams understand what they should do.

packages/web-components/src/components/va-search/va-search.tsx
```diff
@@ -15,159 +16,348 @@ import {
...
-    this.buttonKeyDownEvent.emit(event);
+  /**
+   * Focuses a suggestion, sets its aria-selected attribute to true, updates aria-activedescendant on input
+   * and removes aria-selected from previously selected option if it exists
+   */
+  private selectSuggestion = (suggestion: HTMLLIElement) => {
+    const selectedSuggestion = this.el.shadowRoot.querySelector(
+      '[aria-selected="true"]',
+    );
+    if (selectedSuggestion) selectedSuggestion.removeAttribute('aria-selected');
```
bkjohnson, line 269: Could also use optional chaining here.

packages/web-components/src/components/va-search/va-search.tsx
```diff
@@ -15,159 +16,348 @@ import {
...
+  /**
+   * Sets search input value to the suggestion clicked,
+   * removes aria-selected from previously selected suggestion if it exists,
+   * closes the listbox and fires a submit event.
+   */
+  private handleListboxClick = (index: number) => {
+    const suggestion = this.el.shadowRoot.getElementById(
+      `listbox-option-${index}`,
+    );
+    if (!suggestion) return;
```
bkjohnson, line 163: If this accepted the event argument could we use `event.target` for the suggestion instead of using the index to get it?

packages/storybook/stories/va-search.stories.jsx
```diff
@@ -54,3 +52,34 @@ WithButtonText.args = {
...
+    <VaSearch
+      buttonText={buttonText}
+      value={value}
+      label={label}
+      onInput={e => console.log(e)}
+      onSubmit={e => console.log(e)}
+      suggestions={suggestions}
+    />
+  </div>
+);
```
bkjohnson, line 72: I wanted to see if I could make this be more like a real app and I'm not sure how an engineer is supposed to get the value in order to update `suggestions`. Using `e.target.value` in `onInput` doesn't seem to reflect the changes from typing. Am I missing something?

bkjohnson (commented):
A couple of minor suggestions and a few questions.

packages/storybook/stories/va-search.stories.jsx
andresriveratoro: I'm able to access the value of the input when it fires using `e.nativeEvent.path[0].value` in Storybook.

packages/web-components/src/components/va-search/va-search.tsx
```diff
@@ -15,159 +16,348 @@ import {
...
    */
-  @Prop() label: string = 'Search';
+  private handleSubmit = () => {
+    this.el.dispatchEvent(
+      new CustomEvent('submit', {
+        bubbles: true,
+        cancelable: true,
+        detail: { value: this.inputRef.value },
+      }),
+    );
```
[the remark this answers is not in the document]
andresriveratoro: We do. Good catch!

packages/web-components/src/components/va-search/va-search.tsx
```diff
@@ -15,159 +16,348 @@ import {
...
+   */
+  private handleListboxClick = (index: number) => {
+    const suggestion = this.el.shadowRoot.getElementById(
+      `listbox-option-${index}`,
+    );
+    if (!suggestion) return;
+    const selectedSuggestion = this.el.shadowRoot.querySelector(
+      '[aria-selected="true"]',
+    );
+    if (selectedSuggestion) selectedSuggestion.removeAttribute('aria-selected');
```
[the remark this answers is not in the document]
andresriveratoro: You're correct. This is no longer needed since clicking on a suggestion will cause it to close the listbox.

packages/web-components/src/components/va-search/va-search.tsx
andresriveratoro: This can be removed as it's not really doing anything.

packages/storybook/stories/va-search.stories.jsx
andresriveratoro: This is a screenshot of the component being used in the Stencil dev server. I'm using `event.composedPath()[0].value` to retrieve the value of the search input field.

Screen Shot 2022-04-20 at 15 33 04

packages/web-components/src/components/va-search/va-search.tsx
andresriveratoro: This has been updated so that the suggestion formatting is updated on user input to provide the "creeping" you described.

packages/storybook/stories/va-search.stories.jsx
andresriveratoro: This is what I had in mind:
```jsx
const handleInput = (event) => {
  const inputValue = event.composedPath()[0].value;
// const inputValue = event.nativeEvent.composedPath()[0].value // React
// we want to start offering suggestions when the user has entered at least 3 characters
  if (inputValue.length > 2) fetchSuggestions();
}

const handleSubmit = (event) => {
  const inputValue = event.detail.value;
  // const inputValue = event.nativeEvent.detail.value; // React
  // prevent submitting if input is empty (optional)
  if (!inputValue) return;
  submitQuery(inputValue);
}
```

where `fetchSuggestions` updates the suggestions array and `submitQuery` navigates the user to the results page or updates results for the current page.

pdavies88 (approved):

packages/storybook/stories/va-search.stories.jsx
andresriveratoro: I've updated the `With Suggestions` example to showcase how this component works and to aid engineers in implementing their own handlers.

andresriveratoro:
Updated component to emit input event on programmatic input value change

packages/storybook/stories/va-search-input.stories.jsx
```diff
@@ -0,0 +1,126 @@
...
+};
+
+const TypeaheadTemplate = ({ value, suggestions }) => {
+  const [text, setText] = useState(value);
+  const [latestSuggestions, setLatestSuggestions] = useState(suggestions);
+
+  const handleInput = e => {
+    console.log(e);
+    // event.composedPath()[0].value outside of React
+    setText(e.nativeEvent.composedPath()[0].value);
```
bkjohnson, line 59: This feels awkward to me. Could we update `value` on the component as we go so that `e.target.value` will work?

packages/web-components/src/components/va-search-input/va-search-input.tsx
```diff
@@ -0,0 +1,402 @@
...
+  /**
+   * An array of strings containing suggestions to be displayed in listbox.
+   * This component displays up to 5 suggestions.
+   */
+  @Prop() suggestions?: any;
+
+  /**
+   * The value of the input field
+   */
+  @Prop() value?: string = '';
```
bkjohnson, line 50: Having this be `mutable` could make the input handler more intuitive.

packages/web-components/src/components/va-search-input/va-search-input.tsx
```diff
@@ -0,0 +1,402 @@
...
+    this.updateSuggestions(this.suggestions);
+  }
+
+  /**
+   * If suggestions are provided, then format suggestions and open the listbox.
+   * Limits suggestions to 5 and sorts them.
+   */
+  @Watch('suggestions')
+  watchSuggestionsHandler(newSuggestions: string[]) {
+    if (!Array.isArray(newSuggestions)) return;
```
bkjohnson, line 67: I think the typechecking on `newSuggestions: string[]` would prevent this from happening.

packages/web-components/src/components/va-search-input/va-search-input.tsx
```diff
@@ -0,0 +1,402 @@
...
+   * Fires a submit event
+   */
+  private handleSubmit = () => {
+    this.el.dispatchEvent(
+      new CustomEvent('submit', {
+        bubbles: true,
+        cancelable: true,
+        composed: true,
+        detail: { value: this.inputRef.value },
+      }),
```
bkjohnson, line 90: Could we use a [`SubmitEvent`](https://developer.mozilla.org/en-US/docs/Web/API/SubmitEvent) for this?

packages/web-components/src/components/va-search-input/va-search-input.tsx
```diff
@@ -0,0 +1,402 @@
...
+    switch (event.key) {
+      case 'ArrowDown':
+        // Option doesn't exist if suggestions aren't provided
+        if (!firstOption) return;
+        this.selectSuggestion(firstOption);
+        break;
+      case 'ArrowUp':
+        // Option doesn't exist if suggestions aren't provided
+        if (!lastOption) return;
+        this.selectSuggestion(lastOption);
```
bkjohnson, line 135: Would these if statements work the same if they were checking:

```js
if (!options) return;
```
Having them selectively look at the first or the last option is a bit confusing to me because I'm not understanding why those particular items are important.

packages/storybook/stories/va-search-input.stories.jsx
```diff
@@ -0,0 +1,126 @@
...
+        setTimeout(() => {
+          setLatestSuggestions(['forms']);
+        }, 1000);
+        break;
+      default:
+        setTimeout(() => {
+          setLatestSuggestions([]);
+        }, 1000);
+        break;
+    }
```
bkjohnson, line 100: I know this is just a storybook thing, but we may be able to get rid of the case statement and make this easier to maintain with something like:

```js
const suggestions = [
  'form', 
  'form finder', 
  'form search', 
  'foreign study', 
  'forever gi bill'
];

setLatestSuggestions(suggestion.filter(s => s.includes(text)));
```

packages/web-components/src/components/va-search-input/va-search-input.tsx
andresriveratoro: I believe we could update `value` to be reflective and mutable, but this requires the consumer to handle `value`'s value.

packages/web-components/src/components/va-search-input/va-search-input.tsx
andresriveratoro: This check is required because TypeScript doesn't check types on runtime.

packages/web-components/src/components/va-search-input/va-search-input.tsx
andresriveratoro: You're correct. It doesn't matter if there's a first or last option - what matters is that there's at least 1 option available.

packages/web-components/src/components/va-search-input/va-search-input.tsx
andresriveratoro: When I tried `value` as `reflective` (required to be able to use event.target.value or this value never updates) and `mutable` (this meant updating all value changes to `this.value` instead of `this.inputRef.value`, and also firing events because of the programmatic value changes to the input field), I experienced some weird behavior when NOT managing the state of value in the consumer. If we're okay with having the consumer manage this value so that it's synced, then I can make those changes. I know retrieving the values from non-standard paths like `event.composedPath()[0].value` isn't ideal, but this makes it easier for consumers to use this component without having to manage the state of `value`.

packages/storybook/stories/va-search-input.stories.jsx
andresriveratoro: See comment in the `value` thread.

packages/web-components/src/components/va-search-input/va-search-input.tsx
andresriveratoro: Some of the weird behavior included having an initial value of `for`, deleting the text using the Escape key to clear out the input field, and then typing anything would result in the input field being updated to `for[whatever-key-I-typed]`.

andresriveratoro: I've pushed up the changes that would allow consumers to look for `event.target.value` [here](https://github.com/department-of-veterans-affairs/component-library/pull/391), but `value` isn't being reflected properly.

packages/storybook/stories/va-search-input.stories.jsx
andresriveratoro: This request has been addressed [here](https://github.com/department-of-veterans-affairs/component-library/pull/391).

packages/web-components/src/components/va-search-input/va-search-input.tsx
andresriveratoro: This has been addressed [here](https://github.com/department-of-veterans-affairs/component-library/pull/391) although `value` is still not reflected properly. Consumers can retrieve the value of the input in `event.target.value` for both `input` and `submit` events. 

Note: The `value` property will always be correct, but the attribute (if using getAttribute('value')) will be incorrect.

andresriveratoro:
Updated component to prevent scrolling page when using keyboard navigation in the listbox.

bkjohnson (approved):
Looks good, one minor question.

packages/web-components/src/components/va-search-input/va-search-input.tsx
```diff
@@ -0,0 +1,401 @@
...
+    const inputValue = this.inputRef.value.toLowerCase();
+    if (lowercaseSuggestion.includes(inputValue)) {
+      return (
+        <Fragment>
+          {inputValue}
+          <strong>{lowercaseSuggestion.replace(inputValue, '')}</strong>
+        </Fragment>
+      );
+    }
+    return <strong>{suggestion}</strong>;
```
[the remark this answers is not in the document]
andresriveratoro: Eagle eye. @GnatalieH should the suggestions displayed in the listbox always be lowercase?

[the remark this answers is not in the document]
GnatalieH: Yes @andresriveratoro, it looks like that's how it's currently done on VA.gov global search. Thanks for checking.

PR #357, opened 2022-04-09
Add initial tests for dashboard-cli

bkjohnson:
## Chromatic

https://dashboard-cli-tests--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/620

This adds Jest tests for some parts of the `design-system-dashboard-cli` package. We will increase test coverage with future PRs

## Testing done

- Jest tests

## Screenshots

We're just above 30% test coverage for lines at the moment:

Highlighting the 30.28% line coverage from CI output

## Acceptance criteria
- [ ]

## Definition of done


packages/design-system-dashboard-cli/package.json
```diff
@@ -26,7 +27,9 @@
   "devDependencies": {
     "eslint": "^7.32.0",
     "husky": ">=6",
+    "jest": "^27.5.1",
     "lint-staged": ">=10",
+    "memfs": "^3.4.1",
```
bkjohnson, line 32: This lets us mock a file system in memory so we can be more controlled in our testing.

.github/workflows/node.js.yml
```diff
@@ -24,4 +24,4 @@ jobs:
       working-directory: packages/react-components
     - run: yarn workspace @department-of-veterans-affairs/react-components run build
     - run: yarn workspace @department-of-veterans-affairs/web-components run build
-    - run: yarn workspaces foreach run test
+    - run: yarn workspaces foreach -v run test
```
bkjohnson, line 27: This `verbose` flag puts the workspace package name in front of each line of output so that we can have an easier time finding what's failing.

packages/design-system-dashboard-cli/test/search-files.spec.js
```diff
@@ -0,0 +1,76 @@
...
+    'mock-vw/txt-file.txt': 'Some text', 
+    'mock-cb/liquid-template.liquid': 'A template',
+    'mock-cb/html-layout.html': 'HTML page',
+};
+
+jest.mock('fs');
+memfs.vol.fromJSON(json, '.');
+
+describe('readAllModules', () => {
+  it('reads only js and jsx files', () => {
```
bkjohnson, line 30: I wanted to add a test showing that it excluded `*.unit.jsx` files since that's what the pattern _looks_ like it's doing:

https://github.com/department-of-veterans-affairs/component-library/blob/ffe9df32efc94776cb7141120afa5725c14a61fb/packages/design-system-dashboard-cli/src/search-files.js#L21-L22

But if I add a `'mock-vw/js-file.unit.jsx'` key to the JSON up above, that file gets included in this test. Once we have good coverage to prevent regressions we may want to modify code so that it actually does what we think it does.

packages/design-system-dashboard-cli/.yarnrc.yml
```diff
@@ -0,0 +1 @@
+nodeLinker: node-modules
```
bkjohnson, line 1: This is a yarn 3 workspaces thing. We're telling it to use `node-modules` instead of the newer plug-n-play system they have. The other packages in the monorepo have this as well.

bkjohnson (commented):
I'm not trying to add too much more test coverage to this PR in particular, but let me know if you think I should refactor any of the tests, or if you want to use something other than jest.

pdavies88 (approved):

pdavies88:
Thank you for all the comments they helped me understand what was added and why and I personally like Jest and the coverage can be tracked via it which is nice

PR #364, opened 2022-04-12
Add input datalist to `va-text-input`

Mottie:
## Chromatic

https://39873-input-datalist--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/39873

## Testing done

E2E and storybook

## Screenshots

Storybook 'with datalist options' example with expanded code block

## Acceptance criteria
- [x] `va-text-input` accepts datalist options
- [x] All tests passing

## Definition of done

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@department-of-veterans-affairs/component-library",
   "description": "VA.gov component library. Includes React and web components.",
-  "version": "8.5.0",
+  "version": "8.6.0",
   "license": "MIT",
   "scripts": {
     "build": "webpack"
--- a/packages/storybook/stories/va-text-input.stories.jsx
+++ b/packages/storybook/stories/va-text-input.stories.jsx
@@ -64,13 +64,11 @@ const Template = ({
   inputmode,
   type,
   'aria-describedby': ariaDescribedby,
-  list,
 }) => {
   return (
     <va-text-input
       name={name}
       label={label}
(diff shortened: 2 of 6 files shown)


pdavies88:
@k80bowman @Mottie General question is there a use case that brought about this change? Or is it more of a nice to have enhancement?

We are working on adding in typeahead for the `va-search` component but the mocks that @GnatalieH had designed rely more on keypress then just giving a user all options at once.

Related tickets on typeahead: 
https://github.com/department-of-veterans-affairs/va.gov-team/issues/37974
https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/637
https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/641

To be aware I'm not saying no I'm just trying to figure out if it has a use case where it would be useful. A lot of times we add these enhancements and nobody uses them and then the component becomes bloated dead code that like in theory could be used but nobody does. So personally I would only like to add these enhancements if there is an actual team need for them.

Mottie:
I was thinking about Josh's request to change the date inputs (month, day & year) to all be text inputs instead of a month & day select with a year text input. Using a datalist would allow us to dynamically add options to the text input, including month translations. I'm sure typeahead would accomplish the same thing, but using a browser built-in method seems like a better/easier solution in this case.

This change may not be as useful for something like search, although it would work with dynamic options, but I'll leave that up to your team's discretion.

PR #366, opened 2022-04-14
Bump Formation v7.0.1

pdavies88:
## Closes
https://github.com/department-of-veterans-affairs/va.gov-team/issues/39750

## Testing done
Tested locally and all yarn install, yarn build, yarn test and yarn storybook commands work properly and storybook looks and works correctly

## Acceptance criteria
- [X] Update Formation Dependency to latest version

### DIFF
--- a/packages/web-components/package.json
+++ b/packages/web-components/package.json
@@ -54,7 +54,7 @@
     "typescript": "^4.3.5"
   },
   "peerDependencies": {
-    "@department-of-veterans-affairs/formation": "^7.0.1"
+    "@department-of-veterans-affairs/formation": "^7.0.1",
     "i18next": "*",
     "i18next-browser-languagedetector": "*"
   }
--- a/packages/web-components/package.json
+++ b/packages/web-components/package.json
@@ -37,7 +37,7 @@
   "devDependencies": {
     "@axe-core/puppeteer": "^4.4.0",
     "@babel/core": "^7.12.13",
-    "@department-of-veterans-affairs/formation": "^6.17.5",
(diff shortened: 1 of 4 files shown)


packages/web-components/package.json
```diff
@@ -52,8 +52,5 @@
     "puppeteer": "10.0.0",
     "react-dom": "^16.7.0",
     "typescript": "^4.3.5"
-  },
-  "peerDependencies": {
-    "@department-of-veterans-affairs/formation": "^6.14.0"
```
bkjohnson, line 57: I think we still want to list `formation` as a peer dependency since some of our components rely on those fonts and global styles in order to look right.

bkjohnson: Yeah, things will still look fine because [storybook is building with `formation`](https://github.com/department-of-veterans-affairs/component-library/blob/6e3b36cccb5d6745acb2705d4b67366abdcc35df/packages/storybook/.storybook/preview.js#L1-L3). If I remove `formation` from storybook yarn warns me about it:

peer dependency warning in storybook

And if I actually run storybook without `formation`, it doesn't look right:

image

If I remove `formation` from `vets-website` I get a similar warning:

peer dependency warning in vets-website

So having the peer dependency in place will give us warnings when they aren't met and can be useful for debugging if something isn't working as expected.

pdavies88: @bkjohnson thanks for the info that makes a lot of sense I appreciate it. I added back in the peerDependency and bumped the version to the latest to keep everything in line.

bkjohnson (approved):

Issue #361, opened 2022-04-11
Add internationalization capability - va-accordion


k80bowman:
## Description
This ticket is dependent on #343.

Once we have received translations, we need to update the va-accordion component to accommodate internationalization.

## Details
See the [i18next demo PR](https://github.com/department-of-veterans-affairs/component-library/pull/346) for internationalization infrastructure and implementation examples.

## Acceptance Criteria
- [ ] va-accordion component correctly displays translated text when appropriate
- [ ] tests have been written/updated to cover this functionality

PR #367, opened 2022-04-14
Translate accordion

bkjohnson:
## Chromatic

https://translate-accordion--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/component-library/issues/361

We're still waiting on translations, but this gets everything else in place. This would make storybook less useful for React components due to how the `i18next` dependency setup works combined with us having imports like these in storybook:

https://github.com/department-of-veterans-affairs/component-library/blob/84c7a4a35a78fb3db72b79e794fb6b923dbee993/packages/storybook/stories/Date.stories.jsx#L2

There's also some awkwardness with tying the component's definition to the tests, but that shouldn't affect any team other than the maintainers of the component library.

## Testing done

- Storybook :eyes: 

## Screenshots

Button triggering a `lang` change on `<main>` which triggers the component to re-render with different text

## Acceptance criteria
- [ ]

## Definition of done


packages/web-components/src/components/va-accordion/va-accordion.tsx
```diff
@@ -8,8 +8,16 @@ import {
...
+  forceUpdate
 } from '@stencil/core';
 import { getSlottedNodes } from '../../utils/utils';
+import i18next from 'i18next'
+import { Build } from '@stencil/core';
+
+if (Build.isTesting) {
+  // Make i18next.t() return the key instead of the value
+  i18next.init({ lng: 'cimode' });
+}
```
bkjohnson, line 20: I'm not a huge fan of this but I still haven't found a way around it that will let us at least have something  in the tests.

packages/storybook/stories/va-accordion.stories.jsx
```diff
@@ -73,6 +73,41 @@ const TemplateSubheader = args => {
   );
 };
 
+const I18nTemplate = args => {
+  const { headline, level, ...rest } = args;
+  const [lang, setLang] = useState('en');
+
+  useEffect(() => {
+    document.querySelector('main').setAttribute('lang', lang);
+  }, lang);
```
pdavies88, line 82: Getting a local error that syntax should be `}, [lang]);`

pdavies88:
In SB when clicking on Espanol vs English it changes all the stories with the components Expand and Collapse all text. I know in a prod environment we'd only have a single main tag but do we want to do an e.target.closest for SB so that only the one with the buttons would change?

bkjohnson:
When I test this in storybook I go to the "Canvas" tab and refresh the page so that I know I'm only working with the single iframe. Can you show me what you mean by `e.target.closest`?

pdavies88:
> When I test this in storybook I go to the "Canvas" tab and refresh the page so that I know I'm only working with the single iframe. Can you show me what you mean by `e.target.closest`?

Cool that makes sense then I usually work within the "Docs" tab that has all of the Stories on the same page but yeah if we use "Canvas" it works totally fine. I was thinking if we want to have it work with Docs we could do something like:
```
   const langAdjust = (e, lang) => {
     setLang(lang)
     e.target.closest('main').setAttribute('lang', lang)
   }
  ```
  to just switch the attribute on the story you are working with

pdavies88:
Also getting a little flakiness when trying to use the language switcher right away the buttons don't activate the text. 
Steps Reproduce:
1. Click on Chromatic link: https://60f9b557105290003b387cd5-ohrwnaucjv.chromatic.com/?path=/story/about-introduction--page
2. Go to va-accordion component and scroll to the bottom where the Internationalization lives and click on the buttons (not firing the text change or lang attribute change)
3. Click on Internationalization in the Side Nav and then click on the Canvas tab at the top of the page and then click the buttons (not firing the text change but the lang attribute updates correctly)
4. Refresh your page you are still on Canvas and Internationalization click the buttons and everything works correct

packages/storybook/stories/va-accordion.stories.jsx
bkjohnson: Can you show that screenshot? I'm not getting it so I'm having a hard time understanding what's wrong.

pdavies88: Screen Shot 2022-04-15 at 6 28 31 PM

I don't think the changing language buttons not working is related. This is more of a React useEffect warning that we should adjust https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect

bkjohnson: Ohhhh, I see - so we need `[lang]` instead of `lang` for that second arg. I'll make the update, good catch :+1:

pdavies88 (approved):

andresriveratoro:
@pdavies88 This should [fix](https://github.com/department-of-veterans-affairs/component-library/pull/385) the flakiness. It looks like the main element loads earlier in `Canvas` than it does in `Docs`.

andresriveratoro (approved):

PR #368, opened 2022-04-15
va-modal: Replace focus trap code

Mottie:
## Chromatic

## Description

The `va-modal` was using the `focus-trap` library to keep focus within the modal, but the library does not currently support targeting form elements within the shadow DOM. There is some beta work in progress, but while attempting to implement it, it did not work as expected.

This alternate method may not be ideal as it monitors the keyboard events and then checks the focused element. If the element is outside the modal, it puts focus on the modal close button. This method does not need to search for and detect focusable elements, so using <kbd>Shift</kbd> + <kbd>Tab</kbd> will keep focus on the close button.

This method was tested using Voice over and it does try to start speaking the text of the focused element outside the modal, but then focus is shifted to the close button and that text is immediately read

Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/39081

## Testing done

Updated e2e test

## Screenshots

N/A

## Acceptance criteria
- [x] Focus is trapped inside the open `va-modal`
- [x] All tests pass

## Definition of done


pdavies88:
@k80bowman we should probably loop in @CourtneyChambers77 and have her retest this for accessibility. I want to make sure that this methodology still allows us to pass as I know the whole focus-trap work that you did way back when was because of an accessibility violation.

Mottie:
I was chatting with Andrés about this work... it's not ideal. Using <kbd>Shift</kbd> + <kbd>Tab</kbd> maintains focus on the close button instead of looping. And using <kbd>Tab</kbd> moves focus outside the modal for a split second before returning it to the close button. I'll try to investigate another method.

pdavies88:
> I was chatting with Andrés about this work... it's not ideal. Using Shift + Tab maintains focus on the close button instead of looping. And using Tab moves focus outside the modal for a split second before returning it to the close button. I'll try to investigate another method.

If this method works and is not an accessibility issue then we can most likely move forward. But I'm not an accessibility expert which is why I am thinking we should loop in @SarahKay8 and @CourtneyChambers77 to check it out.

k80bowman:
@Mottie yeah, that does not sound ideal. If you can find an alternative method that works better, that would be great. I'd recommend checking out Ionic's [modal web component](https://github.com/ionic-team/ionic-framework/tree/main/core/src/components/modal) to see what they are doing.

Mottie:
Oof, yeah, it looks like [ionic is getting all the focusable items](https://github.com/ionic-team/ionic-framework/blob/main/core/src/utils/overlays.ts#L85) but they are including their own `.ion-focusable` class.

I think I'll try using this [trap tab method from a11y-dialog](https://github.com/KittyGiraudel/a11y-dialog/blob/main/a11y-dialog.js#L338-L358) which looks like it will need some modification to work with shadow DOM.

SarahKay8:
Currently taking a look

SarahKay8:
Apologies for the delayed response team 

After checking the link you sent to me, I believe the  issue with shift + Tab is the same Tab moving the focus out of the modal momentarily, then focusing the modal content immediately.

An approach I have used in the past, is to use place an empty div before and after the the modal contents. These divs have a tabIndex of 0, when they're focused, I manually move focus to the expected focusable element using a combination of http://event.target/ and event.relatedTarget

It was a poor man’s focus trap

I'll also be off on vacation the 18th-21st just a heads up..




@Mottie @pdavies88 @k80bowman

SarahKay8:
@shiragoodman

Mottie:
I have found a fix, and I'll replace the code in this PR after I've verified and updated testing.

Mottie:
I've updated this branch (forced push).

I dropped the `focus-trap` dependency and implemented a combination of the pattern from [ionic's overlay `trapKeyboardFocus` code](https://github.com/ionic-team/ionic-framework/blob/main/core/src/utils/overlays.ts#L133), which is called on `focus`, and the [`trapTabKey` code from a11y-dialog](https://github.com/KittyGiraudel/a11y-dialog/blob/main/a11y-dialog.js#L338).

The only "minor" issue I encountered with this implementation is that focus does move outside the modal for a instance, but not long enough for the screenreader to pick it up, but it does scroll the background.

k80bowman:
Thank you @Mottie! I'll take a look this afternoon.

andresriveratoro:
This seems to be working perfectly. I've tried it out with VoiceOver and it's only reading out elements inside the modal. I haven't noticed the minor issue you described.

packages/web-components/src/utils/utils.ts
```diff
@@ -48,3 +48,18 @@ export function consoleDevError(message: string): void {
...
+ * https://github.com/KittyGiraudel/focusable-selectors/blob/main/index.js
+ */
+export const focusableQueryString = [
+  '.hydrated:not([tabindex^="-"])',
+  '[tabindex]:not([tabindex^="-"])',
+  'input:not([type=hidden]):not([tabindex^="-"])',
+  'textarea:not([tabindex^="-"])',
+  'button:not([tabindex^="-"])',
+  'select:not([tabindex^="-"])',
+].join(',');
```
k80bowman, line 65: Could you please move this to a modal.ts or modal-focus.ts file in the utils directory? I'd like to keep this file reserved for functions that may be used by multiple components.

k80bowman:
This works great for me, too. @SarahKay8 could we get you to retest this and make sure it functions well from an accessibility standpoint?

packages/web-components/src/utils/utils.ts
Mottie: Done.

SarahKay8:
> This works great for me, too. @SarahKay8 could we get you to retest this and make sure it functions well from an accessibility standpoint?

Will do! Will have feedback (if any) by midday today

SarahKay8:
@k80bowman @Mottie Everything looks good to go from an A11y standpoint.

pdavies88 (approved):

k80bowman:
Thank you @SarahKay8! Good work @Mottie!

k80bowman (approved):

andresriveratoro (commented):
This is minor, but could we remove the `closeButtonContainer` variable and the `<div>` around the closeButton? We don't need those anymore with your solution. Thanks for your work on this @Mottie !

packages/web-components/src/index.html
```diff
@@ -18,6 +18,43 @@
 
   </head>
   <body>
-    <my-component first="Stencil" last="'Don't call me a framework' JS"></my-component>
```
andresriveratoro, line 21: I apologize for not mentioning this earlier, but could we revert the changes to this file? 😅

Mottie: oops! sorry! Done

andresriveratoro (approved):
Legend!

Mottie:
Should I bump the versions now?

Issue #358, opened 2022-04-11
Add internationalization capability - Select and va-select


k80bowman:
## Description
This ticket is dependent on #343.

Once we have received translations, we need to update the Select and va-select components to accommodate internationalization.

## Details
See the [i18next demo PR](https://github.com/department-of-veterans-affairs/component-library/pull/346) for internationalization infrastructure and implementation examples.

## Acceptance Criteria
- [x] Select component correctly displays translated text when appropriate
- [x] va-select component correctly displays translated text when appropriate
- [x] tests have been written/updated to cover this functionality

PR #389, opened 2022-04-22
Add i18next to Select and va-select

andresriveratoro:
## Chromatic

https://358-select-i18n--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/component-library/issues/358

## Screenshots
Screen Shot 2022-04-22 at 12 13 34

## Acceptance criteria
- [x] Select component correctly displays translated text when appropriate
- [x] va-select component correctly displays translated text when appropriate
- [x] tests have been written/updated to cover this functionality

## Definition of done


pdavies88 (approved):
Thanks for doing the SB fixes for this one

packages/storybook/.storybook/preview-body.html
```diff
@@ -0,0 +1,2 @@
+<!-- Reliable element for i18next mutation observer -->
+<main lang="en"></main>
```
bkjohnson, line 2: I'm not sure that we want this because our actual story doesn't go inside of it:

empty `<main>` element

packages/web-components/src/components/va-select/va-select.tsx
```diff
@@ -3,13 +3,21 @@ import {
...
   State,
   h,
 } from '@stencil/core';
+import i18next from 'i18next';
+import { Build } from '@stencil/core';
 import { getSlottedNodes } from '../../utils/utils';
 
+if (Build.isTesting) {
+  // Make i18next.t() return the key instead of the value
+  i18next.init({ lng: 'cimode' });
```
bkjohnson, line 18: We could either remove this or add a test that checks for the `required` key.

packages/web-components/src/components/va-select/va-select.tsx
```diff
@@ -71,6 +79,12 @@ export class VaSelect {
 
   @State() options: Array<Node>;
 
+  connectedCallback() {
+    i18next.on('languageChanged', () => {
```
bkjohnson, line 83: To match the React component we could have a `i18next.off('languageChanged')` in a `disconnectedCallback`.

packages/react-components/src/components/Select/Select.unit.spec.jsx
```diff
@@ -200,7 +200,8 @@ describe('<Select>', () => {
         onValueChange={() => {}}
       />,
     );
-    expect(tree.find('label').text()).to.equal('my label(*Required)');
+    expect(tree.find('label').text()).to.include('my label(*');
```
bkjohnson, line 203: We may want to have the dependency initialized to the `cimode` language in a `before` hook so this is a bit less weird.

packages/storybook/.storybook/preview-body.html
andresriveratoro: This is what removes the flakiness observed in `Docs` pages. Without this, sometimes Storybook doesn't attach the i18next event listener because it doesn't find a main element.

packages/web-components/src/components/va-select/va-select.tsx
andresriveratoro: Good call 👍

packages/react-components/src/components/Select/Select.unit.spec.jsx
```diff
@@ -200,7 +213,8 @@ describe('<Select>', () => {
         onValueChange={() => {}}
       />,
     );
-    expect(tree.find('label').text()).to.equal('my label(*Required)');
+    expect(tree.find('label').text()).to.include('my label(*Required)');
```
bkjohnson, line 216: Now that we're doing the `i18next.init` we can probably leave the `equal`:

```suggestion
    expect(tree.find('label').text()).to.equal('my label(*Required)');
```

packages/storybook/stories/Select.stories.jsx
```diff
@@ -67,3 +67,29 @@ WithAnalytics.args = {
...
+    <div>
+      <button onClick={e => setLang('es')}>Español</button>
+      <button onClick={e => setLang('en')}>English</button>
+      <Select {...args} value={value} onValueChange={onValueChange} />
+    </div>
+  );
+};
+
+export const Internationalization = I18nTemplate.bind({});
+Internationalization.args = {
```
bkjohnson, line 92: This may be happening due to the `Select` that is being imported. I haven't found a way for us to have the prop docs _and_ the internationalization functionality for the React components.



"undefined" value in UI

bkjohnson, line 92: This may be happening due to the `Select` that is being imported. I haven't found a way for us to have the prop docs _and_ the internationalization functionality for the React components.



"undefined" value in UI

andresriveratoro: This issue might be resolved after we use a bundler for `react-components`. I can create a ticket to re-add it once that work has been completed.

bkjohnson: I think removing it would be good. Since we've got `<va-select>` we may be able to get the remaining 7 `<Select>` components migrated and be able to remove these stories completely.

packages/web-components/src/components/va-select/va-select.tsx
bkjohnson: This is the last thing I think we should address before merging. Right now it doesn't look like we need it since there are no changes to the e2e.ts file - should we remove it or add a test where it will be necessary?

bkjohnson (commented):
One last thing.

bkjohnson (approved):

PR #391, opened 2022-04-22
Update va-search-input to emit the correct target value

andresriveratoro:
## Chromatic

https://typeahead-value--60f9b557105290003b387cd5.chromatic.com

## Description
These changes allow consumers to retrieve the updated value on `input` and `submit` events from `event.target.value`.

## Screenshots

https://user-images.githubusercontent.com/36863582/165602408-81424ce7-fbe6-4fd3-8d94-371b070f6cde.mov

## Acceptance criteria
- [x] `va-search-input` `input` and `submit` events emit value of input field in `event.target.value` 

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-search-input/va-search-input.tsx
+++ b/packages/web-components/src/components/va-search-input/va-search-input.tsx
@@ -149,6 +149,8 @@ export class VaSearchInput {
         this.isListboxOpen = false;
         break;
       default:
+        if (event.key.length !== 1) return;
+        this.value = this.value + event.key;
         break;
     }
   };
--- a/packages/web-components/src/components/va-search-input/va-search-input.tsx
+++ b/packages/web-components/src/components/va-search-input/va-search-input.tsx
@@ -48,6 +48,8 @@ export class VaSearchInput {
    * The value of the input field
    */
   @Prop({ mutable: true, reflect: true }) value?: string = '';
+  // $('va-search-input').value will be correct
+  // $('va-search-input').getAttribute('value') will be incorrect
 
   /**
    * If suggestions are provided, then format suggestions and open the listbox.
--- a/packages/web-components/src/components/va-search-input/va-search-input.tsx
+++ b/packages/web-components/src/components/va-search-input/va-search-input.tsx
@@ -59,6 +59,15 @@ export class VaSearchInput {
   }
 
   /**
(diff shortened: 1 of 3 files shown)


packages/web-components/src/components/va-search-input/va-search-input.tsx
```diff
@@ -253,6 +255,8 @@ export class VaSearchInput {
         );
         break;
       default:
+        if (event.key.length !== 1) return;
+        this.value = this.value + event.key;
```
andresriveratoro, line 259: I don't like this approach because it requires us to code additional functionality to keep the value of `value` in sync with what the user is seeing. For example, if a user types `for`, selects `for` with mouse, presses backspace and `x`, it causes the input field value to change to `forx`.

andresriveratoro: We would also need to handle key combinations like command+delete and any other combinations that affect the input field.

pdavies88 (approved):

packages/web-components/src/components/va-search-input/va-search-input.tsx
```diff
@@ -58,6 +60,15 @@ export class VaSearchInput {
     this.updateSuggestions(this.suggestions);
   }
 
+  /**
+   * Fixes issue where submit event dispatches the initial value of value
+   * instead of the current value of the input field.
+   */
+  @Watch('value')
+  watchValueHandler() {
+    this.value = this.inputRef.value;
```
bkjohnson, line 69: How is this different than the `this.value` assignment in the `handleInput` function?

andresriveratoro: This is a very hacky solution that reconciles `this.value` with what's actually present in the input field. When this component re-renders, `this.value` resets to the initial value of `value`. You can observe this in the DOM by inspecting the element, typing "for" and wait for the component to receive a list of new suggestions. When the component receives the new suggestions, `event.target.value` is reset. This has been the only way that I've been able to make sure that `event.target.value` is correct for the `submit` event. I believe this is happening because we're using `value` like a state variable when it's really a prop, but I haven't found any documentation to really explain it. The assignment in `handleInput` happens before it calls `updateSuggestions` that triggers a re-render when `formattedSuggestions` are updated.

bkjohnson (approved):

PR #393, opened 2022-04-25
Update success status for va-text-input

andresriveratoro:
## Chromatic

https://39929-status--60f9b557105290003b387cd5.chromatic.com

## Description
Part of department-of-veterans-affairs/vets-design-system-documentation#773

## Testing done
E2E, Storybook

## Screenshots
Screen Shot 2022-04-25 at 11 21 40
Screen Shot 2022-04-25 at 11 21 33

## Acceptance criteria
- [x] Replace `status` string prop with `success` boolean prop

## Definition of done


pdavies88 (approved):

packages/web-components/src/components/va-text-input/va-text-input.css
```diff
@@ -58,6 +58,6 @@ input:not([disabled]):focus {
   border: 2px solid var(--color-secondary-dark);
 }
 
-:host([status='success']) input {
+:host([success='true']) input {
```
bkjohnson, line 61: Could we use the [`:valid` pseudo-selector](https://developer.mozilla.org/en-US/docs/Web/CSS/:valid) to apply the styling?

bkjohnson (commented):
Can we document when it's a good idea to use `success`? My guess is that it would only be used for more complex validation that depends on other data - things that `required`, `minlength`, and `pattern` might not be able to cover.

packages/web-components/src/components/va-text-input/va-text-input.css
andresriveratoro: We could update it to `:valid`, but if the input isn't required it will display a green border. This would be a design change for any `va-text-input` that is optional. If we do want this, can we implement it or do we have to run it by DSC?

https://css-tricks.com/almanac/selectors/v/valid/#:~:text=The%20%3Avalid%20selector%20allows%20you,an%20evaluation%20of%20user%20input.

andresriveratoro: It seems like we may want to add an additional indicator of the status besides color - maybe an icon?

https://developer.mozilla.org/en-US/docs/Web/CSS/:valid#accessibility_concerns

GnatalieH: Interesting, @andresriveratoro. I assume you are referring to the success state? The error state has the error message which is an indicator.

andresriveratoro: Exactly. Our error states (in this component and others) seem fine to me because they include text/icon in addition to the border color change, but for the success state we have just the border.

bkjohnson: We don't have to deal with `:valid` here since it sounds like there's more design discussion to be had.

k80bowman:
@andresriveratoro is there work that still needs to be done on this?

andresriveratoro:
> @andresriveratoro is there work that still needs to be done on this?

I'm so sorry for letting this fall through the cracks. Our documentation needs to be updated to provide guidance for when to use `success`.

PR #396, opened 2022-04-27
Fix for JSDoc usage of <input> in docs

pdavies88:
## Chromatic

https://jsdoc-input-element-fix--60f9b557105290003b387cd5.chromatic.com

## Description
When using `<input>` in JSDoc it converts from a code block to a HTML Element. Changing our comments for the docs to type out input element instead of putting the comment in as a `<input>`.

## Fixes
Screen Shot 2022-04-27 at 9 39 45 AM

## Acceptance criteria
- [X] Fix all places where we use `<input>` in our JSDoc comments

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@department-of-veterans-affairs/component-library",
   "description": "VA.gov component library. Includes React and web components.",
-  "version": "8.6.2",
+  "version": "8.6.3",
   "license": "MIT",
   "scripts": {
     "build": "webpack"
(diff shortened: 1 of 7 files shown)


k80bowman (approved):
Thanks for fixing this! Are there any other element references like this?

pdavies88:
@k80bowman these were the only 2 that I could find remaining

k80bowman:
Perfect. Thank you!

pdavies88:
I also added a fix for va-accordion-item as well just in case in the future we decide to give it it's own page in the docs we would not run into an issue then

PR #404, opened 2022-05-03
Add va-alert-expandable component

longmd:
## Description
This PR creates an expandable, mobile-friendly alert that combines the "Additional Info" component with the "Background-Only Alert" color schemes, and includes an icon based on the status type. This component had previously been created and used by another team, but never finished the process of becoming a mature component and also posed some questions regarding color contrast and consistent mark-up with that of currently-accepted component. This PR takes the accepted VA-Additional-Info component mark-up and styling and adds the accepted color scheme and icons of the background-only styling of the VA-Alert component. 

## Original Issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/39536

## Testing done
- [x] e2e Tests

## Screenshots
- Screen Shot 2022-05-03 at 9 34 18 AM
- Screen Shot 2022-05-03 at 9 35 04 AM

## Acceptance criteria
- [ ] New component successfully combines already-accepted components to satisfy laid out specs for the component 

## Definition of done


pdavies88:
@longmd can we add a Storybook story for this component as well.

packages/web-components/src/components/va-alert-expandable/va-alert-expandable.tsx
```diff
@@ -0,0 +1,150 @@
...
+export class VaAlertExpandable {
+  @Element() el: HTMLElement;
+
+  @State() open: boolean;
+
+  /**
+   * Determines the icon and background color.
+   * One of `info`, `error`, `success`, `warning`, or `continue`
+   */
+  @Prop() status: string = 'info';
```
bkjohnson, line 28: This isn't crucial (and I know the `<va-alert>` doesn't do this yet), but instead of a plain `string` type can we define those options like this?

https://github.com/department-of-veterans-affairs/component-library/blob/6ee45b200d8be4a507b38dd1b2951d41f81e38c9/packages/web-components/src/components/va-modal/va-modal.tsx#L120

packages/web-components/src/components/va-alert-expandable/va-alert-expandable.tsx
```diff
@@ -0,0 +1,150 @@
...
+    // the additional 2em is #alert-body margin-top and margin-bottom when open
+    bodyElm.style.setProperty(
+      '--calc-max-height',
+      'calc(' + contentHeight + ' + 2rem)',
+    );
+  }
+
+  componentDidLoad() {
+    this.updateAlertBodyMaxHeight();
+    this.vaComponentDidLoad.emit();
```
bkjohnson, line 112: Why is this happening?

packages/web-components/src/components/va-alert-expandable/va-alert-expandable.tsx
```diff
@@ -0,0 +1,150 @@
...
+    const alertClasses = classnames('alert-expandable', status, { 'hide-icon': !showIcon });
+    const bodyClasses = classnames('alert-expandable-body', { 'open': open, 'closed': !open });
+    const role = status === 'error' ? 'alert' : null;
+    const ariaLive = status === 'error' ? 'assertive' : null;
+
+    if (!visible) return <div aria-live="polite" />;
+
+    return (
+      <Host>
+        <div role={this.el.getAttribute('data-role') || role} aria-live={ariaLive} class={alertClasses}>
```
bkjohnson, line 126: Do we need to set the role with `data-role`? If someone were to do this:

```html
<va-alert-expandable role="something"></va-alert-expandable>
```

That `role` attribute would essentially end up on the `<Host>`

So we might be able to do something like this to override it when we have an error:

```tsx

<Host role={status === 'error' ? 'alert' : this.el.getAttribute('role')}>

Comment (longmd) [reply packages/web-components/src/components/va-alert-expandable/va-alert-expandable.tsx:126]: I'll let you all guide this. This setup is currently how `va-alert` currently operates. I would assume whatever is done on one component should be done on the other. If we did move the role up to the host, we would need to move the aria-live attribute, as well.

File: packages/web-components/src/components/va-alert-expandable/va-alert-expandable.tsx (hunk 2 above)
Comment (longmd) [reply packages/web-components/src/components/va-alert-expandable/va-alert-expandable.tsx:112]: This emit is in `va-alert` as well, so I added it in here.

File: packages/web-components/src/components/va-alert-expandable/va-alert-expandable.tsx (hunk 3 above)
Comment (pdavies88) [reply packages/web-components/src/components/va-alert-expandable/va-alert-expandable.tsx:126]: @bkjohnson yes this was a specific use case due to banner vs alert.
Here's the details, for all Alerts we have an aria-role of alert however with the Banner component we wanted to change that to aria-role of banner since it is a better indicator of what the component actually is from an accessibility perspective. Since we nest a va-alert inside of va-banner I needed a way to pass down the correct role depending on the component we were using. 
Moral of all this is we don't need this.el.getAttribute('data-role') here instead just set the aria-role we want to use here ie: role={role}

**Comment by longmd** [comment]
@pdavies88 , Storybook story has been added.

**Review by pdavies88 — APPROVED** [review]
Thanks for updating!

File: packages/web-components/src/components/va-alert-expandable/va-alert-expandable.tsx [hunk 4]
```diff
@@ -0,0 +1,138 @@
...
+
+  /**
+   * If true, doesn't fire the CustomEvent which can be used for analytics tracking.
+   */
+  @Prop() disableAnalytics: boolean = false;
+
+  /**
+   * If false, the status icon is removed.
+   */
+  @Prop() showIcon: boolean = true;
```
bkjohnson, line 43: I think this default should be flipped. We want people to be using the components like this:

```html
<va-alert-expandable show-icon /> <!-- With icon -->
<va-alert-expandable /> <!-- Without icon -->
```

Rather than:

```html
<va-alert-expandable /> <!-- With icon -->
<va-alert-expandable show-icon="false" /> <!-- Without icon -->
```

This is about what is considered a [valid HTML boolean attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#boolean_attributes):

> To be clear, the values "true" and "false" are not allowed on boolean attributes. To represent a false value, the attribute has to be omitted altogether.

If the icon should be part of the default look, then the prop name can change to reflect the default falsey behavior, like:

```html
<va-alert-expandable /> <!-- With icon -->
<va-alert-expandable iconless /> <!-- Without icon -->
<!-- Where `iconless` or whatever the prop is named is defaulted to false -->
```

packages/web-components/src/components/va-alert-expandable/va-alert-expandable.tsx
```diff
@@ -0,0 +1,138 @@
...
+    );
+  }
+
+  componentDidLoad() {
+    this.updateAlertBodyMaxHeight();
+  }
+
+  render() {
+    const { status, open, showIcon, visible } = this;
+    const alertClasses = classnames('alert-expandable', status, { 'hide-icon': !showIcon });
```
bkjohnson, line 105: This boolean logic may have to change as well.

bkjohnson (changes requested):
The PR template should have added a link to chromatic but for some reason it's not there. It also looks like the branch name is too long for it to work anyway, so for future contributions can you [choose a branch name that is less than 37 characters](https://github.com/department-of-veterans-affairs/component-library#branch-naming)? That chromatic link gives us an easy way to see the most up-to-date changes from the PR in storybook.

The "Changes requested" is about how `showIcon` is being setup and used. Removing the `visible` prop would just be a "nice to have"

bkjohnson (approved):

k80bowman (approved):

packages/web-components/src/components/va-alert-expandable/va-alert-expandable.tsx
```diff
@@ -0,0 +1,131 @@
...
+  toggleOpen(): void {
+    // Conditionally track the event.
+    if (!this.disableAnalytics) {
+      this.componentLibraryAnalytics.emit({
+        componentName: 'va-alert-expandable',
+        action: !this.open ? 'expand' : 'collapse',
+        details: {
+          triggerText: this.trigger,
+        },
+      });
```
andresriveratoro, line 70: I believe we need to reach out to Brian to track this and add it on vets-website [component-library-analytics-setup](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/platform/site-wide/component-library-analytics-setup.js).

longmd: @andresriveratoro , is that something you all do as part of onboarding a new component, or is it something that I need to reach out and do prior to?

pdavies88: @k80bowman @bsmartin-ep what is the way that you would like to have this event added to GTM? Should we do this as a separate ticket?

andresriveratoro: If we end up leaving this work to the contributor, we should update our README or include this information somewhere more easily accessible than our Confluence.

k80bowman: @longmd @andresriveratoro can help with adding analytics. Everything else should be good to go here. Thank you for working on this!

longmd: @andresriveratoro , does Brian need to do the setup on their end before I add to the events setup object in `vets-website`?

andresriveratoro: @longmd Brian has to setup the analytics in Google as well for it to work properly, but you're good to go ahead and merge this PR. I'll create a ticket to update vets-website and assign it to you. Thanks!

PR #408, opened 2022-05-05
Fix for va-table not handling data updates properly

jerekshoe:
## Chromatic

## Description
Fixes an issue with the va-table not accounting for data updating: https://codesandbox.io/s/goofy-fire-y7599h

## Testing done
Tests still pass

### Before
Screen Shot 2022-05-05 at 2 59 25 PM

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-table/va-table.tsx
+++ b/packages/web-components/src/components/va-table/va-table.tsx
@@ -4,6 +4,11 @@ import ascendingIcon from '../../assets/sort-arrow-up.svg?format=text';
 import descendingIcon from '../../assets/sort-arrow-down.svg?format=text';
 import { quicksort, reverseQuicksort } from '../../utils/dom-sort';
 
+// For IE11 compatibility. `el.children` renders booleans instead of html elements,
+// so instead we use `el.childNodes` and filter out nodes that aren't a proper tag
+const elementChildren = (el: HTMLElement) =>
+  Array.from(el.childNodes).filter(node => node.nodeName !== '#text');
+
 /**
  * This component expects `<va-table-row>` elements as children.
  * Children of each row element should be `<span>` elements. Table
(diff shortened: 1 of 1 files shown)


packages/web-components/src/components/va-table/va-table.tsx
```diff
@@ -132,6 +124,25 @@ export class VaTable {
...
 
+  private handleSlotChange(e): void {
+    const rows = e.target.assignedElements();
+
+    rows.forEach(row => {
+      const cells = elementChildren(row);
+
+      cells.forEach((cell: HTMLSpanElement, colNum) => {
+        if (isNumeric(cell.textContent)) {
+          cell.classList.add('medium-screen:vads-u-text-align--right');
```
bkjohnson, line 135: Duplicating this isn't great because it open up the door for a cell and header on the same column to have different alignment. The `alignment` object existed to ensure that cells and headers had the same alignment.

jerekshoe: Yeah the problem I have run into is that `handleSlotChange` is triggered on first load of the component, but it is fired before `componentDidLoad`, so even if I moved `alignment` like I did to the `columns` array, it would not be populated when `handleSlotChange` is first ran. It may be possible to make a separate handler for the `onSlotChange` of the 'headers' slot that would populate both `alignment` and `columns`, but I'm not sure if there is a possibility of there being a timing issue there.

bkjohnson: Having a `slotchange` handler to handle the pagination use case does seem like the right direction, but I haven't dug deeper into this to see what other changes would need to happen.

k80bowman: @bkjohnson do you think this is something that warrants a ticket so we have dedicated time to figure out a solution and test it?

micahchiang:
@jamigibbs - Do you know if this PR can be closed? Wondering if the work done in #717 and #729 has addressed the original issue surfaced here.

Issue #343, opened 2022-04-01
Audit components to identify internationalization needs for the Check In team and request translations


bkjohnson:
## Description

Our components have some hard-coded English text. The Check In team has requested that we enable translations for several of our components, so we need to identify everything internal to those components which will need a Spanish translation.

Once we have determined what text needs to be translated, we need to reach out to the Check In team to get the translations from them.

## Details
List of components
- Date (and va-date, since that is in development)
    - Select (and va-select) - used as part of Date (and va-date)
    - NumberInput (and va-number-input, since that is in development)
- va-text-input
- va-loading-indicator
- va-alert
- va-telephone
- va-accordion

It might be helpful to put all of this in a spreadsheet or Confluence table.

## Acceptance criteria
- [ ] Obtain a list of all hard-coded English text for sighted users
- [ ] Obtain a list of all hard-coded `aria-label`s for screen readers
- [ ] Get translations from Check In team

k80bowman:
Good catch I missed that one. Let me ask them how they would like to handle that situation, because it is the same with the va-alert component.

k80bowman:
They are just going to override the value for va-loading-indicator and va-alert. I let them know which props they would need to pass in and made sure they understood that if they don't the text will be in English.

k80bowman:
Spreadsheet with hard-coded values: https://docs.google.com/spreadsheets/d/11a5nfSNhFrju_qLRWNM71cb86oZVe4L4K826xNDDDU8/edit#gid=0

k80bowman:
[Passed](https://dsva.slack.com/archives/C01DBGX4P45/p1649860309441689?thread_ts=1649781669.566199&cid=C01DBGX4P45) spreadsheet to the check-in team.

caw310:
Translations expected 4/27 at the earliest.

k80bowman:
Received initial set of [translations](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/products/health-care/checkin/translations/SPA-FINAL_CIE%20in-app%20content.json). Closing

PR #411, opened 2022-05-06
Added translation to alert aria-label on close button.

brianseek:
## Chromatic

https://translate-alert--60f9b557105290003b387cd5.chromatic.com

## Description
Closes [va.gov-team/39886](https://app.zenhub.com/workspaces/check-in-experience-61fc23a2cb8a14001132e102/issues/department-of-veterans-affairs/va.gov-team/39886)

Is there a DS ticket related to this?

## Testing done
No testes needing updates
## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/storybook/stories/va-alert.stories.jsx
+++ b/packages/storybook/stories/va-alert.stories.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState, useRef } from 'react';
 import { generateEventsDescription } from './events';
 import { VaAlert } from '@department-of-veterans-affairs/web-components/react-bindings';
 import { getWebComponentDocs, propStructure } from './wc-helpers';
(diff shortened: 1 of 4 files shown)


brianseek:
@bkjohnson I ran into some issues with this one. Could be just me and my local, but the component in the story didn't seem to want to update on language change. Can you check to see if I am missing anything here?

packages/storybook/stories/va-alert.stories.jsx
```diff
@@ -177,6 +177,47 @@ const BackgroundOnlyTemplate = ({
...
+  const thisTemplate = useRef(null);
+  const [lang, setLang] = useState('en');
+
+  return (
+    <>
+      <button
+        onClick={() => {
+          const newLang = lang === 'en' ? 'es' : 'en';
+          setLang(newLang);
+          thisTemplate.current.closest('main').setAttribute('lang', newLang);
```
andresriveratoro, line 199: You might have to update this to `document.querySelector('main').setAttribute('lang', lang);` instead of targeting the closest `main` element in Storybook. It looks like it's possible where multiple `main` elements exist on a page in Storybook when there are multiple examples for a component. I suspect our [mutation observer](https://github.com/department-of-veterans-affairs/component-library/blob/master/packages/core/src/i18n/i18n-setup.js) on `main` is different than the closest `main`.

packages/web-components/src/components/va-alert/va-alert.tsx
```diff
@@ -49,7 +57,7 @@ export class VaAlert {
   /**
    * Aria-label text for the close button.
    */
-  @Prop() closeBtnAriaLabel: string = 'Close notification';
+  @Prop() closeBtnAriaLabel: string = i18next.t('close-notification');
```
bkjohnson, line 60: I don't think this is something that we want right now unless things have changed since [this comment](https://github.com/department-of-veterans-affairs/component-library/issues/343#issuecomment-1097210319).

> They are just going to override the value for va-loading-indicator and va-alert. I let them know which props they would need to pass in and made sure they understood that if they don't the text will be in English.

cc @k80bowman for awareness.

k80bowman: No, let's not do this one right now. We don't have the translated text for it.

brianseek: @bkjohnson & @k80bowman closing this PR for now then.

PR #419, opened 2022-05-10
Update stories for Button

andresriveratoro:
## Chromatic

https://39874-buttons--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/39874

## Screenshots
Screen Shot 2022-05-13 at 09 24 42
Screen Shot 2022-05-11 at 18 38 28
Screen Shot 2022-05-11 at 18 38 38

## Acceptance criteria
- [x] Storybook examples match the styling as seen in Sketch mockups
- [x] Add button pairs example

## Definition of done

### DIFF
--- a/packages/storybook/stories/buttons.stories.mdx
+++ b/packages/storybook/stories/buttons.stories.mdx
@@ -1,11 +1,12 @@
 import { Meta, Canvas } from '@storybook/addon-docs/blocks';
+import './buttons.style.scss';
 
 <Meta title="Components/Buttons" />
 
-## Default buttons
+## Primary buttons
 
 <Canvas>
-  <Story name="Default">
+  <Story name="Primary">
     <div>
       <button type="button" className="usa-button">
         Button
@@ -15,18 +16,6 @@ import { Meta, Canvas } from '@storybook/addon-docs/blocks';
       </button>
       <button type="button" className="usa-button usa-button-hover">
         Hover
-      </button>
-    </div>
-  </Story>
-</Canvas>
-
-## Primary buttons
-
-<Canvas>
-  <Story name="Primary">
-    <div>
-      <button type="button" class="usa-button-primary va-button-primary">
-        Button
       </button>
     </div>
   </Story>
@@ -44,19 +33,33 @@ import { Meta, Canvas } from '@storybook/addon-docs/blocks';
   </Story>
 </Canvas>
 
+## Continue buttons
+
+<Canvas>
+  <Story name="Continue">
+    <div>
+      <button type="button" class="usa-button-primary va-button-primary">
+        Button
+      </button>
+      <button class="usa-button-primary va-button-primary">Active</button>
+      <button class="usa-button-primary va-button-primary">Hover</button>
+    </div>
+  </Story>
+</Canvas>
+
 ## Big buttons
 
 <Canvas>
   <Story name="Big">
     <div>
       <button type="button" class="usa-button usa-button-big">
-        Default button
+        Primary button
       </button>
       <button
         type="button"
         class="usa-button-primary va-button-primary usa-button-big"
       >
-        Button
+        Continue button
       </button>
       <button class="usa-button-secondary usa-button-big">
         Secondary button
--- a/packages/storybook/stories/buttons.stories.mdx
+++ b/packages/storybook/stories/buttons.stories.mdx
@@ -1,5 +1,4 @@
 import { Meta, Canvas } from '@storybook/addon-docs/blocks';
-import './buttons.style.scss';
 
 <Meta title="Components/Buttons" />
 
@@ -26,9 +25,9 @@ import './buttons.style.scss';
 <Canvas>
   <Story name="Secondary">
     <div>
-      <button class="usa-button-secondary">Button</button>
-      <button class="usa-button-secondary usa-button-active">Active</button>
-      <button class="usa-button-secondary usa-button-hover">Hover</button>
+      <button className="usa-button-secondary">Button</button>
+      <button className="usa-button-secondary usa-button-active">Active</button>
+      <button className="usa-button-secondary usa-button-hover">Hover</button>
     </div>
   </Story>
 </Canvas>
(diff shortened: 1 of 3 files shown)


bkjohnson (commented):
I feel like this creates a pretty confusing situation. Our "Primary" button is using the default button class from USWDS:

"Primary" button in storybook

But the "Continue" button is using classes with `primary` in the name.

"Continue" button in storybook

 Do we have a plan for dealing with that? This means that a comment like "use the primary button" is very ambiguous. A developer may be talking about `va-button-primary`, but a designer may mean the "Primary" (default) button.

andresriveratoro:
I don't think we have a plan for this currently, but I will update the original issue to include this information. 

I also don't like that ProgressButton is nested inside Buttons. It's easily missed because it's not part of the same page and you can only reach it by clicking on the side navigation. 

The main difference, besides the obvious native element vs React component, between our regular buttons and ProgressButton is that the latter is potentially accessible IF an engineer provides values for aria-label or aria-describedby - there are no default values for these props. I know a lot of design systems include buttons with nothing more than styling including USWDS, but is this really what we want in ours? Accessibility should be baked in for buttons. A lot of things feel wrong here.

GnatalieH:
@bkjohnson now I see where the naming issues came from with buttons (the USWDS names). Default is not a good name for a button, because it's a state of a button (as well as a state for other elements). And Primary is not a good name for the green button, because we want to deprecate most use cases other than using a green button to sign in. I suggested the "Continue" button name because the green modal with the lock icon is called the Continue modal. I'm open to suggestions if @humancompanion-usds wants to change the names of the buttons.

bkjohnson:
>  I know a lot of design systems include buttons with nothing more than styling including USWDS, but is this really what we want in ours? Accessibility should be baked in for buttons.

I agree. I think there are going to situations where a plain `<button>Click me</button>` will be accessible in a given context, but if we know that the situation calls for accessibility enhancements or consideration then we should bake those into a design system component. The ProgressButton is kind of awkward (and is missing some defaults like you pointed out), but hard-coding `aria-hidden="true"` onto the character/icon indicating Next/Previous is a good example of this kind of enhancement. Not including that attribute means that a screen reader will say "left double angle bracket" or "right double angle bracket" which isn't very helpful.

bkjohnson:
Since we're headed in the web components direction, would an ideal interface look something like this?

- `<va-button>`
    - General purpose button
    - Blue background, white text
    - Possibly no specific aria attributes, developers can add their own as they see fit
- `<va-button next>`
    - Intended to be used for navigating to the next part of some workflow (forms)
    - Blue background, white text
    - icon
    - `aria-hidden` on the icon and defaults for `aria-label` and maybe `aria-describedby`
- `<va-button previous>`
    - Intended to be used for navigating to the previous part of some workflow (forms)
    - Blue _border_  and text, transparent background
    - icon
    - `aria-hidden` on the icon and defaults for `aria-label` and maybe `aria-describedby`
- `<va-button secondary>`
  - Same styling as the `next` variant
  - No icon or aria attributes
  
Even though I typically don't associate a form with the concept of pagination, the `next`/`previous` button variants seem similar to the buttons that we have in `<va-pagination>`, so maybe we could keep them in sync in terms of a11y if not visually.
    
I still don't quite understand the use cases for the "continue"/"sign in" button. Is it meant to go in the header here?

"Sign in" button in header

Or is it only meant to be in a "call to action" like when you're on a page that requires authentication to proceed?

"Sign in" button as a CTA

GnatalieH:
@bkjohnson no, the green Continue button is not meant to appear in the header. That's a good case for keeping the name of this button Continue, and not "Sign in". I named the Green button Continue because it is used for a CTA any time the user has to complete an action like signing in to continue with a process. This aligns with our Continue modal and our Continue alert.

packages/storybook/stories/buttons.stories.mdx
```diff
@@ -20,26 +21,14 @@ import { Meta, Canvas } from '@storybook/addon-docs/blocks';
...
-
 ## Secondary buttons
 
 <Canvas>
   <Story name="Secondary">
-    <div>
-      <button class="usa-button-secondary">Button</button>
-      <button class="usa-button-secondary usa-button-active">Active</button>
-      <button class="usa-button-secondary usa-button-hover">Hover</button>
+    <div name="Secondary Buttons">
```
bkjohnson, line 28: Adding the `name` here on the div might not be necessary. In the sass sample I posted I had the `[name]` selector because it looked like storybook was adding that already.

`name="Secondary"` attribute in markup for targeting

packages/storybook/stories/buttons.style.scss
```diff
@@ -0,0 +1,26 @@
+[name='Secondary Buttons'] {
+  & button:nth-child(2) {
+    background-color: #dce4ef;
+  }
+  & button:nth-child(3):focus {
```
bkjohnson, line 5: Can you add comments to these describing what's going on, so that people know that the order matters?

packages/storybook/stories/buttons.style.scss
```diff
@@ -0,0 +1,26 @@
...
+    border-color: #0071bb;
+    box-shadow: inset 0 0 0 2px #0071bb;
+    color: #0071bb;
```
bkjohnson, line 8: Can we try an import like this:

https://github.com/department-of-veterans-affairs/component-library/blob/239d71d99b99d586d7cab1ac9769f79eedb7e40d/packages/storybook/.storybook/style.scss#L4

And then use the sass variables for the colors? It looks like `$color-light-blue`.

bkjohnson (commented):
Some minor comments.

packages/storybook/stories/buttons.stories.mdx
andresriveratoro: This was required to get the styling to work properly on design.va.gov. Without it, it renders a div without a name and the styling is never applied.

packages/storybook/stories/buttons.stories.mdx
```diff
@@ -1,12 +1,13 @@
...
 
 <Meta title="Components/Buttons" />
 
-## Default buttons
+## Primary buttons
 
 <Canvas>
-  <Story name="Default">
-    <div>
+  <Story name="Primary">
```
andresriveratoro, line 9: This `Story` div doesn't render in iframes on design.va.gov

andresriveratoro: I know it shows up in Storybook, but it doesn't in the iframe 😢

packages/storybook/stories/buttons.style.scss
andresriveratoro: Added

packages/storybook/stories/buttons.style.scss
andresriveratoro: Updated to use sass variables

packages/storybook/stories/buttons.stories.mdx
bkjohnson: Oh good catch - I didn't check it in the iframe.

bkjohnson (approved):

PR #425, opened 2022-05-12
Focus only radio button for va-radio

andresriveratoro:
## Chromatic

https://40279-radio-halo--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/40279

## Testing done
E2E, Storybook

## Screenshots
Screen Shot 2022-05-13 at 01 12 35

## Acceptance criteria
- [x] Focus halo is visible only on radio button

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-radio-option/va-radio-option.css
+++ b/packages/web-components/src/components/va-radio-option/va-radio-option.css
@@ -1,34 +1,39 @@
-@import '../../mixins/inputs.css';
-
 :host {
+  --visible-radio-button-size: 1.4rem;
   display: block;
   margin-top: 1rem;
 }
 
+input {
+  height: var(--visible-radio-button-size);
+  margin: 0.5rem 0 0 2.4rem;
+  opacity: 0;
+  position: absolute;
+  width: var(--visible-radio-button-size);
+}
+
 label {
-  box-sizing: border-box;
-  outline: none;
-  display: inline-block;
-  margin: 0;
-  width: 100%;
-  padding-left: 2.4rem;
+  margin-left: 2.4rem;
 }
 
 label::before {
-  box-sizing: border-box;
+  background-color: var(--color-white);
   border-radius: 100%;
   box-shadow: 0 0 0 2px var(--color-white), 0 0 0 3px var(--color-gray-medium);
-  height: 1.4rem;
-  line-height: 1.4rem;
-  margin-right: 1rem;
-  width: 1.4rem;
-  background: var(--color-white);
-  background-color: var(--color-white);
   content: '\A0';
   display: inline-block;
+  height: var(--visible-radio-button-size);
+  line-height: var(--visible-radio-button-size);
+  margin-right: 1rem;
+  width: var(--visible-radio-button-size);
 }
 
-:host([aria-checked='true']) label::before {
+input:not([disabled]):focus + label::before {
+  outline: 2px solid var(--color-gold-light);
+  outline-offset: 6px;
+}
+
+:host([checked]) label::before {
   background-color: var(--color-primary);
   box-shadow: 0 0 0 2px var(--color-white), 0 0 0 4px var(--color-primary);
 }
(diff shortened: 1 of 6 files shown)


andresriveratoro:
I'm having an issue where `vaValueChange` is being emitted twice when selecting with a mouse.

packages/web-components/src/components/va-radio-option/va-radio-option.tsx
```diff
@@ -42,25 +45,24 @@ export class VaRadioOption {
...
-  }
+  };
 
   render() {
-    const ariaDescribedby = (this.checked && this.ariaDescribedby) || null;
+    const { checked, handleChange, label, name, value } = this;
+
     return (
-      <Host
-        aria-checked={this.checked ? `${this.checked}` : 'false'}
```
andresriveratoro, line 53: We shouldn't need [aria-checked](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-checked) now that we have a radio input, but I'll check with Sarah.

packages/web-components/src/components/va-radio-option/va-radio-option.tsx
```diff
@@ -42,25 +45,24 @@ export class VaRadioOption {
...
-        aria-describedby={ariaDescribedby}
```
andresriveratoro, line 54: See our [Slack discussion](https://dsva.slack.com/archives/G01BJ3ESXL4/p1650291953580479) about this attribute.

packages/web-components/src/components/va-radio/va-radio.tsx
```diff
@@ -82,6 +85,9 @@ export class VaRadio {
         nextNode = currentNode;
         this.selectNextNode(currentNode);
         break;
+      case 'ArrowDown':
+        // prevent scrolling page
+        event.preventDefault();
```
andresriveratoro, line 90: This was an opportunistic fix to prevent scrolling the page when using the Down and Up arrow keys.

packages/web-components/src/components/va-radio-option/va-radio-option.css
```diff
@@ -1,34 +1,39 @@
-@import '../../mixins/inputs.css';
```
bkjohnson, line 1: Were some of the styles in here part of the problem?

packages/web-components/src/components/va-radio-option/va-radio-option.css
```diff
@@ -1,34 +1,39 @@
...
   display: inline-block;
+  height: var(--visible-radio-button-size);
+  line-height: var(--visible-radio-button-size);
+  margin-right: 1rem;
+  width: var(--visible-radio-button-size);
+}
+
+input:not([disabled]):focus + label::before {
+  outline: 2px solid var(--color-gold-light);
+  outline-offset: 6px;
```
bkjohnson, line 33: This looks like the "main" change, is that right? I was able to get what looks like the same styling without changing all of the other styling & markup, so do we need the rest of the changes?

focus styling only applied to button

packages/web-components/src/components/va-radio/va-radio.tsx
bkjohnson: Could we use fallthrough semantics to group the one below with this one, like:

```js
      case 'ArrowDown':
      case 'ArrowUp':
        // prevent scrolling page
        event.preventDefault();
      case 'ArrowDown':
      case 'ArrowRight':
        if (currentNodeIndex === radioOptionNodes.length - 1) {
          nextNode = radioOptionNodes[0];
          this.deselectCurrentNode(currentNode);
          this.selectNextNode(nextNode);
        } else {
          nextNode = radioOptionNodes[currentNodeIndex + 1];
          this.deselectCurrentNode(currentNode);
          this.selectNextNode(nextNode);
        }
        break;
```

bkjohnson (commented):
A few questions.

packages/web-components/src/components/va-radio-option/va-radio-option.css
andresriveratoro: You're correct. We don't need any of those other changes. Thanks! I'm closing this PR for a new one since I had issues reverting.

andresriveratoro:
Closing for https://github.com/department-of-veterans-affairs/component-library/pull/428

PR #427, opened 2022-05-13
40076 aria describedby remove from va-text-input

pdavies88:
## Chromatic

https://40076-aria-describedby-remove--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/40076

## Testing done - Screenshots
Screen Shot 2022-05-13 at 1 13 10 PM
Screen Shot 2022-05-13 at 1 14 53 PM

## Acceptance criteria
- [x] va-text-input remove ariaDescribedby prop, remove describedBy variable, remove aria-describedby on input
- [x]  va-text-input Storybook remove all references to ariaDescribedby and its related Story

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-text-input/va-text-input.tsx
+++ b/packages/web-components/src/components/va-text-input/va-text-input.tsx
@@ -162,7 +162,7 @@ export class VaTextInput {
   }
 
   render() {
-    const describedBy = `${this.error ? 'error-message' : ''}` || null; // Null so we don't add the attribute if we have an empty string
+    const describedBy = this.error ? 'error-message' : null; // Null so we don't add the attribute if we have an empty string
     const inputMode = this.inputmode ? this.inputmode : null; // Null so we don't add the attribute if we have an empty string
     const type = this.getInputType();
 
--- a/packages/web-components/src/components/va-text-input/va-text-input.tsx
+++ b/packages/web-components/src/components/va-text-input/va-text-input.tsx
@@ -190,6 +190,7 @@ export class VaTextInput {
           minlength={this.minlength}
           pattern={this.pattern}
           name={this.name}
+          required={this.required || null}
           part="input"
         />
         {this.maxlength && this.value?.length >= this.maxlength && (
--- a/packages/web-components/src/components/va-text-input/va-text-input.tsx
+++ b/packages/web-components/src/components/va-text-input/va-text-input.tsx
@@ -162,7 +162,7 @@ export class VaTextInput {
   }
 
   render() {
-    const describedBy = `${this.error ? 'error-message' : ''}`.trim() || null; // Null so we don't add the attribute if we have an empty string
+    const describedBy = `${this.error ? 'error-message' : ''}` || null; // Null so we don't add the attribute if we have an empty string
     const inputMode = this.inputmode ? this.inputmode : null; // Null so we don't add the attribute if we have an empty string
     const type = this.getInputType();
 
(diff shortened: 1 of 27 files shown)


packages/web-components/src/components/va-text-input/va-text-input.tsx
```diff
@@ -167,9 +162,6 @@ export class VaTextInput {
   };
 
   render() {
-    const describedBy =
-      `${this.ariaDescribedby} ${this.error ? 'error-message' : ''}`.trim() ||
-      null; // Null so we don't add the attribute if we have an empty string
```
bkjohnson, line 172: Do we want to remove the `aria-describedby` even when there is an error message? This seems like it could be left while removing the external-facing prop.

pdavies88: @bkjohnson I have added that portion back in

packages/web-components/src/components/va-text-input/va-text-input.tsx
```diff
@@ -160,16 +155,14 @@ export class VaTextInput {
...
   disconnectedCallback() {
     i18next.off('languageChanged');
-  };
+  }
 
   render() {
-    const describedBy =
-      `${this.ariaDescribedby} ${this.error ? 'error-message' : ''}`.trim() ||
-      null; // Null so we don't add the attribute if we have an empty string
+    const describedBy = `${this.error ? 'error-message' : ''}`.trim() || null; // Null so we don't add the attribute if we have an empty string
```
bkjohnson, line 165: Now that we don't have the chance of `my-custom-describedby-id `  with the dangling space for when `this.error` is undefined, I don't think we need to `.trim()`.
```

pdavies88: PR has been updated and `.trim()` removed

bkjohnson: I think this could be cleaned up a bit more to remove some things that are no longer necessary after removing the `ariaDescribedby` prop.

pdavies88: Ok what else do you want me to remove?

k80bowman: If we're not using that aria-describedby prop to construct a string, do we need the ternary with the empty string here?

pdavies88: @k80bowman the code is currently the following `const describedBy = `${this.error ? 'error-message' : ''}` || null;`
Which is just saying if someone sets an error message add the attribute of `aria-describedby='error-message'` to the input element otherwise if it is null don't set this attribute. By setting the attribute it will then correspond with this line via the id https://github.com/department-of-veterans-affairs/component-library/blob/f7aabf936611d78338ff35fb649553437449ba8c/packages/web-components/src/components/va-text-input/va-text-input.tsx#L180 and will then read out the error message to the user is focused on the input field

pdavies88: Oh I see what you are getting at I will update it to `const describedBy = ${this.error ? 'error-message' : null};`

pdavies88: Just updated the PR with that fix

k80bowman: So I wrote my own test to check this and it does work, but would you mind please adding a test to check that this does not add an aria-describedby when an empty string is passed as the error message prop?

k80bowman (approved):

PR #429, opened 2022-05-17
Update padding and border width for va-alert

andresriveratoro:
## Chromatic

https://658-va-alert--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/658 and partly closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/810

## Screenshots
Screen Shot 2022-05-20 at 13 21 09
Screen Shot 2022-05-20 at 13 21 52
Screen Shot 2022-05-20 at 13 21 22
Screen Shot 2022-05-20 at 13 21 39

## Acceptance criteria
- [x] Right padding has been updated to 1.6rem (16px)
- [x] Top, bottom and left padding has been updated to 1.6rem (16px) 
- [x] Left border has been changed to 8px
- [x] Top border has been changed to 8px (full-width variant)

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-alert/va-alert.css
+++ b/packages/web-components/src/components/va-alert/va-alert.css
@@ -14,7 +14,7 @@
 :host([full-width='true']),
 :host([full-width='']) {
   border-top-style: solid;
-  border-top-width: 10px;
+  border-top-width: 8px;
   display: flex;
   align-items: flex-start;
   justify-content: center;
@@ -40,7 +40,7 @@
   max-width: 1000px;
 }
 
-@media screen and (min-width:1024px) {
+@media screen and (min-width: 1024px) {
   :host([full-width='true']) .va-alert-close,
   :host([full-width='']) .va-alert-close {
     position: relative;
@@ -56,7 +56,7 @@
 
 div.alert {
   display: table;
-  padding: 3.2rem 6.4rem 3.2rem 2.4rem;
+  padding: 1.6rem;
   width: 100%;
   vertical-align: middle;
   box-sizing: border-box;
@@ -65,7 +65,7 @@ div.alert {
 div.alert:not(.bg-only) {
   border-color: inherit;
   border-left-style: solid;
-  border-left-width: 10px;
+  border-left-width: 8px;
 }
 
 div.alert > i::before {
--- a/packages/web-components/src/components/va-alert/va-alert.css
+++ b/packages/web-components/src/components/va-alert/va-alert.css
@@ -56,10 +56,16 @@
 
 div.alert {
   display: table;
-  padding: 1.6rem;
+  padding: 4rem;
   width: 100%;
   vertical-align: middle;
   box-sizing: border-box;
+}
+
+@media screen and (min-width: 481px) {
+  div.alert {
+    padding: 4.8rem;
+  }
 }
 
 div.alert:not(.bg-only) {
--- a/packages/web-components/src/components/va-alert/va-alert.css
+++ b/packages/web-components/src/components/va-alert/va-alert.css
@@ -157,9 +157,12 @@ div.bg-only {
 
 i {
   font-family: 'Font Awesome 5 Free';
+  font-size: 2.4rem;
   font-style: normal;
   font-weight: 900;
+  line-height: 1;
 }
+
 i.fa-times-circle::before {
   content: '\F057'; /* fa-times-circle*/
 }
--- a/packages/web-components/src/components/va-alert/va-alert.css
+++ b/packages/web-components/src/components/va-alert/va-alert.css
@@ -56,16 +56,10 @@
 
 div.alert {
   display: table;
-  padding: 4rem;
+  padding: 1.6rem 2rem 1.6rem 1.6rem;
   width: 100%;
   vertical-align: middle;
   box-sizing: border-box;
-}
-
-@media screen and (min-width: 481px) {
-  div.alert {
-    padding: 4.8rem;
-  }
 }
 
 div.alert:not(.bg-only) {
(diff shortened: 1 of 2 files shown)


k80bowman (approved):

packages/web-components/src/components/va-alert/va-alert.css
```diff
@@ -56,7 +56,7 @@
 
 div.alert {
   display: table;
-  padding: 3.2rem 6.4rem 3.2rem 2.4rem;
+  padding: 1.6rem;
```
bkjohnson, line 59: I dumped some sample text into the alert body to test something out. Here's what we currently have - note how the "Close" icon always stands alone with nothing below it:

Current `<va-alert>` with padding highlighted

Here is what's on the chromatic link:

image

It looks like that extra padding was [originally added](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/pull/8) to fix a problem with long header text running into the icon, but it looks like that's no longer an issue:

Long headline text remaining separated from close icon

However, is there any kind of usability/design/accessibility concern with the body text being underneath the icon?

andresriveratoro: Tagging @GnatalieH @SarahKay8 for feedback. Is having text below the close button in these screenshots a design or accessibility problem?

Screen Shot 2022-05-17 at 16 03 27
Screen Shot 2022-05-17 at 16 03 40

GnatalieH: @andresriveratoro no, there is no design issue. Looks fine to me.

humancompanion-usds: I don't think it's a problem but it wouldn't be a bad idea to add padding to the right so that the button "sits" in a distinct "column" by itself. It would certainly make it look more neat and tidy.

humancompanion-usds: Given that's what we have now as Brooks pointed out, it would also be a good idea to stay consistent. My intent was just to get us closer to the USWDS version but they don't show a dismissible icon so '"Close" icon always stands alone with nothing below it' should still be true.

bkjohnson: It looks like `4rem` does the trick:

image

k80bowman:
How does 4rem look on mobile? The original [issue](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/658) was that there was too much padding on the right for mobile screens. We had 6.4rem, so 4rem is definitely a decrease, but the request was for 2rem (2.4rem if that was too small).

packages/web-components/src/components/va-alert/va-alert.css
k80bowman: How does 4rem look on mobile? The original https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/658 was that there was too much padding on the right for mobile screens. We had 6.4rem, so 4rem is definitely a decrease, but the request was for 2rem (2.4rem if that was too small).

andresriveratoro: This is how it looks like on mobile with 4rem padding on right side.
Screen Shot 2022-05-17 at 16 31 42

SarahKay8: @andresriveratoro Having user information beneath the close button isn't an A11y issue. I see no issue

andresriveratoro: If we want to have equal padding for the alert body on both sides, we could go to `4.8rem`
We could go to 4.8rem if we want to have equal padding for the text on both sides, but like Katy mentioned, the original issue is to address the excessive padding on the right side.

This is how it looks like with 4.8rem padding on right side.
Screen Shot 2022-05-17 at 16 42 33

andresriveratoro: ### With 1.6rem:

Screen Shot 2022-05-17 at 16 52 50

### With 4rem:

Screen Shot 2022-05-17 at 16 53 00

### With 4.8rem:

Screen Shot 2022-05-17 at 16 53 08

GnatalieH: 4rem looks best, to me, because it prevents the headline from overcrowding the close icon without adding excessive space.

andresriveratoro: Padding set to 4rem on screen width under 481px:
Screen Shot 2022-05-18 at 11 21 45

Padding set to 4.8rem on screen width 481px and higher:
Screen Shot 2022-05-18 at 11 22 02

k80bowman: I feel like having the 4rem padding on the left on small screens (under 481px) will end up causing the same problem we're trying to solve.

GnatalieH: I think you are right, @k80bowman. I wonder if we could left-align the body text with the left edge of the "i" icon so the column width of the body text can increase and minimize the huge left margin. How might that look?

andresriveratoro:
Padding was updated to bring it more inline with USWDS [Alert](https://designsystem.digital.gov/components/alert/).
Right padding is 2rem; top, bottom and left padding is 1.6rem.

Screen Shot 2022-05-19 at 17 25 20
Screen Shot 2022-05-19 at 17 25 10

packages/web-components/src/components/va-alert/va-alert.css
andresriveratoro: @GnatalieH I tried moving the body text to begin at the same indentation as the icon, but it didn't look right. USWDS has it at the same indentation as the heading.

k80bowman (approved):

PR #434, opened 2022-05-19
Va alert/no headline spacing issue

mdewey:
## Chromatic

https://va-alert&#x2F;no-header--60f9b557105290003b387cd5.chromatic.com

## Description
Our Design has created a new alert format with no headline. Currently, if there is no header/headline, the component renders out extra space at the top of the component (see before screenshot). This PR conditionally hides the headline slot and updates the styling to not have the extra padding if the headline does not exist. 

[Mocks from Designer](https://www.sketch.com/s/1a920e73-1dcb-47c4-aae8-08656756c131/a/zxWvzOy#Inspector)

## Testing done
- Added unit tests
- Visual 

## Screenshots
Before: 
Screen Shot 2022-05-19 at 11 20 39 AM
After:
Screen Shot 2022-05-19 at 11 20 56 AM

### DIFF
--- a/packages/storybook/stories/va-alert.stories.jsx
+++ b/packages/storybook/stories/va-alert.stories.jsx
@@ -104,6 +104,7 @@ const BackgroundOnlyTemplate = ({
 }) => {
   return (
     <>
+     
       <va-alert
         status="info"
         background-only={backgroundOnly}
@@ -247,4 +248,18 @@ export const NotVisible = Template.bind({});
 NotVisible.args = {
   ...defaultArgs,
   visible: false,
+};
+
+
+export const noHeaderMessage = ({ }) => {
+  return (
+    <>
+      <va-alert
+        status="warning"
+      >
+        <div>Lorem ipsum dolor sit amet consectetur adipisicing elit. Id quod corrupti reiciendis nesciunt unde rem! Amet quos porro repellat, debitis explicabo temporibus commodi consequatur expedita, tempora corrupti, vel perferendis soluta.</div>
+       
+      </va-alert>
+    </>
+  );
 };
--- a/packages/web-components/src/components/va-alert/test/va-alert.e2e.ts
+++ b/packages/web-components/src/components/va-alert/test/va-alert.e2e.ts
@@ -11,6 +11,25 @@ describe('va-alert', () => {
     expect(element).toEqualHtml(`
       <va-alert class="hydrated" status="info">
         <mock:shadow-root>
+          <div class="alert info no-headline">
+            <i aria-hidden="true" role="img"></i>
+            <div class="body" role="presentation">
+              <slot></slot>
+            </div>
+          </div>
+        </mock:shadow-root>
+      </va-alert>
+    `);
+  });
+
+  it('renders a headline when present', async () => {
+    const page = await newE2EPage();
+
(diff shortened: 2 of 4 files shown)


GnatalieH:
@mdewey we usually use the background alert with an icon for system errors like this. It has no headline. I wonder if your Designer tried that component, first? CCing @humancompanion-usds for awareness since alerts/notifications have come up a lot recently.

mdewey:
Hey @k80bowman , @humancompanion-usds , @GnatalieH . I did speak with my designer about that and they mentioned they have reasons. I will not try to sum it up, but @andaleliz can explain better how she and other designers came to this. She is out of the office today, but I let her know, and she will reach out in the coming days :-)

GnatalieH:
Thanks @mdewey. I believe the background only alert with the yellow warning color didn't test well, as users thought it was soothing looking and didn't recognize it as an alert (I remember Liz and Angela talking about this). So, I'm wondering if the red (error) background only alert would work better for critical alert messages. Liz can chime in when she returns.

andaleliz:
Hi all, here's some background on the design decisions behind these alerts:
- In profile, we use the background only alerts to give user feedback on their action, such as a successful save for their profile, or an error when we're unable to save their changes.  It felt like using the other alert pattern for "system" alerts would help differentiate the two.
- The message content is short, and the heading felt like we were repeating content without adding more meaning, so we felt it worked better without it. It also adds more vertical height which is something I try not to do if I don't have to for mobile devices.
- These alerts were originally error messages, and at the time the design and content work was done (Sept 2021), the guidance on design.va.gov for error messages didn't align with our use case. 

I see now that the error message guidance for content has been revised and includes suggestions headings can work in our use case without being redundant. Our team will figure out how to make this work w/ the heading + the new guidance available.  

@mdewey I'm sorry you went through the trouble of working on this update 😞  I hate to have you waste your time.

GnatalieH:
@andaleliz IIRC you want to remove the header from the alert, right (like your example)? I'm cc'ing @humancompanion-usds so he can weigh in on this example. I agree, the color and icon are not really appropriate for the messaging. This is interesting, because it's essentially a "sign out" alert. We have sign in alerts, but haven't documented something like this, yet.

humancompanion-usds:
@andaleliz - My thoughts are as follows:

First, two wrongs don't make a right. The sign-out alert should have a header as well. I don't see their message in our message dictionary so I suspect they went rogue.

Secondly, the scenario you have seems very close to the "Unable to complete a user-initiated task, can still proceed" warning alert that has the following content in our [messages dictionary](https://design.va.gov/patterns/messaging-dictionary#system-downtime):

> We’ve run into a problem
> We’re sorry. Something went wrong on our end. You can try again now, or move on to the next step and come back later to complete this.

I consulted with @DanielleThierryUSDSVA and she agrees this is the closest message to use and that you may customize as necessary. If you want assistance with the content reach out to Danielle.

Finally, a header is superior for a11y, visual hierarchy, and further differentiation from your background-only alerts. I value those more than a bit more additional height to the alert.

andaleliz:
Thanks for weighing in @humancompanion-usds, and that all makes sense! We will update alerts accordingly.

micahchiang:
Closing this because the [original design issue](https://github.com/department-of-veterans-affairs/va.gov-team/issues/43586) associated with it has been closed, and all subsequent work that came out of the linked issue has also been resolved by the Authenticated Experience team.

PR #438, opened 2022-05-23
843 va-date Text Inputs

pdavies88:
## Chromatic

https://843-va-date-text-inputs--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/843

## Related PR
https://github.com/department-of-veterans-affairs/vets-design-system-documentation/pull/905

## Testing done
Screen Shot 2022-06-01 at 11 53 20 AM

## Screenshots
Screen Shot 2022-06-01 at 11 56 25 AM

## Acceptance criteria
- [x] Component is written and added to Storybook
- [x] Component has had design reviews
- [x] Component has had accessibility reviews
- [x] Component should support single digit days and months
- [x] Valid month entry is 1-12
- [x] Valid day entry is dependent on month and will be between 1-29, 1-30, or 1-31 depending on the month
- [x] Invalid data entry should trigger the Please enter a valid date error message
- [x] Valid year entry should be limited from 1900 through 2122
- [x] Add in a `slot` to the `va-date-text-input` component so that developers can add html between the label and input fields
- [x] Add additional tests to cover these new features
- [x] Move onBlur event handler to component level from children to fire once component blur fires instead of each field
- [x] Required validation error should only fire onBlur
- [x] All validation for the component should only occur onBlur
- [x] Handles validation for Leap Years
- [x] Acceptable keys in the input fields are the following: 0123456789, Delete, Backspace, Tab, ArrowLeft, ArrowRight

## Definition of done


pdavies88:
@k80bowman do you want me to add `<slot>` to this component so someone can add Hint Text like in: https://designsystem.digital.gov/components/memorable-date/

k80bowman:
@pdavies88 yes, let's do that and keep it as close to the USWDS implementation as possible.

k80bowman (changes requested):
Is it possible to change the month and day inputs so they accept up to a two-digit input as the [guidance](https://designsystem.digital.gov/components/memorable-date/) from USWDS states:
> The year input is set to accept only a four-digit number while the month and day inputs can accept up to a two-digit number.

pdavies88:
@k80bowman so it is already locked down to 2 digits for month and day and 4 digits for year. I'm using the same pattern that USWDS does to lock it down and also setting a min and max length. I also have tests proving this assertion as well.

In regards to negative numbers I am going to update the code to target specific keycodes since that seems to be the only way to truly bulletproof it from taking in negatives so we will end up having something like:
```
// Listen for input event on numInput.
number.onkeydown = function(e) {
    if(!((e.keyCode > 95 && e.keyCode < 106)
      || (e.keyCode > 47 && e.keyCode < 58) 
      || e.keyCode == 8)) {
        return false;
    }
}
```

k80bowman:
@pdavies88 It looks to me like the USWDS component allows 1 digit for the month and day as well, is that not the case?

pdavies88:
@k80bowman so the USWDS does allow for 1 digit however we specifically call out in multiple places that we need things to be in MM DD YYYY format since that is how the current application teams are expecting the data to be provided to them. 

Now I could do the following if we want to support single digits:
* Bake in additional logic that will format a number less than 10 automatically to put a 0 in front of it onBlur (we wouldn't want to do this prior to blur since a user might be typing in 11 and having it onInput would cause them to get 01 before they could get 11)
* Change minLength to 1 character instead of 2

The caveat to all of this is that we do have a message telling users the format that they need right now as well as error handling for less than 2 characters. Plus there is already baked in ways for a user to break the system like they could put in 99 99 9999 and that's a valid date, so we are already putting some burden on them expecting them to have to read or listen and follow the rules.

k80bowman:
@pdavies88 are the instructions for entering the date properly something that the frontend engineers add? Because in our component I see it as an aria-label, but not displayed visually: 
image

pdavies88:
@k80bowman change has been made and should be available to be seen here for the hint text: https://843-va-date-text-inputs--60f9b557105290003b387cd5.chromatic.com/?path=/docs/components-va-date-text-input--default

Mottie:
Thanks for working on this @pdavies88!

I'm trying to figure out what the `custom-validation-message` property does? If I set the `custom-validation-boolean`, I would expect the message in `custom-validation-message` to show, but it shows the `error` message instead.
va-date not showing the custom-validation-message, but does show the error message when added

pdavies88:
@Mottie so I have everything for validation firing on blur. I can adjust it so that it fire before then. Basically my thought process was to do what we have with the validation object from the React component and streamline it so if you have any logic you want you do that and if what the user gives you is not correct you set that custom-validation-boolean to true and when the fields are blurred it fires up that custom-validation-message instead of the other error messages.

But if you want me to have it fire outside of the blur so like anytime that custom-validation-boolean prop is set to true I can definitely adjust to that. I would just move that snippet of code outside of the onBlur method. If that's what you are thinking let me know.

Mottie:
Actually, I think that makes sense. One would either use `error` or `custom-validation-message`, but not necessarily both at the same time.

packages/web-components/src/components/va-date-text-input/va-date-text-input.tsx
```diff
@@ -0,0 +1,256 @@
...
+    const dayNum = parseInt(day);
+    // Use a leading zero for numbers < 10
+    const numFormatter = new Intl.NumberFormat('en-US', {
+      minimumIntegerDigits: 2,
+    });
+    this.value = `${year}-${month ? numFormatter.format(monthNum) : ''}-${
+      day ? numFormatter.format(dayNum) : ''
+    }`;
+
+    const daysForSelectedMonth = monthNum > 0 ? days[monthNum] : [];
```
Mottie, line 83: I feel like we could add some checks here for leap years and only allow 29 days in February then, instead of always.

packages/web-components/src/components/va-date-text-input/va-date-text-input.tsx
```diff
@@ -0,0 +1,256 @@
...
+      '5',
+      '6',
+      '7',
+      '8',
+      '9',
+      'Backspace',
+      'ArrowRight',
+      'ArrowLeft',
+      'Tab',
+    ];
```
Mottie, line 142: Include `'Delete'` too, please?

packages/web-components/src/components/va-date-text-input/va-date-text-input.tsx
```diff
@@ -0,0 +1,256 @@
...
+    // Error attribute should be leveraged for custom error messaging
+    // Fieldset has an implicit aria role of group
+    return (
+      <Host
+        value={value}
+        error={error || this.error}
+        onBlur={handleDateBlur}
+        aria-describedby={ariaDescribedby}
+      >
+        <fieldset aria-label="Input month and day fields as two digit XX and four digit year format XXXX">
```
Mottie, line 191: Does the `XX` and `XXXX` need to be read out by screenreaders? Probably something Angela Fowler can answer for us.

Mottie (approved):
Nice work, thanks!

packages/web-components/src/components/va-date-text-input/va-date-text-input.tsx
pdavies88: So this was originally provided to me by our former accessibility specialist Courtney, but @SarahKay8 I'd be interested to hear your opinion on what you think the aria-label should read

packages/web-components/src/components/va-date-text-input/va-date-text-input.tsx
pdavies88: @Mottie I have updated to add Delete

packages/web-components/src/components/va-date-text-input/va-date-text-input.tsx
pdavies88: @Mottie I have added a utility function to check for leap years and throw an error message if the date is invalid ie: February 29 1999 is invalid but February 29 2000 is valid

packages/web-components/src/components/va-date-text-input/test/va-date-text-input.e2e.ts
```diff
@@ -0,0 +1,332 @@
...
+    await handleYear.press('0');
+    await handleYear.press('2');
+    await handleYear.press('2');
+    await handleYear.press('3');
+
+    await page.waitForChanges();
+    expect(elementYear.getAttribute('value')).toBe('2022');
+  });
+
+  it('day and month input only allows for 2 characters to be used', async () => {
```
k80bowman, line 136: Is this test still valid? It looks like you're allowing 1 and 2 characters now (from the test on line 242)

k80bowman (commented):
Thank you for adding all the validation and tests, this is great.

My only real feedback is wondering if you checked in with Natalie about how this looks with the label, the guidance, the hint text, and the field labels. That's a lot of text above the input fields and there is no hierarchy to it. I know Natalie is out this week, would you mind having @humancompanion-usds take a look?

image

packages/web-components/src/components/va-date-text-input/va-date-text-input.tsx
SarahKay8: Hey @pdavies88 so a couple of things to note 

 1. To label a fieldset, use the legend element.
2. Each input field must be individually labelled. It does not show here if they are.
3. If they are individually labelled, your fieldset should say what the date is, for example “Birthday” or “Order date”.
See also for reference 

https://www.w3.org/WAI/tutorials/forms/grouping/#related-fields

pdavies88:
@k80bowman Sketch File Screenshot:
Screen Shot 2022-06-02 at 11 56 31 AM

packages/web-components/src/components/va-date-text-input/va-date-text-input.tsx
pdavies88: From @SarahKay8 not sure why it is not showing up here but - Hey [@pdavies88](https://github.com/pdavies88) so a couple of things to note

1. To label a fieldset, use the legend element.
2. Each input field must be individually labelled. It does not show here if they are.
3. If they are individually labelled, your fieldset should say what the date is, for example “Birthday” or “Order date”.
4. See also for reference - https://www.w3.org/WAI/tutorials/forms/grouping/#related-fields

SarahKay8:
Hey Preston so a few things... 
1. To label a fieldset, use the legend element.
2. Each input field must be individually labelled. It does not show here if they are.


Also, While certain screen readers can read the aria-label as the name of the grouping I would suggest moving this information to a normal <p> or similar element so that it would be also available for sighted users? And then just link that visual text to related fields with the aria-describedby attribute. Maybe it would be also more clear if there were separate instructions for each field instead of such that teach all formats. So that way you can get rid of the (XX and XXXX) not the biggest fan. Only because not every screen reader would read that correctly...

If you wanna have one general instruction set for the whole fieldset here's something for reference: https://blog.tenon.io/accessible-validation-of-checkbox-and-radiobutton-groups/

packages/web-components/src/utils/date-utils.ts
```diff
@@ -128,3 +128,12 @@ export const days = {
...
   return /\d{4}-\d{1,2}-\d{1,2}/.test(date);
 }
+
+export function checkLeapYear(year) {
+  //three conditions to find out the leap year
+  if ((0 == year % 4 && 0 != year % 100) || 0 == year % 400) {
+    return true;
+  } else {
+    return false;
+  }
```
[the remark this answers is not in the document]
pdavies88: Updated

packages/web-components/src/components/va-date-text-input/va-date-text-input.tsx
```diff
@@ -0,0 +1,256 @@
...
+          <legend>
+            {label} {required && <span class="required">(*Required)</span>}
```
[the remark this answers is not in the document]
pdavies88: Updated

packages/web-components/src/components/va-date-text-input/test/va-date-text-input.e2e.ts
```diff
@@ -0,0 +1,332 @@
...
+    const month = await page.find('va-date-text-input >>> .input-month');
+    const day = await page.find('va-date-text-input >>> .input-day');
+    const year = await page.find('va-date-text-input >>> .input-year');
+
+    expect(month.getAttribute('value')).toBe('05');
+    expect(day.getAttribute('value')).toBe('03');
+    expect(year.getAttribute('value')).toBe('1999');
+  });
+
+  it('updates date based select and input fields', async () => {
```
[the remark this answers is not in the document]
pdavies88: Updated

packages/web-components/src/components/va-date-text-input/va-date-text-input.tsx
```diff
@@ -0,0 +1,256 @@
...
+    } = this;
+
+    const [year, month, day] = (value || '').split('-').map(val => val);
+
+    // Error attribute should be leveraged for custom error messaging
+    // Fieldset has an implicit aria role of group
+    return (
+      <Host
+        value={value}
+        error={error || this.error}
```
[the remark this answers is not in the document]
pdavies88: Updated

packages/storybook/stories/additional-docs.js
```diff
@@ -151,6 +151,12 @@ export const additionalDocs = {
     maturityCategory: USE,
     maturityLevel: DEPLOYED,
   },
+  'va-date-text-input': {
+    guidanceHref: 'form/date-text-input',
+    guidanceName: 'Date Text Input',
```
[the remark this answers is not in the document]
pdavies88: Updated

packages/web-components/src/components/va-date-text-input/va-date-text-input.tsx
```diff
@@ -0,0 +1,256 @@
...
+   * The error message to render (if any)
+   * This prop should be leveraged to display any custom validations needed for this component
+   */
+  @Prop() error: string;
+
+  /**
+   * The aria-describedby attribute lists the ids of the elements that describe the object.
+   * It is used to establish a relationship between widgets or groups and the text that describes them
+   */
+  @Prop() ariaDescribedby: string;
```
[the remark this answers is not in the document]
Mottie: I think @bkjohnson is saying we can add `aria-describedby` to the host, so it doesn't need to be a prop?

packages/web-components/src/components/va-date-text-input/va-date-text-input.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+  /**
+   * Set to true if custom validation logic does not meet expected criteria for valid input
+   */
+  @Prop() customValidationBoolean: boolean;
+
+  /**
+   * Set message to display if custom validation boolean is true
+   */
+  @Prop() customValidationMessage: string;
```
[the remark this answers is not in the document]
Mottie: I think I could just use the `error` prop, but I'd need some way to over-ride the built-in message.

packages/web-components/src/components/va-date-text-input/va-date-text-input.tsx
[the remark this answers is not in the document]
pdavies88: @Mottie yes I believe so this would be something you would implement instead of having us bake it in. If you are ok with doing that please let me know and I will remove.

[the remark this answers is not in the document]
Mottie: Yes, that sounds good.

packages/web-components/src/components/va-date-text-input/va-date-text-input.tsx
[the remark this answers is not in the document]
pdavies88: @Mottie so in our current version of the `va-date` component where are doing that with our Storybook example here: https://github.com/department-of-veterans-affairs/component-library/blob/master/packages/storybook/stories/va-date.stories.jsx#L41-L65
Basically setting the `error` prop to 'Date must be in the future' now you could add an else statement there and set the `error` prop to an empty string and that will have the error state not show. You can also override the `error` messaging all together by defaulting it to an empty string and that will not have the error messaging show.

If that works please let me know and I will remove the extra props and adjust.

packages/web-components/src/components/va-date-text-input/va-date-text-input.tsx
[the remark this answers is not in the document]
pdavies88: Prop has been removed

packages/web-components/src/components/va-date-text-input/va-date-text-input.tsx
```diff
@@ -0,0 +1,279 @@
...
+      label,
+      name,
+      error,
+      handleDateBlur,
+      handleDateChange,
+      handleDateKey,
+      value,
+    } = this;
+
+    const [year, month, day] = (value || '').split('-').map(val => val);
```
bkjohnson, line 205: This is a two-parter. It looks like the three consts here would be strings or `undefined`. If the map function was something like `val => val || ''`...

packages/web-components/src/components/va-date-text-input/va-date-text-input.tsx
```diff
@@ -0,0 +1,279 @@
...
+            <va-text-input
+              label="Month"
+              name={`${name}Month`}
+              maxlength={2}
+              minlength={2}
+              pattern="[0-9]*"
+              aria-describedby="dateHint"
+              // Value must be a string
+              // if NaN provide empty string
+              value={month ? month.toString() : ''}
```
bkjohnson, line 235: Would we be able to simplify this? And if the NaN is leftover from an earlier `parseInt()` call, would things still work if the value we passed in was `undefined`? If so then we might be able to get rid of the `.map()` in my previous comment.

bkjohnson (commented):
Some minor comments. I know we're still waiting on the `customValidation` things.

packages/web-components/src/components/va-date-text-input/va-date-text-input.tsx
```diff
@@ -0,0 +1,279 @@
...
+          <legend>
+            {label}{' '}
+            {required && (
+              <span class="required">(*{i18next.t('required')})</span>
+            )}
+            <div id="dateHint">Use the following format: MM DD YYYY</div>
+          </legend>
+          {error && (
+            <span class="error-message" role="alert">
+              <span class="sr-only">Error</span> {error}
```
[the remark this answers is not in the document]
pdavies88: Updated and now using i18next

packages/web-components/src/components/va-date-text-input/va-date-text-input.tsx
pdavies88: See comment below

pdavies88:
Per conversation with @SarahKay8 changing date format prompt text to: Please enter two digits for the month and day and four digits for the year.

packages/web-components/src/components/va-date-text-input/va-date-text-input.tsx
bkjohnson: Do we still need the ternary?

pdavies88: Good call updated it to `month?.toString()`

packages/storybook/stories/va-date-text-input.stories.jsx
```diff
@@ -0,0 +1,128 @@
...
+      required={required}
+      error={error}
+      value={value}
+    >
+      <div className="vads-u-margin-bottom--1">This is example hint text</div>
+    </va-date-text-input>
+  );
+};
+
+export const Default = Template.bind({});
```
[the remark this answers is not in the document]
pdavies88: For sure just updated with those changes

k80bowman (approved):

PR #443, opened 2022-05-25
add padding to alert for close button

k80bowman:
## Chromatic

https://alert-padding-fix--60f9b557105290003b387cd5.chromatic.com

This fixes a problem we were seeing with closable background-only alerts with longer text. Further examination confirmed that the same problem would occur with the headline on closable default alerts.

## Description
Closes <ticket>

## Testing done
Visual testing, manual

## Screenshots
Background-only alert before:
image

Background-only alert after:
image

Default alert with long headline before:
image

Default alert with long headline after:
image

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-alert/va-alert.css
+++ b/packages/web-components/src/components/va-alert/va-alert.css
@@ -50,7 +50,7 @@
 ::slotted([slot='headline']) {
   font-size: 2rem !important;
   line-height: 26px !important;
-  padding: 0 !important;
+  padding: 0 4rem 0 0 !important;
   margin: 0 !important;
 }
 
@@ -117,7 +117,7 @@ div.bg-only.hide-icon > i::before {
 
 div.bg-only {
   border: none;
-  padding: 2rem;
+  padding: 2rem 4rem 2rem 2rem;
 }
 
 .info.bg-only {


k80bowman:
@GnatalieH and @humancompanion-usds I would love to get your opinions on this fix.

bkjohnson (commented):
Would something like this work if placed below the current `div.alert` and `div.bg-only` selectors in the CSS file?


```css
div.alert,
div.bg-only {
  padding-right: 4rem;
}
```

That might take care of the headline, body content in normal alerts, and background-only alerts as well.

k80bowman:
@bkjohnson that could work, but I'm not sure I like separating out the CSS like this. I also would like to leave the padding for the body content in normal alerts at 2rem, since we just changed that to allow more space for content.

bkjohnson (approved):
Sounds good to me :+1:

GnatalieH:
> @GnatalieH and @humancompanion-usds I would love to get your opinions on this fix.

@k80bowman and @bkjohnson I'm interested in seeing what the dismissible background alert looks like in a 320px viewport. What is the final padding -- 2 rem on all sides of the close icon? I'm creating the Sketch asset, and I don't think the Storybook example is fixed yet. Thx.

k80bowman:
@GnatalieH if you go to this link (from the top of the PR description) you should be able to see the updated version: https://alert-padding-fix--60f9b557105290003b387cd5.chromatic.com/

Here is a screenshot at 320px:
image

The padding for the dismissable background alert is on the alert text (rather than the close button) and is 2rem on the top, left, and bottom and then 4rem on the right.
image

micahchiang:
Closing since this was addressed in #495

PR #447, opened 2022-05-31
Add button web component

andresriveratoro:
## Chromatic

https://817-button--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/817

## Testing done
E2E and Storybook

## Screenshots
Screen Shot 2022-06-02 at 13 26 52

## Acceptance criteria
- [x] va-button has been created
- [x] va-button has a Storybook page with stories

## Definition of done


packages/storybook/stories/additional-docs.js
```diff
@@ -156,6 +154,12 @@ export const additionalDocs = {
     maturityCategory: USE,
     maturityLevel: DEPLOYED,
   },
+  'va-button': {
+    guidanceHref: 'button',
+    guidanceName: 'Button',
+    maturityCategory: USE,
+    maturityLevel: BEST_PRACTICE,
```
k80bowman, line 161: Can we please set this to "use with caution: candidate" until it goes through staging review?

packages/storybook/stories/additional-docs.js
```diff
@@ -21,10 +21,8 @@ export const additionalDocs = {
   },
   // MDX
   'Button': {
-    guidanceHref: 'button',
-    guidanceName: 'Button',
-    maturityCategory: USE,
-    maturityLevel: BEST_PRACTICE,
+    maturityCategory: DONT_USE,
+    maturityLevel: DEPRECATED,
```
k80bowman, line 25: Let's not deprecate this until the web component goes through staging review

packages/web-components/src/components/va-button/va-button.css
```diff
@@ -0,0 +1,61 @@
...
+  margin: 0.5em 0.5em 0.5em 0;
+}
+
+.va-button:active,
+.va-button:hover {
+  background-color: var(--color-primary-darker);
+}
+
+.va-button.secondary {
+  background-color: var(--color-white);
```
k80bowman, line 23: Addresses https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/196 ✅

k80bowman (commented):
Overall this looks really great. I went through the issues Carol collected and the only outstanding ones were related to button variants we'll handle separately. I just have a couple of small comments.

packages/web-components/src/components/va-button/va-button.tsx
```diff
@@ -0,0 +1,124 @@
...
+        // TODO: add analytics event details
+      },
+    });
+  };
+
+  render() {
+    const {
+      back,
+      big,
+      continue: _continue,
```
andresriveratoro, line 93: I'm not a fan of this because it could cause engineers confusion, but it's a workaround because [continue](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/continue) is a JS statement.

packages/web-components/src/components/va-button/va-button.css
```diff
@@ -0,0 +1,61 @@
+@import '../../mixins/buttons.css';
+
+:host {
+  display: inline-block;
+}
+
+.va-button {
```
bkjohnson, line 7: Since we're working in the shadow DOM can we just style the `button` element directly and not use the class?

packages/web-components/src/components/va-button/va-button.tsx
bkjohnson: We can add a comment in the code explaining this if it would help bring clarity.

packages/web-components/src/components/va-button/va-button.tsx
```diff
@@ -0,0 +1,124 @@
...
+      secondary: secondary || back,
+    });
+    const type = submit ? 'submit' : 'button';
+
+    return (
+      <Host>
+        <button
+          aria-label={label}
+          class={styles}
+          disabled={disabled}
```
bkjohnson, line 113: Isn't this something that we decided against, or has that not happened yet?

packages/web-components/src/components/va-button/va-button.tsx
```diff
@@ -0,0 +1,124 @@
...
+          onClick={handleClick}
+          type={type}
+        >
+          {back && <i aria-hidden="true" class="fa fa-angles-left" />}
+          {text}
+          {_continue && <i aria-hidden="true" class="fa fa-angles-right" />}
```
bkjohnson, line 119: For the `back` and `continue` icons I think it would be good if we enforced that they are mutually exclusive, so something like:

```jsx
  {back && !_continue (<i aria-hidden="true" class="fa fa-angles-left" />)}
```

packages/web-components/src/components/va-button/va-button.tsx
```diff
@@ -0,0 +1,124 @@
...
+      <Host>
+        <button
+          aria-label={label}
+          class={styles}
+          disabled={disabled}
+          onClick={handleClick}
+          type={type}
+        >
+          {back && <i aria-hidden="true" class="fa fa-angles-left" />}
+          {text}
```
bkjohnson, line 118: Related to consistency, do we want to override the text if this is a `back` or `continue` button? It's easy to imagine teams creating back buttons that say "Go back a page", "Previous page", etc.

bkjohnson (commented):
I remember hearing that there was an accessibility ticket describing something that should be on all buttons, which is why we wanted a button web component. I don't know what the ticket is, but is that fix in here? I only see mapping the `label` prop to `aria-label`.

packages/web-components/src/components/va-button/va-button.tsx
andresriveratoro: I think that's a great idea.

packages/web-components/src/components/va-button/va-button.tsx
andresriveratoro: From our last DSC, it looks like there's at least 1 valid use case to have a disabled button: to prevent the user from submitting multiple times. I will leave `disabled` in, and update the disabled form state ticket to update `va-button` if a decision is made to disallow disabled.

k80bowman (approved):

PR #448, opened 2022-06-02
877 va-date Web Component Enhancements

pdavies88:
## Chromatic

https://877-va-date-component-enhancements--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/877

## Testing done
Screen Shot 2022-06-08 at 8 21 46 AM

## Screenshots
Screen Shot 2022-06-08 at 8 19 18 AM

## Acceptance criteria
- [x] Invalid data entry should trigger the Please provide a valid date error message
- [x] Valid year entry should be limited from 1900 through 2200
- [x] Add in ariadescribedBy prop on `va-date` component
- [x] Detail how the prop should be used in our docs by following guidance from: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby
- [x] Add in a `slot` to the `va-date` component so that developers can add html between the label and input fields
- [x] Add additional tests to cover these new features
- [x] Move onBlur event handler to component level from children to fire once component blur fires instead of each field
- [x] Required validation error should only fire onBlur
- [x] Add in custom validation props to handle via component onBlur
- [x] All validation for the component should only occur onBlur
- [x] Handles validation for Leap Years
- [x] Acceptable keys in the input fields are the following: 0123456789, Delete, Backspace, Tab, ArrowLeft, ArrowRight

## Definition of done


pdavies88:
Updates are dependent on decisions made in #438

packages/web-components/src/components/va-date/va-date.tsx
```diff
@@ -52,6 +79,39 @@ export class VaDate {
...
+      dayNum > daysForSelectedMonth.length ||
+      day === '' ||
+      month === '' ||
+      year === '' ||
+      (!leapYear && monthNum === 2 && dayNum > 28) ||
+      (this.required && !isFullDate(this.value))
+    ) {
+      this.error = 'Please enter a valid date';
+    } else if (this.error !== 'Please enter a valid date') {
+      this.error;
```
bkjohnson, line 111: What's this doing?

packages/storybook/stories/va-date.stories.jsx
```diff
@@ -40,70 +40,85 @@ const Template = ({ label, name, required, error, value }) => {
...
-        changes made to the select and input fields. If the criteria below is
-        not met an error message will show:
+        changes made to the input fields that fire when the component is blurred
+        ie: focus is removed from the component. If the criteria below is not
+        met an error message will show:
         <ul>
           <li>Cannot have blank values</li>
-          <li>The Year falls outside of the range of 1900-2122</li>
+          <li>Month and Day are not valid</li>
+          <li>The Year falls outside of the range of 1900-2200</li>
```
andresriveratoro, line 76: Should this be updated to current year + 100? Using a year of 2199 is showing a "Please enter a valid date" error.

pdavies88: Good catch yes the date utility it year + 100 need to update the verbiage to match that

packages/web-components/src/components/va-date/va-date.tsx
```diff
@@ -67,21 +127,40 @@ export class VaDate {
...
+      'ArrowRight',
+      'ArrowLeft',
+      'Tab',
+      'Delete',
+    ];
+    if (validKeys.indexOf(event.key) < 0) {
+      event.preventDefault();
+      return false;
+    }
+  };
```
bkjohnson, line 162: Are these functions separate for a reason? If not, could it live in the `date-utils` file and be shared by both components?

https://github.com/department-of-veterans-affairs/component-library/blob/4d6fe7b8883f254b08b0770dc14eca35efb461a5/packages/web-components/src/components/va-date-text-input/va-date-text-input.tsx#L139-L162

packages/storybook/stories/va-date.stories.jsx
pdavies88: Updated thanks!

packages/web-components/src/components/va-date/va-date.tsx
```diff
@@ -52,6 +79,39 @@ export class VaDate {
...
+    if (
+      yearNum < minYear ||
+      yearNum > maxYear ||
+      monthNum < minMonths ||
+      monthNum > maxMonths ||
+      dayNum < minMonths ||
+      dayNum > daysForSelectedMonth.length ||
+      day === '' ||
+      month === '' ||
+      year === '' ||
```
bkjohnson, line 105: If we're calling `parseInt` on the year, month, & day anyway, why not do it in the `.map()` function? I see that the non-parsed values are used down near the bottom of this if statement, but if we were dealing with `NaN` I think that something like `!!year` could work

packages/web-components/src/components/va-date/va-date.tsx
```diff
@@ -67,21 +127,40 @@ export class VaDate {
...
-      minimumIntegerDigits: 2,
-    });
 
-    // If Month or Date Select are chosen as empty return ''
-    // Otherwise convert numbers less than 10 to have a leading 0
-    this.value = `${currentYear}-${
-      currentMonth ? numFormatter.format(parseInt(currentMonth)) : ''
-    }-${currentDay ? numFormatter.format(parseInt(currentDay)) : ''}`;
+    this.value = `${currentYear}-${currentMonth ? currentMonth : ''}-${
+      currentDay ? currentDay : ''
```
bkjohnson, line 132: Since the current year, month, & day are coming from `target.value`, in what situations would they be falsey? All these ternaries make the expression harder to read, and I'm wondering if we could get away with something like:

```js
this.value = `${currentYear}-${currentMonth}-${currentDay}`;
```

packages/web-components/src/components/va-date/va-date.tsx
```diff
@@ -106,44 +195,39 @@ export class VaDate {
...
-    const daysForSelectedMonth = month > 0 ? days[month] : [];
-
-    // Check validity of date if invalid provide message and error state styling
-    const dateInvalid =
-      required && (!isFullDate(value) || day > daysForSelectedMonth.length)
-        ? 'Please provide a valid date'
-        : null;
+    const [year, month, day] = (value || '').split('-').map(val => val);
+    const monthNum = parseInt(month);
+    const dayNum = parseInt(day);
```
bkjohnson, line 204: Why is the `.map()`/`parseInt()` behavior changing here?

packages/web-components/src/components/va-date/va-date.tsx
pdavies88: It fixes the issue we were running into with someone setting the error prop elsewhere crashing with baked in validation that was causing me to have to add in that `invalid` prop originally. So now if someone uses custom validation like the final example in SB and then we also have default validation elsewhere they won't override each other

packages/web-components/src/components/va-date/va-date.tsx
pdavies88: Yeah good call I'll move it into date-utils

packages/web-components/src/components/va-date/va-date.tsx
```diff
@@ -106,44 +195,39 @@ export class VaDate {
...
     return (
-      <Host value={value} invalid={dateInvalid}>
-        <fieldset aria-label="Select Month and two digit day XX and four digit year format XXXX">
+      <Host value={value} error={error} onBlur={handleDateBlur}>
+        <fieldset aria-label="Select month and day fields are in two digit format XX and input year field is in four digit format XXXX">
           <legend>
-            {label} {required && <span class="required">(*Required)</span>}
+            {label}{' '}
+            {required && (
+              <span class="required">(*{i18next.t('required')})</span>
```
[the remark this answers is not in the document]
pdavies88: @k80bowman the only other hardcoded values are "Month", "Day", and "Year" do we not have translations for those?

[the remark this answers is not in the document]
pdavies88: @k80bowman we have the translations for Month, Day, and Year as well as the individual months so I applied them via this PR. All hardcoded text is now covered.

packages/web-components/src/components/va-date/va-date.tsx
pdavies88: Yes you are correct and I have updated accordingly to fix this issue.

packages/web-components/src/components/va-date/va-date.tsx
pdavies88: It should not you are correct and I have updated accordingly to fix this issue.

packages/web-components/src/components/va-date/va-date.tsx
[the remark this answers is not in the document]
k80bowman: What about `'Please enter a valid date'` from line 118 and `"Please enter two digits for the month"` from line 221 and `"Please enter two digits for the day"` from line 238 and `"Please enter four digits for the year"` from line 260?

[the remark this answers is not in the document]
pdavies88: @k80bowman "Please enter a valid date" is a specific case because we decided to bake in validation to these 2 components. I'm open to whatever but @bkjohnson had originally suggested putting i18next in place. I'll do whatever you all would like. Just let me know if I should leave it in or rip it out for both components.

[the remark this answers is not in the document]
pdavies88: Per standup today i18next has been remove from `va-date` and `va-date-text-input`

k80bowman (approved):

PR #452, opened 2022-06-02
Demo CSS utilities based on USWDS 3

bkjohnson:
## Chromatic

https://css-utilities-demo--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/859

#### What kind of configuration is necessary?

A: This PR represents the necessary SASS configuration for using USWDS3 to generate VADS utility classes.

#### Can we limit the colors available to the ones we have listed on design.va.gov?

A. Yes, we can define a set of color tokens to use

#### Can we use class names that match the class names in our current CSS class library?

A. Yes, but we would also have to include `!important` on each declaration in order to match our current utility classes

#### What are the differences between the CSS classes that are generated through USWDS v3 and our current CSS color-related classes?

A. We can make them identical if we want, but at the moment they have shorter names which are still descriptive, and they have minimal specificity.

For a file size comparison I made some slight modifications to `formation`:

```diff
diff --git a/packages/formation/sass/base/_b-variables.scss b/packages/formation/sass/base/_b-variables.scss
index ca6dcf594..34053c922 100644
--- a/packages/formation/sass/base/_b-variables.scss
+++ b/packages/formation/sass/base/_b-variables.scss
@@ -47,6 +47,7 @@ $color-red-light:            #e59393;
 $color-red-dark:             #cd2026;
 $color-blue:                 #0071BB;
 $color-blue-darkest:         #112e51;
+$color-transparent:         transparent;
 
 $color-base:                 #212121;
 
diff --git a/packages/formation/sass/css-library.scss b/packages/formation/sass/css-library.scss
new file mode 100644
index 000000000..944892dd1
--- /dev/null
+++ b/packages/formation/sass/css-library.scss
@@ -0,0 +1,20 @@
+@charset "UTF-8";
+// Note: if you add variable files in here, make sure you updated shared-variables.scss
+
+// ============ SETTINGS ============//
+// Order matters here! Please don't change it.
+// @import '~uswds/src/stylesheets/core/variables';
+
+// A VA partial. Here so that we can override the
+// USWDS breakpoint and grid variables from above before
+// the grid, alerts, etc are loaded.
+@import 'base/b-variables';
+
+// ----- VA BASE (VARIABLES/HELPERS) ---- //
+@import 'base/b-element-overrides';
+@import 'base/b-functions';
+
+
+// ----- VA UTILITIES (UTILITIY OOCSS) ---- //
+@import 'utilities/background-color';
+@import 'utilities/color';
```

And then ran:

```
npx sass --no-source-map --style compressed sass/css-library.scss formation-color-utilities.css
```

The `formation-color-utilities.css` file uncomressed was 7.2K, while an identically built `utilities.css` stylesheet from this new library was 4.3K. That's a **40% smaller file size just from having smaller class names and not putting `!important` on every declaration.**

## Acceptance criteria
- [ ]

## Definition of done


packages/css-library/src/tokens/color.scss
```diff
@@ -0,0 +1,67 @@
+// Our colors
+$tokens-color-basic: (
+  "base": "#212121",
+  "white": "#ffffff",
+  "black": "#000000",
+  "orange": "#eb7f29",
```
humancompanion-usds, line 6: Is orange used anywhere? This one seems like an outlier.

packages/css-library/src/tokens/color.scss
```diff
@@ -0,0 +1,67 @@
...
+  "green-light": "#4aa564",
+  "green-lighter": "#94bfa2",
+  "green-lightest": "#e7f4e4",
+  "cool-blue": "#205493",
+  "cool-blue-light": "#4773aa",
+  "cool-blue-lighter": "#8ba6ca",
+  "cool-blue-lightest": "#dce4ef",
+);
+
+// TODO: Consider removing these since they aren't used
```
humancompanion-usds, line 55: I agree they are not in use but, they should be and those are the right colors for those hub sections. For example, the icon in use in the disability section should use the hub-disability color token.

PR #464, opened 2022-06-15
41381 Storybook Side Navigation Groupings

pdavies88:
## Chromatic

https://41381-storybook-side-nav--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/41381

## Testing done / Screenshots
Screen Shot 2022-06-15 at 8 13 15 AM

## Acceptance criteria
- [X] Rework titles in storybook files to match hierarchy listed in ticket
- [X] Add Permalinks to storybook files to help us from needing to update design.va.gov with our new url paths
- [ ] After changing titles and adding permalinks put an announcement in #vfs-frontend #vfs-all-teams #vfs-engineers and #design

## Definition of done

### DIFF
--- a/packages/storybook/stories/DropDownPanel.stories.jsx
+++ b/packages/storybook/stories/DropDownPanel.stories.jsx
@@ -2,7 +2,8 @@ import React, { useState } from 'react';
 import DropDownPanel from '../../react-components/src/components/DropDownPanel/DropDownPanel';
 
 export default {
-  title: 'Components/DropDownPanel',
+  title: 'Components/Drop Down Panel/React Component/Drop Down Panel',
+  id: 'Components/DropDownPanel',
   component: DropDownPanel,
 };
 
--- a/packages/storybook/stories/additional-docs.js
+++ b/packages/storybook/stories/additional-docs.js
@@ -5,12 +5,22 @@ const { AVAILABLE, BEST_PRACTICE, CANDIDATE, DEPLOYED, DEPRECATED, PROPOSED } =
   level;
 
 export const additionalDocs = {
+  'Accept terms prompt': {
+    maturityCategory: CAUTION,
+    maturityLevel: DEPLOYED,
+  },
+  'Accordion': {
+    maturityCategory: USE,
+  },
   // MDX
   'Action link': {
     guidanceHref: 'link/action',
     guidanceName: 'Action link',
     maturityCategory: USE,
     maturityLevel: DEPLOYED,
+  },
+  'Additional info': {
+    maturityCategory: USE,
   },
   // MDX
   'Address block': {
@@ -19,16 +29,27 @@ export const additionalDocs = {
     maturityCategory: USE,
     maturityLevel: DEPLOYED,
   },
+  'Alert': {
+    maturityCategory: USE,
+  },
+  'Alert - expandable': {
+    maturityCategory: CAUTION,
+  },
+  'Back to top': {
+    maturityCategory: USE,
+  },
+  'Banner': {
+    maturityCategory: USE,
+  },
+  'Breadcrumbs': {
+    maturityCategory: USE,
+  },
   // MDX
   'Button': {
     guidanceHref: 'button',
     guidanceName: 'Button',
     maturityCategory: USE,
     maturityLevel: BEST_PRACTICE,
-  },
-  'Breadcrumbs': {
-    maturityCategory: DONT_USE,
-    maturityLevel: DEPRECATED,
   },
   'Checkbox': {
     maturityCategory: DONT_USE,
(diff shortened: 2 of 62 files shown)


pdavies88:
@k80bowman updates have been made
Screen Shot 2022-06-16 at 12 38 17 PM

k80bowman (approved):

k80bowman (dismissed):
I just realized there is one thing we need to change on this. Several of the React components now do not have any maturity scale information on their Storybook pages. I'll see what I can do about that.

packages/storybook/stories/DropDownPanel.stories.jsx
```diff
@@ -2,7 +2,8 @@ import React, { useState } from 'react';
 import DropDownPanel from '../../react-components/src/components/DropDownPanel/DropDownPanel';
 
 export default {
-  title: 'Components/DropDownPanel',
+  title: 'Components/Drop Down Panel/React Component/Drop Down Panel',
```
andresriveratoro, line 5: Should these be updated to sentence case?

`Components/Drop down panel/React component/Drop down panel`

packages/storybook/stories/additional-docs.js
```diff
@@ -45,26 +70,79 @@ export const additionalDocs = {
...
+    maturityCategory: USE,
+  },
+  'File Input': {
+    maturityCategory: CAUTION,
+    maturityLevel: AVAILABLE,
+  },
+  'Form': {
+    maturityCategory: USE,
+    maturityLevel: BEST_PRACTICE,
+  },
```
andresriveratoro, line 91: I know it's like this on design.va.gov, but should we display maturity scale for parents? Form appears with a green icon in Storybook, but several of its nested components have use with caution or don't use.

packages/storybook/stories/va-accordion.stories.jsx
```diff
@@ -10,7 +10,8 @@ const accordionDocs = getWebComponentDocs('va-accordion');
 const accordionItem = getWebComponentDocs('va-accordion-item');
 
 export default {
-  title: 'Components/va-accordion',
+  title: 'Components/Accordion/va-accordion',
+  id: 'Components/va-accordion',
```
bkjohnson, line 14: How is this ID being used? I looked at the [storybook documentation](https://storybook.js.org/docs/react/api/csf) and it wasn't obvious.

packages/storybook/stories/additional-docs.js
```diff
@@ -108,6 +215,7 @@ export const additionalDocs = {
     maturityCategory: DONT_USE,
     maturityLevel: DEPRECATED,
   },
+  // Web Component Tag Linkage
```
bkjohnson, line 218: Are these still being used?

packages/storybook/stories/additional-docs.js
```diff
@@ -5,31 +5,56 @@ const { AVAILABLE, BEST_PRACTICE, CANDIDATE, DEPLOYED, DEPRECATED, PROPOSED } =
...
+  },
+  'Back to top': {
+    maturityCategory: USE,
+  },
+  'Banner': {
+    maturityCategory: USE,
+  },
+  'Breadcrumbs': {
+    maturityCategory: USE,
+  },
```
bkjohnson, line 46: Is there a reason why these don't have a `maturityLevel`?

bkjohnson (commented):
I'm not a huge fan of redundant nesting like this:

va-telephone item inside Telephone

Also, this is taken from the chromatic instance for this PR:

image

I much prefer what's currently on design.va.gov since the indicators more clearly indicate what to use:

TextInput react component & va-text-input highlighted in storybook sidebar with the maturity indicators on each of them individually

Can we preserve that?

packages/storybook/stories/va-accordion.stories.jsx
andresriveratoro: This id is used as a [permalink](https://storybook.js.org/docs/react/configure/sidebar-and-urls#permalink-to-stories) to prevent having to update links if the component structure changes in Storybook.

packages/storybook/stories/additional-docs.js
k80bowman: Those are only used for the sidebar indicators, so a `maturityLevel` is not needed.

andresriveratoro:
> I'm not a huge fan of redundant nesting like this:
> 
> va-telephone item inside Telephone
> 
> Also, this is taken from the chromatic instance for this PR:
> 
> image
> 
> I much prefer what's currently on design.va.gov since the indicators more clearly indicate what to use:
> 
> TextInput react component & va-text-input highlighted in storybook sidebar with the maturity indicators on each of them individually
> 
> Can we preserve that?

I think the intent with these changes was to more closely resemble the structure that we're using on design.va.gov for consistency, but it'll be slightly different since we don't display React components on design.va.gov when the web component exists for it. I don't think that we need to explicitly organize them by framework.

Would something like this work?
Screen Shot 2022-06-24 at 17 04 06

k80bowman:
So, yes, I am also not much of a fan of the nesting, especially for things like the form controls (I'm actually not a fan of this on design.va.gov either). We have a separate ticket to handle this, but I'm wondering if maybe we can try something different (and easier). I'll bring this up in standup so we can discuss.

For right now, because we have multiple versions of the same components (while we are in the process of deprecating the React components), I do think it is useful to indicate which framework the components are designed for. We could do this in a couple of ways. First, we could do something like this:
image

We could just separate out the deprecated components as well:
image

Taking that a step further, we could group the components by maturity scale level. I know this gets us further away from what we have on design.va.gov, but it could be useful:
image

k80bowman:
@humancompanion-usds I'd love to get your feedback on this as well.

k80bowman:
We talked about this as a team today and decided on a new way to structure the sidebar. I'm going to go ahead and close this PR and update the ticket with the new structure.

PR #467, opened 2022-06-21
change name of va-date-text-input to va-memorable-date

k80bowman:
## Chromatic

https://date-text-name-change--60f9b557105290003b387cd5.chromatic.com

## Description
Changing the name of the va-date-text-input component to va-memorable-date to match [USWDS](https://designsystem.digital.gov/components/memorable-date/).

I realize this will break design.va.gov storybook links. I will update those as soon as this gets merged.

## Acceptance criteria
- [ ]

## Definition of done


packages/core/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@department-of-veterans-affairs/component-library",
   "description": "VA.gov component library. Includes React and web components.",
-  "version": "10.6.2",
+  "version": "11.0.0",
```
k80bowman, line 4: No one is using this component yet, so I waffled on whether to make this a breaking change, but a component name change should be a breaking change whether anyone is using it or not.

packages/storybook/stories/additional-docs.js
```diff
@@ -155,7 +155,9 @@ export const additionalDocs = {
     maturityCategory: USE,
     maturityLevel: DEPLOYED,
   },
-  'va-date-text-input': {
+  'va-memorable-date': {
+    guidanceHref: 'form/memorable-date',
+    guidanceName: 'Memorable Date',
```
andresriveratoro, line 160: We don't need to specify a guidance name for web components unless we want it set to something other than the component name.

https://github.com/department-of-veterans-affairs/component-library/blob/master/packages/storybook/stories/wc-helpers.jsx#L230-L235

packages/storybook/stories/va-memorable-date.stories.jsx
```diff
@@ -1,17 +1,19 @@
 import React, { useState } from 'react';
-import { VaDateTextInput } from '@department-of-veterans-affairs/web-components/react-bindings';
+import { VaMemorableDate } from '@department-of-veterans-affairs/web-components/react-bindings';
 import { getWebComponentDocs, propStructure, StoryDocs } from './wc-helpers';
 
-VaDateTextInput.displayName = 'VaDateTextInput';
+console.log(VaMemorableDate);
```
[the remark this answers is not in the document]
k80bowman: gah, thank you!

packages/storybook/stories/va-memorable-date.stories.jsx
```diff
@@ -1,17 +1,19 @@
...
-const dateTextInputDocs = getWebComponentDocs('va-date-text-input');
+VaMemorableDate.displayName = 'VaMemorableDate';
+
+const memorableDateInputDocs = getWebComponentDocs('va-memorable-date');
 
 export default {
-  title: 'Components/va-date-text-input',
+  title: 'Components/va-memorable-date',
   parameters: {
     componentSubtitle: `Date web component`,
```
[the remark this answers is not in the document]
k80bowman: Yes it should, thank you.

packages/core/package.json
bkjohnson, line 4: Do we want to use major changes even when the component is considered experimental or in the "use with caution" phase?

bkjohnson (approved):
One small question, but things look good.

packages/core/package.json
k80bowman: As I wrote in a comment above, I waffled on this. No one is using it yet, and you're right it is an experimental component, but I'm not sure it's good practice to change the name of a component and _not_ have it be a major change. I could go either way.

andresriveratoro: I'm in favor of using the correct versioning regardless of usage for consistency. Is there a reason NOT to treat this as a breaking change?

PR #468, opened 2022-06-23
Display selected option when value is changed programmatically

andresriveratoro:
## Chromatic

https://932-select--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/932

## Testing done
Storybook

## Acceptance criteria
- [x] va-select displays correct value when set programmatically

## Definition of done


andresriveratoro:
I haven't been able to add a test for programmatic changes to `value` because the `selected` attribute isn't reflecting in the DOM. My goal was to look for the `option` element with the selected attribute. A possible workaround could be adding the `aria-selected` attribute and querying for that since it shows in the DOM. Thoughts?

```jsx
 <option
  value={node.value}
  selected={value === node.value}
  aria-selected={value === node.value}
 >
  {node.text}
</option>
```

bkjohnson:
Could a test look something like this?

1. Initial setup, maybe a basic `expect` assertion
1. Update children with something like:
    ```
    await page.$eval('va-select', (element: HTMLVaSelectElement) => element.innerHTML = `<option value="new">New opt</option>`;
    ```
1. Update the `value` prop
1. Check to see that the options have changed

My [understanding of the issue](https://dsva.slack.com/archives/CBU0KDSB1/p1654646401906479?thread_ts=1654629669.256369&cid=CBU0KDSB1) is that it's not really about `selected`, and more about the `<option>` children not changing after that initial render. So I'm not sure if including `selected` in the test would be necessary, but I might be missing something.

k80bowman (approved):

PR #472, opened 2022-06-28
Add va-button-pair component

andresriveratoro:
## Chromatic

https://937-button-pair--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/937

## Screenshots
Screen Shot 2022-07-01 at 11 48 43
Screen Shot 2022-07-01 at 11 48 31

## Acceptance criteria
- [x] Component is written and added to Storybook
- [ ] Component has had accessibility and design reviews

## Definition of done


andresriveratoro:
va-button-pair displays Back/Continue as default and Confirm/Cancel when `confirm` is true.

packages/web-components/src/components/va-button/va-button.tsx
```diff
@@ -20,47 +20,47 @@ export class VaButton {
...
    * If `true`, the button will submit form data when clicked.
    */
-  @Prop() submit?: boolean = false;
+  @Prop({ reflect: true }) submit?: boolean = false;
 
   /**
    * The text displayed on the button. If `continue` or `back` is true, the value of text is ignored.
    */
-  @Prop() text?: string;
+  @Prop({ reflect: true }) text?: string;
```
bkjohnson, line 63: Why do we need all of these props reflected?

andresriveratoro: I've updated it to only reflect props that are needed for styling. Without reflecting props like `continue`, styling rules such as `:host([continue]:not([continue='false']))` don't get applied because the prop isn't reflected in the DOM.

packages/web-components/src/components/va-button-pair/va-button-pair.tsx
```diff
@@ -0,0 +1,144 @@
+import { Component, Event, EventEmitter, Host, h, Prop } from '@stencil/core';
+
+@Component({
+  tag: 'va-button-pair',
+  styleUrl: 'va-button-pair.css',
+  shadow: true,
+})
+export class VaButtonPair {
+  /**
+   * If `true`, button pair will use Confirm and Cancel for button text.
```
bkjohnson, line 10: Can we add something here describing the Yes/No text for when `confirm` and `continue` are both false? I don't want that to be unexpected.

bkjohnson (approved):
One small comment, things look good though.

packages/web-components/src/components/va-button-pair/va-button-pair.tsx
bkjohnson: In storybook I see that the default story has the Yes/No text, so this isn't as unexpected as I thought :upside_down_face:

k80bowman:
I'm not sure if something changed in the design somewhere along the way, but it looks like in Sketch the update/cancel and yes/no variations are also side-by-side on larger screens. Looks like you might need to update the CSS to reflect this.
image

andresriveratoro:
I'll confirm with @GnatalieH as those may not be the latest mockups. Update/Cancel should be Confirm/Cancel too.

andresriveratoro:
Updated confirm pair to Update/Cancel and fixed styles

PR #476, opened 2022-06-29
Discovery for design token library

bkjohnson:
## Chromatic

https://style-dictionary-discovery--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/934

This is discovery for an initial design token library using style dictionary. I did a brief search to see what other tools were available for this kind of thing, but none of them served the same purpose as style dictionary or had the same community size.

#### What would the LOE be to complete this work?

I think we could get all of our color, spacing, & font tokens into the style dictionary config and have it setup to work with our new CSS utilities in a few days of focused work.

#### Is [Style Dictionary](https://amzn.github.io/style-dictionary/#/) the best tool for this or are there other options we should consider?

It appears that Style Dictionary is the leader. The next best tool I found was [Theo](https://github.com/salesforce-ux/theo) which is from Salesforce, but this hasn't been updated since last October and it has fewer stars on Github, so a smaller community to submit bug reports and feature PRs.

#### What are the benefits of having a VADS token library? For our team? For other teams?

We would be able to maintain a single source of truth for design tokens and we can provide [formats](https://amzn.github.io/style-dictionary/#/formats?id=pre-defined-formats) that will work in SASS, CSS variables, ES modules for JS, etc. There are even formats for exporting tokens as resources for android & iOS. We can also write our own [custom formats](https://amzn.github.io/style-dictionary/#/formats?id=custom-formats). This would allow us to accommodate a wide variety of teams and tools where Design System usage would be beneficial. This may allow us to export tokens that could be imported into Sketch, Figma, & UXPin for example.

#### What would be the benefits to beginning our work to transition to USWDS v3 with the development of a design token library?

Paired with the [results of the early discovery on a CSS utility library usind USWDS v3](https://github.com/department-of-veterans-affairs/component-library/tree/8f90c0c97c5ebf4aead0467f6530f9d7e4a6da84/packages/css-library), we could use the output of the style dictionary build to provide tokens for the utility classes. 

#### Would there be any drawbacks to doing this work first (as opposed to maybe doing it later)? What are they?

We may end up with a configuration setup that isn't ideal and difficult to change.

## Testing done

Local builds and :eyes: 

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/storybook/stories/wc-helpers.jsx
+++ b/packages/storybook/stories/wc-helpers.jsx
@@ -84,13 +84,23 @@ export const propStructure = comp => {
     propObj[prop.attr] = {
       description: prop.docs,
       required: prop.required,
-      defaultValue: { value: prop.default },
+      defaultValue: {
+        value: prop.default,
+      }, // Used in subcomponents
       type: {
-        name: prop.type,
+        name: prop.type, // Used in subcomponents
+        required: prop.required,
       },
       // Assigns the argType to the Properties category
       table: {
         category: 'Properties',
+        defaultValue: {
+          summary: prop.default,
+        },
+        type: { summary: prop.type },
+      },
+      control: {
+        type: prop.values[0].type === 'string' ? 'text' : prop.values[0].type,
       },
     };
     return propObj;
--- a/packages/web-components/src/components.d.ts
+++ b/packages/web-components/src/components.d.ts
@@ -10,47 +10,47 @@ export namespace Components {
         /**
           * Whether or not the accordion items will have borders
          */
-        "bordered": boolean;
+        "bordered"?: boolean;
         /**
-          * If true, doesn't fire the CustomEvent which can be used for analytics tracking.
+          * If `true`, doesn't fire the CustomEvent which can be used for analytics tracking.
          */
-        "disableAnalytics": boolean;
+        "disableAnalytics"?: boolean;
         /**
           * True if only a single item can be opened at once
          */
-        "openSingle": boolean;
+        "openSingle"?: boolean;
         /**
           * Optional accordion section heading text. Only used in analytics event. Default is null.
          */
-        "sectionHeading": string;
+        "sectionHeading"?: string;
     }
     interface VaAccordionItem {
         /**
           * The accordion item header text
          */
-        "header": string;
+        "header"?: string;
         /**
           * Header level for button wrapper. Must be between 1 and 6
          */
-        "level": number;
+        "level"?: number;
         /**
           * True if the item is open
          */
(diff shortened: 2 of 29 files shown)


packages/css-library/src/tokens/color.scss
```diff
@@ -1,27 +1,27 @@
...
 
 // Our colors
 $tokens-color-basic: (
-  "base": "#212121",
-  "white": "#ffffff",
-  "black": "#000000",
+  "base": $color-base,
+  "white": $color-white,
+  "black": $color-black,
   "orange": false,
```
bkjohnson, line 8: This is just one possible way to use the tokens. We could also have style dictionary build a [SASS map format](https://amzn.github.io/style-dictionary/#/formats?id=scssmap-flat), but that would put all tokens (colors, spacing, etc.) into the same map which we may not want. A custom formatter may allow us to have multiple SASS maps build in the same file.

If we went in that direction we may be able to have our custom formatter recognize [custom token attributes](https://amzn.github.io/style-dictionary/#/tokens?id=design-token-attributes) to allow us to do something like:

```json
"orange": { "value": "#eb7f29", "attributes": { "utilityClass": false } }
```

packages/css-library/config.json
```diff
@@ -0,0 +1,13 @@
+{
+  "source": ["tokens/**/*.json"],
+  "platforms": {
+    "scss": {
+      "transformGroup": "scss",
+      "buildPath": "src/tokens/",
+      "files": [{
+        "destination": "_variables.scss",
```
bkjohnson, line 8: We can name the filename and path to whatever we want and make sure that those files are included in the npm package.

packages/css-library/tokens/color/base.json
```diff
@@ -0,0 +1,36 @@
...
+    "primary-alt-light": { "value": "#9bdaf1"},
+    "primary-alt-lightest": { "value": "#e1f3f8"},
+    "secondary": { "value": "#e31c3d"},
+    "secondary-dark": { "value": "#cd2026"},
+    "secondary-darkest": { "value": "#981b1e"},
+    "secondary-light": { "value": "#e59393"},
+    "secondary-lightest": { "value": "#f9dede"},
+    "gray": {
+      "gray": { "name": "color-gray", "value": "#5b616b"},
+      "dark": { "value": "#323a45"},
```
bkjohnson, line 25: I left the primary & secondary groups of colors un-nested just to show some different options for structuring the tokens. This line right here would give us a `color-gray-dark` token.

PR #481, opened 2022-07-08
Add va-link component

andresriveratoro:
## Chromatic

https://974-va-link--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/974

## Acceptance criteria
- [x] va-link component has been created
- [x] va-link exists in Storybook

## Definition of done


packages/storybook/stories/va-link.stories.jsx
```diff
@@ -0,0 +1,159 @@
...
+      filename={filename}
+      video={video}
+    >
+      Download VA form 10-10EZ
+      <dfn>
+        (<abbr title="Portable Document Format">PDF</abbr>, 5 pages)
+      </dfn>
+    </va-link>
+  );
+};
```
andresriveratoro, line 97: I opted for creating multiple templates to have better `Show Code` examples without unnecessary fragments in them.

packages/storybook/stories/va-link.stories.jsx
```diff
@@ -0,0 +1,159 @@
...
+      channel={channel}
+      download={download}
+      href={href}
+      filename={filename}
+      video={video}
+    >
+      Download VA form 10-10EZ
+      <dfn>
+        (<abbr title="Portable Document Format">PDF</abbr>, 5 pages)
+      </dfn>
```
bkjohnson, line 94: Since this looks like a standard format, is there a way we can parse the file and extract the page count in order to keep things consistent and accurate? I can imagine if someone updates the PDF asset but not the link that this bit of content could get out of date and be misleading. And since we want that content there by default, automatically including it could make the component easier to use and more reliable.

Link guidance on documentation with the structure "(PDF, page count)" highlighted

packages/web-components/src/components/va-link/va-link.tsx
```diff
@@ -0,0 +1,115 @@
...
+@Component({
+  tag: 'va-link',
+  styleUrl: 'va-link.css',
+  shadow: true,
+})
+export class VaLink {
+  /**
+   * If `true`, the anchor text will be bolded and include a right arrow icon.
+   */
+  @Prop() active?: boolean = false;
```
bkjohnson, line 20: I'm a bit confused about this prop. Is it meant to be different than the notion of `:active` in CSS?

bkjohnson (commented):
I see that there's some [existing guidance](https://design.va.gov/components/link/) on links and their variants, such as:

image

For the media-oriented links, I'm wondering if we should build for that structure instead of expect people to always use slots to follow it. For example, we could parse the `href` and use it to make an [API call to tell us what the channel's name is](https://developers.google.com/youtube/v3/code_samples/code_snippets?apix=true&apix_params=%7B%22part%22%3A%5B%22snippet%2CcontentDetails%2Cstatistics%22%5D%2C%22id%22%3A%5B%22UCBvOzPLmbzjtpX-Htstp2vw%22%5D%7D) (uncheck Google OAuth 2.0 and click "Execute")

Google's API explorer showing us JSON representing info on the VA's youtube channel, with the channel title highlighted

The same thing could be done for video titles.

packages/web-components/src/components/va-link/va-link.tsx
andresriveratoro: This prop is meant to be used when the [Active link variant](https://design.va.gov/components/link/#active-link) is required.

packages/storybook/stories/va-link.stories.jsx
```diff
@@ -0,0 +1,159 @@
...
+      active={active}
+      channel={channel}
+      download={download}
+      href={href}
+      filename={filename}
+      video={video}
+    >
+      Download VA form 10-10EZ
+      <dfn>
+        (<abbr title="Portable Document Format">PDF</abbr>, 5 pages)
```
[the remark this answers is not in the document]
andresriveratoro: Good catch. I've been unable to get the `abbr` element to use the correct color based on the status of the anchor. It seems I am not able to rely on the `:visited` selector to update the color of the nested `abbr`. Any ideas @bkjohnson ?

This works:
```css
va-link abbr {
  color: var(--color-link-default) !important;
}
```

This doesn't work:
```css
va-link a:visited abbr {
  color: var(--color-link-visited) !important;
}
```

```css
a:visited ::slotted(*) {
  color: var(--color-link-visited) !important;
}
```

Help! 🆘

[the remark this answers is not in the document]
andresriveratoro: As a workaround, we could do something like this:

```diff
diff --git a/packages/web-components/src/components/va-link/va-link.tsx b/packages/web-components/src/components/va-link/va-link.tsx
index 3c19d8e5..2cb1f14a 100644
--- a/packages/web-components/src/components/va-link/va-link.tsx
+++ b/packages/web-components/src/components/va-link/va-link.tsx
@@ -44,6 +44,16 @@ export class VaLink {
    */
   @Prop() filename?: string;
 
+  /**
+   * The type of the file. Only valid if download is `true`.
+   */
+  @Prop() filetype?: 'PDF';
+
+  /**
+   * The number of pages of the file. Only valid if download is `true`.
+   */
+  @Prop() pages?: number;
+
   // @Prop() type?: string;
 
   /**
@@ -74,15 +84,23 @@ export class VaLink {
     });
   };
 
+  private getAbbreviationTitle = () => {
+    const { filetype } = this;
+    if (!filetype) return;
+    if (filetype === 'PDF') return 'Portable Document Format';
+  };
+
   render() {
     const {
       active,
       channel,
       download,
+      filetype,
       filename,
+      getAbbreviationTitle,
       handleClick,
       href,
-      // type
+      pages,
       video,
     } = this;
 
@@ -91,8 +109,6 @@ export class VaLink {
         <a
           href={href}
           download={download ? filename : undefined}
-          // Property 'type' does not exist on type 'AnchorHTMLAttributes<HTMLAnchorElement>'
-          // type={download ? type : undefined}
           onClick={handleClick}
           rel={channel || video ? 'noopener' : undefined}
           target={channel || video ? '_blank' : undefined}
@@ -108,6 +124,12 @@ export class VaLink {
               <slot></slot>
             </Fragment>
           )}
+          {download && filetype && (
+            <dfn>
+              (<abbr title={getAbbreviationTitle()}>{filetype}</abbr>
+              {pages && { pages }})
+            </dfn>
+          )}
         </a>
       </Host>
     );
```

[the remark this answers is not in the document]
bkjohnson: This looks like a Chrome bug, because in firefox it looks like how I expect it to:

"PDF" abbreviation is the same color as the surrounding text

However, since the problem still exists in Chrome we aren't going to be able to solve it with just CSS due to the `<abbr>` being nested inside of another element in the slot. Solutions I can think of are:

1. Move the `(<abbr title="Portable Document File">PDF</abbr>), N pages)` inside of the component
    - We could give a prop for the page count, or parse the pdf with [a dependency like this](https://github.com/adrienjoly/npm-pdfreader) to get the count ourselves. I saw your earlier comment about not wanting to have a user's browser download a pdf just to show the page count though, and I kind of agree with that.
1. Use some JS to add a class to the nested `<abbr>` to achieve the desired styling like we're doing elsewhere:
    - https://github.com/department-of-veterans-affairs/component-library/blob/45608acf85449782307945c77e56994a3ceaacb8/packages/web-components/src/components/va-modal/va-modal.tsx#L292

The simplest solution may be the extra prop.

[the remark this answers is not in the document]
k80bowman: I agree about having the slot or the props, but not both.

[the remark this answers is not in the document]
andresriveratoro: Agreed, I will remove the slot to make it easier to use and bake in the `<dfn>` and `<abbr>` elements.

packages/web-components/src/components/va-link/va-link.tsx
```diff
@@ -0,0 +1,169 @@
...
+    if (video) {
+      return (
+        <Host>
+          <a href={href} onClick={handleClick} rel="noopener" target="_blank">
+            <i aria-hidden="true" />
+            {text} <dfn>on YouTube</dfn>
+          </a>
+        </Host>
+      );
+    }
```
k80bowman, line 158: It looks like the only difference between this variant and the channel variant is whether it says "on YouTube" or just "YouTube". Could these two variants be combined and the text determined by a conditional?

packages/web-components/src/components/va-link/va-link.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+  /**
+   * The suggested filename. Only valid if download is `true`.
+   */
+  @Prop() filename?: string;
+
+  /**
+   * The type of the file. Only displayed if download is `true`.
+   */
+  @Prop() filetype?: 'PDF';
```
k80bowman, line 42: If this only accepts a PDF type, should we set a default?

packages/web-components/src/components/va-link/va-link.tsx
```diff
@@ -0,0 +1,169 @@
...
+        <Host>
+          <a href={href} download={filename} onClick={handleClick}>
+            <i aria-hidden="true" />
+            {text}{' '}
+            {filetype && (
+              <dfn>
+                (<abbr title={getAbbreviationTitle()}>{filetype}</abbr>
+                {pages &&
+                  typeof pages === 'number' &&
+                  `, ${pages.toString()} pages`}
```
k80bowman, line 139: could you add a test for this logic? Something like "does not report number of pages if typeof pages is not number" or something to that effect.

packages/web-components/src/components/va-link/va-link.tsx
```diff
@@ -0,0 +1,169 @@
...
+  @Event({
+    bubbles: true,
+    composed: true,
+    eventName: 'component-library-analytics',
+  })
+  componentLibraryAnalytics: EventEmitter;
+
+  private handleClick = (e: MouseEvent): void => {
+    // TODO: remove console.log
+    console.log(e);
```
k80bowman, line 71: Just highlighting this, though I'm sure you're aware since you left a TODO.

packages/web-components/src/components/va-link/va-link.tsx
andresriveratoro: Our [guidance](https://design.va.gov/components/link/#download-link) hints that it's not limited to PDFs, but I haven't found a download link with a different file type. I could make this the default if we want that.

packages/web-components/src/components/va-link/va-link.tsx
andresriveratoro: This is here to allow the web component to build. I can remove the function parameter or leave the console log until we're using the event in the analytics event.

packages/web-components/src/components/va-link/va-link.tsx
k80bowman: Maybe making it a default isn't the answer. I'm a little concerned that limiting the type to PDF will cause problems, since the guidance is not limited to PDF. It just means that if someone wants to use this component with a different file type we'll have to update it, right?

andresriveratoro: I added a prop named `abbrTitle` to get around this. 
```jsx
<va-link 
  abbr-title="Microsoft Word Document"
  download
  text="Download VA form 10-10EZ"
  filename="10-10EZ.docx"
  filetype="DOCX"
  pages="2"
/>
```

andresriveratoro: This could possibly lead to inconsistent titles used for the `abbr` title, but I think it's an acceptable solution since we're not limiting the file extensions that can be used.

k80bowman (approved):

packages/web-components/src/components/va-link/va-link.tsx
```diff
@@ -0,0 +1,159 @@
...
+@Component({
+  tag: 'va-link',
+  styleUrl: 'va-link.css',
+  shadow: true,
+})
+export class VaLink {
+  /**
+   * The title used in the abbr element. If filetype is PDF, the abbr title will be Portable Document Format.
+   */
+  @Prop() abbrTitle?: string;
```
bkjohnson, line 12: Should we default this to PDF?

packages/web-components/src/components/va-link/va-link.css
```diff
@@ -0,0 +1,57 @@
+@import '../../mixins/links.css';
+
+:host {
+  display: block;
```
bkjohnson, line 4: It looks like at least our default links are meant to be inline-block rather than full block.

Link examples inline

And the [guidance at the top says](https://design.va.gov/components/link/)  says:

>  A link is a navigation element that can appear alone, inline (embedded), or in a group with other links.

I don't know if there are any production examples of inline variants like the download link, but we may want to display the component as inline-block.

bkjohnson (commented):
Mostly minor comments, one potentially larger one about the CSS `display` declaration. If that gets solved or it's fine as is then I'm fine with everything else.

packages/web-components/src/components/va-link/va-link.css
bkjohnson: On that note, maybe we should display our links inline in some of the storybook examples.

bkjohnson (approved):
:+1:

PR #482, opened 2022-07-11
Update Storybook sidenav

k80bowman:
## Chromatic

## Description
Closes department-of-veterans-affairs/va.gov-team#41381

## Acceptance criteria
- [ ]

## Definition of done


packages/storybook/stories/va-table.stories.jsx
```diff
@@ -2,13 +2,14 @@ import React from 'react';
...
 
 export default {
-  title: 'Components/va-table',
+  title: `Components/${componentName}`,
   parameters: {
-    componentSubtitle: 'Table web component',
+    componentSubtitle: 'va-table',
     docs: {
-      page: () => <StoryDocs data={tableDocs} />,
+      page: () => <StoryDocs componentName={componentName} data={tableDocs} />,
```
bkjohnson, line 12: I don't think this `componentName` is really doing anything that the default doesn't do. Is there a reason we have it and others like it?

bkjohnson (commented):
We should make sure this gets merged around the same time that we merge a PR in the documentation repo which [updates all of the storybook links](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/74e59f4772bae22f70c25fe23f69405db6692d75/src/_components/additional-info.md?plain=1#L21).

packages/storybook/stories/va-table.stories.jsx
k80bowman: Yes, this is how I am adding the sidebar maturity scale indicator for the web components.

I could just change the tag in the component-docs.json and it would accomplish the same thing, but I wasn't sure what else that would break.

packages/storybook/stories/buttons.stories.mdx
```diff
@@ -3,17 +3,17 @@ import { Guidance, MaturityScale } from './wc-helpers';
 import { additionalDocs } from './additional-docs';
 import './buttons.style.scss';
 
-<Meta title="Components/Buttons/Button" />
+<Meta title="Components/Buttons/Button - CSS" />
```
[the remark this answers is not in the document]
k80bowman: Yes, it should. I didn't do it here because I thought I might want to do it separately, but I guess it probably doesn't matter much.

andresriveratoro (commented):
This looks great overall!

The following component stories are broken:
NumberInput - React
RadioButtons - React
TextInput - React

It looks like Divider, Sidenav and Tag subtitles should be updated to include CSS like we're doing for other CSS components. ProgressButton, the Icon components and OMBInfo are missing React in the subtitle as well.

k80bowman:
Thanks @andresriveratoro. I'll make sure to fix those component stories and update the subtitles.

humancompanion-usds:
This looks fine to me. Just a few questions:

* Buttons seems odd and I don't see Button pairs in there. It's under ProgressButton. Perhaps I'm just not caught up properly? I checked out this branch. The screenshot has ButtonPair but still a folder for Buttons. What is in that folder?
* I'm noticing a few variations in Storybook that are not in design.va.gov. For example, Accordion > Internationalization. When this is done we may need to comb thru quickly and make sure we're aligned.
* The icons folder is fine but why do we only have a few of our icons in there? I'm confused by this.

packages/storybook/stories/va-loading-indicator.stories.jsx
```diff
@@ -3,13 +3,14 @@ import React, { useState } from 'react';
...
 
 export default {
-  title: 'Components/va-loading-indicator',
+  title: `Components/${componentName}`,
   parameters: {
-    componentSubtitle: `Loading indicator web component`,
+    componentSubtitle: `va-loading-indicator web component`,
     docs: {
-      page: () => <StoryDocs data={loadingIndicatorDocs} />,
+      page: () => <StoryDocs componentName={componentName} data={loadingIndicatorDocs} />,
```
andresriveratoro, line 13: If we modify each story default export to include an `id` property, we won't have to update links on design.va.gov.

```jsx
export default {
  id: 'components/va-loading-indicator', // results in http://design.va.gov/storybook/?path=/docs/components-va-loading-indicator
  title: `Components/${componentName}`,
  parameters: {
    componentSubtitle: `va-loading-indicator web component`,
    docs: {
      page: () => <StoryDocs componentName={componentName} data={loadingIndicatorDocs} />,
...
```

packages/storybook/stories/va-table.stories.jsx
bkjohnson: If we leave the object properties as they are in `additional-docs.js`, do we need to do this?

bkjohnson (commented):
I'm wondering if we can get the same effect with fewer changes.

andresriveratoro:
We could update each story with an `id` and check item.id instead of item.name in `packages/storybook/.storybook/manager.js`, remove the `components-` prefix and apply the maturity scale icons that way without having to pass in `componentName` to StoryDocs.

bkjohnson (approved):

k80bowman:
I've been fooling with this one and unfortunately I've found that the `item.id` is not pulled from the [CSF](https://storybook.js.org/docs/react/api/csf) object in the story, it is constructed from the component name. So, for example, I set this as my CSF object for the accordion web component:
```
export default {
  id: 'components-va-accordion',
  title: 'Components/Accordion',
  subcomponents: componentStructure(accordionItem),
  parameters: {
    componentSubtitle: `va-accordion web component`,
    docs: {
      page: () => <StoryDocs data={accordionDocs} />,
    },
  },
};
```
The item object that I have access to in the manager file looks like this:
image

So, you see, `item.id` equals `components-accordion` and not `components-va-accordion` as I've set in the CSF object.

I'm a little frustrated, because what I would really like to have is one variable that contains the component name that gets used to look up the maturity scale information for both the sidebar indicator and the label on the storybook component page. The PR as it stands was my attempt at handling this by adding the componentName variable to the web components, but I agree it's clunky. I'm open to other suggestions, though.

bkjohnson:
In the future (once we no longer have React components) I'm imagining that we won't have the `additional-docs.js` file and we can indicate the maturity of our components using a comment like this in the source:

https://github.com/department-of-veterans-affairs/component-library/blob/87342464332c7ef70517f617197d6a891bf8804a/packages/web-components/src/components/va-button/va-button.tsx#L11-L14

I think that will simplify things and make documentation integration easier, so I'm not too worried about the interim. I'd just like to not spend lots of time dealing with React things and make progress on removing those components for good.

k80bowman:
@bkjohnson I was able to get the maturity scale tag and sidebar indicator to load from the web component docs as you suggested! 🎉 

I tried it out with button, but I'd appreciate any feedback you have before I update the rest of the web components and remove their data from the additional-docs file.

packages/storybook/stories/va-button.stories.jsx
```diff
@@ -4,9 +4,9 @@ import { getWebComponentDocs, propStructure, StoryDocs } from './wc-helpers';
 const buttonDocs = getWebComponentDocs('va-button');
 
 export default {
-  title: 'Components/va-button',
+  title: 'Components/Button',
   parameters: {
-    componentSubtitle: `Button web component`,
+    componentSubtitle: `va-button web component`,
```
bkjohnson, line 9: :+1:

packages/web-components/src/components/va-button/va-button.tsx
```diff
@@ -9,7 +9,10 @@ import {
 } from '@stencil/core';
 
 /**
+ * @componentName Button
  * @nativeHandler onClick
+ * @maturityCategory caution
+ * @maturityLevel candidate
```
bkjohnson, line 15: :+1:

packages/storybook/stories/wc-helpers.jsx
```diff
@@ -234,16 +250,16 @@ export function StoryDocs({ componentName, data, children }) {
...
+  const maturityLevel = maturityScale.level ?? componentDocs?.maturityLevel;
   // This feels a bit awkward, but I didn't want to use a magic number
   const _componentName = data?.tag?.slice('va-'.length);
   // Default the guidance values to be based on the web component's
   // tag name where possible
-  const {
-    guidanceHref = _componentName,
-    guidanceName = capitalize(_componentName)?.replaceAll('-', ' '),
-  } = componentDocs;
+  const guidanceHref = componentDocs?._componentName;
```
bkjohnson, line 261: I haven't followed all the commits so I don't know if it's related to the recent `@maturityCategory` changes or not, but I'm not sure if this will do what we think it will.

packages/storybook/stories/wc-helpers.jsx
```diff
@@ -129,6 +130,21 @@ export function Guidance({ href, name }) {
   );
 }
 
+const getDocsTagValue = (tagName, docsTags = []) => {
+  return docsTags
+    .filter(item => item.name === tagName)
+    .map(item => item.text)
+}
```
bkjohnson, line 137: Can we also make use of this function here:

https://github.com/department-of-veterans-affairs/component-library/blob/ccc6d2d88c567db08e6e17d512230dea8a76f7f4/packages/storybook/stories/wc-helpers.jsx#L196-L198

packages/storybook/stories/wc-helpers.jsx
```diff
@@ -234,16 +250,16 @@ export function StoryDocs({ componentName, data, children }) {
   const componentDocs = additionalDocs?.[component];
   const componentData = data ? { ...data, ...componentDocs } : null;
 
-  const maturityCategory = componentDocs?.maturityCategory;
-  const maturityLevel = componentDocs?.maturityLevel;
+  const maturityScale = getMaturityScale(data.docsTags);
+
+  const maturityCategory = maturityScale.category ?? componentDocs?.maturityCategory;
+  const maturityLevel = maturityScale.level ?? componentDocs?.maturityLevel;
```
bkjohnson, line 256: :+1:

packages/storybook/.storybook/manager.js
```diff
@@ -1,20 +1,40 @@
...
+  
+  // get the maturity category for the component
+  const maturityCategory = componentDocs[0]?.docsTags
+    .filter(item => item.name === 'maturityCategory')
+    .map(item => item.text)
+    .toString()
+    .toUpperCase();
+  
+  return category[maturityCategory];
+}
```
bkjohnson, line 21: Would it make sense to reuse helpers for this? There's that `getDocsTagValue` that you wrote as well as this:

https://github.com/department-of-veterans-affairs/component-library/blob/ccc6d2d88c567db08e6e17d512230dea8a76f7f4/packages/storybook/stories/wc-helpers.jsx#L16-L20

That function could probably be refactored to use the `componentName` like we have here.

bkjohnson (commented):
The button component looks good - I left some comments about helpers and some other changes that might be unrelated.

k80bowman:
Ok, this one is ready to be re-reviewed! Some things to note:
- I moved the `guidanceHref` for web components to the stencil files and was able to delete all references to the web components in `additional-docs.js`
- I checked all the links locally from design.va.gov to Storybook and everything works.
- This includes fixes to links from Storybook to design.va.gov. Some of these were broken due to new nesting in the components section of the site.

packages/storybook/stories/ProgressButton.stories.jsx
```diff
@@ -1,56 +0,0 @@
-import React, { Fragment } from 'react';
-import ProgressButton from '../../react-components/src/components/ProgressButton/ProgressButton';
-import { StoryDocs } from './wc-helpers';
-
-export default {
-  title: 'Components/Buttons/ProgressButton',
```
bkjohnson, line 6: Not that I'm complaining, but why are we deleting this story but leaving others as "Deprecated"?

packages/web-components/src/components/va-on-this-page/va-on-this-page.tsx
```diff
@@ -5,6 +5,12 @@ import { consoleDevError } from '../../utils/utils';
  * This component will render links based on the content around it. It scans the document for any `<h2>`
  * elements that contain an `id` inside of an `<article>` and will create a list of links to the headings.
  */
+
+/**
+ * @componentName On this page
```
bkjohnson, line 10: Do things still work if this block is moved to the existing doc block above?

packages/storybook/stories/wc-helpers.jsx
```diff
@@ -232,18 +246,18 @@ function CanvasLink() {
...
   // This feels a bit awkward, but I didn't want to use a magic number
   const _componentName = data?.tag?.slice('va-'.length);
   // Default the guidance values to be based on the web component's
   // tag name where possible
-  const {
-    guidanceHref = _componentName,
-    guidanceName = capitalize(_componentName)?.replaceAll('-', ' '),
-  } = componentDocs;
+  const guidanceHref = getDocsTagValue('guidanceHref', data?.docsTags)[0] ?? (componentDocs?.guidanceHref ?? _componentName);
+  const guidanceName = componentDocs?.guidanceName ?? capitalize(_componentName)?.replaceAll('-', ' ') ?? componentName;
```
bkjohnson, line 260: Is the last `componentName` nullish coalescing necessary? I'm reading this as: If there isn't a `guidanceName` in the component's `additional-docs` object, then try to capitalize the web component's tag name, and if that isn't available then use the `componentName` prop. But they don't look like they would be part of a URL:

ripgrep output for "componentName" on storybook story files. We see `componentName` props like "Modal - React" and "Icon - user"

packages/web-components/src/components/va-on-this-page/va-on-this-page.tsx
k80bowman: I kind of like the idea of keeping it separate, just for clarity. Do we gain anything by combining the blocks?

packages/storybook/stories/ProgressButton.stories.jsx
k80bowman: I decided to go ahead and delete it because we really don't want people using it and I don't feel like we need to maintain the documentation for it the way we would with a React component.

packages/storybook/stories/wc-helpers.jsx
k80bowman: That's needed for the React components.

packages/storybook/stories/ProgressButton.stories.jsx
bkjohnson: It is a [React component](https://github.com/department-of-veterans-affairs/component-library/blob/9b58a8c0216c9215dd8b5b31fb3d90dea9f29227/packages/react-components/src/components/ProgressButton/ProgressButton.jsx) though.

Report of `ProgressButton` component usage in vets-website. It is used 12 times total - 10 times in platform-ish code, once in the "covid-vaccine-expansion" application, and once in the "Coronavirus Research - Volunteer" application.

packages/web-components/src/components/va-on-this-page/va-on-this-page.tsx
bkjohnson: There's no real gain, it's just a subjective thing. I'm more used to any JSdoc or TSdoc tags being in the same doc block as the string description, like this:

https://github.com/department-of-veterans-affairs/component-library/blob/9b58a8c0216c9215dd8b5b31fb3d90dea9f29227/packages/web-components/src/components/va-file-input/va-file-input.tsx#L98-L102

packages/storybook/stories/wc-helpers.jsx
bkjohnson: I did some debugging and found that for a React component, `guidanceName` is defined, but `guidanceHref` isn't (one uses the undefined `_componentName` which may be an accident).

Screenshot of the browser's debugger showing an undefined `guidanceHref` and a defined `guidanceName`. `componentDocs` is missing both properties.

Since the href is undefined, this means we aren't seeing the `<Guidance>` component:

https://github.com/department-of-veterans-affairs/component-library/blob/8a145b1b5b66ff3009442555aab15c7a82eabb8a/packages/storybook/stories/wc-helpers.jsx#L267

https://github.com/department-of-veterans-affairs/component-library/blob/8a145b1b5b66ff3009442555aab15c7a82eabb8a/packages/storybook/stories/wc-helpers.jsx#L118-L119

If not seeing that is intentional, then I don't think we need the nullish coalescing.

But even if `guidanceHref` is defined so that it uses `componentName` instead of `_componentName`, like this:

```js
  const guidanceHref = getDocsTagValue('guidanceHref', data?.docsTags)[0] ?? (componentDocs?.guidanceHref ?? componentName);
```

Since the `guidanceHref` and `guidanceName` are defined, we see the `Guidance` component and it's using a broken URL:

Screenshot of the Guidance component appearing and using a broken link based on the "Expanding group" `componentName`

Screenshot of a "Forbidden" error when clicking on the link provided in the `Guidance` component

bkjohnson (changes requested):
If we're fine with the output as is then I don't think we need the nullish coalescing. If we do want it, then I think we need to make some changes related to the `componentName` value.

packages/storybook/stories/wc-helpers.jsx
k80bowman: I'm not sure I follow what you are doing there, but I went ahead and removed that last bit from the guidanceName and instead specified names for the two React components that needed it in the additional-docs file.

bkjohnson (approved):

PR #483, opened 2022-07-12
va-button uses aria-label only when a value is given

andresriveratoro:
## Chromatic

https://941-aria-label--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/941

## Testing done
E2E

## Acceptance criteria
- [x] va-button only uses aria-label when a value is given

## Definition of done


packages/web-components/src/components/va-button/va-button.tsx
```diff
@@ -126,7 +126,7 @@ export class VaButton {
       <Host>
         <button
           aria-disabled={ariaDisabled}
-          aria-label={label || buttonText}
+          aria-label={label}
```
bkjohnson, line 129: Can we add a test to show that the `label` prop results in an `aria-label` on the button?

bkjohnson (approved):
One small comment, looks good though.

bkjohnson (approved):

PR #489, opened 2022-07-15
Add aria-invalid to most form components

bkjohnson:
## Chromatic

https://aria-invalid--60f9b557105290003b387cd5.chromatic.com

## Description
Part of:

- https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/844
- https://github.com/department-of-veterans-affairs/va.gov-team/issues/39727

This adds the `aria-invalid` attribute to the `<input>`/`<select>` part of components as well as to the `radiogroup` role for radio buttons. This attribute defaults to false and is set to true when there is an error message. The `<va-checkbox-group>` component doesn't have this update because it looked like the `group` role didn't match the [associated roles of the aria attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-invalid#associated_roles).

The form components which haven't received `aria-invalid` are:

- `va-checkbox-group`
- `va-radio-option`

## Testing done

- Storybook :eyes: 
- E2E tests

### va-date

[aria-invalid appearing on year field](https://user-images.githubusercontent.com/2008881/179854632-860c8327-6e09-413e-b04a-7d57d2864332.webm)

### va-memorable-date

[aria-invalid appears on each text-input component when the the memorable date component is blurred. If all fields are missing then they all receive aria-invalid. If only the year is missing then the year will be marked as invalid while month and day are left alone.](https://user-images.githubusercontent.com/2008881/179855401-2bf7cd3c-4c1a-43bb-9ea9-659f4b2ae889.webm)

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-date/va-date.tsx
+++ b/packages/web-components/src/components/va-date/va-date.tsx
@@ -4,6 +4,7 @@ import {
   EventEmitter,
   Host,
   Prop,
+  State,
   h,
   Element,
 } from '@stencil/core';
@@ -11,8 +12,8 @@ import {
 import {
   months,
   days,
-  isFullDate,
-  isMonthYearDate,
+  // isFullDate,
+  // isMonthYearDate,
   checkLeapYear,
   maxMonths,
   maxYear,
@@ -83,6 +84,10 @@ export class VaDate {
   })
   dateBlur: EventEmitter;
 
+  @State() invalidDay: boolean = false;
+  @State() invalidMonth: boolean = false;
+  @State() invalidYear: boolean = false;
+
   /**
    * Set the value prop as an ISO-8601 date using provided arguments.
    * Strips trailing hyphens and sets date to be null if the
@@ -113,23 +118,36 @@ export class VaDate {
     const daysForSelectedMonth = month > 0 ? days[month] : [];
     const leapYear = checkLeapYear(year);
 
-    const expectedDateFormat = this.monthYearOnly
-      ? isMonthYearDate
-      : isFullDate;
+    this.invalidYear = (
+      !year ||
+      year < minYear ||
+      year > maxYear
+    );
 
-    // Check validity of date if invalid provide message and error state styling
+    this.invalidMonth = (
+      !month ||
+      month < minMonths ||
+      month > maxMonths
+    );
+
+    this.invalidDay = (
+      !this.monthYearOnly && (
+        !day ||
+        day < minMonths ||
+        day > daysForSelectedMonth.length)
+    );
+
+    if (!leapYear && month === 2 && day > 28) {
+      this.invalidYear = true;
+      this.invalidMonth = true;
+      this.invalidDay = true;
+    }
+
     if (
       !this.error &&
-      (year < minYear ||
-        year > maxYear ||
-        month < minMonths ||
-        month > maxMonths ||
-        (!this.monthYearOnly &&
-          (day < minMonths || day > daysForSelectedMonth.length || !day)) ||
-        !month ||
-        !year ||
-        (!leapYear && month === 2 && day > 28) ||
-        (this.required && !expectedDateFormat(this.value)))
+      (this.invalidYear ||
+        this.invalidMonth ||
+        this.invalidDay)
     ) {
       this.error = 'Please enter a valid date';
     } else if (this.error !== 'Please enter a valid date') {
@@ -226,6 +244,7 @@ export class VaDate {
               // Value must be a string
               value={month?.toString()}
               onVaSelect={handleDateChange}
+              invalid={this.invalidMonth}
               class="select-month"
               aria-label="Please enter two digits for the month"
             >
@@ -244,6 +263,7 @@ export class VaDate {
                 // Value must be a string
                 value={daysForSelectedMonth.length < day ? '' : day?.toString()}
                 onVaSelect={handleDateChange}
+                invalid={this.invalidDay}
                 class="select-day"
                 aria-label="Please enter two digits for the day"
               >
@@ -263,6 +283,7 @@ export class VaDate {
               // Value must be a string
               // Checking is NaN if so provide empty string
               value={year ? year.toString() : ''}
+              invalid={this.invalidYear}
               onInput={handleDateChange}
               onKeyDown={handleDateKey}
               class="input-year"
--- a/packages/web-components/src/components/va-date/va-date.tsx
+++ b/packages/web-components/src/components/va-date/va-date.tsx
@@ -12,8 +12,6 @@ import {
 import {
   months,
   days,
-  // isFullDate,
-  // isMonthYearDate,
   checkLeapYear,
   maxMonths,
   maxYear,
--- a/packages/web-components/src/components/va-date/va-date.tsx
+++ b/packages/web-components/src/components/va-date/va-date.tsx
@@ -118,29 +118,31 @@ export class VaDate {
     const daysForSelectedMonth = month > 0 ? days[month] : [];
     const leapYear = checkLeapYear(year);
 
-    this.invalidYear = (
-      !year ||
-      year < minYear ||
-      year > maxYear
-    );
+    if (this.required) {
+      this.invalidYear = (
+        !year ||
+        year < minYear ||
+        year > maxYear
+      );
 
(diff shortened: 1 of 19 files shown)


packages/web-components/src/components/va-memorable-date/va-memorable-date.tsx
```diff
@@ -94,19 +99,36 @@ export class VaMemorableDate {
...
+
+    this.invalidDay =  (
+      !day ||
       dayNum < minMonths ||
-      dayNum > daysForSelectedMonth.length ||
-      day === '' ||
-      month === '' ||
-      year === '' ||
-      (!leapYear && monthNum === 2 && dayNum > 28) ||
-      (this.required && !isFullDate(this.value))
```
bkjohnson, line 108: I'm not quite sure why we were checking `this.required && !isFullDate(this.value)` in the first place since missing `day`, `month`, _or_` year` would result in the same thing happening regardless of whether or not it was required. While reviewing, try to think if not checking `required` would cause any regressions.

packages/web-components/src/components/va-date/va-date.tsx
```diff
@@ -113,23 +118,36 @@ export class VaDate {
...
-      (year < minYear ||
-        year > maxYear ||
-        month < minMonths ||
-        month > maxMonths ||
-        (!this.monthYearOnly &&
-          (day < minMonths || day > daysForSelectedMonth.length || !day)) ||
-        !month ||
-        !year ||
-        (!leapYear && month === 2 && day > 28) ||
-        (this.required && !expectedDateFormat(this.value)))
```
bkjohnson, line 132: Same story here.

packages/web-components/src/components/va-radio/va-radio.tsx
```diff
@@ -172,7 +172,7 @@ export class VaRadio {
   render() {
     const { label, required, error } = this;
     return (
-      <Host role="radiogroup">
+      <Host role="radiogroup" aria-invalid={error ? 'true' : 'false'}>
```
bkjohnson, line 175: In addition to [`radiogroup` being an associated role for `aria-invalid`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-invalid#associated_roles), it seems like [this pattern has been seen elsewhere](https://dsva.slack.com/archives/C023ZKDJJK1/p1657907719138989?thread_ts=1657892156.507549&cid=C023ZKDJJK1).

packages/web-components/src/components/va-memorable-date/va-memorable-date.tsx
```diff
@@ -76,8 +77,12 @@ export class VaMemorableDate {
   })
   dateBlur: EventEmitter;
 
+  @State() invalidDay: boolean = false;
+  @State() invalidMonth: boolean = false;
+  @State() invalidYear: boolean = false;
```
bkjohnson, line 82: We may eventually want these to be props so that custom validation can take advantage of the `aria-invalid` attribute.

packages/web-components/src/components/va-memorable-date/va-memorable-date.tsx
```diff
@@ -76,8 +77,12 @@ export class VaMemorableDate {
...
+
   private handleDateBlur = (event: FocusEvent) => {
-    const [year, month, day] = (this.value || '').split('-').map(val => val);
+    const [year, month, day] = (this.value || '').split('-');
```
bkjohnson, line 85: I removed the `.map()` because it wasn't doing anything.

bkjohnson (commented):
I've got a question about one of the removed pieces of validation.

packages/web-components/src/components/va-date/va-date.tsx
```diff
@@ -4,15 +4,16 @@ import {
...
   Element,
 } from '@stencil/core';
 
 import {
   months,
   days,
-  isFullDate,
-  isMonthYearDate,
+  // isFullDate,
+  // isMonthYearDate,
```
k80bowman, line 16: Did you mean to leave these commented out?

packages/web-components/src/components/va-memorable-date/va-memorable-date.tsx
k80bowman: Why is this something we would do eventually and not now?

packages/web-components/src/components/va-memorable-date/va-memorable-date.tsx
k80bowman: Are you saying that if a memorable date component was not required and someone left it blank, they would get the same error messages?

packages/web-components/src/components/va-date/va-date.tsx
bkjohnson: For now, yes. I left a comment wondering about why they were used in the first place considering the rest of the boolean expression, so I'm not totally sure that removing them doesn't break things. If we decide that they don't need to be there then I'll actually remove the functions.

packages/web-components/src/components/va-memorable-date/va-memorable-date.tsx
k80bowman: I was curious, so I added a couple of tests around the required prop just to see what happens.

On va-date, I found that even when a date is not required, we show an error message if someone clicks into the component and then tabs out without entering anything. This test fails.
```
it('does not display an error message when required is false and input is left blank', async () => {
      await page.setContent(
        '<va-date name="test" />',
      );
      const date = await page.find('va-date');
      const handleYear = await page.$('pierce/[name="testYear"]');
      await handleYear.click({ clickCount: 1 });
      // Trigger Blur
      await handleYear.press('Tab');
  
      await page.waitForChanges();
      expect(date.getAttribute('error')).toEqual('');
    })
```

I also added this test to the memorable date component and it failed, meaning we're not doing any checking when a date is required on this component, only when it is incorrect.
```
  it('displays an error message when required is true and input is left blank', async() => {
    await page.setContent(
      '<va-memorable-date name="test" required="true" />',
    );
    const elementYear = await page.find('va-memorable-date >>> .input-year');
    const handleYear = await page.$('pierce/[name="testYear"]');
    await handleYear.click({ clickCount: 1 });
    // Trigger Blur
    await handleYear.press('Tab');

    await page.waitForChanges();
    expect(elementYear.getAttribute('error')).toEqual('Please enter a valid date');
  });
```

It would be a good idea to add a couple of tests around the required prop for both of these components, one testing what happens when required is false and the input is left blank and one testing what happens when required as true and the input is left blank.

packages/web-components/src/components/va-memorable-date/va-memorable-date.tsx
bkjohnson: I was just trying to limit the scope creep.

packages/web-components/src/components/va-memorable-date/va-memorable-date.tsx
bkjohnson: This commit changes the behavior a bit and adds some tests: https://github.com/department-of-veterans-affairs/component-library/pull/489/commits/6718483cf48118b6cb1a6ece5bd98746e8342c66

packages/web-components/src/components/va-memorable-date/test/va-memorable-date.e2e.ts
```diff
@@ -184,6 +202,135 @@ describe('va-memorable-date', () => {
...
+      invalidYear = await handleYear.evaluate(getAriaInvalid);
+      invalidMonth = await handleMonth.evaluate(getAriaInvalid);
+      invalidDay = await handleDay.evaluate(getAriaInvalid);
+
+      expect(invalidYear).toEqual('false');
+      expect(invalidMonth).toEqual('false');
+      expect(invalidDay).toEqual('false');
+    });
+
+    it.only('correctly indicates an invalid day', async () => {
```
k80bowman, line 291: Probably should remove the `only` here

bkjohnson: Good catch :sweat_smile:

k80bowman (approved):
This looks great Brooks. Thank you for doing all of this work!

PR #492, opened 2022-07-21
Add tty prop to telephone component

bkjohnson:
## Chromatic

https://telephone-tty--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/775

This allows consumers of the telephone component to indicate that it is for a TTY number. This will add some `TTY` text to the `aria-label` and the link text, but not to the `href`.

## Testing done

- Unit tests
- E2E test
- Storybook :eyes: 

## Screenshots

A screenshot of storybook where a telephone component with the tty prop has TTY in its aria-label and in the actual link text

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-telephone/va-telephone.tsx
+++ b/packages/web-components/src/components/va-telephone/va-telephone.tsx
@@ -57,6 +57,15 @@ export class VaTelephone {
   })
   componentLibraryAnalytics: EventEmitter;
 
+  static splitContact(contact: string) :string[] {
+    if (contact.length === 10) {
+      const regex = /(?<area>\d{3})(?<local>\d{3})(?<last4>\d{4})/g;
+      const { area, local, last4 } = regex.exec(contact).groups;
+      return [area, local, last4];
+    }
+    return [contact];
+  }
+
   /**
    * Format telephone number for display.
    * `international` and `extension` args only work on 10 digit contacts
@@ -70,8 +79,7 @@ export class VaTelephone {
   ): string {
     let formattedNum = num;
     if (num.length === 10) {
-      const regex = /(?<area>\d{3})(?<local>\d{3})(?<last4>\d{4})/g;
-      const { area, local, last4 } = regex.exec(num).groups;
+      const [ area, local, last4 ] = VaTelephone.splitContact(num);
       formattedNum = `${area}-${local}-${last4}`;
       if (international) formattedNum = `+1-${formattedNum}`;
       if (extension) formattedNum = `${formattedNum}, ext. ${extension}`;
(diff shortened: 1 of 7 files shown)


packages/web-components/src/components/va-telephone/va-telephone.tsx
```diff
@@ -72,12 +79,15 @@ export class VaTelephone {
         formattedNum = `${area}-${local}-${vanity} (${last4})`;
       }
     }
+    if (tty) {
+      formattedNum = `TTY: ${formattedNum}`;
```
bkjohnson, line 83: I didn't see any guidance that a TTY number _couldn't_ be a 10 digit number, so that's why this isn't inside an else block.

packages/web-components/src/components/va-telephone/va-telephone.tsx
```diff
@@ -87,9 +97,14 @@ export class VaTelephone {
...
       ? number.split(/[^\d]+/)
       : number.split(/[^\d\w]+/);
+
     return numberType
       .filter(n => n)
-      .map(chunk => (chunk === 'ext' ? 'extension' : chunk.split('').join(' ')))
+      .map(chunk => {
+        if (chunk === 'ext')
+          return 'extension';
+        return chunk === 'TTY' ? 'TTY' : chunk.split('').join(' ');
```
bkjohnson, line 106: If we keep adding more variants we may want to refactor all of these formatting functions because they're kind of getting unwieldy.

k80bowman: Is it worth doing that refactor now?

packages/web-components/src/components/va-telephone/va-telephone.tsx
k80bowman: Would you mind checking with Sarah to verify this?

bkjohnson: @SarahKay8 do you know if TTY numbers are always 3 digits or if it's possible for them to be 10 digits?

packages/web-components/src/components/va-telephone/va-telephone.tsx
bkjohnson: I mean, we already have it inside a named function:

https://github.com/department-of-veterans-affairs/component-library/blob/a17e132af6ba21a1727e50359d881739930d106a/packages/web-components/src/components/va-telephone/va-telephone.tsx#L88-L109

I could give that map function a name but I don't know if it would really do much. As we add more features, things like this get more unwieldy as well:

https://github.com/department-of-veterans-affairs/component-library/blob/a17e132af6ba21a1727e50359d881739930d106a/packages/web-components/src/components/va-telephone/va-telephone.tsx#L73-L80

packages/web-components/src/components/va-telephone/va-telephone.tsx
bkjohnson: It looks like 10 digit TTY numbers [are possible](https://github.com/department-of-veterans-affairs/vets-website/blob/5a07edbdeb4987b90716bce44d0b4256ee4b4c9b/src/applications/static-pages/health-care-manage-benefits/secure-messaging-page/components/UnauthContent/index.js#L163-L165).

k80bowman (approved):
This looks great! Once it's merged, please add the TTY example to design.va.gov.

PR #493, opened 2022-07-22
Uswds textarea discovery

bkjohnson:
## Chromatic

https://uswds-textarea-discovery--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1036

## Testing done

Storybook :eyes: 

## Acceptance criteria
- [ ]

## Definition of done


packages/web-components/stencil.config.ts
```diff
@@ -1,14 +1,21 @@
 import { Config } from '@stencil/core';
 import { postcss } from '@stencil/postcss';
+import { sass } from '@stencil/sass';
 import url from 'postcss-url';
+import postcssSass from '@csstools/postcss-sass';
```
bkjohnson, line 5: This plugin [claims that its options are forwarded to the `sass` dependency](https://github.com/csstools/postcss-sass#options).

packages/web-components/stencil.config.ts
```diff
@@ -1,14 +1,21 @@
...
   globalStyle: 'src/global/main.css',
   plugins: [
+    sass({ includePaths: ["../../node_modules/@uswds/uswds/packages/"],
+      injectGlobalPaths: [
+        'src/global/settings.scss',
+      ]
+    }),
     postcss({
-      plugins: [url({ url: 'inline' })],
+      plugins: [postcssSass({ verbose: true, loadPaths: ["../../node_modules/@uswds/uswds/foo"] }), url({ url: 'inline' })],
```
bkjohnson, line 18: The [documentation for the JS api](https://sass-lang.com/documentation/js-api/interfaces/Options#loadPaths) shows the `loadPaths` option that is an array of strings, but I couldn't get this to work. I don't know where the bug is, but without this working I was getting errors like this:

Build error on a css file with the message "File to import not found: `@import 'uswds-helpers/src/styles/_usa-sr-only`

Within the CSS that we control I could change the import locations to be something like `../../node_modules/@uswds/uswds/packages/...`, but within USWDS sass files there are package imports that the compiler wasn't resolving. With the `@stencil/sass` plugin we can set `includePaths` and don't have to deal with these issues.

bkjohnson (commented):
I first tried a postcss plugin for handling SASS files, but this didn't work. That's why I added the regular sass plugin.

bkjohnson (commented):
I think we can get away with using tokens to adjust the default font size, but for focus styles we're going to have to use our own CSS.

packages/web-components/src/components/va-textarea/va-textarea.tsx
```diff
@@ -126,19 +126,22 @@ export class VaTextarea {
...
         </label>
         {error && (
-          <span id="error-message" role="alert">
-            <span class="sr-only">{i18next.t('error')}</span> {error}
+          <span id="error-message" class="usa-error-message" role="alert">
+            <span class="usa-sr-only">{i18next.t('error')}</span> {error}
           </span>
         )}
         <textarea
+          class="usa-textarea"
```
bkjohnson, line 144: The `usa-textarea` package [doesn't have error state styling](https://github.com/uswds/uswds/blob/6dfc4689fda25ebeb7a6c6dc1f68a50c810c0d29/packages/usa-textarea/src/styles/_usa-textarea.scss#L3-L10). The [`usa-input` package does](https://github.com/uswds/uswds/blob/6dfc4689fda25ebeb7a6c6dc1f68a50c810c0d29/packages/usa-input/src/styles/_usa-input.scss#L8-L12), but the height of that is off. I feel like it would be strange to mix classes like `usa-textarea usa-input--error`, and in order to do that we would have to import CSS that will go unused.

bkjohnson (commented):
We may want to open an issue and PR with USWDS to allow `usa-textarea` to have error styling. Right now it doesn't so we would have to use our own CSS or use existing USWDS styles awkwardly.

packages/web-components/src/components/va-textarea/va-textarea.tsx
k80bowman: This is an interesting finding, I'm glad you made note of it. This could definitely affect our adoption of USWDS CSS for our components, especially if we need to go back and forth with them on requested changes.

bkjohnson: As a worst-case scenario, we could use what we can from USWDS and add whatever CSS we needed to our `va-textarea.scss` file.

packages/web-components/src/global/settings.scss
```diff
@@ -0,0 +1,9 @@
+@use "sass:map";
+@use "uswds-core" with ( 
+    $theme-body-font-size: 10
+);
+
+// Sass maps are immutable, so we need to re-assign the original variable
+// after setting it
+uswds-core.$all-color-shortcodes:  map.set(uswds-core.$all-color-shortcodes, "error-dark", #cd2026);
```
bkjohnson, line 8: After our conversation in standup I went back to do some more discovery and found that this pattern allows us to set the tokens to be whatever we want and avoids the USWDS validation of them. However, I don't recommend this for regular use since it skips some intermediary steps (the `$tokens-color-state` map isn't updated, for example). But this could be used as a stepping stone to get us to USWDS3 while avoiding dealing with color changes, and then once we're ready we could remove it.

cc @k80bowman @humancompanion-usds

PR #495, opened 2022-07-27
fix: text does not overlap icon in alerts

andresriveratoro:
## Chromatic

https://1040-va-alert-icons--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1040

## Testing done
Storybook

## Screenshots
Screen Shot 2022-07-27 at 09 49 20

## Acceptance criteria
- [x] Text in alerts does not overlap the close icon

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-alert/va-alert.css
+++ b/packages/web-components/src/components/va-alert/va-alert.css
@@ -166,3 +166,7 @@ i {
 i.fa-times-circle::before {
   content: '\F057'; /* fa-times-circle*/
 }
+
+:host([closeable]:not([closeable='false'])) .alert {
+  padding-right: 5.7rem;
+}
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@department-of-veterans-affairs/component-library",
   "description": "VA.gov component library. Includes React and web components.",
-  "version": "11.2.0",
+  "version": "11.2.1",
   "license": "MIT",
   "scripts": {
     "build": "webpack"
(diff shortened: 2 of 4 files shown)


packages/web-components/src/components/va-alert/va-alert.css
```diff
@@ -166,3 +166,7 @@ i {
 i.fa-times-circle::before {
   content: '\F057'; /* fa-times-circle*/
 }
+
+:host([closeable]:not([closeable='false'])) .alert {
+  padding-right: 5.7rem;
```
andresriveratoro, line 171: This ends up being 16px away from the close icon.

bkjohnson: @humancompanion-usds seemed [interested in having the padding be equal on both sides](https://github.com/department-of-veterans-affairs/component-library/pull/429#discussion_r875850031). Is that still something that we want?

k80bowman: We can try it, but I worry that we're going to end up adding back the [problem we were trying to solve](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/658) with that earlier PR. If we add too much padding, alerts with a lot of text (which maybe shouldn't be a thing, but it is) get really long on smaller screens.

andresriveratoro: This padding only takes effect if there's a close icon and it maintains the same padding on both sides relative to the icons. The `Dismissable Background Only Icon` left icon is 20px from the edge. I could update that icon to be 16px from the edge if we want.

bkjohnson (approved):

andresriveratoro:
Update Sketch to use 16px margin between the close icon and the body container for designs that include the close icon.

k80bowman (approved):

PR #503, opened 2022-08-02
chore: add v3 border, display and font utilities

andresriveratoro:
## Chromatic

https://1027-bdf-utilities--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1027

## Acceptance criteria
- [x] `css-library` has v3 utilities for border, display, font-family, font-size, font-style, font-weight properties

## Definition of done

### DIFF
--- a/packages/css-library/src/tokens/border.scss
+++ b/packages/css-library/src/tokens/border.scss
@@ -1,0 +1,18 @@
+@use './_variables' as *;
+
+$tokens-border: (
+  '0': '0',
+  '1px': '1px solid',
+  '2px': '2px solid',
+  '3px': '3px solid',
+  '4px': '4px solid',
+  '5px': '5px solid',
+  '7px': '7px solid',
+  '10px': '10px solid',
+);
+
+$tokens-border-style: (
+  'solid': 'solid',
+  'dashed': 'dashed',
+  'dotted': 'dotted',
+);
--- a/packages/css-library/src/utilities/border.scss
+++ b/packages/css-library/src/utilities/border.scss
@@ -6,54 +6,16 @@
 $u-border-base: (
   border: (
     base: 'vads-border',
-    modifiers: null,
+    modifiers: (
+      noModifier: '',
+      'top': '-top',
+      'right': '-right',
+      'bottom': '-bottom',
+      'left': '-left',
+    ),
     values: map-collect($tokens-border),
     settings: $responsive,
     property: 'border',
-    type: 'utility',
-  ),
-);
-
-$u-border-top: (
-  border-top: (
-    base: 'vads-border-top',
-    modifiers: null,
-    values: map-collect($tokens-border),
-    settings: $color-settings,
-    property: 'border-top',
-    type: 'utility',
-  ),
-);
-
-$u-border-right: (
-  border-right: (
-    base: 'vads-border-right',
-    modifiers: null,
-    values: map-collect($tokens-border),
-    settings: $color-settings,
-    property: 'border-right',
-    type: 'utility',
-  ),
-);
-
-$u-border-bottom: (
-  border-bottom: (
-    base: 'vads-border-bottom',
-    modifiers: null,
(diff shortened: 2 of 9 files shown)


packages/css-library/src/tokens/border.scss
```diff
@@ -0,0 +1,18 @@
...
+  '2px': '2px solid',
+  '3px': '3px solid',
+  '4px': '4px solid',
+  '5px': '5px solid',
+  '7px': '7px solid',
+  '10px': '10px solid',
+);
+
+$tokens-border-style: (
+  'solid': 'solid',
```
bkjohnson, line 15: Would it make sense to use Style Dictionary for things like this instead?

bkjohnson: We talked about this and decided that SD tokens aren't that helpful here at the moment.

packages/css-library/src/utilities/border.scss
```diff
@@ -0,0 +1,51 @@
...
+    values: map-collect($tokens-border),
+    settings: $responsive,
+    property: 'border',
+    type: 'utility',
+  ),
+);
+
+$u-border-style: (
+  border-style: (
+    base: 'vads-border-style',
```
bkjohnson, line 25: One thing that we haven't explicitly talked about is naming patterns for this new library. With the color utilities I dropped the `color` from the name since a utility class like `vads-gray` is unambiguously for a color, and [shortened `background` to `bg`](https://github.com/department-of-veterans-affairs/component-library/blob/6bcfbcb543d06b3183c72365b72c79de9f642635/packages/css-library/src/utilities/background-color.scss#L8).

This will gives us `vads-border-style-dashed`, when `vads-border-dashed` can communicate the same thing with fewer characters, and thus a smaller stylesheet.

This can be applied to other classes - `vads-font-md` and `vads-font-2xl` seems like they can be interpreted as changing the `font-size` property even without explicitly mentioning size in the name. However, there are cases where removing that extra word makes the classes become ambiguous. Font style and font weight are examples of this.

With the current PR we would get `vads-font-style-normal` and `vads-font-weight-normal`. The font style would have additional `-italic` and `-bold` utilities, and the font weight style would have `bold` as well.

My question for us and @k80bowman is, do we want to document how we choose the names for these new utilities? Should we go for performance & usability when we can, or should we be as specific as possible no matter what?

bkjohnson (commented):
I've got a question about utility naming in general.

packages/css-library/src/utilities/border.scss
k80bowman: I would agree that performance and usability are our primary concerns with naming these utilities. I would also say yes, we should absolutely document how we choose the names, if for no other reason than we can point to it when people ask. I could see this actually being part of our [engineering best practices doc](https://vfs.atlassian.net/wiki/spaces/DST/pages/2176516116/Design+System+Engineering+Best+Practices).

andresriveratoro: @bkjohnson That makes sense. I'll update utilities to use fewer characters without being ambiguous.

k80bowman (approved):

packages/css-library/src/utilities/display.scss
```diff
@@ -0,0 +1,14 @@
+@use 'uswds-core/src/styles/functions' as *;
+@use '../tokens/display' as *;
+@use './settings' as *;
+
+$u-display: (
+  display: (
+    base: 'vads',
```
bkjohnson, line 7: Just having the `vads` base with tokens like `none` and `flex` means we will have utility names like this for setting `display`:

- `vads-none`
- `vads-flex`

We'll also have utilities like `vads-flex-1` for setting the `flex` property, but it seems fine to have the similar names since they're related, and I don't think things will be overridden which is good.

bkjohnson (approved):
Looks good, just left a comment for future awareness.

PR #504, opened 2022-08-03
chore: add flex, position and visibility v3 utilities

andresriveratoro:
## Chromatic

https://1028-utilities--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1028

## Acceptance criteria
- [x] `css-library` has v3 utilities for flex, position and visibility properties

## Definition of done


packages/css-library/src/tokens/visibility.scss
```diff
@@ -0,0 +1,39 @@
...
+      height: 1px;
+      margin: -1px;
+      overflow: hidden;
+      padding: 0;
+      position: absolute;
+      width: 1px;
+      word-wrap: normal;
+    }
+  }
+}
```
bkjohnson, line 39: There's a lot of [extra things in here coming from some blog post](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/a20b0958ebf8ac9bb6bd1630638678297eb83cfe/packages/formation/sass/utilities/_visibility.scss#L9-L23). Do we need all of them, or could we [`@extend usa-sr-only`](https://github.com/uswds/uswds/blob/846e45b50662851a6d37f20726d9d20dd415a3fa/packages/uswds-helpers/src/styles/_usa-sr-only.scss#L3-L5). The internal mixin looks [much simpler](https://github.com/uswds/uswds/blob/846e45b50662851a6d37f20726d9d20dd415a3fa/packages/uswds-core/src/styles/mixins/general/screen-reader.scss#L2-L6).

Also, we might not need this to be responsive. If something should only be visible to screen readers in a desktop view, I'm guessing the same would be true for tablet, mobile, etc.

andresriveratoro: Good call on this not needing to be responsive for the screen reader utility, and thank you for those helpful links.

packages/css-library/src/tokens/flexbox.scss
```diff
@@ -0,0 +1,95 @@
+@use './_variables' as *;
+
+$tokens-flex: (
+  1: (
+    slug: '1',
+    content: 1 0 0%,
+    extend: (
+      'flex-grow': 1,
+    ),
+  ),
```
bkjohnson, line 10: Looks like we might not need to specify flex-grow [according to the docs](https://developer.mozilla.org/en-US/docs/Web/CSS/flex).

> ```css
>    /* Three values: flex-grow | flex-shrink | flex-basis */
>    flex: 2 2 10%;
>  ```
Our [current approach seems redundant](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/a831cc59059d2ef01c161f03e4fdb3caa5c5a8d4/packages/formation/sass/utilities/_flexbox.scss#L38) since we specify `1` as the first part of the base `flex` property.

packages/css-library/src/utilities/position.scss
```diff
@@ -0,0 +1,20 @@
...
+    base: 'vads-position',
+    modifiers: null,
+    values:
+      map-collect(
+        (
+          absolute: absolute,
+          fixed: fixed,
+          relative: relative,
+        )
+      ),
```
bkjohnson, line 15: May want these to be defined as a SASS variable to be consistent with others.

bkjohnson (commented):
Some minor changes to be made, but things look good.

packages/css-library/src/tokens/flexbox.scss
andresriveratoro: The flex-grow property has been removed from the flex utility classes.

packages/css-library/src/utilities/position.scss
andresriveratoro: These now exist in `src/tokens/position.scss`.

bkjohnson (approved):

PR #507, opened 2022-08-04
Create a File Input web component

jamigibbs:
## Chromatic

https://966-file-input-web-component--60f9b557105290003b387cd5.chromatic.com

## Description
Closes
https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/966
https://github.com/department-of-veterans-affairs/va.gov-team/issues/27411
https://github.com/department-of-veterans-affairs/va.gov-team/issues/27414#issuecomment-949698486

## Testing done
- Storybook
- Chrome, Firefox, Safari, Edge (latest versions)
- Browserstack (iOS, Android)
- VoiceOver

## Screenshots

Screen Shot 2022-08-12 at 12 08 24 PM
Screen Shot 2022-08-12 at 12 08 29 PM
Screen Shot 2022-08-12 at 12 08 34 PM

## Acceptance criteria
- [x] Review DST backlog for outstanding issues with this component, if necessary
- [x] Create web component and add to Storybook
- [x] Write any necessary tests
- [x] Add Chromatic link to [#967](https://app.zenhub.com/workspaces/platform-design-system-5f8de67192551b0012ebb802/issues/department-of-veterans-affairs/vets-design-system-documentation/967) and request review from an accessibility specialist
- [x] Ping designer for design review
- [x] Display the appropriate [maturity scale](https://design.va.gov/about/maturity-scale) option in Storybook (once this feature is available)
- [x] If this is a new component that has not gone through Staging Review, it should be labeled "Use with Caution: Candidate"
- [ ] Merge component
- [ ] Create a new release of component-library
- [ ] Update component-library dependency in vets-design-system-documentation to have the updated component-docs.json

## Definition of done


packages/web-components/src/components.d.ts
```diff
@@ -286,6 +286,24 @@ export namespace Components {
     }
     interface VaFeaturedContent {
     }
+    interface VaFileInput {
+        /**
+          * The text displayed on the button.
+         */
+        "buttontext"?: string;
```
jamigibbs, line 293: @andresriveratoro I can't seem to get these comments to display as the description on the Storybook page (or any default values for that matter). Do you have any insight on how to get that to pull into the page?

packages/web-components/src/components/va-button/va-button.tsx
```diff
@@ -11,6 +11,10 @@ import {
 /**
  * @nativeHandler onClick
  */
+
+/**
+ * @nativeHandler onKeyPress
```
jamigibbs, line 16: @andresriveratoro I needed to expose this native event in `va-button` so that the file upload button (which uses `va-button`) could be activated by keyboard for a11y. Are there any issues with this that you can think of or anything else I need to consider by adding this here?

andresriveratoro (commented):
It looks like enableAnalytics prop is missing from

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -0,0 +1,174 @@
...
+        <span id="error-message" role="alert">
+          {error && (
+            <Fragment>
+              <span class="sr-only">{i18next.t('error')}</span> 
+              {error}
+            </Fragment>
+          )}
+        </span>
+        <va-button
+          onKeyPress={this.handleButtonKeyPress}
```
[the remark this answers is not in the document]
jamigibbs: @andresriveratoro Good catch! Not sure how that slipped past me. 👍

packages/storybook/stories/va-file-input.stories.jsx
```diff
@@ -0,0 +1,135 @@
...
+      label={label}
+      name={name}
+      buttontext={buttontext}
+      accept={accept}
+      required={required}
+      multiple={multiple}
+      error={error}
+      enable-analytics={enableAnalytics}
+    />
+  );
```
andresriveratoro, line 65: I like what you're doing here, but I think it makes it slightly more difficult for VFS teams to use. We can use the [bindings](https://stenciljs.com/docs/framework-bindings) created by Stencil using `yarn build-bindings` for this component here instead. A benefit of using bindings is VFS engineers using this component in React applications can use `onVaChange` as a prop and pass `handleChangeEvent` to it without having to add and remove event listeners.

andresriveratoro: Here's an [example](https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/storybook/stories/va-modal.stories.jsx#L27-L74) using bindings in Storybook.

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+  /**
+   * The name for the input element.
+   */
+  @Prop() name?: string;
+
+  /**
+   * The text displayed on the button.
+   */
+  @Prop() buttontext?: string;
```
andresriveratoro, line 42: Should we give this a default value of "Upload Document"?

andresriveratoro: This is minor, but how about if we update this to use camel case or rename to `text`?

andresriveratoro: If we don't supply a default value, we should make it a required prop and reflect that in our Storybook.

jamigibbs: @andresriveratoro There is some discussion about if we should provide this as an attribute at all and rather set a static text value. If the decision has been made to allow this text to be dynamic, I'll definitely come back around and address your suggestions.

packages/storybook/stories/va-file-input.stories.jsx
jamigibbs: @andresriveratoro Thank you for the direction here! I'll check this out.

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+  /**
+   * The name for the input element.
+   */
+  @Prop() name?: string;
+
+  /**
+   * The text displayed on the button (required)
+   */
+  @Prop() buttonText!: string;
```
andresriveratoro, line 42: Should this be required? I see you're providing a fallback value of `i18next.t('Upload file')` for the button.

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -0,0 +1,169 @@
...
+          aria-label={i18next.t(label)}
+          onClick={this.handleButtonClick}
+          secondary
+          text={buttonText ? i18next.t(buttonText) : i18next.t('Upload file')}
+          aria-describedby={error ? 'error-message' : undefined}
+        />
+        <input
+            id="fileInputField"
+            multiple={multiple}
+            style={{ display: 'none' }}
```
andresriveratoro, line 159: Could we have these styles live in `va-file-input.css`?

packages/web-components/src/components/va-file-input/va-file-input.tsx
jamigibbs: @andresriveratoro It does seem redundant now (the fallback request was part of the design discussion). I'll clean that up.

packages/web-components/src/components/va-file-input/va-file-input.tsx
jamigibbs: @andresriveratoro That's a good question. I wasn't completely sure myself. The React version had it inline so I was just pulling it over here. I leaned towards keeping it because I felt like it made it a little more readable for what was going but I'm happy to move it into the styles though and just have a `.hidden` class.

packages/web-components/src/components/va-file-input/va-file-input.css
```diff
@@ -0,0 +1,28 @@
...
+@import '../../mixins/form-field-error.css';
+
+:host {
+  display: block;
+  font-family: var(--font-source-sans);
+}
+
+label[part="label"] {
+  display: block;
+}
```
bkjohnson, line 11: Is this a leftover change? I couldn't find `part="label"` anywhere.

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -0,0 +1,168 @@
...
+          {error && (
+            <Fragment>
+              <span class="sr-only">{i18next.t('error')}</span> 
+              {error}
+            </Fragment>
+          )}
+        </span>
+        <va-button
+          id="fileInputButton"
+          aria-label={i18next.t(label)}
```
bkjohnson, line 149: I don't think we should pass a prop like this to i18next because we don't know if we're going to have a translation for it.

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -0,0 +1,168 @@
...
+          onClick={this.handleButtonClick}
+          secondary
+          text={buttonText ? i18next.t(buttonText) : i18next.t('Upload file')}
```
bkjohnson, line 152: Same here. There's a good change we aren't going to have a translation for whatever the `buttonText` prop happens to be. And I also don't think we have a translation for "Upload file".

A screenshot of the internationalized story for the file input component. "Required" is in Spanish as "Requerido", but the aria-label and text of the button are in English.

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  /**
+   * Set the input to required and render the (Required) text.
+   */
+  @Prop() required?: boolean = false;
+
+  /**
+   * Allow the input to accept multiple files.
+   */
+  @Prop() multiple?: boolean = false;
```
bkjohnson, line 52: Should this prop change the user-facing text at all? Like `Upload file(s)`?

bkjohnson (commented):
A few comments, mostly about translation.

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -0,0 +1,168 @@
...
+            <Fragment>
+              <span class="sr-only">{i18next.t('error')}</span> 
+              {error}
+            </Fragment>
+          )}
+        </span>
+        <va-button
+          id="fileInputButton"
+          aria-label={i18next.t(label)}
+          onClick={this.handleButtonClick}
```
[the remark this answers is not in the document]
jamigibbs: @bkjohnson I originally had a handler for `keypress` but we discovered it was firing both click and keypress events making the upload interface display twice. If focus is on the button though and space or enter are pressed, it will fire the click event (as a browser default behavior) so this should handle both mouse and keyboard events. Let me know if maybe you've discovered a bug with just having it like this though.

packages/web-components/src/components/va-file-input/va-file-input.tsx
jamigibbs: @bkjohnson I might have just misunderstood how the translations will map. If there is no translation available, would it default to the provided label text (english)? Alternately, if there is no label, would it just not populate with anything? Is there any harm in setting it up as "translation-ready"?

packages/web-components/src/components/va-file-input/va-file-input.tsx
jamigibbs: I don't know the answer to this. @GnatalieH Do you know if we want to pluralize the button text when the component allows for multiple files? I did not see this in the original React component. Looking at [USWDS](https://designsystem.digital.gov/components/file-input/), they have two different types of text in the examples but no explicit guidance on pluralizing.

packages/web-components/src/components/va-file-input/va-file-input.tsx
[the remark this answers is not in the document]
bkjohnson: Oh, that's nice. I didn't test it out, I just mentioned it because of [this recent issue](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1096).

packages/web-components/src/components/va-file-input/va-file-input.tsx
bkjohnson: The `t()` function takes a key as its first argument and uses it to look up values in a [language-specific translation object like this](https://github.com/department-of-veterans-affairs/component-library/blob/3fc3b7d33d9cfe5ac51a54d7aed0b445c26cfe05/packages/core/src/i18n/translations/en.js#L1-L26). For template strings I've seen it return `undefined` if the value isn't there, so I'm not totally sure why we still see text at all when we're in Spanish mode.

As far as making something translation-ready, I asked about [something similar in the past](https://github.com/department-of-veterans-affairs/component-library/pull/438#discussion_r888220759), and @k80bowman has said that [we should only provide translations for a component if we can get all of the values](https://github.com/department-of-veterans-affairs/component-library/pull/448#discussion_r891584148). However, that was said before we had linting for string literals, so now there is less of a chance of us forgetting to translate a piece of text, but partially translated components still might not be a good UX :shrug:.

packages/web-components/src/components/va-file-input/va-file-input.tsx
jamigibbs: @bkjohnson On further thought here, from a technical perspective, would we even want to control the text displayed within the component itself based on another prop? Shouldn't a pluralization be something the developer implementing it be responsible for updating (by changing the `buttonText` prop value itself)? Otherwise, we'd probably need to add an entirely new prop dedicated to pluralization text and I'm just not sure that's really necessary. This seems more like design guidelines than something we need to restrict within the component itself.  cc @GnatalieH

packages/web-components/src/components/va-file-input/va-file-input.tsx
jamigibbs: That makes sense and thanks! I'll make some updates here.

packages/web-components/src/components/va-file-input/va-file-input.tsx
GnatalieH: @jamigibbs we won't have an instance where bulk file uploading will be an option, in fact we are discouraging that pattern because it's not mobile-friendly, and lots of errors can occur. So I don't think we need to worry about making the text plural. CC @bkjohnson

bkjohnson: It's unclear to me what the `buttonText` decision was. I see that there are lots of resolved comments which I haven't clicked on and opened, but [the original issue](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/966#issuecomment-1206886490) mentioned that a decision would be made:

> Button text (whether to lock it for editing or not): we will have a decision from Danielle and Beth next week about this

And I haven't seen anything else about that called out in the code or the PR description. So, if we're going to internalize that text, I think pluralizing it could be a reasonable thing to do.

---

>  we won't have an instance where bulk file uploading will be an option

Does this mean that we don't need `multiple` at all?

jamigibbs: @bkjohnson The "button text decision" was about if we should allow the text to be dynamic or a fixed static value (removing the prop). The decision was to allow it to be dynamic and default to "File upload". The conversation happened in [Slack](https://dsva.slack.com/archives/C01K37HRUAH/p1660240015362249?thread_ts=1659547593.525989&cid=C01K37HRUAH) so I'll update the ticket itself with that info.

packages/web-components/src/components/va-file-input/va-file-input.tsx
jamigibbs: @bkjohnson I have removed `i18next` so I think that will resolve the translation concerns now.  https://github.com/department-of-veterans-affairs/component-library/pull/507/commits/1d8ec3cec7cde235e66aebd083f4e07dc98c29f0

caw310:
Decision during stand up is to remove the multiple file upload prop.

packages/web-components/src/components/va-file-input/va-file-input.css
```diff
@@ -0,0 +1,24 @@
...
+
+.required {
+  color: var(--color-secondary-dark);
+  margin-left: 0.4rem;
+}
+
+.hidden {
+  display: none;
+}
```
bkjohnson, line 24: Would adding the [`hidden` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) to the `<input>` element do the same thing?

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  /**
+   * Unable to set a default text value when the prop is marked
+   * as required so generating the default text with this utility.
+   * @returns 
+   */
+  private getButtonText = (): string => {
+    return this.buttonText ? this.buttonText : 'Upload file';
+  };
```
bkjohnson, line 112: If we are defaulting the value anyway, why do we want to mark the prop as required?

jamigibbs: You're right, the prop does not need to be marked as required (I will update that). But after testing setting a default value on it, I see that we will still need this utility so that it always defaults to "Upload file" if no other text is provided.

packages/web-components/src/components/va-file-input/va-file-input.tsx
GnatalieH: @bkjohnson and @jamigibbs the button text should default to "Upload file" rather than "File upload". The Content team recommends starting button text with a verb. Thanks.

packages/web-components/src/components/va-file-input/va-file-input.css
jamigibbs: I was concerned that `hidden` would affect screen readers in some way although now that I'm testing both ways, I can't find a functional difference so I think that's a good callout and streamlines it a bit more.

packages/web-components/src/components/va-file-input/va-file-input.tsx
k80bowman: Yeah, I'm not a fan of partially translated components if we can avoid it. There may be times we can't (say the component has already been translated, but we add a new error message), but if we don't have any translations yet I'd rather not enable i18next.

packages/web-components/src/components/va-file-input/va-file-input.tsx
jamigibbs: Resolved here: https://github.com/department-of-veterans-affairs/component-library/pull/507/commits/89adf95016f7a8b180d8cda20c7743e31b611206

packages/web-components/src/components/va-file-input/va-file-input.css
jamigibbs: Resolved here: https://github.com/department-of-veterans-affairs/component-library/pull/507/commits/89adf95016f7a8b180d8cda20c7743e31b611206

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+  /**
+   * The name for the input element.
+   */
+  @Prop() name?: string;
+
+  /**
+   * The text displayed on the button.
+   */
+  @Prop() buttonText: string;
```
bkjohnson, line 37: Why won't this work?

```suggestion
  @Prop() buttonText: string = "Upload file";
```

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -0,0 +1,166 @@
...
+  private handleButtonClick = () => { 
+    this.el.shadowRoot.getElementById('fileInputField').click();
+  }
+
+  /**
+   * Makes sure the button text always has a value.
+   * @returns string - Button text to render.
+   */
+  private getButtonText = (): string => {
+    return this.buttonText ? this.buttonText : 'Upload file';
```
bkjohnson, line 105: I'm still not sure why we need this function.

bkjohnson (commented):
I apologize if I'm overlooking something obvious, I'm just trying to understand this from your comment above:

> But after testing setting a default value on it (buttontext), I see that we will still need this utility so that it always defaults to "Upload file" if no other text is provided.

packages/web-components/src/components/va-file-input/va-file-input.tsx
jamigibbs: @bkjohnson Sorry if I didn't explain completely. I'm seeing that there is a difference between setting a "default" and "requiring a default". My requirement is to make sure that the button _always_ has a specific default text but also allow it to be dynamic.

If I just did `@Prop() buttonText: string = "Upload file";`, then the prop could be set to blank and the button would have no text like this:

Screen Shot 2022-08-17 at 1 03 26 PM

The utility function allows the value to be dynamic while also enforcing a default value.

packages/web-components/src/components/va-file-input/va-file-input.tsx
jamigibbs: See my comment here: https://github.com/department-of-veterans-affairs/component-library/pull/507#discussion_r948267986

packages/web-components/src/components/va-file-input/va-file-input.tsx
bkjohnson: Gotcha :+1:

bkjohnson (approved):

bkjohnson (approved):

PR #508, opened 2022-08-05
chore: add height, line-height, text-align, text-decoration and width utilities

andresriveratoro:
## Chromatic

https://1029-utilities--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1029

## Acceptance criteria
- [x] `css-library` has v3 utilities for height, line-height, max-height, max-width, min-height, min-width, text-align, text-decoration and width properties

## Definition of done


packages/css-library/src/tokens/line-height.scss
```diff
@@ -0,0 +1,8 @@
+$tokens-line-height: (
+  1: 1,
+  2: 1.15,
+  3: 1.35,
+  4: 1.5,
+  5: 1.62,
+  6: 1.75,
+);
```
bkjohnson, line 8: This seems a bit strange to me because I don't know what the numbers mean, but they [match what we already have](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/a831cc59059d2ef01c161f03e4fdb3caa5c5a8d4/packages/formation/sass/utilities/_line-height.scss#L5-L12). Adding a comment explaining what they mean might be nice.

We also may want these values to be tokens in StyleDictionary, but we can do that as a follow-up.

packages/css-library/src/utilities/text-align.scss
```diff
@@ -0,0 +1,14 @@
+@use 'uswds-core/src/styles/functions' as *;
+@use '../tokens/text-align' as *;
+@use './settings' as *;
+
+$u-text-align: (
+  text-align: (
+    base: 'vads-text',
```
bkjohnson, line 7: This is the same base as the utility for `text-decoration`, but all the tokens are different so I don't _think_ there would be ambiguity. We end up with:

- `vads-text-left`
- `vads-text-right`
- `vads-text-center`
- `vads-text-justify`

packages/css-library/src/utilities/text-decoration.scss
```diff
@@ -0,0 +1,14 @@
+@use 'uswds-core/src/styles/functions' as *;
+@use '../tokens/text-decoration' as *;
+@use './settings' as *;
+
+$u-text-decoration: (
+  text-decoration: (
+    base: 'vads-text',
```
bkjohnson, line 7: And then:

- `vads-text-none`
- `vads-text-underline`
- `vads-text-strikethrough`

packages/css-library/src/utilities/width.scss
```diff
@@ -0,0 +1,38 @@
...
+    values: map-collect($tokens-width),
+    settings: $width-settings,
+    property: 'width',
+    type: 'utility',
+  ),
+);
+
+$u-min-width: (
+  min-width: (
+    base: 'vads-min-width',
```
bkjohnson, line 18: These are different from the USWDS pattern of `minw-*` but I think our approach is better from a readability standpoint.

bkjohnson (approved):
Looks good to me. Left a comment about possibly using StyleDictionary tokens for line-height values since it isn't really obvious what the numbers represent, but it's not urgent.

PR #513, opened 2022-08-17
Add analytics details for web components

k80bowman:
## Chromatic

https://1110-component-analytics--60f9b557105290003b387cd5.chromatic.com

## Description
Adds or updates analytics for the following web components:
-  va-banner
-  va-button
-  va-button-pair
- va-date
- va-link
- va-memorable-date
- va-number-input
- va-pagination

https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1110

## Testing done
Locally

## Acceptance criteria
- [ ] Analytics details have been added to the components listed above.

## Definition of done


packages/web-components/src/components/va-button-pair/va-button-pair.tsx
```diff
@@ -72,26 +72,31 @@ export class VaButtonPair {
...
-    });
+    if (!this.disableAnalytics) {
+      const detail = {
+        componentName: 'va-button-pair',
+        action: 'click',
+        details: {
+          type: 'primary',
+        },
+      };
+      this.componentLibraryAnalytics.emit(detail);
```
bkjohnson, line 83: This seems like it might be redundant considering the `va-button` will be emitting its own event as well and the `button-type` value will either be `primary` or `secondary`. Do we need the pair to emit its own event?

packages/web-components/src/components/va-button/va-button.tsx
```diff
@@ -76,17 +76,20 @@ export class VaButton {
...
+  private handleClick = (): void => {
+    if (!this.disableAnalytics) {
+      const detail = {
+        componentName: 'va-button',
+        action: 'click',
+        details: {
+          event: 'cta-button-click',
+          'button-type': this.secondary ? 'secondary' : 'primary',
+          'button-click-label': this.getButtonText(),
+          'button-background-color': this.secondary ? 'white' : '#0071bb',
```
bkjohnson, line 88: I'm not sure why we want to send the background color, but hardcoding a hex here seems like something that will bite us in the future if the color changes.

packages/web-components/src/components/va-button/va-button.tsx
```diff
@@ -76,17 +76,20 @@ export class VaButton {
...
-    });
+  private handleClick = (): void => {
+    if (!this.disableAnalytics) {
+      const detail = {
+        componentName: 'va-button',
+        action: 'click',
+        details: {
+          event: 'cta-button-click',
+          'button-type': this.secondary ? 'secondary' : 'primary',
+          'button-click-label': this.getButtonText(),
```
bkjohnson, line 87: I think it would be good if we standardized around a set vocabulary for common things. We're using `button-click-label` here to represent the "label" or text of the button, and in the radio component we're just using the simpler `label` to match the prop name:

https://github.com/department-of-veterans-affairs/component-library/blob/bb1bcb44d2eb724fbbe70da96f8e3bb06f7eb551/packages/web-components/src/components/va-radio/va-radio.tsx#L161-L165

To make things worse, the `label` prop on the button is meant to be a pass-through for an `aria-label`:

https://github.com/department-of-veterans-affairs/component-library/blob/bb1bcb44d2eb724fbbe70da96f8e3bb06f7eb551/packages/web-components/src/components/va-button/va-button.tsx#L130-L133

So I'm not saying "this needs to have the name 'label'", but I'm hoping we can avoid getting into a situation where our analytics events are using many different names to represent the same thing, and I'm not sure that the `button-` prefix adds anything here since we already know it's a button from the `componentName`.

packages/web-components/src/components/va-button/va-button.tsx
```diff
@@ -76,17 +76,20 @@ export class VaButton {
...
-        // TODO: add analytics event details
-      },
-    });
+  private handleClick = (): void => {
+    if (!this.disableAnalytics) {
+      const detail = {
+        componentName: 'va-button',
+        action: 'click',
+        details: {
+          event: 'cta-button-click',
```
bkjohnson, line 85: What does this do for us?

packages/web-components/src/components/va-link/va-link.tsx
```diff
@@ -79,14 +79,16 @@ export class VaLink {
...
-      details: {
-        // TODO: add analytics event details
-      },
-    });
+    if (!this.disableAnalytics) {
+      const detail = {
+        componentName: 'va-link',
+        action: 'click',
+        details: {
+          'link-click-label': this.text,
```
bkjohnson, line 87: Same thing here regarding the `link-click-label` naming.

packages/web-components/src/components/va-memorable-date/va-memorable-date.tsx
```diff
@@ -101,6 +101,20 @@ export class VaMemorableDate {
     // will happen afterwards
     validate(this, yearNum, monthNum, dayNum);
     this.dateBlur.emit(event);
+
+    if (this.enableAnalytics) {
+      const detail = {
+        componentName: 'va-memorable-date',
+        action: 'blur',
+        details: {
+          label: this.label,
```
bkjohnson, line 110: And here it's just `label` instead of `date-blur-label` or something like that.

bkjohnson (commented):
Seeing all these analytics events in the same diff makes me realize we lack a common interface for the names of the event details. I know that changing the names of these details can be risky for the ones that are already in use, but do you think there's anything we can do to help the names be consistent? For the common pieces of data, I think ideally the same name would get you the same piece of data no matter what component the event is for.

packages/web-components/src/components/va-button-pair/va-button-pair.tsx
jamigibbs: The analytics ticket wanted an event sent for `va-button-pair` which I think the reason is so that GA knows `va-button-pair` was used. I don't know the answer to if we should remove this because of a `va-button` redundancy. I would lean towards having unique analytics events for individual components just based on my very limited knowledge of GA (even if those components contain children). I am not sure if this will cause "double logging" in GA though or if that would even be an issue with them.

packages/web-components/src/components/va-button/va-button.tsx
jamigibbs: The GA event logs as a button click for `va-button` which they are labeling as a "call to action button click (`cta-button-click`). These are the details I received from the analytics team.

packages/web-components/src/components/va-button/va-button.tsx
jamigibbs: @bkjohnson This was added to the PR before I took over the analytics ticket so I don't know the reason or the history behind the default color but following the tickets related to this, I can see that background color is part of the data layer specs here:

https://depo-platform-documentation.scrollhelp.site/analytics-monitoring/design-system-components

packages/web-components/src/components/va-button/va-button.tsx
jamigibbs: I dug into this a little more and it looks like we should probably set the key value at this point as just "label" because [in the vets-website script](https://github.com/department-of-veterans-affairs/vets-website/blob/77e40f5431f228ffba7a159dd377f5a09d3e4d9b/src/platform/site-wide/component-library-analytics-setup.js#L350), it's transformed to `${action.prefix}-${key}`. For example, in the va-radio example, the actual datalayer key that will be sent for label after it's transformed is `button-label`. 

Note: I think the use of the value `action` is kind of a poor name in this case because there is also an `action` within the component event details. "action" in this case is just the found component info object in vets-website like below:

```
  'va-button': [
    {
      action: 'click',
      event: 'cta-button-click',
      prefix: 'button',
    },
  ],
```

packages/web-components/src/components/va-memorable-date/va-memorable-date.tsx
jamigibbs: I've updated this to just `label` because of the way the vets-website script is transforming it for the datalayer.

packages/web-components/src/components/va-link/va-link.tsx
jamigibbs: I've updated this to just `label` because of the way the vets-website script is transforming it for the datalayer.

packages/web-components/src/components/va-button/va-button.tsx
bkjohnson: Can't that be added in the `vets-website` mapping where it's more specific to the GA implementation?

packages/web-components/src/components/va-button-pair/va-button-pair.tsx
bkjohnson: It would cause double logging because we don't stop the analytics event for the button from propagating.

packages/web-components/src/components/va-button/va-button.tsx
```diff
@@ -76,17 +76,20 @@ export class VaButton {
...
+          type: this.secondary ? 'secondary' : 'primary',
+          label: this.getButtonText(),
+          'background-color': this.secondary ? 'white' : '#0071bb',
```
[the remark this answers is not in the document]
jamigibbs: To make this dynamic, we'd probably want to use [getComputedStyle](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle) but if that's put in the click handler (where the analytics event is triggered), then the element background color would be different than it's non-clicked state (I have tested this). 

Screen Shot 2022-11-14 at 10 26 57 AM
Screen Shot 2022-11-14 at 10 26 59 AM

If we want to get the background color of the non-click state, that means we'd probably want to generate the computed style within a component lifecycle which also means we'd probably want to create a private variable to store it.

So we're talking about adding a dozen additional lines of code to make the color value dynamic just for a piece of analytics data that we don't even know why it's necessary or needed. If we can't verify the reason this value is needed or don't want to hard code a value then I don't think there's enough reason to add any additional code for it and I propose we just remove it from the analytics event entirely.

packages/web-components/src/components/va-button/va-button.tsx
jamigibbs: I understand what you mean now. The part `event: 'cta-button-click',` [is in vets-website](https://github.com/department-of-veterans-affairs/vets-website/blob/77e40f5431f228ffba7a159dd377f5a09d3e4d9b/src/platform/site-wide/component-library-analytics-setup.js#L180-L185) already. I'm not sure why it was added here. Removing.

packages/web-components/src/components/va-button/va-button.tsx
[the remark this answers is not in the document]
bkjohnson: Since it was part of the PR when you picked it up, maybe it would be good if we asked Michelle if it's truly needed or not.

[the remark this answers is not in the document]
jamigibbs: @michelle-dooley We're looking at the `va-button` GA event `cta-button-click`. We're not sure if the detail `background-color` is actually necessary for the design system component. I think it might have come out of a conversation that happened in the analytics request ticket here but I don't know the context (if any): https://github.com/department-of-veterans-affairs/va.gov-team/issues/45510#issuecomment-1216867680 

Do you remember if this was a specific request from Katy or is that something that you require on your end? We are considering removing it because we're not sure that the additional code needed to make the color value dynamic is worth the value of collecting the data.

packages/web-components/src/components/va-button-pair/va-button-pair.tsx
jamigibbs: @michelle-dooley We sometimes use our own design system components in our other design system components. For example in this case, we are using the `va-button` inside of `va-button-pair` and there are separate GA events for those components: `int-button-pair-click` and `cta-button-click`.

So when `va-button` is clicked, both `int-button-pair-click` and `cta-button-click` fire. This _seems_ like it would be an issue for accurate reporting in GA but I just wanted to confirm with you before we dug into a technical solution on our end. The solution may end up being that we remove the GA event for `va-button-pair` completely and just rely on the `va-button` event itself.

packages/web-components/src/components/va-button/va-button.tsx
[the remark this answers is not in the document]
jamigibbs: Removing `background-color` after [confirming with Michelle](https://github.com/department-of-veterans-affairs/component-library/pull/513#issuecomment-1315799366).

packages/web-components/src/components/va-button-pair/va-button-pair.tsx
jamigibbs: @bkjohnson I think after the [verification by Michelle](https://github.com/department-of-veterans-affairs/va.gov-team/issues/45510#issuecomment-1320211727) that we would be seeing double logging here and it would be a bit redundant as you mentioned originally, I'm ok with removing this from `va-button-pair` if you're comfortable with that too.

bkjohnson: We could stop the propagation of the analytics event coming from the button and just let the one from the button pair bubble up.

jamigibbs: How would we know when to stop it? Would we be checking if `va-button-group` is a parent element of `va-button` from within `va-button`?

bkjohnson: We would know when to stop it if the `<va-button-pair>` receives an analytics event bubbling up from its button children.

jamigibbs: @bkjohnson I think I'm on the same page now. I've updated va-radio-button to handle that analytics events like this: https://github.com/department-of-veterans-affairs/component-library/pull/513/commits/87a38facab601f4902c6584eb0ddd050425f3d34

Andrew565 (approved):
LGTM, though I think it needs a quick merge update.

PR #515, opened 2022-08-19
Add grid utilities

andresriveratoro:
## Chromatic

https://1032-grid--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1032

## Acceptance criteria
- [x] The grid utility classes based on USWDS v3 have been added to the new CSS library

## Definition of done


bkjohnson (commented):
For these grid utilities, particularly the float grid, would it be possible to import `usa-layout-grid` and possibly change the prefix? It looks like the `usa-width-*` is specific to USWDS1.

andresriveratoro:
@bkjohnson I don't believe we can override the variables we need because `$namespace-grid` isn't [configurable](https://sass-lang.com/documentation/variables#default-values) and I haven't been able to override ` $theme-namespace`.

Here's what I'm getting when trying to override `$theme-namespace`.
Screen Shot 2022-08-25 at 09 59 52

andresriveratoro:
Do we want to use `usa-layout-grid` with our breakpoints or USWDS?

packages/css-library/src/utilities/flex-grid.scss
```diff
@@ -0,0 +1,53 @@
...
+    padding-right: 3rem;
+  }
+}
+
+.vads-grid-container-full {
+  margin-left: auto;
+  margin-right: auto;
+  max-width: 100%;
+  padding: 0;
+}
```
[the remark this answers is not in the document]
andresriveratoro: `vads-grid-container-full` is from our [Flexbox grid](https://design.va.gov/foundation/layout/flexbox-grid). Should we discard it?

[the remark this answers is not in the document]
andresriveratoro: Importing `usa-layout-grid` is also bringing the `grid-container` and `row` utilities.

bkjohnson:
Could we try to override [this value in the Sass map](https://github.com/uswds/uswds/blob/846e45b50662851a6d37f20726d9d20dd415a3fa/packages/uswds-core/src/styles/_properties.scss#L156-L162)?

andresriveratoro:
I tried doing that using a map merge, but I'm getting the following errors:

Screen Shot 2022-08-26 at 11 40 37
Screen Shot 2022-08-26 at 11 40 21


```jsx
@use 'sass:map';
@use 'uswds-core/src/styles/properties' as *;
$system-properties: map.merge(
  $system-properties,
  (
    breakpoints: (
      standard: (
        'large': (
          'xsmall-screen': spacing-multiple(40),
          'small-screen': 481px,
        ),
        'larger': (
          'medium-screen': 768px,
        ),
        'largest': (
          'small-desktop-screen': 1008px,
          'large-screen': 1201px,
        ),
      ),
    ),
  )
);
@use 'usa-layout-grid';
```

packages/css-library/src/utilities/flex-grid.scss
```diff
@@ -0,0 +1,53 @@
+@use 'uswds-core/src/styles/functions' as *;
+@use '../tokens/flex-grid' as *;
+@use './settings' as *;
+
+.vads-grid-container {
```
bkjohnson, line 5: USWDS3 has [flex utilities](https://designsystem.digital.gov/utilities/flex/) and [layout grid utilities](https://designsystem.digital.gov/utilities/layout-grid/), but not a [flex grid like we currently have](https://design.va.gov/foundation/layout/flexbox-grid). If we think that the layout grid utilities from USWDS can be a sufficient replacement, then we probably don't want these.

packages/css-library/src/stylesheets/utilities.scss
```diff
@@ -1,3 +1,26 @@
...
+    "card": false,
+    "card-lg": false,
+    "mobile": true,
+    "mobile-lg": false,
+    "tablet": true,
+    "tablet-lg": false,
+    "desktop": false,
+    "desktop-lg": true,
+    "widescreen": false
+  ),
```
andresriveratoro, line 22: The breakpoints here are used in utilities that have `responsive: true` in their settings. Please note that these breakpoints settings are being ignored by the `usa-layout-grid` import.

packages/css-library/src/utilities/float-grid.scss
```diff
@@ -0,0 +1 @@
+@use 'usa-layout-grid';
```
andresriveratoro, line 1: These utilities are ignoring the breakpoints settings we specify in `packages/css-library/src/stylesheets/utilities.scss`.

packages/css-library/src/utilities/flex-grid.scss
andresriveratoro: I ended up leaving these in because they're not 1:1 replacements.

Our existing flex grid utility:
```
.vads-flex-col-9 {
  box-sizing: border-box;
  max-width: 75%;
  min-height: 1px;
  min-width: 0;
  position: relative;
  width: 100%;
  flex: 0 0 75%;
}
```

USWDS grid utility:
```
.vads-grid-col-9 {
  flex: 0 1 auto;
  width: 75%;
}
```

packages/css-library/src/stylesheets/utilities.scss
k80bowman: Does this mean we are setting up to use two different sets of breakpoints, one for `usa-layout-grid` and one for everything else?

andresriveratoro: This set of breakpoints affects utilities when they are generated like we have for all other utilities except `usa-layout-grid`.

Here's a [margin utility](https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/css-library/src/utilities/margin.scss#L10) using [responsive settings](https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/css-library/src/utilities/settings.scss#L82-L89).

Since we're importing `usa-layout-grid` and not generating them, I believe this [import](https://github.com/uswds/uswds/blob/develop/packages/usa-layout-grid/src/styles/_usa-layout-grid.scss#L3) is overriding our custom uswds-core settings.

packages/css-library/src/utilities/flex-grid.scss
```diff
@@ -0,0 +1,54 @@
...
+  padding: 0;
+}
+
+.vads-flex-row {
+  display: flex;
+  flex-wrap: wrap;
+  min-width: 100%;
+}
+
+$u-flex-grid-col: (
```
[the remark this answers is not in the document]
andresriveratoro: These were given a `vads-flex` prefix to make it explicit which utilities are our [flex grid](https://design.va.gov/foundation/layout/flexbox-grid) and `usa-layout-grid`.

k80bowman (approved):
This looks good for now. I'm a little concerned about having two different sets of breakpoints, but we should do a little discovery around how this affects implementation as part of a separate ticket. Thanks for working through this, Andrés!

PR #517, opened 2022-08-23
Fix va-modal for Cypress

andresriveratoro:
## Chromatic

https://1113-modal-cypress--60f9b557105290003b387cd5.chromatic.com

## Description
Part of https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1113

## Acceptance criteria
- [x] Cypress tests utilizing va-modal pass 

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-modal/va-modal.tsx
+++ b/packages/web-components/src/components/va-modal/va-modal.tsx
@@ -355,58 +355,55 @@ export class VaModal {
 
     return (
       <Host
-      aria-label={ariaLabel}
-      aria-modal="true"
-      role={ariaRole(status)}
+        aria-label={ariaLabel}
+        aria-modal="true"
+        role={ariaRole(status)}
+        visible={visible}
       >
-        <div
-          class="va-modal"
-        >
-          <div class={wrapperClass} tabIndex={-1}>
-            <button
-              aria-label={btnAriaLabel}
-              class="va-modal-close"
-              onClick={e => this.handleClose(e)}
-              ref={el => (this.closeButton = el as HTMLButtonElement)}
-              type="button"
-            >
-              <i aria-hidden="true" />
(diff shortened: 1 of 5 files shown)


andresriveratoro:
I believed I fixed the previous error, but now I'm getting a new one. 🙃 
Screen Shot 2022-08-23 at 17 23 07

andresriveratoro:
The current changes pass the Cypress tests 🎉 
The error above was due to an incorrect usage of the web component. Removing `role="alertdialog"` gets rid of the error. Our modal components sets the correct role for a given `status`.
Screen Shot 2022-08-23 at 17 34 03

bkjohnson (commented):
Do these changes mess up the OMB Info component at all?

packages/web-components/src/components/va-modal/va-modal.tsx
```diff
@@ -355,58 +355,55 @@ export class VaModal {
 
     return (
       <Host
-      aria-label={ariaLabel}
-      aria-modal="true"
-      role={ariaRole(status)}
+        aria-label={ariaLabel}
+        aria-modal="true"
+        role={ariaRole(status)}
+        visible={visible}
```
andresriveratoro, line 361: @bkjohnson This change `visible={visible}` reflects in the DOM allowing the styles to be applied properly when using va-modal inside other web components.

bkjohnson: Could we have the `visible` prop have `{reflect: true}` and get the same result?

andresriveratoro: You're correct. Updating to that.

bkjohnson (approved):
The changes look good to me. I think this should have a `patch` label for publishing though.

PR #518, opened 2022-08-25
Fix alert expandable component calc-max-height

jamigibbs:
## Chromatic

https://1093-alert-height-defect--60f9b557105290003b387cd5.chromatic.com

## Description
Closes 
https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1093
https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1078

## Testing done
- Local testing in Storybook
- Browserstack mobile (Android)

## Screenshots

I think what the issue ended up being is that padding was getting applied on a dynamic class `.alert-expandable-body.open` so that wasn't getting added as part of the initially loaded content height calculation.

Additionally, if I added it to just `.alert-expandable-body` then the background color would be visible too soon (before the content was triggered to display). So I added another wrapping element around the `slot` identified as `#slot-wrap` where our padding could be applied and calculated correctly as part of the content display.

I have also tested the display with a large image to make sure it was also adjusting for that type of content too.

```
<div id="alert-body" class={bodyClasses}>
  <div id="slot-wrap">
    <slot></slot>
  </div>
</div>
```

Screen Shot 2022-08-30 at 1 50 01 PM

A similar effect was applied to `va-additional-info` except that it didn't require any additional wrapping elements since this component does not have a background color.

Screen Shot 2022-08-30 at 1 50 16 PM

Screen Shot 2022-08-30 at 3 35 55 PM

Screen Shot 2022-08-30 at 3 36 17 PM

## Acceptance criteria
- [ ]  Height is calculated correctly in the Alert - Expandable, Alert, and Additional Info components

## Definition of done


packages/storybook/.storybook/style.scss
```diff
@@ -33,3 +33,93 @@ a.sbdocs-a:hover {
...
+  width: 74.41059%;
+}
+
+.region-grid:not(:only-child):last-child {
+  width: 32%;
+}
+
+.region-grid {
+  margin-right: 12px;
+  padding-right: 12px;
```
jamigibbs, line 118: The right padding of this wrapping class seems to be causing the `scrollHeight` of the `alert-body` to calculate incorrectly.

jamigibbs:
@andresriveratoro This issue is reproduced here on Storybook: https://1093-alert-height-defect--60f9b557105290003b387cd5.chromatic.com/?path=/docs/components-va-alert-expandable--default  

The `right-padding` from `.region-grid` seems to be affecting the content width which in turn is causing the height calculations to be incorrect.

andresriveratoro:
@jamigibbs If it helps you troubleshoot it, I found that removing the `vamc-facility-expandable-alert` class from va-alert-expandable gets rid of the component growing to the right.

jamigibbs:
> @jamigibbs If it helps you troubleshoot it, I found that removing the `vamc-facility-expandable-alert` class from va-alert-expandable gets rid of the component growing to the right.

Good eye. Yes, I noticed that as well but forgot to mention that it's a custom class/styling being added by the content team that I don't think we can necessarily control. I will investigate it a little more though. Ideally, the component height should not be affected by any custom classes or styling but maybe there's something else we can do to mitigate it.

k80bowman (approved):

PR #519, opened 2022-08-25
Fix va-modal teardown crash

andresriveratoro:
## Chromatic

https://1113-teardown--60f9b557105290003b387cd5.chromatic.com

## Description
Part of https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1113

## Testing done
Cypress on vets-website locally

## Acceptance criteria
- [x] va-modal passes Cypress tests on vets-website

## Definition of done


packages/web-components/src/components/va-modal/va-modal.tsx
```diff
@@ -318,7 +318,7 @@ export class VaModal {
     clearAllBodyScrollLocks();
     this.undoAriaHidden?.();
 
-    document.body.classList.remove('modal-open');
+    document.body?.classList?.remove('modal-open');
```
bkjohnson, line 321: If `modal-open` _isn't_ removed, does scrolling get messed up at all? I'm looking at this:

https://github.com/department-of-veterans-affairs/component-library/blob/313f689371d692a74b6566470524875cadec16b7/packages/web-components/src/components/va-modal/va-modal.tsx#L294-L297

andresriveratoro: I'm not sure how to explain the behavior we saw during the huddle where `document.body` didn't exist resulting in an error when trying to access the classList property. If the modal-open class isn't removed, then the [programmatic scrolling](https://github.com/department-of-veterans-affairs/vets-website/pull/11712/files) doesn't happen.

```jsx
const position = errorEl.getBoundingClientRect().top + currentPosition;
if (!document.body.classList.contains('modal-open')) {
  Scroll.animateScroll.scrollTo(position - 10, getScrollOptions());
}
```

bkjohnson: I think it may be a good idea if the component-library doesn't add a class like that just for a `vets-website` utility. Instead, maybe the utility could have a condition like:

```js
const position = errorEl.getBoundingClientRect().top + currentPosition;
if (!document.body.classList.contains('modal-open') || !document.querySelectorAll('va-modal[visible]') {
  Scroll.animateScroll.scrollTo(position - 10, getScrollOptions());
}
```

I also noticed that there's a [bit of styling associated with `modal-open`](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/8bfacf86f340fae3b5111dec669bda7ed69fef6d/packages/formation/sass/base/_va.scss#L19-L21), but it isn't clear if it's actually necessary. It also seems like that's being overriden from somewhere:

Screenshot of a modal on staging.va.gov. In the inspector, the body tag has the `modal-open` class, but its styles are being overriden by an inline stylesheet

Additionally, `modal-open` is being used incorrectly to [prevent scrolling on things that aren't modals](https://github.com/department-of-veterans-affairs/vets-website/pull/14737/files#diff-3047b268882a0cdef2aa02ff1c1e53d85232162dfc020428b000812357da0846R10).

bkjohnson (commented):
That comment is a bit long-winded, but I think we should see if we can safely remove `modal-open` without breaking things, since that will simplify the component.

andresriveratoro:
@bkjohnson I've removed the modal-open class logic from va-modal and updated the modal check in this [vets-website PR](https://github.com/department-of-veterans-affairs/vets-website/pull/22014) to:

```jsx
    const isShadowRootModalOpen = Array.from(
      document.querySelectorAll('va-omb-info')
    ).some(ombInfo =>
      ombInfo.shadowRoot.querySelector(
        'va-modal[visible]:not([visible="false"])',
      ),
    );

    const isModalOpen =
      document.body.classList.contains('modal-open') ||
      document.querySelectorAll('va-modal[visible]:not([visible="false"])') ||
      isShadowRootModalOpen;

    if (!isModalOpen) {
      Scroll.animateScroll.scrollTo(position - 10, getScrollOptions());
    }
```

andresriveratoro:
@bkjohnson The query selectors were updated [here](https://github.com/department-of-veterans-affairs/vets-website/pull/22014/commits/895630e5b366bb707f79f073b5e954d6a23a5721).

bkjohnson (approved):
I think this should be good once we bump the version again for a release.

Issue #333, opened 2022-03-23
React components to remove in version 11


bkjohnson:
## Description

Due to migration to web components, we've got a few React components that are now unused:

### Migration completed or Web Component available

- [x] `Banner`
- [x] `PromoBanner`
- [x] `CollapsiblePanel`
- `CheckboxGroup`
- [x] `TextArea`

### Unused
- `HelpMenu`
- `IconHelp`
- `AcceptTermsPrompt`
- `PrivacyAgreement`

This is to keep track of which React components are unused. Bundling these kinds of removals together into a single version release can keep our major version creep low, which indicates a more stable package. It will also mean less chore work when bumping `component-library` versions in sibling repos.

## Tasks
- [ ] Remove listed components in next major version of the component library

k80bowman:
It looks like `Banner` and `CollapsiblePanel` are no longer used as well!

bkjohnson:
The description has been edited to reflect recent changes.

humancompanion-usds:
This work is being tracked in an epic: 
https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1530

PR #524, opened 2022-09-06
Privacy Agreement Web Component 

jamigibbs:
## Chromatic

## Description
Closes <ticket>

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.css
+++ b/packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.css
@@ -1,0 +1,10 @@
+@import '../../mixins/accessibility.css';
+@import '../../mixins/form-field-error.css';
+
+:host {
+  display: block;
+}
+
+a {
+  color: var(--color-link-default);
+}
--- a/packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.tsx
+++ b/packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.tsx
@@ -1,0 +1,90 @@
+/* eslint-disable i18next/no-literal-string */
+import { Component, Host, h, Prop, Event, EventEmitter, Fragment } from '@stencil/core';
+
+/**
+ * @componentName Privacy Agreement
+ * @maturityCategory caution
+ * @maturityLevel candidate
+ */
+
+@Component({
+  tag: 'va-privacy-agreement',
+  styleUrl: 'va-privacy-agreement.css',
+  shadow: true,
+})
+export class VaPrivacyAgreement {
+  /**
+   * Whether the checkbox is checked or not.
+   */
+  @Prop() checked?: boolean = false;
+
+  /**
+   * Whether to display the error message or not.
+   */
+  @Prop() isError?: boolean = false;
+
+  /**
+   * Emit component-library-analytics events on the blur event.
+   */
+  @Prop() enableAnalytics?: boolean = false;
+
+  /**
+   * The event used to track usage of the component. This is emitted when the
+   * input is blurred and enableAnalytics is true.
+   */
+  @Event({
+    eventName: 'component-library-analytics',
+    composed: true,
+    bubbles: true,
+  })
+  componentLibraryAnalytics: EventEmitter;
+
+  /**
+   * The event emitted when the checkbox changes.
+   */
+  @Event() vaChange: EventEmitter;
+
+  private handleBlur = (e: Event) => {
+    this.checked = (e.target as HTMLInputElement).checked;
+    this.vaChange.emit({ checked: this.checked });
+
+    if (this.enableAnalytics) {
+      this.componentLibraryAnalytics.emit({
+        componentName: 'va-privacy-agreement',
+        action: 'blur',
+        details: {
+          checked: this.checked
+        },
+      });
+    }
+  };
+
+  render() {
+    return (
+      <Host>
+        <span>
+          Please read and accept the&nbsp;
+          <a target="_blank" href="/privacy/">
+            privacy policy
+          </a>
+        </span>
+        <span id="error-message" role="alert">
+          {(this.isError && !this.checked) && (
+            <Fragment>
+              <span class="sr-only">error</span>
+              You must accept the privacy policy before continuing
+            </Fragment>
+          )}
+        </span>
+        <va-checkbox
+          required
+          id="checkbox"
+          label='I have read and accept the privacy policy'
+          checked={this.checked}
+          onBlur={this.handleBlur}
+        />
+      </Host>
+    );
+  }
+
+}
--- a/packages/web-components/src/mixins/form-field-error.css
+++ b/packages/web-components/src/mixins/form-field-error.css
@@ -4,14 +4,16 @@
     margin-bottom: 1.2rem;
 }
 
-:host([error]:not([error=''])) {
+:host([error]:not([error=''])), 
+:host([iserror="true"]) {
     border-left: 0.4rem solid  var(--color-secondary-dark);
     padding-left: 2rem;
     position: relative;
 }
 
 @media screen and (min-width: 1008px) {
-    :host([error]:not([error=''])) {
+    :host([error]:not([error=''])), 
+    :host([iserror="true"]) {
         margin-left: -2.4rem; /* padding left + border left */
     }
 }
@@ -21,7 +23,8 @@
 }
 
 :host([error]:not([error=''])) label,
-:host([error]:not([error=''])) span {
+:host([error]:not([error=''])) span, 
+:host([iserror="true"]) span {
     font-weight: 700;
 }
 
--- a/packages/storybook/stories/PrivacyAgreement.stories.jsx
+++ b/packages/storybook/stories/PrivacyAgreement.stories.jsx
@@ -3,13 +3,13 @@ import PrivacyAgreement from '../../react-components/src/components/PrivacyAgree
 import { StoryDocs } from './wc-helpers';
 
 export default {
-  title: 'Components/Privacy agreement',
+  title: 'Components/Privacy agreement - React',
   component: PrivacyAgreement,
   id: 'components/privacyagreement',
   parameters: {
     componentSubtitle: 'Privacy agreement React component',
     docs: {
-      page: () => <StoryDocs componentName="Privacy agreement" />,
+      page: () => <StoryDocs componentName="Privacy agreement - React" />,
     },
   },
 };
--- a/packages/storybook/stories/additional-docs.js
+++ b/packages/storybook/stories/additional-docs.js
@@ -104,7 +104,7 @@ export const additionalDocs = {
     maturityCategory: DONT_USE,
     maturityLevel: DEPRECATED,
   },
-  'Privacy agreement': {
+  'Privacy agreement - React': {
     maturityCategory: CAUTION,
     maturityLevel: AVAILABLE,
   },
--- a/packages/storybook/stories/va-privacy-agreement.stories.jsx
+++ b/packages/storybook/stories/va-privacy-agreement.stories.jsx
@@ -1,0 +1,48 @@
+import React, { useState } from 'react';
+import { VaPrivacyAgreement } from '@department-of-veterans-affairs/web-components/react-bindings';
+import { getWebComponentDocs, propStructure, StoryDocs } from './wc-helpers';
+
+VaPrivacyAgreement.displayName = 'VaPrivacyAgreement';
+
+const privacyAgreementDocs = getWebComponentDocs('va-privacy-agreement');
+
+export default {
+    title: 'Under development/Privacy agreement',
+    id: 'components/va-privacy-agreement',
+    parameters: {
+        componentSubtitle: 'va-privacy-agreement web component',
+        docs: {
+            page: () => <StoryDocs data={privacyAgreementDocs} />,
+        },
+    },
+};
+
+const defaultArgs = {
+    checked: false,
+    isError: false
+};
+
+const Template = ({
+    checked,
+    isError
+}) => {
+    const [isChecked, setIsChecked] = useState(false);
+    const handler = (event) => setIsChecked(event?.detail?.checked);
+
+    if (isChecked) {
+        isError = false;
+    }
+    return (
+        <VaPrivacyAgreement
+            checked={isChecked}
+            isError={isError}
+            onVaChange={handler}
+        />
+    );
+};
+
+export const Default = Template.bind(null);
+Default.args = { ...defaultArgs };
+
+export const WithError = Template.bind(null);
+WithError.args = { ...defaultArgs, isError: true };
--- a/packages/web-components/src/components.d.ts
+++ b/packages/web-components/src/components.d.ts
@@ -535,6 +535,20 @@ export namespace Components {
          */
         "showLastPage"?: boolean;
     }
+    interface VaPrivacyAgreement {
+        /**
+          * Whether the checkbox is checked or not.
+         */
+        "checked"?: boolean;
+        /**
+          * Emit component-library-analytics events on the blur event.
+         */
+        "enableAnalytics"?: boolean;
+        /**
+          * Whether to display the error message or not.
+         */
+        "isError"?: boolean;
+    }
     interface VaProcessList {
     }
     interface VaProgressBar {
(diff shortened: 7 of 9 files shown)


jamigibbs:
Hi @GnatalieH 👋 Do you mind giving your design review on this when you have a chance?

packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.tsx
```diff
@@ -0,0 +1,89 @@
...
+export class VaPrivacyAgreement {
+  /**
+   * Whether the checkbox is checked or not.
+   */
+  @Prop({ mutable: true }) checked?: boolean = false;
+
+  /**
+   * Whether to display the error message or not.
+   */
+  @Prop() isError?: boolean = false;
```
bkjohnson, line 23: Do we need an external prop for this? I'm wondering if a State variable could work.

packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.tsx
```diff
@@ -0,0 +1,89 @@
...
+      });
+    }
+  }
+
+  render() {
+    return (
+      <Host>
+        <span>
+          Please read and accept the&nbsp;
+          <a target="_blank" href="/privacy/">
```
bkjohnson, line 66: Is this the link we want to use? https://va.gov/privacy redirects to the [OPRM subdomain](https://www.oprm.va.gov/).

Screenshot of oprm.va.gov

jamigibbs: Interesting. I think the correct link right now is https://www.va.gov/privacy-policy/ but this was taken directly from the [React version](https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/react-components/src/components/PrivacyAgreement/PrivacyAgreement.jsx#L21) of the component so I guess that link changed at some point.

@k80bowman This could be related to what we were talking about before with making the link a prop (defaulting to `/privacy-policy`) instead of hard-coded. If we still want to keep it on lockdown, I can just update this to the current working privacy link but I'm wondering if the old version of this component is still being used somewhere with an incorrect hard-coded link.

packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.tsx
jamigibbs: I considered that but there were two things that made me pause; we would probably still need to communicate back what the error state is for any kind of form or page validation handling and second thing is that it would cause a rerender so I wasn't sure if that was worth doing. 

I do like the idea of delegating error management to the component but wouldn't the developer implementing it still need to know the error state either way?

bkjohnson: You're right about the `@State` triggering a rerender - we could just have it be a class property instead if we wanted to avoid that. 

We've already got the `checked` prop, so could that be used to communicate an error state?

packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.tsx
bkjohnson: The React version of this component isn't being used anywhere. [This issue](https://github.com/department-of-veterans-affairs/component-library/issues/333) hasn't been updated in a while, but the four components under "Unused" are still unused.

packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.css
```diff
@@ -0,0 +1,14 @@
...
+:host {
+  display: block;
+}
+
+a {
+  color: var(--color-link-default);
+}
+
+#error-message {
+  margin-bottom: -1.2rem;
```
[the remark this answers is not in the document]
jamigibbs: Yes. The privacy agreement component needed to manage its own error display and I wanted to use the available error mixin stylesheet but the compromise for doing that was that I needed to override the checkbox components margin like this.

packages/web-components/src/components/va-checkbox/va-checkbox.tsx
```diff
@@ -136,8 +136,7 @@ export class VaCheckbox {
           onChange={this.handleChange}
         />
         <label htmlFor="checkbox-element">
-          {label}
-          {required && <span class="required">(*Required)</span>}
+          <span>{label} {required && <span class="required">(*Required)</span>}</span>
```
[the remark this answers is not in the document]
jamigibbs: This was necessary to align the label along the right edge of the checkbox instead of the text wrapping as described [here](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/991#issuecomment-1240758320). Flexbox is being used on the `label` element to solve for that and in doing so, each of the elements  display as columns (3 columns originally):

Screen Shot 2022-09-12 at 3 01 11 PM

Wrapping the`{label}` and "required" together with a span made sure that the "required" text doesn't go into its own column and positions awkwardly to the right (like shown above).

Screen Shot 2022-09-12 at 3 01 47 PM

packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.tsx
jamigibbs: Gotcha. I'll just make sure the link is updated to `/privacy-policy` then. 👍

packages/web-components/src/components/va-checkbox/va-checkbox.tsx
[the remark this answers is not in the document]
bkjohnson: Since these changes aren't strictly related to the privacy agreement component, could we move the wrapping fixes to their own PR? I think that will help with release notes and searches in the future.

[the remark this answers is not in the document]
jamigibbs: @bkjohnson I suppose I could but I did document this update [in the development ticket](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/991#issuecomment-1240758320) and ultimately decided to add it here so that it could pass design review.

packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.tsx
```diff
@@ -0,0 +1,89 @@
...
+            </Fragment>
+          )}
+        </span>
+        <va-checkbox
+          required
+          id="checkbox"
+          label="I have read and accept the privacy policy"
+          checked={this.checked}
+          onVaChange={this.handleCheckboxChange}
+        />
```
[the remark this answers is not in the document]
jamigibbs: I did not use the checkbox's error prop because there were some differences with the error styling that I couldn't overcome. I believe it was that the checkbox label is bold and because of the shadow dom, I could not override it:

**Checkbox error**

Screen Shot 2022-09-13 at 12 48 31 PM

**Privacy error**

Screen Shot 2022-09-13 at 12 49 14 PM

packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.tsx
k80bowman: Yep, this is the link to use. Thank you for updating it.

packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.tsx
```diff
@@ -0,0 +1,89 @@
+/* eslint-disable i18next/no-literal-string */
```
bkjohnson, line 1: Also, can we remove this? Letting the warnings appear helps us know what will need translation.

packages/web-components/src/components/va-checkbox/va-checkbox.tsx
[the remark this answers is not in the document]
bkjohnson: I don't think that having a separate PR would result in a change to the design review - the fix PR could be merged into `main` and we could pull it into this one.

Here's an example of why I think that could be good. If we look at the git history for a component we will see all the commits to it:

Git history of the table web component

Having a fix to an existing component in the same PR (and eventually squashed commit) where we add a new component muddies things a bit. If we were to look at the history for `va-checkbox` we wouldn't see a commit like "Fix styles when label text gets wrapped" - we would just see "Privacy Agreement Web Component" and would have to dig a bit deeper to realize that there was more going on.

That will also make the git blame more specific. If I look at the blame while on this PR branch, it gives me a good idea that the span was added for a label thing:

Git blame for the PR branch. The line with the span is highlighted and the blame reads "Fix va-checkbox label allignment"

If we were to merge this PR as-is, then the squash would result in the blame showing the commit message for adding the privacy agreement web component, so again we would have to dig a bit deeper to understand the real reason behind the change. And since this fix is for a visual bug that only happens in very specific situations, if someone were to remove this wrapping span in the future as part of a refactor for simplifying the markup or something like that, they wouldn't know that they would be causing a regression unless they happened to be testing with a really long label or happened to dig deep enough to realize that the PR that added the privacy agreement component was also fixing a bug in the checkbox component.

I think it's ok to link the PRs together and make it clear that one resulted from a bug during work on the other, but having a more specific git history seems like the kind of thing that could make our jobs easier in the future when we're looking back to get an idea of what was done and why.

packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.tsx
```diff
@@ -0,0 +1,89 @@
...
+      });
+    }
+  }
+
+  render() {
+    return (
+      <Host>
+        <span>
+          Please read and accept the&nbsp;
+          <a target="_blank" href="/privacy-policy/">
```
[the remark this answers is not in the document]
jamigibbs: I just checked some online references about this and I think you're right! I'll confirm though in channel. Good catch.

packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.tsx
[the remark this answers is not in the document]
bkjohnson: @GnatalieH should our checkbox labels be bolded when there is an error? I would imagine that we want both of these to be consistent.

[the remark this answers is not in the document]
bkjohnson: Related issue: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1115

packages/web-components/src/components/va-checkbox/va-checkbox.tsx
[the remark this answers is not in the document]
jamigibbs: 👍  That commit is reverted and a separate PR is up: https://github.com/department-of-veterans-affairs/component-library/pull/530

packages/web-components/src/mixins/form-field-error.css
```diff
@@ -4,14 +4,16 @@
...
+:host([showerror="true"]) {
     border-left: 0.4rem solid  var(--color-secondary-dark);
     padding-left: 2rem;
     position: relative;
 }
 
 @media screen and (min-width: 1008px) {
-    :host([error]:not([error=''])) {
+    :host([error]:not([error=''])), 
+    :host([showerror="true"]) {
```
bkjohnson, line 16: Now that we're using the `error` prop on the checkbox component, do we still need these changes?

jamigibbs: Yes, we still need that to display error styling:

Screen Shot 2022-09-15 at 8 42 28 AM

jamigibbs:
@bkjohnson @k80bowman Just a note that this PR should not be approved until the `va-checkbox` PR is merged into it: https://github.com/department-of-veterans-affairs/component-library/pull/532

packages/web-components/src/mixins/form-field-error.css
bkjohnson: If we force the `error` prop to [reflect into the DOM](https://stenciljs.com/docs/properties#reflect-properties-values-to-attributes-reflect) like this:

```diff
diff --git a/packages/web-components/src/components/va-checkbox/va-checkbox.tsx b/packages/web-components/src/components/va-checkbox/va-checkbox.tsx
index c856c76..ccc8c60 100644
--- a/packages/web-components/src/components/va-checkbox/va-checkbox.tsx
+++ b/packages/web-components/src/components/va-checkbox/va-checkbox.tsx
@@ -32,7 +32,7 @@ export class VaCheckbox {
   /**
    * The error message to render.
    */
-  @Prop() error?: string;
+  @Prop({ reflect: true }) error?: string;
 
   /**
    * The description to render. If this prop exists, va-checkbox will render it
```
Then it seems to look like we would expect without these CSS changes:

The browser's inspector tab is opened on a local storybook instance for the Privacy Agreement web component. The `error` prop is seen on the checkbox component, and the error styles apply appropriately.

jamigibbs: Ok so do you want another PR to update the `va-checkbox` prop? Or are you fine with that being in this PR?

bkjohnson: I think that's fine to include in this PR since it's not a fix for anything independent of this PR.

packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.css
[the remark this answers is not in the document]
bkjohnson: Now that things have been refactored so that the privacy agreement is just a special use of the checkbox component, do we need to style the margins any differently? [This comment](https://github.com/department-of-veterans-affairs/component-library/pull/532/files#r971258403) on the other PR makes me think that the margins are fine.

bkjohnson (commented):
Adding a top-level comment to make the other comment appear at the bottom. I'm wondering if the refactoring means that we can simplify the CSS changes in the other related PR.

jamigibbs:
https://github.com/department-of-veterans-affairs/component-library/pull/524#discussion_r973202341

@bkjohnson Sorry, I can't nest a response to your top-level comment: 

> Now that things have been refactored so that the privacy agreement is just a special use of the checkbox component, do we need to style the margins any differently? [This comment](https://github.com/department-of-veterans-affairs/component-library/pull/532/files#r971258403) on the other PR makes me think that the margins are fine.

This will likely be removed but I want to get [the other PR merged](https://github.com/department-of-veterans-affairs/component-library/pull/532) into this one before I remove it.

jamigibbs:
@bkjohnson Now that the spacing PR is merged https://github.com/department-of-veterans-affairs/component-library/pull/532, I think it's clearer why I had adjusted the error message margin that was pointed out here https://github.com/department-of-veterans-affairs/component-library/pull/532#discussion_r971226675

With `<span id="error-message" role="alert">` having the bottom-margin `#error-message { margin-bottom: 1.2rem }`, it creates too large of a space when there is no error message present:

Screen Shot 2022-09-19 at 9 03 05 AM

This is the [sketch design](https://www.sketch.com/s/610156b6-f281-4497-81f3-64454fc72156/a/l13oDA7/#Inspect) that we're trying to achieve:


Screen Shot 2022-09-19 at 9 01 44 AM


If we adjust the `va-checkbox` margin to be applied to the input label instead of the error message element, this should be resolved for both default and error states:
 
```
#error-message {
  margin-bottom: 0;
}

:host([error]:not([error=""])) input + label {
  margin-top: 1.2rem;
}
```
The ` margin-bottom: 0;` is applied to override the `/mixins/form-field-error.css` general styles.

The result of that update in `va-checkbox` matches the sketch designs for both [va-privacy-agreement](https://www.sketch.com/s/610156b6-f281-4497-81f3-64454fc72156/a/l13oDA7/#Inspect) and [va-checkbox](https://www.sketch.com/s/610156b6-f281-4497-81f3-64454fc72156/symbols?g=Form%252Fcheckbox):

**va-privacy-agreement: Default**
Screen Shot 2022-09-19 at 9 09 10 AM

**va-privacy-agreement: Error**
Screen Shot 2022-09-19 at 9 09 12 AM

**va-checkbox: With Description**
Screen Shot 2022-09-19 at 9 11 55 AM


**va-checkbox: Error**
Screen Shot 2022-09-19 at 9 11 59 AM

bkjohnson (commented):
Regarding the margin changes - I didn't see a checkbox with a "description" example in Sketch, so is the goal to make changes just for the privacy agreement checkbox but not checkboxes in general? Or do we want to change the margin for all checkboxes?

jamigibbs:
> Regarding the margin changes - I didn't see a checkbox with a "description" example in Sketch, so is the goal to make changes just for the privacy agreement checkbox but not checkboxes in general? Or do we want to change the margin for all checkboxes?

@bkjohnson I think the goal is to make it so that the `va-checkbox` spacing is consistent in any scenario it's being used. I just observed that there is a "phantom" margin applied to the `#error-message` element even if there isn't any error message displayed. It made more sense to me to have a margin applied to an element that's actually visible (the label) instead of the error message that may or may not be displayed.

Screen Shot 2022-09-19 at 4 07 17 PM

packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.css
```diff
@@ -0,0 +1,29 @@
+@import '../../mixins/accessibility.css';
+@import '../../mixins/form-field-error.css';
```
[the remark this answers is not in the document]
jamigibbs: Darn, I thought I removed that already. Thank you!

bkjohnson (commented):
The proposed margin changes seem fine.

bkjohnson (approved):

PR #526, opened 2022-09-07
Add translation custom doc tags

bkjohnson:
## Chromatic

https://component-translation-indicators--60f9b557105290003b387cd5.chromatic.com

## Description
Part of https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1101

This adds JSDoc tags to web components which have i18n support that indicate which languages have a translation available. To make sure I didn't forget to add the docs to any files with i18n support, I gathered the list of components that would need the tags with this command:

` rg "i18next\.t" -l src/components/`

Output of the ripgrep command. There are 6 files which call the `i18next.t` function. Number input, Accordion, Text input, Textarea, Select, and On this page.

This approach doesn't address the `<Date>` React component which has Spanish translations available, but the hope is that that will be getting removed soon.

## Testing done

Local testing in `vets-design-system-documentation` repo.

## Screenshots

Screenshot of the documentation site locally with a new label at the top of the "On this page" component page. The label has a globe icon and reads "English, Spanish, Tagalog".

## Acceptance criteria
- [ ]

## Definition of done


packages/web-components/src/components/va-on-this-page/va-on-this-page.tsx
```diff
@@ -25,6 +25,9 @@ if (Build.isTesting) {
  * @componentName On this page
  * @maturityCategory use
  * @maturityLevel best_practice
+ * @translations English
+ * @translations Spanish
+ * @translations Tagalog
```
bkjohnson, line 30: Right now this is the only component with Tagalog translations.

packages/web-components/src/components/va-accordion/va-accordion.tsx
```diff
@@ -42,6 +42,8 @@ if (Build.isTesting) {
  * @accordionItemToggled This event is fired when an accordion item is opened or closed
  * @maturityCategory use
  * @maturityLevel best_practice
+ * @translations English
+ * @translations Spanish
```
bkjohnson, line 46: How do we want to provide these language indicators? Do we want to directly provide an English user-facing string like this? Do we want to provide the name of the language in the language itself (i.e. "Español")? Or do we want to provide the language code (`es`, `en`) and leave the presentation up to whatever consumes this data?

Using the language codes might be best for consistency since it could be easy to do this in one file:

```js
/**
 * @translations English
 * @translations Spanish
 */
```

And this in another:

```js
/**
 * @translations English
 * @translations Español
 */
```

bkjohnson (commented):
Another question I have is the tag name itself. Is `translations` the best option or is there something else that would be clearer?

packages/web-components/src/components/va-accordion/va-accordion.tsx
k80bowman: I don't have a clear preference on this. All of our documentation is currently in English and I don't see that changing any time soon, so I'm not sure it would be useful to provide the name of the language in the language itself, as English-speakers may not know what that means. Whatever option you feel is best here is fine with me.

k80bowman (approved):

packages/web-components/src/components/va-accordion/va-accordion.tsx
bkjohnson: I'll go with this for now and if we want to change it later we can.

PR #533, opened 2022-09-14
Fix month validation

bkjohnson:
## Chromatic

https://month-date-validation-fix--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1173

This updates the date validation so that if the day is invalid and we don't have a month value, we override the invalid day message with the invalid month message.

The valid day range was being calculated based on the month, since for some months 31 will be valid and for others it won't.

## Testing done

- Storybook :eyes: 

## Screenshots

The memorable date component on storybook has its day filled out but not its month. There is an error message saying "Please enter a month between 1 and 12".

## Acceptance criteria
- [ ]

### DIFF
--- a/packages/web-components/src/utils/date-utils.ts
+++ b/packages/web-components/src/utils/date-utils.ts
@@ -158,7 +158,8 @@ export function validate(
   const leapYear = checkLeapYear(year);
   const daysForSelectedMonth = leapYear && month == 2 ? 29 : days[month]?.length || 0;
 
-  // Begin built-in validation
+  // Begin built-in validation.
+  // Empty fields are acceptable unless the component is marked as required
   if (year && (year < minYear || year > maxYear)) {
     component.invalidYear = true;
     component.error = `Please enter a year between ${minYear} and ${maxYear}`;
@@ -177,6 +178,8 @@ export function validate(
     component.invalidDay = false;
   }
 
+  // The month error message will trigger if the month is outside of the acceptable range,
+  // but also if the day is invalid and there isn't a month value.
   if ((month && (month < minMonths || month > maxMonths)) ||
       (!month && component.invalidDay)) {
     component.invalidMonth = true;
--- a/packages/web-components/src/utils/date-utils.ts
+++ b/packages/web-components/src/utils/date-utils.ts
@@ -158,14 +158,10 @@ export function validate(
   const leapYear = checkLeapYear(year);
   const daysForSelectedMonth = leapYear && month == 2 ? 29 : days[month]?.length || 0;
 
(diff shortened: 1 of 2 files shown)


packages/web-components/src/utils/date-utils.ts
```diff
@@ -171,15 +175,22 @@ export function validate(
...
   if (!monthYearOnly && (day < minMonths || day > daysForSelectedMonth)) {
     component.invalidDay = true;
-    component.error = `Please enter a day between ${minMonths} and ${daysForSelectedMonth}`;
+    component.error = dayError;
+
+    // We need a valid month in order to determine the day
+    if (!month || month < minMonths || month > maxMonths) {
+      component.invalidMonth = true;
+      component.error = monthError;
+    }
```
bkjohnson, line 184: Having this extra check inside the day check is the main change here. The other changes are refactoring to use shared consts.

packages/web-components/src/utils/date-utils.ts
```diff
@@ -177,7 +177,8 @@ export function validate(
     component.invalidDay = false;
   }
 
-  if (month && (month < minMonths || month > maxMonths)) {
+  if ((month && (month < minMonths || month > maxMonths)) ||
+      (!month && component.invalidDay)) {
```
bkjohnson, line 181: Override the invalid day message if we don't have a month.

k80bowman: Could you please add some tests around the logic in this file? Right now it looks like we only have a test for the leap year logic.

packages/web-components/src/utils/date-utils.ts
```diff
@@ -177,7 +177,8 @@ export function validate(
     component.invalidDay = false;
   }
 
-  if (month && (month < minMonths || month > maxMonths)) {
+  if ((month && (month < minMonths || month > maxMonths)) ||
```
jamigibbs, line 180: Is it possible to add some clarifying comments like the other if statements have in this function?

packages/web-components/src/utils/date-utils.ts
bkjohnson: Tests have been added.

packages/web-components/src/utils/date-utils.ts
bkjohnson: I added some comments which should help - let me know what you think.

k80bowman (approved):

jamigibbs (approved):

PR #535, opened 2022-09-19
Add radio hint prop & tweak aria label

Mottie:
## Chromatic

https://1204-radio-hint--60f9b557105290003b387cd5.chromatic.com?path=/docs/components-va-radio--default

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1204

## Testing done

Updated `va-radio` and `va-radio-option` e2e tests

## Screenshots

Hint text - added via `va-radio` `hint` prop
Screenshot of the hint text story for the radio component. It is slightly gray.

## Acceptance criteria
- [x] Add `va-radio` hint text
- [x] Fix fallback ID and label `for` reference to ID - this was out-of-scope for this PR, but a necessary fix
- [x] All radio e2e tests passing

## Definition of done
- [x] Documentation has been updated, if applicable (storybook)


packages/storybook/stories/va-radio.stories.jsx
```diff
@@ -80,9 +84,35 @@ const IdUsageTemplate = ({
...
+        <div>Option two longer description</div>
+      </va-radio-option>
+    </va-radio>
+  );
+};
+
 const defaultArgs = {
   'enable-analytics': false,
   'label': 'This is a label',
+  'hint': '',
```
jamigibbs, line 114: The Storybook page is displaying `hint` in a strange place above the component properties. 

Screen Shot 2022-09-21 at 9 57 00 AM

Is this because you're passing it as a default argument that isn't mapping to any actual property in the component? Or is Storybook picking up the named slot?

Mottie: I'm not sure how to fix this... the same thing happens with the [accordion headline entry](https://design.va.gov/storybook/?path=/docs/components-va-accordion--default)

packages/storybook/stories/va-radio.stories.jsx
```diff
@@ -93,6 +123,26 @@ Default.args = {
...
+    </div>
+  ),
+}
+
+export const Description = OptionDescriptionTemplate.bind(null);
+
+export const Tile = Template.bind(null);
+Tile.args = {
+  ...defaultArgs,
+  tile: true,
```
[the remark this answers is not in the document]
Mottie: Oh, good call... the tile style doesn't include the description text 😱

Mottie:
> Was there any other testing done for these updates besides Storybook development and e2e tests? Specifically I'm wondering if we've made sure to check cross-browser, mobile/responsive, or any a11y testing? If that was done, do you mind updating the description?

I didn't do any cross-browser testing, but I did manually check the storybook canvas to see how the radio changed at different sizes. As for a11y testing, I didn't actually run a screenreader, but I did:
- Copy the same hint text patterns used by the `va-text-input`
- Included the option description text _within_ the label, same as the USWDS example.

Maybe we can bug @joshkimux to take a peek at the chromatic link (in the first comment) to see if he finds any issues?

packages/storybook/stories/va-radio.stories.jsx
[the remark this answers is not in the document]
Mottie: 3 radio options: option one, two and three with the second one selected and including a long lorem ipsum line of text

packages/web-components/src/components/va-radio-option/va-radio-option.tsx
```diff
@@ -56,19 +56,24 @@ export class VaRadioOption {
 
   render() {
     const { checked, ariaDescribedby, name, value, label } = this;
+    const id = this.el.id || (name + value);
     return (
       <Host
         aria-checked={checked ? `${checked}` : 'false'}
         aria-describedby={(checked && ariaDescribedby) || null}
-        aria-label={name}
```
Mottie, line 63: Removed this `aria-label` as it prevented the actual content in the label & description from being read out by a screen reader. Thanks @joshkimux!

packages/web-components/src/components/va-radio-option/va-radio-option.tsx
```diff
@@ -56,19 +56,24 @@ export class VaRadioOption {
...
         onClick={() => this.handleChange()}
         role="radio"
         value={value}
-        id={this.el.id || label}
+        id={id}
       >
-        <label htmlFor={label}>{label}</label>
+        <div class="tile">
+          <label htmlFor={id} aria-describedby="description">{label}</label>
+          <div id="description">
```
Mottie, line 73: Added the `aria-describedby` pointing to the description per @joshkimux recommendations. Moving the slot inside the label may cause issues if a `va-additional-info` component is used.

packages/web-components/src/components/va-radio/va-radio.tsx
```diff
@@ -177,12 +183,18 @@ export class VaRadio {
 
   render() {
     const { label, required, error } = this;
+    const ariaLabel = label + (required ? ' required' : '');
     return (
-      <Host role="radiogroup" aria-invalid={error ? 'true' : 'false'}>
+      <Host
+        role="radiogroup"
+        aria-invalid={error ? 'true' : 'false'}
+        aria-label={ariaLabel}
```
Mottie, line 191: Added this `aria-label` so that the screen reader includes the content in the legend

packages/storybook/stories/va-radio.stories.jsx
jamigibbs: @bkjohnson I guess this is getting picked up because of the named slot. Is that expected? Are we ok with it displaying like that?

Mottie: [`va-select` has the same pattern](https://design.va.gov/storybook/?path=/docs/components-va-select--default)... the options added to the default args are showing up

bkjohnson: Yeah, Robin is right about the `defaultArgs`. I wouldn't say that it's an ideal situation.

packages/web-components/src/components/va-radio-option/va-radio-option.tsx
```diff
@@ -56,19 +56,24 @@ export class VaRadioOption {
...
+            <slot></slot>
```
bkjohnson, line 74: If we just want the "description" to be text, then I think we should make it be a prop instead of a slot.

packages/web-components/src/utils/utils.ts
```diff
@@ -30,9 +30,11 @@ export function getSlottedNodes(
       (node: any) => node.tagName === 'SLOT',
     ).length > 0;
 
-  // This will only get the first slot on a component
+  // This will check all the slots on a component
   const children = hasShadowDOM
-    ? root.shadowRoot.querySelector('slot').assignedNodes()
+    ? Array.from(root.shadowRoot.querySelectorAll('slot')).map(
```
bkjohnson, line 35: Why are these changes happening?

bkjohnson (commented):
I'm a bit confused about both "hint" and "description" options being added. Based on the council discission it seemed like the need was just to have hint text or a description on an individual radio option. Can you explain a bit more?

packages/web-components/src/components/va-radio-option/va-radio-option.tsx
Mottie: The "hint" is added below the `va-radio` legend whereas the "description" is added inside the option text.

As for the description, I wasn't sure if there'd ever be a use case for adding a web component, like a `va-additional-info` inside an option. If you don't think it's necessary, then I can change this to a prop.

packages/web-components/src/utils/utils.ts
Mottie: The `va-radio` already includes a slot for the `va-radio-option`. When I added the "hint" slot, this util would only return the first (hint) slot, so all the `va-radio-options` did not get returned and thus the arrow keys to move between options did not work.

This util was updated to find _all slots_ and search through them for the matching `nodeName`

packages/web-components/src/components/va-radio-option/va-radio-option.tsx
Mottie: My original question was about adding hint text to the radio **group**. In Matt's response, he said "I don't see why not", but then shared a USWDS screenshot that included the tile variation AND description.

Mottie:
```
Group Label
Group Hint text

⚪ Option one
   Description for option one
⚪ Option two
   Description for option two
```

packages/web-components/src/utils/utils.ts
Mottie: Fixed

packages/web-components/src/components/va-radio-option/va-radio-option.tsx
bkjohnson: I think there was some miscommunication where we thought that the tile + description would solve your problem. If that's not the case, can you remove those changes so they can be added later? I think having a smaller PR would help us understand exactly what is being changed and why.

bkjohnson (commented):
Before merging this I think we should settle on some consistent language between components. In `<va-text-input>` we have "hint text" as a regular slot:

Screenshot of the "text input with hint text" story in storybook. The markup has an additional info component as the hint text.

For `<va-checkbox>` we have a `description` slot/prop depending on if you just want a string or HTML elements:

Screenshot of two back to back stories with hint text - one where it is passed as the `description` prop and another where a `<p>` tag is passed in the description slot.

And this PR would add a `hint` slot. I think having that many interfaces for something that's meant to solve the same kind of problem is confusing and can be hard to maintain.

---

Additionally, @humancompanion-usds has talked about hint text having specific styling like we see [on this form documentation page](https://design.va.gov/components/form/#hint-text):

An example of hint text in a light gray color

This color does not meet color contrast goals which we need to address, but it also makes me think that hint text elsewhere that's a plain string should be styled in the same way.

humancompanion-usds:
I want the Tile variation in for other teams that need it. Please keep that and resolve the hint text issues as well. Thank you.

k80bowman:
@bkjohnson do you have a suggestion for how you'd like the hint text/slot variations to be resolved? I'd like to be able to give @Mottie a concrete recommendation for how to implement that in this PR. And as for the hint text color issues, I feel like that should be resolved in a separate issue, since that would affect other components.

Mottie:
I'm starting 2 weeks of PTO tomorrow morning. So we can either wait until I get back or I'd appreciate it if someone would take over the work in this PR.

bkjohnson:
We talked about this in standup this morning and the consensus was:

- Move the tile/description changes out
    - I spoke with Matt about this and he's ok with it
    - We'll add that functionality in its own PR
- Switch the `hint` slot to a `hint` _prop_
- Style the hint text to match the example in our documentation

We will also write another ticket to audit and improve the consistency of hint text/description values.

While you're gone I'll try to make those changes.

k80bowman:
I created a [ticket](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1249) for improving hint text consistency and fixing the color contrast issues

packages/web-components/src/components/va-radio/va-radio.tsx
bkjohnson: @joshkimux do we want to overwrite a potential `aria-label` that a developer added, like:

```jsx
<va-radio
  required
  hint="Some hint text"
  label="Disability type"
  aria-label="Disability type selection" >
   //...
</va-radio>
```

joshkimux: Hi @bkjohnson , sorry for coming back to this so late 🙏 
I'm struggling to follow this a little bit as a non-technical person. From what I (think I'm) reading...
- (assumption) [We can't guarantee that the `legend` will be the first child of the fieldset](https://www.accessibility-developer-guide.com/examples/forms/grouping-with-fieldset-legend/#legend-must-be-a-direct-child-of-fieldset) which results in the group not being uniquely labelled
- So Robin's [using an `aria-label` here with a `radiogroup` role to manually label the field](https://developer.mozilla.org/en-US/docs/web/accessibility/aria/roles/radiogroup_role#examples)
- And you're pointing out that a developer _may have already added_ an `aria-label` to the radio group, in which case the `aria-label` Robin is proposing may overwrite the developer's `aria-label`

It's been a while since I looked at this, so my immediate thoughts are:
- do we have any examples of when a developer has already added an `aria-label`? Are these referring to short-term manual fixes made to get around the `legend` not being a first child?
- did we explore perhaps using `aria-labelledby` instead where we reuse the contents of the `legend` to label the `fieldset`? That way if any changes is made to the visible label (including the required string), the aural label will stay consistent without a separate manual entry

This is a weakpoint for me, but maybe someone like @davidakennedy could help here!

Mottie: I think what's being missed in this conversation is that the `aria-label` wasn't previously defined, so the developer could add one if needed. In this work, the `aria-label` is being defined in the line above:

```js
const ariaLabel = label + (required ? ' required' : '');
```

it's taking the existing `label` text (which ends up in the `legend`) and adding `required` to the end if it's required. Brooks is asking if we should be taking away the ability for the developer to add an `aria-label`

bkjohnson: Robin is right. Do we want these `aria-label` changes to force the `aria-label` to be something like "school required", or do we only want that to be the default which can still be overridden by developers who use the component?

To answer Josh's first question, I couldn't find any examples where a developer was setting `aria-label` on the `<va-radio>` component.

Also, I'm sorry if I dragged you in unnecessarily - I read [this comment](https://github.com/department-of-veterans-affairs/component-library/pull/535#discussion_r977842267) and assumed that the `aria-label` changes were part of a request in a slack DM or something like that :upside_down_face:

joshkimux: That's a really really tough question 😮 , I can't think of a scenario where a developer would set an `aria-label` on the `<va-radio>` component, and it would definitely be undesired in most cases from an a11y standpoint. 

I'd rather we encourage folks to make changes to the visible label instead of giving options to folks to use what is basically a band aid (`aria` in a nutshell) to make a fix. Will bite my tongue if someone can think of an actual use case where it would be needed, but in most instances I'd hope folks can work within the `label` and the content/design of a page. @davidakennedy keep me honest here in case I'm missing something!

bkjohnson: Sounds good - I'll add in a test to serve as documentation of sorts and then get this merged.

bkjohnson (approved):

packages/web-components/src/components/va-radio/va-radio.tsx
davidakennedy: I agree with @joshkimux here. Let's encourage use of the visible label first. We can always change course if teams being a good use case.

PR #540, opened 2022-09-26
Icon discovery

bkjohnson:
## Chromatic

https://icon-discovery--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/32758

As part of this discovery I've tried a few different approaches:

### Setting the `src` attribute on an `` tag

This can be seen in [the first commit](https://github.com/department-of-veterans-affairs/component-library/pull/540/commits/b1225efeae67a50d96cb80ac7d6a64d3f33e745d). It works, but having an `` makes it not as easy to style the icon to be in different colors.

img tag with src attribute doesn't accept color styling

### Dynamic `import()`

It doesn't look like this will work with SVG files. Here is a diff of local changes made after the second commit:

```diff
diff --git a/packages/web-components/src/components/va-icon/va-icon.tsx b/packages/web-components/src/components/va-icon/va-icon.tsx
index 476b090..9ede06d 100644
--- a/packages/web-components/src/components/va-icon/va-icon.tsx
+++ b/packages/web-components/src/components/va-icon/va-icon.tsx
@@ -1,5 +1,4 @@
 import { Component, Host, Prop, getAssetPath, h } from '@stencil/core';
-import ascendingIcon from '../../assets/sort-arrow-up.svg?format=text';
 
 @Component({
   tag: 'va-icon',
@@ -11,12 +10,13 @@ export class VaIcon {
 
   @Prop() icon!: string;
 
-  render() {
+  async render() {
     const path = getAssetPath(`./assets/${this.icon}.svg`);
-    console.log('PATH', path)
+    const importedIcon = await import(path + "?format=text");
+    console.log('icon', importedIcon)
     return (
       <Host >
-        <div innerHTML={ascendingIcon}></div>
+        <div innerHTML={importedIcon}></div>
     </Host>
     );
   }
```

And here is the output:

Screenshot of some console logs where the svg imports were blocked due to a disallowed MIME type of "image/svg+xml"

This was based off of the static SVG import in the table component:

https://github.com/department-of-veterans-affairs/component-library/blob/ab4a7b6f64c28e610a4c83d1cff5fb652e833dc8/packages/web-components/src/components/va-table/va-table.tsx#L3-L4

### Using a spritesheet for the icons and an inline SVG

This is the way that USWDS uses icons. The markup for icons on their [icon documentation page](https://designsystem.digital.gov/components/icon/) looks like this:

```html
      <svg class="usa-icon" aria-hidden="true" focusable="false" role="img">
        <use xlink:href="/assets/img/sprite.svg#accessibility_new"></use>
      </svg>
```

We can do the same kind of thing, and an inline SVG can be easily styled:

Inline SVG being styled to have a different size and color using CSS

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-icon/va-icon.tsx
+++ b/packages/web-components/src/components/va-icon/va-icon.tsx
@@ -1,5 +1,4 @@
-import { Component, Host, Prop, getAssetPath, h } from '@stencil/core';
-import ascendingIcon from '../../assets/sort-arrow-up.svg?format=text';
+import { Component, Prop, getAssetPath, h } from '@stencil/core';
 
 @Component({
   tag: 'va-icon',
@@ -12,12 +11,12 @@ export class VaIcon {
   @Prop() icon!: string;
 
   render() {
-    const path = getAssetPath(`./assets/${this.icon}.svg`);
+    const path = getAssetPath("./assets/sprite.svg");
     console.log('PATH', path)
     return (
-      <Host >
-        <div innerHTML={ascendingIcon}></div>
-    </Host>
+      <svg class="usa-icon" aria-hidden="true" focusable="false" role="img">
+        <use xlinkHref={`${path}#${this.icon}`}></use>
+      </svg>
     );
   }
 
--- a/packages/web-components/src/components/va-icon/va-icon.tsx
+++ b/packages/web-components/src/components/va-icon/va-icon.tsx
@@ -1,0 +1,22 @@
+import { Component, Prop, getAssetPath, h } from '@stencil/core';
+
+@Component({
+  tag: 'va-icon',
+  styleUrl: 'va-icon.css',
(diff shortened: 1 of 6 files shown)


packages/web-components/src/components/va-icon/va-icon.tsx
```diff
@@ -0,0 +1,23 @@
...
+  styleUrl: 'va-icon.css',
+  // assetsDirs: ["../../../../node_modules/@uswds/uswds"],
+  shadow: true,
+})
+export class VaIcon {
+
+  @Prop() icon!: string;
+
+  render() {
+    const path = getAssetPath("./assets/sprite.svg");
```
bkjohnson, line 14: The stencil copy task will move this file into the build output.

micahchiang:

PR #556, opened 2022-10-19
Add SMS option to va-telephone component

jamigibbs:
## Chromatic

https://1246-sms-option-va-telephone--60f9b557105290003b387cd5.chromatic.com

## Description
This PR adds SMS as a telephone display option to va-telephone in the following format:

**Text/SMS**
`<a href="sms:838255" aria-label="8 3 8 3 5 5.">838255</a>`

Reference this comment for further clarification on the display: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1246#issuecomment-1298566047

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1246

## Testing done
Storybook
Browserstack

## Screenshots
Screen Shot 2022-11-01 at 2 13 19 PM

## Acceptance criteria
- [ ] The SMS option is added to va-telephone and displays as described above.

## Definition of done


packages/web-components/src/components/va-telephone/va-telephone.tsx
```diff
@@ -46,6 +46,12 @@ export class VaTelephone {
    */
   @Prop() tty?: boolean = false;
 
+  /**
+   * Indicates if this is a number meant to be used 
+   * to text.
+   */
+  @Prop() sms?: boolean = false;
```
bkjohnson, line 53: Is there a reason why we would go with a prop for this instead of going into "SMS mode" if the contact length is 5 or 6 digits?

packages/web-components/src/components/va-telephone/test/va-telephone.spec.tsx
```diff
@@ -32,52 +33,90 @@ describe('formatPhoneNumber', () => {
...
+  it('does not create a phone number label if SMS and TTY are false', () => {
+    expect(VaTelephone.createTelLabel(false, false)).toBe(
+      '',
+    );
+  });
+
+  it('does not create a phone number label if SMS and TTY are true', () => {
+    expect(VaTelephone.createTelLabel(true, true)).toBe(
+      '',
+    );
```
bkjohnson, line 99: This is a bit confusing to me, because reading the first test made me think that the next test would be that it _does_ create a phone number label if both of those args are true. I now see that it's about mutual exclusivity - the label is only created if one of the values is true, but not both. Can you make a change to help clear this up? I think even just adding a comment would work.

packages/web-components/src/components/va-telephone/va-telephone.tsx
```diff
@@ -94,7 +101,10 @@ export class VaTelephone {
       }
     }
     if (tty) {
-      formattedNum = `TTY: ${formattedNum}`;
+      formattedNum = `${formattedNum}`;
+    }
+    if (sms) {
+      formattedNum = `${formattedNum}`;
```
bkjohnson, line 107: Since these are doing the same thing do we need to separate them?

packages/web-components/src/components/va-telephone/va-telephone.tsx
```diff
@@ -124,15 +137,33 @@ export class VaTelephone {
...
     // but it seems that using a comma to pause for 2 seconds might be a better
     // solution - see https://dsva.slack.com/archives/C8E985R32/p1589814301103200
-    const href = `tel:${isN11 ? contact : `+1${contact}`}`;
+    let href = `tel:+1${contact}`;
+    if (isN11) {
+      href = `tel:${contact}`;
+    } 
+    if (sms) {
+      href = `sms:${contact}`;
+    }
```
bkjohnson, line 151: I think this is a bit clearer and it has the benefit of keeping the SMS and N11 functionality separate, since I don't think you can text those kinds of numbers.

```suggestion
    let href = null;
    if (sms) {
      href = `sms:${contact}`;
    }
    else if (isN11) {
      href = `tel:${contact}`;
    }
    else {
      href = `tel:+1${contact}`;
    }
```

packages/web-components/src/components/va-telephone/va-telephone.tsx
```diff
@@ -124,15 +137,33 @@ export class VaTelephone {
...
+  static createTelLabel(tty: boolean, sms: boolean): string {
+    let label = '';
+    if (tty) {
+      label = 'TTY: ';
+    }
+    if (sms) {
+      label = 'Text: ';
+    }
+    // If both tty & sms are true, we don't want to display a label.
+    return tty && sms ? '' : label;
```
bkjohnson, line 164: Using the if statements like this makes it less obvious that this is supposed to be mutually exclusive behavior. The return is a bit confusing as well since `label` could be an empty string, so we would end up with:

```js
    // If both tty & sms are true, we don't want to display a label.
    return tty && sms ? '' : '';
```

Can this be changed so that the output of the function is more obvious?

packages/web-components/src/components/va-telephone/va-telephone.tsx
```diff
@@ -145,34 +176,41 @@ export class VaTelephone {
...
     ) : (
-      <a
-        href={VaTelephone.createHref(contact, extension)}
-        aria-label={formattedAriaLabel}
-        onClick={this.handleClick.bind(this)}
-      >
-        {formattedNumber}
-      </a>
+      <Fragment>
+        {telLabel ? <span>{telLabel}</span> : ''}
```
bkjohnson, line 205: I looked at the original issue and it was unclear to me if the feedback was just to add "Text" in front of the number like we do for TTY, or if we should also make "TTY" and "Text" _not_ part of the clickable link. Is this specified somewhere? For background, in May [Randi made "TTY" part of the clickable text in the content style guide](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/pull/800).

The phone numbers section of the content styleguide. The markup for a TTY number is highlighted, with an extra highlight showing that the "TTY" preceding text is inside of the clickable link

jamigibbs: @bkjohnson You can review the discussion about this update in the [Slack thread here](https://dsva.slack.com/archives/G01BJ3ESXL4/p1666708186077849?thread_ts=1666707265.753579&cid=G01BJ3ESXL4) and this was added to the [original ticket as a comment here](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1246#issuecomment-1290671223). If a follow-up ticket needs to be created to update the documentation about this, then we should let the team know.

bkjohnson: > Here is the way we should format SMS:
> 
> Text: [838255](url)
> 
> And we should update TTY to be consistent with the phone number format and SMS format:
> 
> TTY: [711](url)

I read the comment, but I'm saying that without a clear statement like "the type of link needs to be its own element so that it isn't clickable" the hyperlink formatting there could be an accident.

The slack discussion helps a little, but it seems like it's assuming that regular numbers will have "Call:" in front:

>  and do not link Text nor TTY **to be consistent with Call.**

We don't put "Call:" in front of the numbers though. The reason "TTY:" was added is because it is typically paired with a non-tty number, [like this](https://github.com/department-of-veterans-affairs/vets-website/blob/f62cb89618190a6eda0230092b1b782aa86ba29b/src/applications/appeals/995/content/GetFormHelp.jsx#L8-L11).

Screenshot of a TTY number appearing in parentheses after a number meant for a hearing or speech-enabled user.

humancompanion-usds: Thanks for that context @bkjohnson . I should have asked for more examples in the Slack thread. I suspect removing "TTY:" from the hyperlink would make that tap target too small and more difficult to hit. Sometimes we say "call TTY: 711" within a sentence and there having a larger target to see does seem useful. So let's leave TTY alone.

Also, "TTY:" is more of a label that just happens to be the same as the protocol type. To make matters more complicated, users can now switch a phone call to [RTT](https://support.google.com/accessibility/android/answer/9042284?hl=en) (Real-time text) which runs over TTY. In which case, I'm not sure which number they are supposed to call to do that.

Anyway, back to "text". I don't think "SMS: [number]" makes sense as not all users will know that protocol means send a text. Thus "text" is more like "call" which we don't use in the hyperlink, don't inject, and does not use a colon. Though I can find instances where we violate the TTY rule and where a colon could be used:

Screen Shot 2022-10-28 at 11 34 17 AM

At any rate, apologies for pulling a 180 here but let's go with leaving text out. Folks can add "Text" or "text" ahead of the number just as they do with phone numbers.

packages/web-components/src/components/va-telephone/test/va-telephone.spec.tsx
jamigibbs: After reverting the TTY changes, these tests are no longer relevant and have been removed.

packages/web-components/src/components/va-telephone/va-telephone.tsx
jamigibbs: I have updated the condition logic based on this suggestion.

packages/web-components/src/components/va-telephone/va-telephone.tsx
jamigibbs: This is no longer relevant since the TTY changes have been reverted.

bkjohnson (approved):

PR #557, opened 2022-10-20
Fix screenreader not reading va-modal content in va-omb-info

jamigibbs:
## Chromatic

https://a11y-va-omb-info-staging--60f9b557105290003b387cd5.chromatic.com

## Description
During staging review for va-omb-info, an a11y issue was identified where the `va-modal` component was not allowing screenreaders to read the modal content when embeded within va-omb-info. The modal content worked correctly as a standalone component though.

I identified that the imported NPM package in va-modal, [aria-hidden](https://github.com/theKashey/aria-hidden), was applying the aria-hidden attribute to the parent web component (va-omb-info in this case) and blocking the modal content from being read.

After reviewing the aria-hidden NPM script, I noticed that [the `hideOthers` function will accept an array of elements](https://github.com/theKashey/aria-hidden/blob/master/src/index.ts#L134) to keep on the page. 

The solution proposed here is to add a new prop to va-modal that allows for elements to be skipped from being hidden by the aria-hidden NPM script. The most common usage of this would be when va-modal is used as a child component but it might also be useful in other scenarios that have not been considered yet.

Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/48546

## Testing done
Storybook -VoiceOver
Browserstack - NVDA

## Screenshots

**Without the `ariaHiddenNodeExceptions` prop**

va-omb-info top level parent container being hidden (screenreader does not work for modal)

Screen Shot 2022-10-21 at 12 37 36 PM

**With `ariaHiddenNodeExceptions` receiving the va-omb-info element**

Screenreader works as expected in both va-omb-info and va-modal

```
ariaHiddenNodeExceptions={[this.el]}
```

Screen Shot 2022-10-21 at 12 34 44 PM

## Acceptance criteria
- [ ] Modal content is read by screenreaders in both va-omb-info and va-modal components.

## Definition of done


jamigibbs:
I just noticed that the new prop `ariaHiddenNodeExceptions` is showing up as "undefined" in Storybook. Checking on that.

  
Screen Shot 2022-10-25 at 2 24 28 PM

SarahKay8:
Yep will do! @jamigibbs

packages/web-components/src/components/va-modal/va-modal.tsx
```diff
@@ -132,6 +132,11 @@ export class VaModal {
    */
   @Prop({ reflect: true }) visible?: boolean = false;
 
+  /**
+   * DOM-nodes that should not be hidden from screen readers.
+   */
+  @Prop() ariaHiddenNodeExceptions?: HTMLElement[] = [];
```
jamigibbs, line 138: @bkjohnson I wonder if you had any thoughts about this issue. The component docs created via `stencil build --docs-json component-docs.json` don't generate this as a property for displaying in storybook:

Screen Shot 2022-10-26 at 11 26 14 AM

Because it's an array type, it won't ever generate with an attribute name (because it's a property not an attribute) and therefore won't display correctly in our Storybook for it.

Using this new attribute to solve the screenreader issue seemed like the cleanest approach but I'm sure how to solve for it showing up in Storybook like this.

bkjohnson: Maybe we switch `prop.attr` here to be `prop.name`, or do something like `prop.attr || prop.name`:

https://github.com/department-of-veterans-affairs/component-library/blob/ce3b4b70118991bf2b380aac7f498457427a4cb4/packages/storybook/stories/wc-helpers.jsx#L85

image


I couldn't find any instances where we are [setting the `attribute` inside the `@Prop` decorator](https://stenciljs.com/docs/properties#attribute-name-attribute), so in general it looks like the main difference between `name` and `attr` in the JSON docs is camel casing vs kebab casing:

image

jamigibbs: Yes, good suggestion. I can make that update. I wonder though if there are any implications of having an array prop like that though (seeing as how this is the first time we've needed it).

packages/web-components/src/components/va-modal/va-modal.tsx
```diff
@@ -132,6 +132,11 @@ export class VaModal {
    */
   @Prop({ reflect: true }) visible?: boolean = false;
 
+  /**
+   * DOM-nodes that should not be hidden from screen readers.
```
bkjohnson, line 136: Can we change the documentation so that people don't mistakenly think that they need to do something? We're basically adding this prop for ourselves to use, and _maybe_ there will be situations when someone else needs to use it. Something like this might work, but feel free to wordsmith it however you like:

> Additional DOM-nodes that should not be hidden from screen readers. Useful when an open modal shouldn't hide all content behind the overlay.

packages/web-components/src/components/va-modal/va-modal.tsx
bkjohnson: The array prop will mean that it can't (easily) be used in plain HTML the same way that booleans, strings, and numbers can be used. The `<va-search-input>` component is similar with its `suggestions` array prop (that's why the story uses a react binding).

https://github.com/department-of-veterans-affairs/component-library/blob/ce3b4b70118991bf2b380aac7f498457427a4cb4/packages/storybook/stories/va-search-input.stories.jsx#L98

It's like if instead of just being able to do [this](https://github.com/department-of-veterans-affairs/content-build/blob/6dc80d182b656f43e837d0292ad79be698750a9c/src/site/components/informational_alert.drupal.liquid#L2):

```html
  <va-alert status="info" id="va-info-alert">
    <h3 slot="headline">
      Need help after hours?
    </h3>
    <div>
      {{ entity.fieldWysiwyg.0.processed | drupalToVaPath | phoneLinks}}
    </div>
  </va-alert>
```

We had to have an additional `<script>` that did this if we wanted to set the `status` prop:

```js
const alertEl = document.getElementById('va-info-alert');
alertEl.status = "info";
```

Except for the modal situation we'd have to have an array of DOM nodes instead of strings, which would mean even more JS in that extra script.

bkjohnson (approved):
One small comment about clarity on the prop documentation, but everything else looks good.

PR #577, opened 2022-11-16
Add Storybook style overrides for USWDS v3 components

jamigibbs:
## Chromatic

https://1331-storybook-u-s-w-d-s-style--60f9b557105290003b387cd5.chromatic.com

## Description
This PR includes a script that will dynamically update the `body` and `html` tags in Storybook when the URL contains the value "uswds" as well as some global CSS required to reset Formation values to pixels. 

**Note:** We will need to be mindful of the branch names we're creating as we are developing for USWDS v3. If the branch name has "uswds" in it, the Chromatic link will trigger the dynamic CSS update on all pages.

This branch does not include the work in progress for the USWDS v3 va-radio component so testing for this can be done on the Chromatic link generated from the discovery ticket draft PR (which includes the va-radio USWDS section):
https://github.com/department-of-veterans-affairs/component-library/pull/573

If we feel like this would be easier to validate as part of the [va-radio USWDS v3 work](https://github.com/department-of-veterans-affairs/component-library/pull/563), then I can close this PR and integrate it into https://github.com/department-of-veterans-affairs/component-library/pull/563 for overall testing.

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1331

## Testing done
Storybook.

## Screenshots
Screen Shot 2022-11-16 at 1 23 53 PM
Screen Shot 2022-11-16 at 1 23 42 PM

## Acceptance criteria
- [ ] Dynamic and global base styles are added so that Storybook pages are displaying as expected between Formation and USWDS v3 components.

## Definition of done

### DIFF
--- a/packages/storybook/.storybook/preview.js
+++ b/packages/storybook/.storybook/preview.js
@@ -41,18 +41,16 @@ applyPolyfills().then(() => {
  * The native events `hashchange` and `popstate` were not reliable.
  */
  (() => {
-  const oldPushState = history.pushState;
-  history.pushState = function pushState() {
-    const ret = oldPushState.apply(this, arguments);
+  const pushState = history.pushState;
+  history.pushState = function () {
+    pushState.apply(this, arguments);
     window.dispatchEvent(new Event('locationchange'));
-    return ret;
   };
 
-  const oldReplaceState = history.replaceState;
-  history.replaceState = function replaceState() {
-    const ret = oldReplaceState.apply(this, arguments);
+  const replaceState = history.replaceState;
+  history.replaceState = function () {
+    replaceState.apply(this, arguments);
     window.dispatchEvent(new Event('locationchange'));
-    return ret;
   };
 
   window.addEventListener('popstate', () => {
--- a/packages/storybook/.storybook/preview.js
+++ b/packages/storybook/.storybook/preview.js
@@ -44,7 +44,6 @@ applyPolyfills().then(() => {
   const oldPushState = history.pushState;
   history.pushState = function pushState() {
     const ret = oldPushState.apply(this, arguments);
-    window.dispatchEvent(new Event('pushstate'));
     window.dispatchEvent(new Event('locationchange'));
     return ret;
   };
@@ -52,7 +51,6 @@ applyPolyfills().then(() => {
   const oldReplaceState = history.replaceState;
   history.replaceState = function replaceState() {
     const ret = oldReplaceState.apply(this, arguments);
-    window.dispatchEvent(new Event('replacestate'));
     window.dispatchEvent(new Event('locationchange'));
     return ret;
   };
--- a/packages/storybook/.storybook/preview.js
+++ b/packages/storybook/.storybook/preview.js
@@ -36,6 +36,32 @@ applyPolyfills().then(() => {
   window.CustomEvent = CustomEvent;
 })();
 
+/**
+ * The custom event locationchange will dispatch when the URL changes.
+ * The native events `hashchange` and `popstate` were not reliable.
+ */
+ (() => {
+  const oldPushState = history.pushState;
+  history.pushState = function pushState() {
+    const ret = oldPushState.apply(this, arguments);
+    window.dispatchEvent(new Event('pushstate'));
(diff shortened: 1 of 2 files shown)


packages/storybook/.storybook/style.scss
```diff
@@ -33,4 +33,17 @@ a.sbdocs-a:hover {
...
+.usa-label {
+  font-size: 15px!important;
+}
+
+button {
+  padding: 10px 20px!important;
+}
+
+h1 {
+  font-size: 32px!important;
```
bkjohnson, line 48: Can you link to the USWDS3 styles that we are mimicking here?

packages/storybook/.storybook/preview.js
```diff
@@ -36,6 +36,30 @@ applyPolyfills().then(() => {
...
+    window.dispatchEvent(new Event('locationchange'));
+    return ret;
+  };
+
+  const oldReplaceState = history.replaceState;
+  history.replaceState = function replaceState() {
+    const ret = oldReplaceState.apply(this, arguments);
+    window.dispatchEvent(new Event('locationchange'));
+    return ret;
+  };
```
bkjohnson, line 56: Can you leave a comment to explain what's going on with the `.apply()`? Also, I don't understand why we are returning anything - what is that for?

jamigibbs: The approach here is that we are using the native `pushState` and `replaceState` functions so that we can add our own custom event `locationchange` as part of it. So `apply` is making sure the original context and arguments (if any) are still getting used. This solution is based on [the following here](https://stackoverflow.com/questions/6390341/how-to-detect-if-url-has-changed-after-hash-in-javascript#answer-52809105) which I have simplified to only dispatch `locationchange` instead of the other extra events we don't really need. And it's returning the state into the history as it's just kind of a "passthrough" for the history values.

packages/storybook/.storybook/style.scss
jamigibbs: Well now I don't see any difference when I remove it so I guess I can remove it from here.

But the Formation rem style for H1 is coming from here (`h1: 4rem`):
https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/master/packages/formation/sass/utilities/_font-size.scss#L8

And in USWDS, they seem to be setting it with the type scale token "14" which I think is the equivalent of 40px (?). This is the trail I followed to find that:

1. https://github.com/uswds/uswds/blob/develop/packages/uswds-core/src/styles/mixins/typography/headings.scss#L16-L17

```
@mixin h1 {
  @include typeset("heading", $theme-h1-font-size, $theme-heading-line-height);
  font-weight: fw("bold");
}
```

2. https://github.com/uswds/uswds/blob/develop/packages/uswds-core/src/styles/settings/_settings-typography.scss#L400

```
$theme-h1-font-size: "2xl" !default;
```

3. https://github.com/uswds/uswds/blob/develop/packages/uswds-core/src/styles/variables/type-scale.scss#L14

```
"2xl": system-type-scale(settings.$theme-type-scale-2xl),
```

4. https://github.com/uswds/uswds/blob/develop/packages/uswds-core/src/styles/settings/_settings-typography.scss#L308

```
$theme-type-scale-2xl: 14 !default;
```

4. https://github.com/uswds/uswds/blob/develop/packages/uswds-core/src/styles/settings/_settings-typography.scss#L354

```
14:         40px
```

packages/storybook/.storybook/preview.js
bkjohnson: It looks like [`.replaceState()` isn't meant to return anything](https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState#return_value)

jamigibbs: Oh good catch. I didn't notice that in the docs. I think that we can just simplify this even further then to something like this (same for replaceState too):

```js
const pushState = history.pushState;
history.pushState = function () {
  pushState.apply(this, arguments);
  window.dispatchEvent(new Event('locationchange'));
};
```

bkjohnson (approved):

PR #579, opened 2022-11-21
Add va-radio USWDS v3 version

jamigibbs:
## Chromatic

https://1265-us-wds-radio-button--60f9b557105290003b387cd5.chromatic.com/?path=/docs/uswds-va-radio--default

## Description
Adapt the va-radio component to provide a [USWDS Radio buttons](https://designsystem.digital.gov/components/radio-buttons/) variant.

Sketch: https://www.sketch.com/s/610156b6-f281-4497-81f3-64454fc72156/p/447FB6DA-215E-4F15-A71B-94B7686DA614/canvas

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1265

## Testing done
Storybook

## Screenshot

Screen Shot 2022-11-23 at 11 07 02 AM

Screen Shot 2022-11-23 at 11 07 10 AM

Screen Shot 2022-11-23 at 11 10 39 AM

Screen Shot 2022-11-23 at 11 07 19 AM

Screen Shot 2022-11-23 at 11 07 25 AM

## Acceptance criteria
- [x]  Create a version of this component that imports and uses the USWDS CSS
- [x] Make any structural changes necessary and add a prop that determines which version of the component will be used
- [x] Check with Natalie to make sure the component and the design match
- [x] Add a USWDS story to Storybook
- [ ] Update component-library dependency in vets-design-system-documentation to get the updated component-docs.json

## Definition of done


packages/web-components/src/components/va-radio/va-radio.tsx
```diff
@@ -87,7 +93,8 @@ export class VaRadio {
   @Listen('keydown')
   handleKeyDown(event: KeyboardEvent) {
     const currentNode = event.target as HTMLVaRadioOptionElement;
-    const radioOptionNodes = getSlottedNodes(this.el, 'va-radio-option');
+    const radioOptionNodes = (getSlottedNodes(this.el, 'va-radio-option') as HTMLVaRadioOptionElement[])
+      .filter(node => !node.disabled);
```
jamigibbs, line 97: USWDS v3 provides a "disabled" variant for options and this modification is to handle navigating through the options so that it skips over those that are identified as disabled.

packages/web-components/src/utils/utils.ts
```diff
@@ -21,19 +21,8 @@ export function getSlottedNodes(
...
 ): Array<Node> {
-  // If the browser is using the shadowDOM, the childNodes should be an array of two things:
-  // A `<style>` element and a `<slot>` element
-  // Chrome, Firefox, Safari, Edge - literally every modern browser will use shadowDOM.
-  // This is purely for IE compatibility
-  const hasShadowDOM =
-    Array.from(root.shadowRoot.childNodes).filter(
-      (node: any) => node.tagName === 'SLOT',
-    ).length > 0;
-
```
jamigibbs, line 32: This did not account for a slot _not_ being a direct descent of the root `Host` which was the case for the USWDS markup so it was causing issues for that variant.

```html
<Host>
   <fieldset>
      [....]
    <slot></slot>
  </fieldset>
</Host>
```

Ultimately though, this appears to be a relic for IE11 which is no longer supported so I opted to remove it entirely as tech debt.

packages/web-components/stencil.config.ts
```diff
@@ -1,12 +1,16 @@
...
 import url from 'postcss-url';
 import { reactOutputTarget } from '@stencil/react-output-target';
 
 export const config: Config = {
   namespace: 'component-library',
   globalStyle: 'src/global/main.css',
   plugins: [
+    sass({ 
+      includePaths: ["../../node_modules/@uswds/uswds/packages/"],
+    }),
```
jamigibbs, line 13: Because of the way things changed along the way for this component, I ended up adding the new NPM packages to this PR (@stencil/sass & @uswds/uswds) but if we feel like this and the following related files should be its own PR, I can pull it out.

packages/web-components/package.json
```diff
@@ -40,11 +40,13 @@
     "@department-of-veterans-affairs/formation": "^7.0.1",
     "@stencil/postcss": "^2.0.0",
     "@stencil/react-output-target": "^0.0.9",
+    "@stencil/sass": "^2.0.0",
     "@types/jest": "^26.0.20",
     "@types/puppeteer": "^5.4.3",
     "@types/react-dom": "16.0.9",
     "@typescript-eslint/eslint-plugin": "^5.30.5",
     "@typescript-eslint/parser": "^5.30.5",
+    "@uswds/uswds": "^3.2.0",
```
jamigibbs, line 49: Because of the way things changed along the way for this component, I ended up adding the new NPM packages to this PR  but if we feel like this and the related files (stencil config, packag.json, and yarn.lock) should be its own PR, I can pull it out.

packages/web-components/src/components/va-radio-option/va-radio-option.scss
```diff
@@ -1,15 +1,40 @@
...
+:host([disabled='true']:focus) label.usa-radio__label::before {	
+  outline: none;	
+}
+
+:host([uswds]) .usa-radio__label {
+  line-height: 1.5;
+}
+
+.usa-radio__label::before {
+  box-shadow: rgb(113, 118, 122) 0px 0px 0px 2px;
```
bkjohnson, line 26: Why do we need to do this?

packages/web-components/src/components/va-radio/va-radio.scss
```diff
@@ -0,0 +1,43 @@
...
+@import '../../mixins/uswds-error-border.css';
+
+:host {
+  display: block;
+  border: none;
+  padding: 0;
+}
+
+.usa-hint, .usa-error-message {
+  margin-bottom: 4px;
```
bkjohnson, line 18: Same question here.

packages/web-components/src/components/va-radio/va-radio.tsx
```diff
@@ -203,28 +210,60 @@ export class VaRadio {
...
+    if (uswds) {
+      return ( 
+        <Host
+          aria-invalid={error ? 'true' : 'false'}
+          aria-label={ariaLabel}
+        >
+          <fieldset class="usa-form usa-fieldset">
+            <legend class={legendClass}>
+              {label}
+              {required && <span class="required">{i18next.t('required')}</span>}
```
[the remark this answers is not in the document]
jamigibbs: Why yes, yes indeed. I must not have found it during my search. Additionally, `.required` is using the old red color (lighter) and `usa-label--required` has the updated red (darker):

```
.usa-label--required {
  color: color("error-dark");
}
```
Screen Shot 2022-11-22 at 9 25 12 AM

packages/web-components/src/mixins/uswds-error-border.css
```diff
@@ -0,0 +1,11 @@
+:host([error][uswds]:not([error=''])) {
+    border-left: 0.4rem solid #b50909; /* USWDS red-60v, $theme-color-error-dark */
```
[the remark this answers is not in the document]
jamigibbs: Yes. How do we want to handle that with this PR? When we implement the solution from [1322](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1322), should that work also include updating this file?

packages/web-components/src/components/va-radio-option/va-radio-option.scss
jamigibbs: We have a different radio button border color (Sketch link in the description): 

USWDS v3:
Screen Shot 2022-11-22 at 1 02 54 PM

VA:
Screen Shot 2022-11-22 at 1 02 33 PM

packages/web-components/src/components/va-radio/va-radio.scss
jamigibbs: I didn't see an error message example on USWDS for radio and this matches what's in Sketch:

Screen Shot 2022-11-22 at 1 05 51 PM

packages/web-components/src/mixins/uswds-error-border.css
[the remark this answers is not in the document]
bkjohnson: Yeah, that's what I'm thinking.

packages/web-components/src/components/va-radio-option/va-radio-option.scss
humancompanion-usds: I don't think we should. When we see a discrepancy like this it should be raised with Natalie. Not sure why we'd deviate here.

jamigibbs: Ok I'll remove it. There's also the line height difference too you can see in the screenshot above: https://github.com/department-of-veterans-affairs/component-library/blob/74aa2b344e2596394832f358485abd1f5c76807b/packages/web-components/src/components/va-radio-option/va-radio-option.scss#L21-L23

@GnatalieH The USWDS radio default border color is `#1b1b1b` (sketch has `#71767A`) and the default line-height for the radio item text is `line-height: 1.3` (sketch has 1.5).

jamigibbs: Currently I have the line-height 1.5 (24px) CSS override removed but if we want to add it back, the development burden is pretty low. It's just the CSS below. 

Based on Natalie's reasoning above, do we want to make this an exception @humancompanion-usds? 

```css
 .usa-radio__label { 
   line-height: 1.5; 
 } 
```

GnatalieH: @jamigibbs I checked the Chromatic link and all looks great! The only thing I'd change is "Id" usage to "ID usage". Thanks!

packages/web-components/src/components/va-radio/va-radio.scss
jamigibbs: This style was removed.

packages/web-components/src/components/va-radio-option/va-radio-option.scss
jamigibbs: Thanks @GnatalieH! I think we need to display it as "Id" because of the way Storybook parses the story titles. If it was "ID" it would automatically add a space between the letters so it would say "I D Usage" instead. 😞

jamigibbs: @GnatalieH Disregard my last comment about "ID". It looks like it's smart enough to parse this correctly. I'll update! 👍

packages/storybook/stories/va-radio-uswds.stories.jsx
```diff
@@ -0,0 +1,325 @@
...
+};
+
+const vaRadioConst = args => {
+  const {   
+    'enable-analytics': enableAnalytics,
+    error,
+    label,
+    hint,
+    required,
+    tile,
```
jhonnyoddball, line 36: Do we need title in this example?

packages/web-components/package.json
jhonnyoddball: That's fine with me

packages/storybook/stories/va-radio-uswds.stories.jsx
jamigibbs: Yes, I think I can clean that up in a few places from the story. Thanks!

packages/storybook/stories/va-radio-uswds.stories.jsx
```diff
@@ -0,0 +1,325 @@
...
+const ReactBindingExample = ({
+  'enable-analytics': enableAnalytics,
+  error,
+  label,
+  required,
+}) => {
+  return (
+    <>
+    <VaRadio 
+        uswds
```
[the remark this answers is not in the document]
jamigibbs: There might be a way to make this a default but I wanted the `uswds` prop to display in the examples on Storybook and I didn't think it would display any other way. But I'm not a Storybook expert so if you see a way to set this as a default while still showing in the example, I can definitely update it.

Screen Shot 2022-12-01 at 3 44 37 PM

packages/web-components/src/components/va-radio/va-radio.scss
```diff
@@ -1,12 +1,24 @@
...
-@import '../../mixins/form-field-error.css';
+@use 'uswds-helpers/src/styles/usa-sr-only';
+@use 'usa-form/src/styles/usa-form';
+@use 'usa-fieldset/src/styles/usa-fieldset';
+@use 'usa-legend/src/styles/usa-legend';
+@use 'usa-label/src/styles/usa-label';
+@use 'usa-hint/src/styles/usa-hint';
+@use 'usa-error-message/src/styles/usa-error-message';
+
+@import '../../mixins/uswds-error-border.css';
```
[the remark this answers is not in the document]
jamigibbs: Unfortunately, it does not compile if you remove the final path value:

Screen Shot 2022-12-01 at 4 09 32 PM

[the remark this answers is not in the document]
jamigibbs: As for having a global scss file, I think that might make sense at some point but in this specific situation, we only want `usa-sr-only`. We don't necessarily want all of the `uswds-helpers` so I think it might make more sense to keep the import as it is right now where it's very specific about what it needs (instead of a whole bunch of stuff this component isn't using).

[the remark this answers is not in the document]
jhonnyoddball: Since the `scss` files are not in the same project you need to import them

Try:
```
@import 'usa-form/src/styles';
```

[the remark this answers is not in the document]
jamigibbs: You are welcome to experiment with different imports locally and let me know if there is a better way but we went through a discovery about how to import these modules earlier in the development process and the preferred way is with `use`. See this discussion with the Stencil developers:

https://stencil-worldwide.slack.com/archives/C79EANFL7/p1667581320118309

Screen Shot 2022-12-02 at 8 54 52 AM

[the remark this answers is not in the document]
jhonnyoddball: Sounds good!

jhonnyoddball (approved):

PR #581, opened 2022-11-25
Determine how a USWDS3-based web component will get its color

jhonnyoddball:
## Chromatic

https://us-ds-v3-colors--60f9b557105290003b387cd5.chromatic.com

## Description
Closes [1322](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1322)

## Testing done
Locally
Storybook

### Before overriding colors
Screen Shot 2022-11-25 at 3 09 08 PM

### After overriding colors
Screen Shot 2022-11-25 at 3 17 10 PM

## Acceptance criteria
- [ ]

## Definition of done


jhonnyoddball:
Closing after demo

packages/web-components/src/global/variables.css
```diff
@@ -1,37 +0,0 @@
-/* Colors from: */
-/*   https://design.va.gov/design/color-palette */
-/* Fonts from: */
-/*   https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/8d39616e957cb44b6664a658b2c51dca8076fe7e/packages/formation/sass/base/_b-variables.scss#L16-L17 */
-:root {
-  --color-base: #212121;
```
jamigibbs, line 6: Can you document how these variables will still work in other components with the new import method you're recommending here?

packages/web-components/src/global/settings.scss
```diff
@@ -0,0 +1,2 @@
+@import 'uswds-helpers/src/styles';
+@import 'usa-form-group/src/styles';
```
jamigibbs, line 2: Do we need a global import like this? Would it be better for us to simply import single modules that are needed on a component to component basis (ie. less "stuff" will be loading that might not necessarily be needed by the component)?

jamigibbs:
What will we need to do for "mixins" that we might be creating? For example, in the v3 conversion, we have an error stylesheet that will need a v3 error color updated: https://github.com/department-of-veterans-affairs/component-library/pull/579#discussion_r1028582378

packages/web-components/stencil.config.ts
```diff
@@ -7,6 +8,15 @@ export const config: Config = {
   namespace: 'component-library',
   globalStyle: 'src/global/main.css',
   plugins: [
+    sass({ 
+      includePaths: [
+        '../../node_modules/@uswds/uswds/packages/',
+        '../../node_modules/@department-of-veterans-affairs/css-library/src/tokens/',
```
jamigibbs, line 14: Is there any additional build steps that need to happen to get this to compile?

packages/storybook/.storybook/preview.js
```diff
@@ -1,6 +1,6 @@
-import '@department-of-veterans-affairs/formation/dist/formation.min.css';
+// import '@department-of-veterans-affairs/formation/dist/formation.min.css';
 import './style.scss';
-import '@department-of-veterans-affairs/formation/dist/formation';
+// import '@department-of-veterans-affairs/formation/dist/formation';
```
jamigibbs, line 3: For completeness, it might be a good idea to test all of this out with Formation back in place. We have already merged the Storybook/Formation/v2 overrides into the main branch so if this branch is up to date, removing this shouldn't be necessary anymore: https://github.com/department-of-veterans-affairs/component-library/pull/577

jhonnyoddball: I agree

packages/web-components/src/global/settings.scss
jhonnyoddball: We can definitely start there and if we need to refactor to a global import we could in the future

packages/web-components/stencil.config.ts
jhonnyoddball: No, that I can think of

PR #588, opened 2022-12-13
va-radio: add part attr to all elements

Mottie:
## Chromatic

https://49541-req-radio--60f9b557105290003b387cd5.chromatic.com

## Description
Related to [#49539](https://github.com/department-of-veterans-affairs/va.gov-team/issues/49539)

We're using css to enlarge the legend text to `H3` size to match the design of our form. Recently, we decided to make this question required, and thus an `H3` sized required label also appeared; but we can't modify the style using css. 

Legend styled as an h3, but this also enlarges the required label to be the same size

This PR adds a `part` attribute to include the new USWDS elements, and also include hint and error messaging.

## Testing done

N/A

## Screenshots

N/A

## Acceptance criteria
- [x] Add `part` attribute to `va-radio` elements 
- [x] Add `part` attribute to USWDS rendered elements

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-radio/va-radio.tsx
+++ b/packages/web-components/src/components/va-radio/va-radio.tsx
@@ -230,14 +230,14 @@ export class VaRadio {
               {label}
               {required && <span class="usa-label--required" part="required"> {i18next.t('required')}</span>}
             </legend>
-            <span class="usa-error-message" role="alert" part="error">
+            <span class="usa-error-message" role="alert">
               {error && (
                 <Fragment>
                   <span class="usa-sr-only">{i18next.t('error')}</span> {error}
                 </Fragment>
               )}
             </span>
-            {hint && <span class="usa-hint" part="hint">{hint}</span>}
+            {hint && <span class="usa-hint">{hint}</span>}
             <slot></slot>
           </fieldset>
         </Host>
@@ -253,8 +253,8 @@ export class VaRadio {
             {label}
             {required && <span class="required" part="required">{i18next.t('required')}</span>}
           </legend>
-          {hint && <span class="hint-text" part="hint">{hint}</span>}
-          <span id="error-message" role="alert" part="error">
+          {hint && <span class="hint-text">{hint}</span>}
+          <span id="error-message" role="alert">
             {error && (
               <Fragment>
                 <span class="sr-only">{i18next.t('error')}</span> {error}
--- a/packages/web-components/src/components/va-radio/va-radio.tsx
+++ b/packages/web-components/src/components/va-radio/va-radio.tsx
@@ -220,24 +220,24 @@ export class VaRadio {
         'usa-legend': true,
         'usa-label--error': error
       });
-      return ( 
+      return (
         <Host
           aria-invalid={error ? 'true' : 'false'}
           aria-label={ariaLabel}
         >
           <fieldset class="usa-form usa-fieldset">
-            <legend class={legendClass}>
+            <legend class={legendClass} part="legend">
               {label}
-              {required && <span class="usa-label--required"> {i18next.t('required')}</span>}
+              {required && <span class="usa-label--required" part="required"> {i18next.t('required')}</span>}
             </legend>
-            <span class="usa-error-message" role="alert">
+            <span class="usa-error-message" role="alert" part="error">
               {error && (
                 <Fragment>
                   <span class="usa-sr-only">{i18next.t('error')}</span> {error}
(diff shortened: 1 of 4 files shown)


packages/web-components/src/components/va-radio/va-radio.tsx
```diff
@@ -251,10 +251,10 @@ export class VaRadio {
         >
           <legend part="legend">
             {label}
-            {required && <span class="required">{i18next.t('required')}</span>}
+            {required && <span class="required" part="required">{i18next.t('required')}</span>}
           </legend>
-          {hint && <span class="hint-text">{hint}</span>}
-          <span id="error-message" role="alert">
+          {hint && <span class="hint-text" part="hint">{hint}</span>}
```
jamigibbs, line 256: We are currently evaluating making hint text consistent across all of the form components so I think this (and the one above) should not be included at the moment: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1249

packages/web-components/src/components/va-radio/va-radio.tsx
```diff
@@ -220,24 +220,24 @@ export class VaRadio {
...
         >
           <fieldset class="usa-form usa-fieldset">
-            <legend class={legendClass}>
+            <legend class={legendClass} part="legend">
               {label}
-              {required && <span class="usa-label--required"> {i18next.t('required')}</span>}
+              {required && <span class="usa-label--required" part="required"> {i18next.t('required')}</span>}
             </legend>
-            <span class="usa-error-message" role="alert">
+            <span class="usa-error-message" role="alert" part="error">
```
jamigibbs, line 233: Is there a specific scenario that you need to be able to override error style?

packages/web-components/src/components/va-radio/va-radio.tsx
```diff
@@ -220,24 +220,24 @@ export class VaRadio {
...
         <Host
           aria-invalid={error ? 'true' : 'false'}
           aria-label={ariaLabel}
         >
           <fieldset class="usa-form usa-fieldset">
-            <legend class={legendClass}>
+            <legend class={legendClass} part="legend">
               {label}
-              {required && <span class="usa-label--required"> {i18next.t('required')}</span>}
+              {required && <span class="usa-label--required" part="required"> {i18next.t('required')}</span>}
```
jamigibbs, line 231: Is there a specific reason that you need to override the required text style?

Mottie: We're enlarging the `legend` text to match an `H3`, but the required label needs to be "normal" text sized

legend is H3-sized and the required label is not bold and smaller in size

packages/web-components/src/components/va-radio/va-radio.tsx
Mottie: I don't have one, but added this if we need it in the future... I can remove it if you don't think it's needed

packages/web-components/src/components/va-radio/va-radio.tsx
Mottie: Ok, I'll update the PR

packages/web-components/src/components/va-radio/va-radio.tsx
Mottie: Just to note, maybe the stylesheets aren't in place yet? But I added a `uswds` property to the `va-radio` and everything got really tiny

legend and option labels are tiny
font size set to 1.06 rem

jamigibbs: We have run into an issue where Formation sets the root font size to 10px but USWDS v3 is expecting 16px so the variation wasn't rendering the correct font size. We solved for this in Storybook with this [toggle approach](https://github.com/department-of-veterans-affairs/component-library/pull/577/files) but I don't think it has been solved yet for vets-website. Thanks for bringing it up though because I think we missed creating a task for it (which [I've done here](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1402)).

jamigibbs (approved):
LGTM. Before merging, do you mind incrementing the core package version too so that we can generate a release?

Mottie:
Core package minor bump? Also, not sure why the label & test CI checks are stuck

jamigibbs:
@Mottie Minor bump should be fine. I'm not sure why it's stuck either. Maybe the next push will restart it.

Mottie:
whew, there is goes :D

PR #594, opened 2022-12-22
Add variables.scss file to obtain USWDS v3 colors

jhonnyoddball:
## Chromatic

https://1409-add-variables.scss-file-to-obtain-colors--60f9b557105290003b387cd5.chromatic.com

## Description
Closes [1409](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1409)

In this PR the `variables.scss` file is added inside the `web-components` package to be able to access the color variables in any component. 

In order to use the new `variables.scss`. The working file needs to be converted to `.scss` (if needed). Then they can be called by adding:

```
@import `variables`;
```

This will be the same case for all the mixins files

### How to obtain the new tokens from `css-library`

This is a manual process. Inside the `css-library` folder, run command `yarn build:tokens`

### Output
css
✔︎ src/tokens/_variables.css

scss
✔︎ src/tokens/_variables.scss

### Move the `_variables.scss` to the `web-components` package

This process is also manual since we are testing if this approach will work correctly

Copy the contents of the `_variables.scss` from `css-library` to `web-components`

#### Note
This process will be manual until the `css-library` is being published, we automate the creation of the token, we add the `css-library` package as dependency to the `web-components` package and reference the `_variables.scss` file directly from the `css-library` package.

Eventually `variables.css` and `_variables.scss` from `web-components` will be replaced with the files in the `css-library`

## Example
Screen Shot 2022-12-21 at 10 07 22 AM

## Testing done
- Locally
- Storybook

## Screenshots

Screen Shot 2022-12-21 at 10 07 48 AM

Screen Shot 2022-12-21 at 10 08 01 AM

## Acceptance criteria
- [ ]

## Definition of done


jamigibbs (commented):
How would this implementation work with the `uswds-error.border.css` mixin? This is currently being imported into `va-radio` and `va-text-input` for displaying our custom error style: https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/src/mixins/uswds-error-border.css#L2

Would it make sense to make the mixin update to this PR or better to create a separate issue?

packages/web-components/src/global/_variables.scss
```diff
@@ -0,0 +1,85 @@
+
+// Do not edit directly
+// Generated on Thu, 17 Nov 2022 19:14:09 GMT from css-library
```
jamigibbs, line 3: I'm not sure I understand how this file was generated. Can you document how that happened and how it made it here in this global folder?

Also how do we distinguish between `variables.css` vs `_variables.scss` in the same folder?

packages/web-components/stencil.config.ts
```diff
@@ -9,7 +9,10 @@ export const config: Config = {
   globalStyle: 'src/global/main.css',
   plugins: [
     sass({ 
-      includePaths: ["../../node_modules/@uswds/uswds/packages/"],
+      includePaths: [
+        '../../node_modules/@uswds/uswds/packages/',
+        'src/global/'
```
jamigibbs, line 14: Have we explored including the `css-library` as a dependency similar to the way we are consuming the `uswds` packages here? 

So instead of `src/global` it would just import the css library package? '../../node_modules/@department-of-veterans-affairs/css-library'. 

https://www.npmjs.com/package/@department-of-veterans-affairs/css-library

jhonnyoddball:
> How would this implementation work with the `uswds-error.border.css` mixin? This is currently being imported into `va-radio` and `va-text-input` for displaying our custom error style: https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/src/mixins/uswds-error-border.css#L2
> 
We might need to create an error token in the `css-library` [color tokens](https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/css-library/tokens/color.json) since [b50909](https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/src/mixins/uswds-error-border.css#L2) does not exist yet. Then we need to recompile the tokens, obtain a new file and update the `_variables.scss` file I created in this PR with the new tokens.

> Would it make sense to make the mixin update to this PR or better to create a separate issue?
Definitely in a new issue since we need to create the error token before and then apply it.

Now, is it the idea to use the USWDS v3 variables directly like `$theme-color-error-dark: "red-60v" !default;` into our components or maybe create our own `override.scss` file in `web-components` to add these custom variables from USWDS v3? or add these new colors to the `css-library`?

packages/web-components/src/global/_variables.scss
jhonnyoddball: ### How to obtain the new tokens from `css-library`

Inside the `css-library` folder, run command `yarn build:tokens`

### Output
css
✔︎ src/tokens/_variables.css

scss
✔︎ src/tokens/_variables.scss

jhonnyoddball: > I'm not sure I understand how this file was generated. Can you document how that happened and how it made it here in this global folder?

After running the command above you will get the `_variables.css` and you can move the content to the new file in `web-components`.


> Also how do we distinguish between `variables.css` vs `_variables.scss` in the same folder?

Based on [stencil](https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/stencil.config.ts#L9-L17)

```
  plugins: [
    sass({ 
      includePaths: [
        '../../node_modules/@uswds/uswds/packages/',
        'src/global/'
      ],
    }),
    postcss({
      plugins: [url({ url: 'inline' })],
    }),
  ],
```
- `variables.css` gets compiled into `src/global/main.css`
- `_variables.scss` is being included only by the path. The file will need to be imported to be able to use the variables

packages/web-components/stencil.config.ts
jhonnyoddball: Yes, I did it in this test [PR](https://github.com/department-of-veterans-affairs/component-library/pull/581/files)

```
includePaths: [
        '../../node_modules/@uswds/uswds/packages/',
        '../../node_modules/@department-of-veterans-affairs/css-library/src/tokens/',
      ],
```

jhonnyoddball: We agreed to create the file locally in `web-components` first and then it we like it, then we will pull the file directly from `css-library`

packages/web-components/src/global/_variables.scss
jamigibbs: Thanks for outlining those steps. About `variables.css` vs `_variables.scss`,  I guess I was just wondering how a future developer would know what the difference is between them especially if that output doesn't directly map to anything (it's just manually moved at this point). 

Could you update this PR's title and/or description to specifically mention that this is a "USWDS v3" integration? While we have both v1/Formation and v3 stuff together, it can start to get difficult to distinguish things especially when they have similar names.

packages/web-components/stencil.config.ts
jamigibbs: Understood. Is there a ticket already about publishing the `css-library` npm package yet? We might want to add a comment there to follow up on updating the way this is imported to save us the trouble of having to manually generate the variables stylesheet. I think doing that might become tedious at some point especially if we are needing to add additional colors to the library (like the error color).

jamigibbs:
> Now, is it the idea to use the USWDS v3 variables directly like $theme-color-error-dark: "red-60v" !default; into our components or maybe create our own override.scss file in web-components to add these custom variables from USWDS v3? or add these new colors to the css-library?

I think the idea is that any colors that we need to use from the USWDS v3 color palette would be added to the `css-library` and they'd be accessible anywhere that we need them in the web components. I am not sure how exactly that will be solved but we were hoping in would be a part of this PR: https://github.com/department-of-veterans-affairs/component-library/pull/579#discussion_r1028582378

It might be as simple as changing that stylesheet to `.scss` and importing the variables after the color has been added to the css-library.

jhonnyoddball:
[Ticket created](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1417) to add the missing USDWS v3 error colors

jamigibbs (approved):

PR #596, opened 2023-01-05
va-breadcrumbs: Add handling of dynamic slot changes of breadcrumb items.

jamigibbs:
## Chromatic

https://1296-breadcrumbs-rerender--60f9b557105290003b387cd5.chromatic.com/?path=/docs/components-va-breadcrumbs--dynamic-state

## Description
This PR adds the handling of slot changes when the component is used in a SPA where breadcrumb items are being re-rendered dynamically (not necessarily after a full component re-render on a new page). Since the issue was reported as part of a [React implementation](https://github.com/department-of-veterans-affairs/vets-website/blob/main/src/applications/mhv/secure-messaging/components/shared/SmBreadcrumbs.jsx#L95), I approached solving it based on that libraries re-rending behavior with dynamic state.

The bigger challenge here was the way that React keeps a reference to the nodes that are no longer in the DOM tree. Because the `va-breadcrumbs` component is [making direct changes to the DOM,](https://github.com/department-of-veterans-affairs/component-library/blob/2f6708694985ed717a7020a1f2122cce242fca30/packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx#L67-L78) it conflicts with Reacts DOM management and was generating errors when crumb state was updating. This is why the reporter was receiving the error `Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.`

The workaround to allowing for a library like React and our component to work together in a re-rendering scenario is to wrap slot nodes with `<li>` so that nodes referenced by React will stay in the DOM tree. This solution was suggested here: https://github.com/facebook/react/issues/11538#issuecomment-390386520

To avoid having all currently active uses of `va-breadcrumbs` updated with new slot content structure, this update will handle both scenarios where the slotted nodes are all in the current `<a href="...">...</a>` format and also the scenario where dynamic re-rendering is necessary. In those cases, the slotted nodes should be be passed into the slot as `<li><a href="...">...</a></li>`

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1296

## Testing done
Storybook

### Current Slot Content Structure
Screenshot 2023-01-05 at 2 54 49 PM

### Added Slot Content Structure (for re-rendered slot content)
Screenshot 2023-01-05 at 2 54 42 PM

### New Storybook Story (handles dynamic re-rendering)

Screenshot 2023-01-05 at 2 38 47 PM

## Acceptance criteria
- [ ] The va-breadcrumbs component works for static slot content and also when the slot content needs to be re-rendered on the same page.

## Definition of done


packages/storybook/stories/va-breadcrumbs.stories.jsx
```diff
@@ -32,3 +83,6 @@ Default.args = {
   ...defaultArgs,
 };
 Default.argTypes = propStructure(breadcrumbsDocs);
+
+export const DynamicState = DynamicCrumbsTemplate.bind(null);
```
jamigibbs, line 87: Is "Dynamic State" the best way to describe this new Storybook story?

Screenshot 2023-01-05 at 2 43 01 PM

jamigibbs: Also would it make sense to add additional documentation on the Storybook example to point out that the structure for dynamically updating crumbs need to be in the `<li><a href="...">...</a></li>` format?  (For example: "Note that the crumb structure needs to include a list item")

Or is the Storybook example clear enough?
 
Screenshot 2023-01-06 at 10 33 47 AM

jhonnyoddball: I would say... Add more info to make it clear that it's necessary to have it to make it work

jhonnyoddball (approved):
Nice job!

PR #602, opened 2023-01-13
Accessibility updates to va-pagination to support Voice Control

jamigibbs:
## Chromatic

https://1408-pagination-a11y--60f9b557105290003b387cd5.chromatic.com

## Description
This PR addresses some accessibility issues mentioned in the ticket below. The word `prev` was updated to `previous` and the `aria-labels` were tested using Voice Control on mac to make sure that the component could be used entirely by voice control. The `aria-labels` were mostly just refactored and the word "page" added where relevant for using the voice controls.

Voice Commands that are working:

- "Click page 3" ✅  (and other page specific numbers)
- "Click previous page" ✅ 
- "Click next page" ✅ 

The command structure that I was unable to make work:

- "Click 3" ❌ 

It also fixes a bug where the selected page was not losing its active style (making it appears like two pages were selected at the same time). 

Related Issue: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1408

## Testing done
Storybook - Voice Command
Storybook - Voice Over

## Screenshots
Screenshot 2023-01-17 at 2 55 46 PM

## Acceptance criteria
- [ ] The va-pagination component can be used with Voice Control.

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-pagination/va-pagination.tsx
+++ b/packages/web-components/src/components/va-pagination/va-pagination.tsx
@@ -147,16 +147,23 @@ export class VaPagination {
       return <div />;
     }
 
+    const previousAriaLabel = ariaLabelSuffix ? `Previous page ${ariaLabelSuffix}` : 'Previous page';
+    const nextAriaLabel = ariaLabelSuffix ? `Next page ${ariaLabelSuffix}` : 'Next page';
+    const lastPageAriaLabel = ariaLabelSuffix ? `Page ${pages} ${ariaLabelSuffix}` : `Page ${pages}`;
+
     const renderPages = this.pageNumbers().map(pageNumber => {
       const pageClass = classnames({
         'button-active': page === pageNumber,
         'button-inner': true,
       });
 
+      const pageAriaLabel = ariaLabelSuffix ? `Page ${pageNumber} ${ariaLabelSuffix}` : `Page ${pageNumber}`;
+
       return (
         <li>
           <button
             aria-current={page === pageNumber ? 'true' : null}
+            aria-label={pageAriaLabel}
             class={pageClass}
             onClick={() =>
               this.handlePageSelect(pageNumber, 'nav-paginate-number')
@@ -164,8 +171,7 @@ export class VaPagination {
             onKeyDown={e => this.handleKeyDown(e, pageNumber)}
             type="button"
           >
-            <span class="sr-only">Page</span>{pageNumber}
-            {ariaLabelSuffix && (<span class="sr-only"> {ariaLabelSuffix}</span>)}
+            {pageNumber}
           </button>
         </li>
       );
@@ -178,6 +184,7 @@ export class VaPagination {
           {this.page > 1 && (
             <li>
               <button
+                aria-label={previousAriaLabel}
                 class="button-prev"
                 onClick={() =>
                   this.handlePageSelect(this.page - 1, 'nav-paginate-previous')
@@ -185,8 +192,7 @@ export class VaPagination {
                 onKeyDown={e => this.handleKeyDown(e, this.page - 1)}
                 type="button"
               >
-                Previous<span class="sr-only"> Page</span>
-                {ariaLabelSuffix && (<span class="sr-only"> {ariaLabelSuffix}</span>)}
+                Previous
               </button>
             </li>
           )}
@@ -203,6 +209,7 @@ export class VaPagination {
               </li>
               <li>
                 <button
+                  aria-label={lastPageAriaLabel}
                   class="button-inner"
                   onClick={() =>
                     this.handlePageSelect(pages, 'nav-paginate-number')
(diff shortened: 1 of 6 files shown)


jamigibbs:
### Voice Commands

While I was able to see the Voice Command working for the following commands, I do want to mention that the software seems to be very sensitive to the enunciation of the words. Sometimes I needed to say the command multiple times with very clear enunciation before it worked.

Voice Commands that are working:

- "Click page 3" (and other page specific numbers)
- "Click previous page"
- "Click next page"

Command structure that I was unable to make work:

- "Click 3"

I think that this type of command might not be enough context for Voice Command to understand where to click. I was also not able to get this command to work in the [govuk](https://design-system.service.gov.uk/), [uswds](https://designsystem.digital.gov/components/pagination/), or [w3](https://design-system.w3.org/components/pagination.html) design system pagination components.

The use of "page" as part of the voice command seems to be consistent with the other pagination components though and this value helps trigger it.

**Development Note**

I was not able to get voice commands to work when using `aria-label`. I needed to switch to using our `sr-hidden` class where the string "Page" would only be present for accessibility tools:

`<span class="sr-only">Page</span> {pageNumber}`

I'm not sure why this is the case. Maybe the way the Voice Command works with Storybook or with web components?

### Elipses (...)

In our component, we were previously allowing for click behavior to happen within the code (even though it didn't actually do anything). When looking at how other systems were accounting for this type of treatment, I noticed that there was a few things we could do to make it less confusing:

1. Add the `role="presentation"` to the element.
2. Make sure that when tabbing through the list that the elipses were skipped.

This seems to follow what these examples are doing as well although we might want to consider how to avoid using `...` completely:

### [govuk](https://design-system.service.gov.uk/)
Screenshot 2023-01-13 at 3 01 20 PM

### [uswds](https://designsystem.digital.gov/components/pagination/)
Screenshot 2023-01-13 at 3 01 11 PM

### [w3](https://design-system.w3.org/components/pagination.html)
Screenshot 2023-01-13 at 3 01 04 PM

It is possible to right now to remove the ellipses by setting the `show-last-page` to `false`. In this case, `va-pagination` will not show the last page and will simply tab one page at a time with the "next" button.

`show-last-page=true`
Screenshot 2023-01-13 at 3 13 56 PM

`show-last-page=false`
Screenshot 2023-01-13 at 3 13 51 PM

packages/web-components/src/components/va-pagination/va-pagination.tsx
```diff
@@ -70,6 +72,10 @@ export class VaPagination {
 
   private handlePageSelect = (page, eventID) => {
     this.pageSelect.emit({ page });
+
+    // Reset focus to the active page button.
+    (this.el.shadowRoot.activeElement as HTMLElement).blur();
```
jamigibbs, line 77: This fixes a bug where the selected page was not losing its active style making it appears like two pages were selected at the same time.  I don't think there is a specific issue ticket for this but it was mentioned as part of https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1408

joshkimux: I'm so glad you caught this! I thought I just had a super strange bug on my side because I couldn't replicate it.

joshkimux:
Testing the following to confirm fixes:

- [x] "Click page 3" ✅ (and other page specific numbers)
- [x] "Click previous page" ✅
- [x] "Click next page" ✅

There was a bit of lag on my end (took about 6-12 seconds per command when interacting with the pagination, but not for other commands like wake up/go to sleep). @jamigibbs did you experience lag on your end too? I'm not sure what the cause of that might be.

jamigibbs:
> There was a bit of lag on my end (took about 6-12 seconds per command when interacting with the pagination, but not for other commands like wake up/go to sleep). @jamigibbs did you experience lag on your end too? I'm not sure what the cause of that might be.

@joshkimux hmmm 🤔 I just tried it out again and there's maybe a very slight lag? Half a second or so? It is not significant for me though. Which browser are you using? I can only guess that any lag is related to the software processing on the local machine but I'm not entirely sure.

https://user-images.githubusercontent.com/872479/213017110-548a5b60-7362-42b1-a9be-ab27947bebb6.mov

joshkimux:
@jamigibbs I'm using Safari on a Macbook! I'm also assuming it might just be my local machine in this case.

jamigibbs:
@joshkimux I tried this is Safari and I'm wondering if maybe it's something quirky with Storybook. I had to open up the component in its own window before I could get Voice Command to work there (no significant delay still though):

Screenshot 2023-01-18 at 3 14 33 PM

https://user-images.githubusercontent.com/872479/213296269-98465ca2-1c11-4e2c-ab3a-869d171b395f.mov

jhonnyoddball (approved):
Great job, nice demo

jamigibbs:
@joshkimux Before I merge this I just wanted to double check with you about your last comments here and make sure you're 👍 about these updates.

PR #604, opened 2023-01-17
Add USWDS v3 Error color to css-library

jhonnyoddball:
## Chromatic

## Description
Closes [1417](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1417)

## Testing done
Locally
Storybook - Text Input USWDS v3

## Screenshots
Screen Shot 2023-01-17 at 11 36 01 AM

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/css-library/tokens/color.json
+++ b/packages/css-library/tokens/color.json
@@ -38,5 +38,84 @@
       "warm-light": { "value": "#e4e2e0"},
       "cool-light": { "value": "#dce4ef"}
     }
+  },
+  "v3-color": {
+    "base": {
+      "*": { "value": "#71767a"},
+      "dark": { "value": "#565c65"},
+      "darker": { "value": "#3d4551"},
+      "darkest": { "value": "#1b1b1b"},
+      "light": { "value": "#a9aeb1"},
+      "lighter": { "value": "#dfe1e2"},
+      "lightest": { "value": "#f0f0f0"},
+      "ink": { "value": "#1b1b1b"}
+    },
+    "primary": {
+      "*": { "value": "#005ea2"},
+      "dark": { "value": "#1a4480"},
+      "darker": { "value": "#162e51"},
+      "light": { "value": "#73b3e7"},
+      "lighter": { "value": "#d9e8f6"},
+      "vivid": { "value": "#0050d8"}
+    },
+    "secondary": {
+      "*": { "value": "#d83933"},
+      "dark": { "value": "#b50909"},
+      "darker": { "value": "#8b0a03"},
+      "light": { "value": "#f2938c"},
+      "lighter": { "value": "#f8dfe2"},
+      "vivid": { "value": "#e41d3d"}
+    },
+    "accent": {
+      "cool": {
+        "*": { "value": "#00bde3"},
+        "dark": { "value": "#28a0cb"},
+        "darker": { "value": "#07648d"},
+        "light": { "value": "#97d4ea"},
+        "lighter": { "value": "#e1f3f8"}
+      },
+      "warm": {
+        "*": { "value": "#fa9441"},
+        "dark": { "value": "#c05600"},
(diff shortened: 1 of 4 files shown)


packages/css-library/tokens/color.json
```diff
@@ -38,5 +38,84 @@
       "warm-light": { "value": "#e4e2e0"},
       "cool-light": { "value": "#dce4ef"}
     }
+  },
+  "v3-color": {
```
jhonnyoddball, line 42: All these are the [new available colors from USWDS v3 ](https://designsystem.digital.gov/design-tokens/color/theme-tokens/) and the [Tokens from these colors](https://github.com/uswds/uswds/blob/develop/packages/uswds-core/src/styles/settings/_settings-color.scss#L86)

packages/web-components/src/global/variables.css
```diff
@@ -31,6 +31,7 @@
   --color-secondary-dark: #cd2026;
   --color-secondary-light: #e59393;
   --color-secondary-lightest: #f9dede;
+  --v3-color-error-dark: #b50909;
```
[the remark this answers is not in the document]
jamigibbs: Was this added here to avoid converting the `uswds-error-border.css` mixin to an `scss` file?

packages/web-components/src/global/_variables.scss
```diff
@@ -32,6 +30,7 @@ $color-gray-lightest: #f1f1f1;
 $color-gray-warm-dark: #494440;
 $color-gray-warm-light: #e4e2e0;
 $color-gray-cool-light: #dce4ef;
+$v3-color-error-dark: #b50909;
```
jamigibbs, line 33: Was it intentional not to generate this variables file so that it includes all of the other colors added to `packages/css-library/tokens/color.json`?

jhonnyoddball: Correct, you mentioned to only add the error variable to `web-components`

packages/web-components/src/global/variables.css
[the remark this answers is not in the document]
jhonnyoddball: Yes, this file is being using in a lot of places, therefore, we will need to change a lot of `.css` files to `.scss` files. Maybe in the future, but I think it's out of the scope.

packages/web-components/src/global/_variables.scss
jamigibbs: Sorry I meant only adding the v3 colors we needed to the actual css-library (not just to this file that is generated by it). I don't think we should be manually updating this file at all.

I am open to adding all of those other colors to css-library if we feel like we are going to use them. I just wanted to avoid making the library larger than it needed to be pre-maturely.

jhonnyoddball: Ohhh, I see what you are saying. I can remove them all but that one and we can add as we need them then.

packages/web-components/src/global/variables.css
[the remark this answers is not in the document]
jhonnyoddball: There was something, I can't remember now. But, I'll try to change it to see if I remember.

jamigibbs, line 34: Everything looks good to me! Do you think we still need this custom property here though now that we are pulling the error color directly from the css-library variables file?

jamigibbs (approved):

packages/web-components/src/global/variables.css
jhonnyoddball: I just added it there to keep it consistent and to have it available in case we need it in a `.css` file, in case we can't change it to a `.scss` file.

jamigibbs:
@jhonnyoddball Before you merge, do we think ignore is the right label? Or should it be minor or patch? I'm not completely sure.

PR #608, opened 2023-01-23
Add the "Banner - Official Gov" web component

jamigibbs:
## Chromatic

https://1276-official-gov-site-banner--60f9b557105290003b387cd5.chromatic.com/?path=/docs/components-va-official-gov-banner--default

## Description
This component was built based on the [USWDS Banner component](https://designsystem.digital.gov/components/banner/). The style and markup were taken directly from the USWDS example page linked above. 

Even thought the USWDS v3 library is currently available within the component library package, this variation intentionally does not import the modules directly so that the component can be used immediately. USWDS v3 variations of components that use the v3 modules are not currently available for integrating with Formation.

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1276

## Testing done
Storybook

## Screenshots

Screenshot 2023-02-01 at 2 23 55 PM

Screenshot 2023-02-01 at 2 24 08 PM

Screenshot 2023-02-01 at 2 24 19 PM

Screenshot 2023-02-01 at 2 24 28 PM

### Mobile

Screenshot 2023-02-01 at 2 25 58 PM

Screenshot 2023-02-01 at 2 26 04 PM

## Acceptance criteria
- [ ]  Component is written and added to Storybook

## Definition of done


packages/web-components/src/components/va-official-gov-banner/va-official-gov-banner.scss
```diff
@@ -0,0 +1,304 @@
+:host {
+  display: block;
+}
+
+.banner {
+  background-color: #f0f0f0;
```
jamigibbs, line 6: All of the hex colors hardcoded are from the USWDS v3 example. When this component is updated to use the v3 modules (ie. `@use "usa-banner"`), these colors should automatically map to the relevant classes in the markup and can eventually removed. I did not think it was necessary to add these to the `css-library` package for that reason.

packages/web-components/src/components/va-official-gov-banner/va-official-gov-banner.scss
```diff
@@ -0,0 +1,304 @@
...
+@media (max-width: 768px) {
+  header .header-action::after, header.expanded .header-action::after {
+    background-color: #005ea2;
+  }
+
+  header.expanded .header-action::after, header .header-action::after {
+    background: 0 0;
+    background-color: #005ea2;
+    -webkit-mask: url("../../../../../node_modules/@uswds/uswds/dist/img/usa-icons/expand_more.svg") no-repeat center/contain;
+    mask: url("../../../../../node_modules/@uswds/uswds/dist/img/usa-icons/expand_more.svg") no-repeat center/contain;
```
jamigibbs, line 191: Similar to the hardcoded hex values, this direct link to the v3 icons will no longer be necessary once this component has been converted to use the v3 modules (ie. @use "usa-banner").

packages/web-components/src/components/va-official-gov-banner/va-official-gov-banner.scss
```diff
@@ -0,0 +1,304 @@
...
+  }
+
+  header.expanded button::after {
+    -webkit-mask: url("../../../../../node_modules/@uswds/uswds/dist/img/usa-icons/expand_less.svg") no-repeat center/contain;
+    mask: url("../../../../../node_modules/@uswds/uswds/dist/img/usa-icons/expand_less.svg") no-repeat center/contain;
+  }
+}
+
+// Mobile display close icon.
+@media (max-width: 768px) {
```
jamigibbs, line 232: The responsive width throughout is based on the [Formation medium screen breakpoint](https://design.va.gov/foundation/breakpoints). The USWDS v3 examples uses `40rem` for a 16px base font so this would be the equivalent to 640px. I simply picked a breakpoint value from Formation that was closest to that value.

packages/web-components/src/components/va-official-gov-banner/va-official-gov-banner.tsx
```diff
@@ -0,0 +1,209 @@
...
+    html = html.replace('https://', `<strong>https://</strong>`);
+
+    if (el) {
+      el.innerHTML = html;
+    }
+  }
+
+  render() {
+    const { tld } = this;
+    if (tld === 'gov' || tld === 'mil') {
```
jamigibbs, line 138: I experimented with getting the top-level domain value programmatically within the component itself but found that it would be very difficult to test or display the component on sites that weren't actually `.gov` or `.mil` sites (like Chromatic and Storybook). If we're not happy with relying on the developer to set the correct value here manually, then maybe we can discuss alternatives. The default prop value is `.gov`.

packages/web-components/src/components/va-official-gov-banner/va-official-gov-banner.tsx
```diff
@@ -0,0 +1,209 @@
...
+                    <span class="button-text">{i18next.t('gov-site-button')}</span>
+                  </button>
+                </div>
+              </header>
+  
+              <div class="content" id="official-gov-banner" hidden>
+                <div class="grid-row">
+                  <div class="col">
+                    <img 
+                      src="https://s3-us-gov-west-1.amazonaws.com/content.www.va.gov/img/icon-dot-gov.svg"
```
jamigibbs, line 173: This was taken directly from the existing VA.gov implementation. I sometimes see a slight delay with this is loaded because there's a call out to AWS for it so if we're not happy with that behavior, I can see if there is a way to load without hot linking.

Screenshot 2023-02-01 at 3 17 13 PM

Andrew565 (approved):
LGTM

packages/web-components/src/components/va-official-gov-banner/va-official-gov-banner.tsx
jamigibbs: I attempted to get this `png` image to load locally using the [Stencil asset base path](https://stenciljs.com/docs/v2/assets#assets) but wasn't successful. This might be a [limitation of Stencil](https://github.com/ionic-team/stencil/issues/1868). For now we will probably need to keep the AWS link. We are able to import SVGs though.

jamigibbs:
@GnatalieH When you have a moment, do you mind providing design review for this new component?

GnatalieH:
Looks good to me, @jamigibbs! Thanks! I prefer the update, USWDS look/feel in the mobile viewport.

jamigibbs:
> One thing I'm wondering, is should we represent this component in the USWDS section on Storybook? I know we are creating it as a variation of Banner, so I understand why it's categorized as it is. But I feel like we should cross reference it in USWDS components. Thoughts?

@GnatalieH  Good question. I intentionally have kept it out of that section because technically it has not been updated to use the actual USWDS v3 modules like the other components in that section are. We are not able to use those component variations in the USWDS section yet everywhere (there are technical limitations right now) but this component can be used as it is because it's not using their modules yet. So I didn't want to cause any confusion about its availability which is why it's in the regular section. We will still need to come back to this component at some point to make it an official v3 variation that's using the USWDS library directly.

GnatalieH:
@jamigibbs that makes sense. It's good where it is, then. Thanks!

PR #610, opened 2023-01-27
Add va-select USWDS v3 variation

jhonnyoddball:
## Chromatic

https://1444-select-us-wds-v3--60f9b557105290003b387cd5.chromatic.com

## Description
Closes [1444](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1444)

## Testing done
Locally
Storybook

## Screenshots
Screen Shot 2023-01-27 at 11 57 51 AM

Screen Shot 2023-01-27 at 11 57 59 AM

### Select with internationalization
Screen Shot 2023-01-31 at 10 03 38 AM

## Acceptance criteria
- [x] Create a version of this component that imports and uses the USWDS CSS
- [x] Make any structural changes necessary and add a prop that determines which version of the component will be used
- [x] Check with Natalie to make sure the component and the design match
- [x] Add a USWDS story to Storybook

## Definition of done


packages/web-components/src/components/va-select/va-select.scss
```diff
@@ -1,38 +1,54 @@
-@import '../../mixins/inputs.css';
-@import '../../mixins/accessibility.css';
-@import '../../mixins/form-field-error.css';
-@import '../../mixins/hint-text.css';
+@forward 'settings';
```
jhonnyoddball, line 1: Needed to obtain USWDS icons in any component

humancompanion-usds:
@jhonnyoddball - I believe that the USWDS version of Select supports having the first option have an empty option value so that the first item can be placeholder text (i.e "-- Select --"). Is that something our web-component supports? When I tried to change the name and value in the Storybook instance in chromatic, nothing happened (though I was able to change other fields). Otherwise this looks good to me. I checked that the CSS matched :-)

jhonnyoddball:
> @jhonnyoddball - I believe that the USWDS version of Select supports having the first option have an empty option value so that the first item can be placeholder text (i.e "-- Select --"). Is that something our web-component supports? When I tried to change the name and value in the Storybook instance in chromatic, nothing happened (though I was able to change other fields). Otherwise this looks good to me. I checked that the CSS matched :-)

@humancompanion-usds we can definitely have something like this as a first option:

```
<option value>- Select -</option>
```

It'll be an empty value, but this is something always available in drop downs

GnatalieH (approved):
Looks good to me.

jamigibbs:
> 
> @humancompanion-usds we can definitely have something like this as a first option:
> 
> ```
> <option value>- Select -</option>
> ```
> 
> It'll be an empty value, but this is something always available in drop downs

Yes, I think that we should try and match the USWDS example here but the question is if this is something we want to control within the component itself, or, just add that default option to the Storybook example slot manually. So some questions/comments:

- Do we want to allow the text "Select" to have different variations (ie. "choose") , or, do we want to control that as part of the DS?
- Should we allow teams to optionally choose to display `- Select -` or should it always show by default?
- If we are adding this to the component itself to control, we will need to get translations for it and make sure we are managing that.

packages/storybook/stories/va-select-uswds.stories.jsx
```diff
@@ -0,0 +1,129 @@
...
+
+const Template = ({
+  label,
+  name,
+  value,
+  required,
+  error,
+  uswds,
+  hint,
+  'aria-live-region-text': ariaLiveRegionText,
```
jamigibbs, line 55: I'm guessing that this `aria-live-region-text` was added to the story because the original `va-select.stories.jsx` file has it? 

But I'm a little confused about it. It's not a prop in the component itself. It just seems to be added as some kind of guidance in Storybook. And when I research the purpose of it, I can't seem to find any documentation about having this type of aria attribute. Also when I test it with VoiceOver, it does not seem to announce anything when the select changes. 

I'm going to reach out to the #accessibility-help channel to see if we can determine if this is necessary or not.

packages/web-components/src/global/_settings.scss
```diff
@@ -0,0 +1,3 @@
+@use "uswds-core" with (
+  $theme-image-path: '../../../../../node_modules/@uswds/uswds/dist/img/',
+);
```
jamigibbs, line 3: Great job figuring this out. I'm going to add a note here for future us to explore this override method for fonts as well related to this issue:

https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1411

USWDS v3 Reference: https://github.com/uswds/uswds/blob/b715c487ceb7fa18075189ee79861c482c5b61a6/README.md#set-the-base-asset-paths-fonts-and-images

packages/storybook/stories/va-select-uswds.stories.jsx
jamigibbs: @jhonnyoddball Based on this [feedback from Josh](https://dsva.slack.com/archives/C8E985R32/p1675095738560739), I don't think we should include this `aria-live-region-text` attribute to the v3 Storybook examples. I'm unable to find any documented history about this or able to get it to work in a screenreader and his feedback says that it would be redundent.

I'll write up a separate ticket though to follow-up on this for the `va-select.stories.jsx` Storybook file though once we have an a11y specialist.

jamigibbs: Follow up ticket: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1483

jhonnyoddball: Yeah, I duplicated the file and just added what's needed

jhonnyoddball:
> > @humancompanion-usds we can definitely have something like this as a first option:
> > ```
> > <option value>- Select -</option>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It'll be an empty value, but this is something always available in drop downs
> 
> Yes, I think that we should try and match the USWDS example here but the question is if this is something we want to control within the component itself, or, just add that default option to the Storybook example slot manually. So some questions/comments:
> 
> * Do we want to allow the text "Select" to have different variations (ie. "choose") , or, do we want to control that as part of the DS?
> * Should we allow teams to optionally choose to display `- Select -` or should it always show by default?
> * If we are adding this to the component itself to control, we will need to get translations for it and make sure we are managing that.

@humancompanion-usds 

- Should we incorporate `- Select -` header into our component to keep it the same as USWDS?
- Should we allow people to change the value of that `header`? @jamigibbs brings a good point that we will need to offer translation as well for that.

humancompanion-usds:
Yes we should include the "- Select -" option by default and the text should not be alterable. The value is blank and I would think that also would not be allowed to be changed. Yes, we should get it translated.

It should be possible to change the default option that is selected as that is part of the USWDS guidance:

> Use a good default. When most users will (or should) pick a particular option, make it the default: <option selected="selected">Default</option>

I would expect this option to always be present and the behavior to be as follows: If it's a required field and the user selects the blank value then we'd throw an error. If it's an optional field then the blank option provides a way for a user to unselect, to not make a choice or undo a choice they have made. However, we should verify this behavior with how it works today to understand how big of a change that might be so we can communicate appropriately.

packages/storybook/stories/va-select-uswds.stories.jsx
```diff
@@ -0,0 +1,129 @@
...
+    docs: {
+      page: () => <StoryDocs data={selectDocs} />,
+    },
+  },
+};
+
+const defaultArgs = {
+  'label': 'Branch of Service',
+  'name': 'branch',
+  'value': 'army',
```
jamigibbs, line 21: Now that we have the `- select -` default value in the component, can we update the Storybook example so that it displays by default (instead of "army")?

jamigibbs (approved):

PR #615, opened 2023-02-06
Add aria-describedby-message prop to va-text-input

jamigibbs:
## Chromatic

https://24502-input-aria-described-by--60f9b557105290003b387cd5.chromatic.com/?path=/docs/components-va-text-input--default

## Description

[Previously we made an assumption](https://github.com/department-of-veterans-affairs/va.gov-team/issues/39674) that the only text that needed to be read by screenreaders was the label (linked using `for > id`) so the `aria-describedby` was removed for all other optional text except for error in the web component version of va-text-input.

In the React version of this component, [we did identify a need to have other optional text](https://github.com/department-of-veterans-affairs/component-library/pull/96#discussion_r632736947) though so that flexibility was previously available.

This PR adds back the optional text for scenarios like the one brought up in https://github.com/department-of-veterans-affairs/va.gov-team/issues/24502 where some external text outside of the label needed to be associated with the input field.

Note: The reason that we needed to add another prop to the component specifically for passing in an optional aria-describedby message (`aria-describedby-message`) is because the web component's shadow DOM will not allow it to reference an external ID for `aria-describedby`. The element must live within the component itself.

Related issue: https://github.com/department-of-veterans-affairs/va.gov-team/issues/24502

## Testing done
Storybook 
Voiceover

## Screenshots

Screenshot 2023-02-07 at 10 14 56 AM

Screenshot 2023-02-07 at 10 14 40 AM

## Acceptance criteria
- [ ] Optional text is read by a screenreader as part of va-text-input.

## Definition of done


packages/web-components/src/components/va-text-input/test/va-text-input.e2e.ts
```diff
@@ -15,7 +15,7 @@ describe('va-text-input', () => {
             Hello, world
           </label>
           <slot></slot>
-          <span id="error-message" role="alert"></span>
+          <span id="input-error-message" role="alert"></span>
```
jamigibbs, line 18: I updated this id to match what the [USWDS example](https://designsystem.digital.gov/components/text-input/) uses: 

Screenshot 2023-02-07 at 10 40 33 AM

Andrew565 (approved):
LGTM

PR #621, opened 2023-02-16
Formation overrides for va-text-input component

Andrew565:
## Chromatic

https://asteele&#x2F;v3-overrides--60f9b557105290003b387cd5.chromatic.com

## Description
Closes #1492

## Screenshots

Screenshot 2023-02-15 at 2 33 32 PM

### DIFF
--- a/packages/web-components/src/global/_formation_overrides.scss
+++ b/packages/web-components/src/global/_formation_overrides.scss
@@ -1,0 +1,20 @@
+@import '../../global/functions';
+
+.usa-label {
+  font-size: rem-override(1.06rem);
+}
+
+.usa-input {
+  height: rem-override(2.5rem);
+  margin-top: rem-override(0.5rem);
+  max-width: rem-override(30rem);
+  padding: rem-override(0.5rem);
+  font-size: rem-override(1.06rem);
+
+  &--success,
+  &--error {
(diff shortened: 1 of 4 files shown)


packages/web-components/src/global/_formation_overrides.scss
```diff
@@ -0,0 +1,20 @@
+@import '../../global/functions';
+
+.usa-label {
+  font-size: rem-override(1.06rem);
+}
+
+.usa-input {
```
jamigibbs, line 7: Do we really want all of these styles to be loaded into every single component? I'm assuming the approach is to import `@import '../../global/formation_overrides';` into all USWDS v3 component variations? But a component like banner would not need these styles.

packages/web-components/src/global/_functions.scss
```diff
@@ -0,0 +1,15 @@
+@use 'sass:math';
+
+@import 'variables';
+
+// Uses $base-font-size from _variables.scss
+@function rem-override($original) {
```
jamigibbs, line 6: How does this look in Storybook with the custom override we have in place? Do we need to make any adjustments to the changes we made here: https://github.com/department-of-veterans-affairs/component-library/commit/8afb4a34a512725db1baa50b1c81c0d76658a7b2

PR #622, opened 2023-02-16
Formation overrides for va-text-input component

Andrew565:
## Chromatic

https://asteele-v3-overrides--60f9b557105290003b387cd5.chromatic.com

## Description
Closes [#1492](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1492)

## Testing done
Visual test.

## Screenshots
Screenshot 2023-02-15 at 2 33 32 PM


packages/web-components/src/global/_formation_overrides.scss
```diff
@@ -0,0 +1,20 @@
+@import '../../global/functions';
+
+.usa-label {
+  font-size: rem-override(1.06rem);
+}
+
+.usa-input {
```
jamigibbs, line 7: Do we really want all of these styles to be loaded into every single component? I'm assuming the approach is to import `@import '../../global/formation_overrides';` into all USWDS v3 component variations? But a component like banner would not need these styles.

packages/web-components/src/global/_functions.scss
```diff
@@ -0,0 +1,15 @@
+@use 'sass:math';
+
+@import 'variables';
+
+// Uses $base-font-size from _variables.scss
+@function rem-override($original) {
```
jamigibbs, line 6: How does this look in Storybook with the custom override we have in place? Do we need to make any adjustments to the changes we made here: https://github.com/department-of-veterans-affairs/component-library/commit/8afb4a34a512725db1baa50b1c81c0d76658a7b2

Andrew565: Good question, I'll check.

packages/web-components/src/global/_formation_overrides.scss
Andrew565: In this case, even though the file is located within the 'global' folder, it is not automatically imported globally (same for the _functions.scss file). What causes the file to be imported is on [line 8](https://github.com/department-of-veterans-affairs/component-library/pull/622/files#diff-3e956d46d7983277911085179f5eb079e722c72953af3196d26cce81399364dfR8) of the va-text-input.scss file. So, the idea is that we would import the formation_overrides file only where we need it; it's in the 'global' folder only as a signifier that it is 'available' globally, not 'imported' globally. Very good question, thanks for double-checking the intentions!

jamigibbs: What I mean is that is that all of the CSS will be loaded in components that don't necessarily need those styles (not that it's global to the site).

packages/web-components/src/global/_functions.scss
Andrew565: Looks good in Storybook so far!

packages/web-components/src/global/_formation_overrides.scss
Andrew565: Agreed that this is a less-than-ideal solution long term, but an acceptable solution for now. Will revisit if and when this overrides file gets "too big" to afford to import everywhere it's needed.

Andrew565:
@jamigibbs Ready for re-review, thanks!

packages/web-components/src/global/_variables.scss
```diff
@@ -37,6 +37,7 @@ $font-weight-normal: 400;
 $font-weight-bold: 700;
 $font-style-normal: normal;
 $font-style-italic: italic;
+$base-font-size: 16px;
```
jamigibbs, line 40: Can we add an inline comment to this (or rename) that mentions that this is the base font size for USWDS v3? It's a little confusing because the Formation base font size is something different.

jamigibbs:
@Andrew565 Is there a follow-up tasks created yet to remove the Storybook Formation scripting that overrides those styles yet? I think we need to have a clear plan on when we're going to remove that because it's hard to fully test these style overrides if those scripting overrides still exist --> https://github.com/department-of-veterans-affairs/component-library/pull/577

We also have a couple of components getting the v3 variation added to them now so I'm not sure when we need to start importing the Formation style overrides.

packages/web-components/src/global/_variables.scss
Andrew565: I see what you mean, as that's the base font size for USWDS v3 itself. ;)

Andrew565:
> @Andrew565 Is there a follow-up tasks created yet to remove the Storybook Formation scripting that overrides those styles yet?

No, but I will add that alongside the mitigation implementation tickets for radio, radio-option, and select.

packages/web-components/src/global/_formation_overrides.scss
jamigibbs: FYI I'm marking this unresolved not because I think there needs more work in this PR but so that the conversation is easier to reference in the future if we need to come back and reorganize the stylesheet. Thanks for clarifying this! 👍

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@department-of-veterans-affairs/component-library-monorepo",
   "description": "The monorepo containing the component library packages.",
   "private": true,
-  "version": "1.0.0",
+  "version": "1.0.1",
```
jamigibbs, line 5: I think you intended to increment the `packages/core/` version number. This version number is not in use afik.

jamigibbs:
This PR needs a `patch` label. You can also remove this one if you'd like. I don't think we anything but version labels for this repo at the moment:

Screenshot 2023-02-21 at 1 44 50 PM

package.json
Andrew565: Oh, sheesh.

Andrew565: Reverted. Do I need to update the core version number if I didn't change anything in the core package?

jamigibbs: We use the core version for [releases](https://github.com/department-of-veterans-affairs/component-library#releasing) so it should be incremented if we are publishing the NPM package. I think generally we just bump core in the last PR that will be included in the release: https://github.com/department-of-veterans-affairs/component-library/releases

jamigibbs:
@Andrew565 I tested this locally with the Storybook overrides removed and it looks like the translations buttons will also need to be updated in the components Storybook file. The font size is using rem so changing that to 16px should fix it (eventually we shouldn't need to put inline style here at all but we will need it until the Formation overrides are completely removed):

https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/storybook/stories/va-text-input-uswds.stories.jsx#L118-L120

```
<button style={{fontSize: '1rem'}} onClick={e => setLang('es')}>Español</button>
<button style={{fontSize: '1rem'}} onClick={e => setLang('en')}>English</button>
<button style={{fontSize: '1rem'}} onClick={e => setLang('tl')}>Tagalog</button>
```

Screenshot 2023-02-22 at 8 57 42 AM

I think we also missed `margin-top` and `max-width` for `.usa-label`. They are still displaying as rem when I inspect:

Screenshot 2023-02-22 at 9 03 57 AM

jamigibbs:
> @Andrew565 Is there a follow-up tasks created yet to remove the Storybook Formation scripting that overrides those styles yet?

> No, but I will add that alongside the mitigation implementation tickets for radio, radio-option, and select.

I'm linking to those tickets here so this can PR can be cross-referenced when that work is done:

- https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1567
- https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1566
- https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1568

Andrew565:
Also this ticket: [DSP P1 - Remove storybook manual USWDS style override scripts](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1569)

Andrew565:
Also linking this: [DSP P1 - Remove storybook manual USWDS style override scripts](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1569)

Andrew565:
@jamigibbs Ready for re-review

packages/web-components/src/components/va-text-input/va-text-input.tsx
```diff
@@ -229,12 +221,14 @@ export class VaTextInput {
...
-      messageAriaDescribedby
+      messageAriaDescribedby,
     } = this;
     const type = this.getInputType();
     const maxlength = this.getMaxlength();
-    const ariaDescribedbyIds = `${messageAriaDescribedby ? 'input-message' : ''} ${error ? 'input-error-message' : ''}`
-      .trim() || null; // Null so we don't add the attribute if we have an empty string
+    const ariaDescribedbyIds =
+      `${messageAriaDescribedby ? 'input-message' : ''} ${
+        error ? 'input-error-message' : ''
```
jamigibbs, line 230: Are the formatting updates a preference thing or does the team need to align on the prettier and linting settings (and also make sure we have it running locally)? https://github.com/department-of-veterans-affairs/component-library/blob/main/.prettierrc.json

packages/web-components/src/components/va-text-input/va-text-input.tsx
```diff
@@ -251,20 +245,25 @@ export class VaTextInput {
           {label && (
             <label htmlFor="inputField" class={labelClass} part="label">
               {label}
-              {required && <span class="usa-label--required"> {i18next.t('required')}</span>}
+              {required && (
+                <span class="usa-label--required">
+                  {' '}
+                  {i18next.t('required')}
+                </span>
+              )}
```
jamigibbs, line 253: Why is this line changed?

packages/web-components/src/global/_variables.scss
```diff
@@ -31,12 +31,14 @@ $color-gray-warm-dark: #494440;
 $color-gray-warm-light: #e4e2e0;
 $color-gray-cool-light: #dce4ef;
 $v3-color-error-dark: #b50909;
-$font-family-sans: 'Source Sans Pro', 'Helvetica Neue', Helvetica, Roboto, Arial, sans-serif;
+$font-family-sans: 'Source Sans Pro', 'Helvetica Neue', Helvetica, Roboto, Arial,
+  sans-serif;
```
jamigibbs, line 35: Another formatting thing?

jamigibbs (approved):
LGTM but I think we need to follow-up as a team on formatting code in unrelated PRs and making sure we are all aligned on what, when, etc.

packages/web-components/src/components/va-text-input/va-text-input.tsx
Andrew565: I've always considered formatting preferences to be voluntary/arbitrary/optional, personally. I like making use of Prettier in my code, especially since VS Code's Prettier plugin just takes care of it for me. But I don't think it needs to be a required thing.

packages/web-components/src/global/_variables.scss
Andrew565: Yep.

PR #633, opened 2023-02-28
 va-radio-option: Fix long label wrapping

Mottie:
## Chromatic

https://1527-radio-option-wrap--60f9b557105290003b387cd5.chromatic.com

## Description

Long radio option labels would wrap and align with the radio symbol instead of indenting under the beginning of the label text

va-radio with 2 options. The second one has text that is causing it to wrap to the next line, but the wrapped word is left aligned with the button symbol and not the start of the text

See [#1527](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1527)

This fix switched to using flexbox to create columns to maintain the alignment

## Testing done

Updated va-radio-option e2e test

## Screenshots

A second option with a longer label has been included in storybook

A radio group showing two options. The second option has long text that wraps, but now the wrapped text is left aligned with the start of the label above it

<details><summary>A longer wrapping label with a description was also checked</summary>

This change was not included in storybook

A radio group with showing two options. The first option has long text that wraps and includes description text below the label which also wraps. Both the label and description are left aligned.</details>

## Acceptance criteria
- [x] Long option labels wrap and align properly
- [x] All tests passing

## Definition of done


packages/web-components/src/components/va-radio-option/va-radio-option.scss
```diff
@@ -44,6 +45,8 @@
   line-height: 1.4rem;
   margin-right: 1rem;
   width: 1.4rem;
+  min-width: 1.4rem;
+  max-width: 1.4rem;
```
Mottie, line 49: Forcing the width was required because without it, the circle became squished

2 radio options with the second containing 3 lines of wrapped text, but the radio symbol is horizontally compressed making it an oval shape

packages/web-components/src/components/va-radio-option/va-radio-option.tsx
```diff
@@ -124,8 +124,10 @@ export class VaRadioOption {
           id={id}
         >
           <label htmlFor={id} id="option-label">
-            {label}
-            {description && <span class="description" aria-describedby="option-label">{description}</span>}
+            <div>
+              {label}
+              {description && <span class="description" aria-describedby="option-label">{description}</span>}
+            </div>
```
Mottie, line 130: Wrapping the label and description in a `div` otherwise flex box would try to keep all 3 on the same line

packages/web-components/src/components/va-radio-option/va-radio-option.scss
```diff
@@ -62,7 +65,6 @@
 
 :host(:not([uswds])) .description {
   display: block;
-  margin-left: 25px;
```
Mottie, line 65: No longer needed this margin since we're using flexbox

Andrew565 (approved):
LGTM

PR #666, opened 2023-03-31
1606 va-memorable-date update styles

ataker:
## Chromatic

https://1606-memorable-date--60f9b557105290003b387cd5.chromatic.com

## Description
Partial https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1606
Fixed margins that were incorrect. 
Looking for feedback if we should update the colors described in this ticket. The labels and the error colors come from other components, like `va-text-input` so it would need to be updated more globally. If we want to make these changes I think it makes sense to create new tickets to reflect the wider scope of the changes.

## Testing done
Storybook in chrome

## Screenshots
Select bottom margin removed: 
Select before
Select after

Error margin removed:
Error margin before
Error margin after

Top margin set to 16px from 12px:
Top margin before
Top margin after

Error colors updated:
Error before
Error after

## Acceptance criteria
- [ ] 

## Definition of done


packages/web-components/src/global/_formation_overrides.scss
```diff
@@ -119,6 +119,7 @@
 .usa-select {
   background-size: rem-override(1.25rem);
   background-position: right rem-override(0.5rem) center;
+  margin-bottom: 0;
```
jamigibbs, line 122: The only things that should go into this formation override stylesheet are rem to pixel conversion for v3. Everything else should go into component specific stylesheets.

jamigibbs (changes requested):
Do you mind doing the following for this PR?

- [ ] Add screenshots that show which margins were changed (before and after if possible).
- [ ] Add a version patch label to the PR (`major`, `minor`, `patch`, or `ignore-for-release`). 
- [ ] Increment the `package.json` version in the web-components package.

jamigibbs:
I was getting a little turned around about what might need to be updated [from the original ticket](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1606) so I'm writing out a checklist below to make sure we're all on the same page:

- [ ] The `error-message` element between the hint text and labels over the input fields in the non-error states is creating a space when it should not (if there is no error). The margin between hint text and label should be 16px.
- [ ] The labels over the input fields should be #1B1B1B. Storybook is showing #212121 for the labels over the inputs.
- [ ] There is an extra bottom margin in the Select box variations that should not be there. It's 3.392 px.
- [ ] `--v3-color-error-dark: #b50909` should also be used for the input border color in the error states. 
- [ ] The red border on the inputs should match the left red border and error message text in color. The `secondary-dark` color is not correct.

> Looking for feedback if we should update the colors described in this ticket.

Yes we'll want to update the colors here if you're able to confirm from the feedback above. For example, to update item two in the list above (labels over input fields should be #1B1B1B), the corresponding CSS to add to the component stylesheet will probably be something like this (although you should confirm):

```css
.usa-label { color: var(--color-base)  }
```

The `.usa-label` class can be added to the `va-memorable-date` component stylesheet with the color we are wanting. This will not have a global affect (web component styles are limited to their shadow dom elements).

See also for available variables: `/packages/web-components/src/global/variables.css`

@ataker  Do you have access to the Sketch file for this? I think that in order to complete this task with all of the updates that @GnatalieH is requesting, you will need to be able to view and inspect the Sketch file. Then you can see exactly how the design should look and get precise colors and padding/margin values.

This is the Sketch link for `va-memorable-date` - v3 version: https://www.sketch.com/s/610156b6-f281-4497-81f3-64454fc72156/a/AxeE0LR#Inspect

packages/web-components/src/components/va-memorable-date/va-memorable-date.scss
```diff
@@ -50,5 +50,10 @@ slot:not([name])::slotted(*) {
 }
 
 :host(:not([uswds])) #error-message {
-  margin-bottom: 0.8rem;
+  margin-bottom: 0;
 }
+
+:host(:not([uswds])) .date-container,
+:host .usa-memorable-date {
```
jamigibbs, line 57: You don't need to add :host here for `.usa-memorable-date`. We only need to leverage `:host` when we're trying to differentiate styles between the same selector that's being used in different ways (for example a class that's styled one way for v1 and styled another way for v3). In this case, `.usa-memorable-date` is only being used in v3 so it shouldn't need `:host`.

This should also be moved into the "v3 style area" as described in the follow comment.

packages/web-components/src/components/va-select/va-select.scss
```diff
@@ -58,3 +58,7 @@ span.required {
   font-family: var(--font-source-sans);
   padding: 0.8rem 3rem 0.8rem 0.7em;
 }
+
+:host .usa-select {
```
jamigibbs, line 62: We are trying to keep all `v3` specific styles at the top (they go before the comment that says `/** Original Component Style **/`) so do you mind moving those types of styles into that area? Eventually we will need to remove v1 completely and keeping v1 and v3 separate will help us do that.

Also, you don't need to add `:host` here either.

jamigibbs (changes requested):

ataker:
Screenshots in PR updated

packages/web-components/src/components/va-memorable-date/va-memorable-date.scss
```diff
@@ -41,7 +49,7 @@
 
 :host(:not([uswds])) #dateHint,
 #hint {
-  color: var(--color-gray);
+  color: var(--color-gray-medium);
```
jamigibbs, line 52: This changes the v1 version of hint text. I didn't think this ticket involved making any changes to v1 (or changes to any hint text color at all).

packages/web-components/src/components/va-date/va-date.css
```diff
@@ -2,6 +2,21 @@
...
+}
+
+.usa-error-message {
+  color: var(--v3-color-error-dark);
+}
+
+:host([uswds][error]:not([error=''])) va-select::part(select),
+:host([uswds][error]:not([error=''])) va-text-input::part(input) {
+  border: 4px solid var(--v3-color-error-dark);
+}
```
jamigibbs, line 16: Is `va-date` used in `va-memorable-date`? I'm not sure why this component is being updated.

I'm also not sure why we would need to add error styling in a parent component for the children. Are we not able to activate the error state in `va-select` and `va-text-input` (the child components used in `va-memorable-date`) so that it naturally displays the error border for their v3 variation?

jamigibbs: And btw the answer to that question about styling the child components could be yes (like we had to do it that way). I just want to double check that we tried doing it with the child component directly as well.

packages/web-components/src/components/va-memorable-date/va-memorable-date.scss
ataker: Whoops, yeah, must have imagined that one

packages/web-components/src/components/va-select/va-select.scss
```diff
@@ -11,12 +11,16 @@
 @import '../../mixins/uswds-error-border.scss';
 @import '../../global/formation_overrides';
 
-:host {
-  color: var(--color-base);
+:host([uswds]) {
+  color: var(--v3-color-base-darkest);
   display: block;
   font-family: var(--font-source-sans);
 }
```
jamigibbs, line 18: I think you meant to make two separate `:host` styles. Otherwise, as it is now, all of those styles will only apply to v3.

So I believe you want to add one for v3:

```css
:host([uswds]) {
  color: var(--v3-color-base-darkest);
}
```

And one for v1 and v3 (this is what was there originally):

```css
:host {
  display: block;
  font-family: var(--font-source-sans);
}
```

packages/web-components/src/components/va-text-input/va-text-input.scss
```diff
@@ -9,8 +9,8 @@
 @import '../../mixins/uswds-error-border.scss';
 @import '../../global/formation_overrides';
 
-:host {
-  color: var(--color-base);
+:host([uswds]) {
+  color: var(--v3-color-base-darkest);
   display: block;
   font-family: var(--font-source-sans);
 }
```
jamigibbs, line 16: Same comment as above about two separate `:host` style blocks.

packages/web-components/src/components/va-text-input/va-text-input.tsx
```diff
@@ -55,6 +55,11 @@ export class VaTextInput {
    */
   @Prop() error?: string;
 
+  /**
+   * Whether or not to add usa-input--error as class if error message is outside of component
+   */
+  @Prop() reflectInputError?: boolean = false;
```
jamigibbs, line 61: Thank you for adding this! Just for future us to remember why we did this, it was added here the same way that it was in `va-select` which was required to trigger the error state for the children components: https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/src/components/va-select/va-select.tsx#L61-L64

jamigibbs (commented):
Thank you for digging into this one! I left a minor update request and if you could also increment the version in the web-components package.json too before it's merged.

jamigibbs:
And small administrative request before this is merged; can you update the name of this PR so that it at least includes the name of the component (va-memorable-date)? This will make it easier when we create release notes.

PR #651, opened 2023-03-16
add option for currency

it-harrison:
## Chromatic

https://{{head.ref}}--60f9b557105290003b387cd5.chromatic.com
This PR updates the `va-number-input` component to add currency support as described in [this ticket](https://app.zenhub.com/workspaces/platform-design-system-5f8de67192551b0012ebb802/issues/gh/department-of-veterans-affairs/vets-design-system-documentation/1280).
---

## Description
Closes [1280](https://app.zenhub.com/workspaces/platform-design-system-5f8de67192551b0012ebb802/issues/gh/department-of-veterans-affairs/vets-design-system-documentation/1280)

## Testing done
manual testing in updated versions of Chrome, FireFox, Safari, Edge

## Screenshots
Screenshot 2023-03-16 at 12 08 58 PM

Screenshot 2023-03-16 at 12 09 32 PM

## Acceptance criteria
- [ ]

## Definition of done
- [ ] Documentation has been updated, if applicable
- [x] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)

PR #669, opened 2023-04-06
 Add va-number-input USWDS v3 variation

jamigibbs:
## Chromatic

https://1625-uswds-va-number-input--60f9b557105290003b387cd5.chromatic.com/?path=/docs/uswds-va-number-input--default

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1625

Below is the Sketch file for `va-text-input`. This component does not have all of the same states but it inherits the same general style modules for USWDS:
- https://www.sketch.com/s/610156b6-f281-4497-81f3-64454fc72156/a/nRYozkW

## Testing done
Storybook

### Focus
Screenshot 2023-04-07 at 2 18 17 PM

Screenshot 2023-04-07 at 12 48 20 PM

Screenshot 2023-04-07 at 12 48 23 PM

Screenshot 2023-04-07 at 12 48 16 PM

## Acceptance criteria
- [ ] A USWDS v3 variation of va-number-input exists.

## Definition of done

### DIFF
--- a/packages/web-components/src/components.d.ts
+++ b/packages/web-components/src/components.d.ts
@@ -555,6 +555,10 @@ export namespace Components {
          */
         "required"?: boolean;
         /**
+          * Whether or not the component will use USWDS v3 styling.
+         */
+        "uswds"?: boolean;
+        /**
           * The value for the input.
          */
         "value"?: string;
@@ -2052,6 +2056,10 @@ declare namespace LocalJSX {
           * Set the input to required and render the (Required) text.
          */
         "required"?: boolean;
+        /**
+          * Whether or not the component will use USWDS v3 styling.
+         */
+        "uswds"?: boolean;
         /**
           * The value for the input.
          */
--- a/packages/web-components/src/components/va-text-input/test/va-text-input.e2e.ts
+++ b/packages/web-components/src/components/va-text-input/test/va-text-input.e2e.ts
@@ -355,7 +355,7 @@ describe('va-text-input', () => {
     expect(await inputEl.getProperty('autocomplete')).toBe('email');
   });
 
-  // Begin USWDS v3 test
+  // Begin USWDS tests
   it('uswds v3 renders', async () => {
     const page = await newE2EPage();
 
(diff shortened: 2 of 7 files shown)


packages/web-components/src/components/va-text-input/test/va-text-input.e2e.ts
```diff
@@ -355,7 +355,7 @@ describe('va-text-input', () => {
     expect(await inputEl.getProperty('autocomplete')).toBe('email');
   });
 
-  // Begin USWDS v3 test
+  // Begin USWDS tests
```
jamigibbs, line 358: I noticed that the va-text-input component did not have all of the existing tests added for USWDS so I'm adding them to this PR but if we would prefer this submitted as a separate PR, I can do that.

jamigibbs (commented):
I have intentionally left out the `currency` prop that was [added here](https://github.com/department-of-veterans-affairs/component-library/pull/651) in our v1 variation because we eventually want to add [input prefix/suffix](https://designsystem.digital.gov/components/input-prefix-suffix/). But if we'd like to add it to the v3 variation until that happens, then I can add it.

packages/web-components/src/components.d.ts
```diff
@@ -2052,6 +2056,10 @@ declare namespace LocalJSX {
           * Set the input to required and render the (Required) text.
          */
         "required"?: boolean;
+        /**
+          * Whether or not the component will use USWDS v3 styling.
+         */
+        "uswds"?: boolean;
```
Andrew565, line 2062: Do we know why this is being duplicated here? I don't think it's a big deal, just curious.

Andrew565 (approved):
LGTM

packages/web-components/src/components.d.ts
jamigibbs: This file is automatically generated by the Stencil compiler but it looks like it's adding one instance to the [LocalJSX namespace](https://github.com/department-of-veterans-affairs/component-library/blob/7da4d51fd85e258f230a966745ee665e405a2624/packages/web-components/src/components.d.ts#L1386) and the other to the [Components namespace](https://github.com/department-of-veterans-affairs/component-library/blob/7da4d51fd85e258f230a966745ee665e405a2624/packages/web-components/src/components.d.ts#L8).

Which is then referenced by the [va-number-input interface](https://github.com/department-of-veterans-affairs/component-library/blob/7da4d51fd85e258f230a966745ee665e405a2624/packages/web-components/src/components.d.ts#L2628) so this is likely something to do with how the React binding for the web components work. I didn't dig into it further though.

PR #672, opened 2023-04-11
Automate Undeprecated Component Usage Data

Andrew565:
## Chromatic

https://asteele-automate-component-usage-data--60f9b557105290003b387cd5.chromatic.com

---

## Description
Closes #[1647](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1647)

## Testing done
Works on my machine

## Acceptance criteria
- [x] Accurate files are generated

## Definition of done
- [x] Update the script which is run to output a CSV file that can be imported into Excel or Google Sheets or something similar


packages/design-system-dashboard-cli/src/codeowners.json
```diff
@@ -0,0 +1,74 @@
+{
+  "*": ["va-platform-cop-frontend"],
+  "src/applications/_mock-form": ["platform-design-system-fe"],
+  "src/applications/burial-poc-v6": ["platform-design-system-fe"],
```
jamigibbs, line 4: Not blocking but I just want to call out that the [CODEOWNERS file gets updated regularly](https://github.com/department-of-veterans-affairs/vets-website/commits/main/.github/CODEOWNERS) so if these are hardcoded based on that reference, we might just need to keeps tabs on things a little bit or not count on the output as always being completely accurate.

packages/design-system-dashboard-cli/src/write-react-owners-to-csv.js
```diff
@@ -0,0 +1,134 @@
...
+
+        /**
+         * @type {string[]}
+         * @description There can be more than one owner, so this should always be an array
+         */
+        let owners = codeowners[ownerPath];
+
+        // In the event an owner is not listed or is not found, the 'default' is the cop-frontend team
+        if (owners === undefined) {
+          owners = ['va-platform-cop-frontend'];
```
jamigibbs, line 82: 🙏

update-component-usage.sh
```diff
@@ -0,0 +1,14 @@
...
+yarn build
+yarn build-bindings
+cd ../react-components
+yarn install
+yarn build
+cd ../core
+yarn install
+yarn build
+cd ../design-system-dashboard-cli
+yarn report
```
jamigibbs, line 14: Can you update our [CLI readme](https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/design-system-dashboard-cli/README.md) for how to run this report?

Andrew565:
> General admin request: when linking to the related ticket, use the Github link (not Zenhub). We lose all of the automated tracking and linking that happens between PR's and tickets if we don't do that.

Good point, didn't realize that.

packages/design-system-dashboard-cli/src/codeowners.json
Andrew565: Good to note.

update-component-usage.sh
Andrew565: Sure. Forgot I included this script here. :)

jamigibbs (approved):
Giddy up! 🏇

yarn.lock
```diff
@@ -8212,6 +8212,39 @@ __metadata:
   languageName: node
   linkType: hard
 
+"csv-generate@npm:^4.2.2":
+  version: 4.2.2
+  resolution: "csv-generate@npm:4.2.2"
+  checksum: f262698c92376261c928779335850c583f5d6ea9f2e476ba476ece7c9fb70bdafe634a4f3024721d44e46b7c50d2135d5932fa383f4fbe48dcaa642635052b2f
+  languageName: node
```
jamigibbs, line 8219: Is this the `yarn.lock` file for the `core` package?  Should new packages instead be added just to the `design-system-dashboard-cli` package.json? 

`core` is essentially independent from `design-system-dashboard-cli`. Just theorizing why the tests are failing.

Andrew565: Hmm, good point, didn't realize which yarn it had installed into.

Andrew565: Ah, it added it to the yarn.lock at the root only because of the way the workspaces feature is setup. The actual dependency is still in the dashboard-cli folder.

PR #674, opened 2023-04-12
1679 accordion header icon

mdewey:
## Chromatic

---
- [ ] Link to any related issues in the description so they can be cross-referenced.
- [ ] Under **Testing done**, be specific about how this update was tested. 
    - Examples: Storybook, Chrome, Browserstack, Mobile/Responsive, etc.
- [ ] Add the appropriate version patch label (`major`, `minor`, `patch`, or `ignore-for-release`). 

## Description
Closes [Ticket](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1679)

## Testing done
- Chrome
- Storybook
- Various Screen sizes
- updated unit testing

## Screenshots
image
image
image

## Acceptance criteria
- [x] Icons and styling are in the header
- [x] Allows passing and render of headers and subheaders as slots for flexibility

## Definition of done


README.md
```diff
@@ -79,10 +86,46 @@ This will allow you to run Storybook locally to view all components
...
     ```
     <body>
         <va-progress-bar label="Add a label here" percent={35}></va-progress-bar>
         <va-segmented-progress-bar current={2} total={6}></va-segmented-progress-bar>
     </body>
     ```
+
 3. Run `yarn serve` inside `packages/web-components/` to start the Stencil Dev Server
+
+## Testing
```
jamigibbs, line 99: Thanks for this addition! 👍

packages/web-components/src/components/va-accordion/va-accordion-item.tsx
```diff
@@ -92,38 +57,22 @@ export class VaAccordionItem {
...
-      this.el.querySelector('[slot="headline"]')?.innerHTML;
-
-    const Header = () =>
-      h(
-        // TODO: Remove level & header prop & let the slot do its work
-        /* eslint-disable-next-line i18next/no-literal-string */
-        this.slotTag || ieSlotCheckTag || `h${this.level}`,
-        null,
+    const headerText = this.el.querySelector('[slot="headline"]').innerHTML.replace(/(<([^>]+)>)/ig, '');
+    return (
```
jamigibbs, line 61: What is this regex doing?

packages/web-components/src/components/va-accordion/va-accordion-item.tsx
```diff
@@ -92,38 +57,22 @@ export class VaAccordionItem {
...
-      );
-    return (
-      <Host>
-        <Header />
-        <slot name="headline" onSlotchange={() => this.populateStateValues()} />
-        <div id="content">
+          <slot name="headline" />
+          {this.subheader ? <p id='subheader'>{this.subheader}</p> : <slot name="subheader" />}
+        </button>
+        <div tabIndex={0} id="content">
```
jamigibbs, line 75: Maybe I am misunderstanding what we are trying to solve for here but my understanding of the issue is that we want to add support just for icons. 

This opens up the slot to display _any_ content. Are we all aligned with that being ok from the design system's perspective?

Screenshot 2023-04-12 at 3 31 50 PM

humancompanion-usds: Not super keen. We have to retain the header integrity. Not sure why those tests are coming out. The idea here was not to make the header and sub-header a free for all for any content.

jamigibbs: Is it possible to have a named slot only for displaying an icon (while also maintaining the existing header integrity)? 

```
Usage: <i slot="icon" class="fas fa-paperclip"></i>

Component: <slot name="icon"></slot>
```
If we also need to allow styling of the icon, we could add a `part` attribute which will pierce the shadow dom: 

```
<slot name="icon" part="icon"></slot>
```

We have done something similar to check for slot nodes in other components using the `getSlottedNodes` [utility function](https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/src/utils/utils.ts#L20). va-breadcrumbs also [handles slotted node elements](https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx#L95).

jamigibbs: Alternatively, if we are confident that we will only support font awesome icons, would adding a new prop for an icon name be a reasonable approach? With this direction, a font name would simply be passed into `va-accordion` and displayed within the component itself. Then there's no need to keep track of slot nodes.

```
<va-accordion-item icon="paperclip" ... >

@Prop() icon?: string = null;

<i class={`fas fa-${icon}`}></i>
```

humancompanion-usds: We're moving to USWDS icons from FontAwesome. So this would work temporarily but not in future.

packages/web-components/src/components/va-accordion/va-accordion-item.tsx
mdewey: That regex is removing the HTML tags from the header so that the header text can be used in the `aria-label` below

packages/web-components/src/components/va-accordion/va-accordion-item.tsx
mdewey: So the full context is that we are looking to support the use cases in this design

mdewey: Hey! Thank you for the feedback! 

So the full context for this update is that we are updating the component to 

1) be more accessible 
2) Support our use cases in this design https://www.sketch.com/s/32bdb345-fb1f-409b-985f-ff627cb60032

On the accessibility front, the tabbing was not working. A user can not tab into the content of the accordion. 

For the use cases, we need the header to support icons in both the headline and subheader and custom styling (bolding/unbolding) in both the header and subheader. 

Opening up the component like this allows for more options, but teams can use it to suit their needs. This also allows teams not to be tied to Font Awesome, which I see USWDS is moving away from in the long run.   

From an engineering side, this approach allows for flexibility. At the end of the day, for this specific use case, we need to 

- Use the component for its accessible features
- Be able to have an icon in both the sub/header
- Be able to style the sub/header  (bolded/unbolded is the first use case we have)  

The current component doesn't allow for any of those. We could do specific updates like a `<slot name='icon">` or `isBold` prop, which introduces unnecessary complexity.  

I am more than willing to hop on a call to understand a path forward better. 

Tagging @coultonbunney-usds  and @rtwell as designers from my team.

jamigibbs: Thanks for the Sketch link, @mdewey! That does add a lot more context and I can undersatnd now why this isn't just adding an icon. All of the elements in the head appear to need flexible styling or interactions.

I'm not sure how we'd like to move forward here but I want to also mention that we have a lot of other instances of this component in vets-website that are using the props that this PR is removing so it will create some additional work for other teams to update all the current instances to instead pass content into a slot instead:

Screenshot 2023-04-18 at 9 27 27 AM

I'm wondering if the best solution for your team in order to accomplish the Sketch design is to fork this component and bring it into your app so that it can be customized specifically for your needs?

@humancompanion-usds @Andrew565 I'm not sure if either of you have any other thoughts about this or if we'd like to put some time on the calendar to discuss.

mdewey: Mm... Okay. Thanks for your insight! 

Using the ' slot ' idea, I will work through a possible backward-compatible solution. 

In the meantime, what does `forking the component` practically look like? Do we want multiple versions of the design system wondering around?

mdewey:
Hey gang, 

I have created a new solution using slots and that still maintains the integrity of the header. 

https://github.com/department-of-veterans-affairs/component-library/pull/679

I am going to move the conversion over to that PR.

I am seeking feedback on the approach before adding unit testing and polishing off the code.

mdewey:
Hey @humancompanion-usds and @jamigibbs Check out the linked PR about the new approach

PR #678, opened 2023-04-19
 Add va-privacy-agreement USWDS v3 variation

harshil1793:
## Chromatic

https://1678-privacy-agreement-v3--60f9b557105290003b387cd5.chromatic.com

---

## Description
This pull request adds a USWDS V3 version of `privacy-agreement` component by adding a new prop `uswds`.

In addition, this PR adds a USWDS story to Storybook, allowing users to view and interact with the new version of the component.

## Changes:
- Created a new version of the component that uses the USWDS V3 checkbox
- Added a `uswds` prop that determines which version of the component will be used
- Added this component to USWDS Storybook section
- Added a CSS property to increase the size of the check icon for improved visibility and usability
- Added class to change checkbox width to 100% for only privacy agreement
- For consistent font size, added conversion for privacy text 

Closes: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1678

## Testing done
- Storybook
- Chrome/Mozilla/Edge

## Screenshots
**Privacy Agreement USWDS V3 unchecked:**
v3

**Privacy Agreement USWDS V3 checked:**
v3 checked

**USWDS Section:**
Privacy v3 nav

## Acceptance criteria
- [ ] USWDS version has been created and added to Storybook
- [ ] Design.va.gov has the latest version of component-library

## Definition of done


packages/web-components/src/components/va-checkbox/va-checkbox.scss
```diff
@@ -95,6 +95,7 @@
   background-image: url('../../assets/check.svg');
   background-position: 50%;
   background-repeat: no-repeat;
+  background-size: cover;
```
jamigibbs, line 98: This is for `:host(:not([uswds])) input:checked+label::before` which is the v1 version. I'm not sure why it is needed if this PR is for v3.

packages/web-components/src/components/va-privacy-agreement/test/va-privacy-agreement.e2e.ts
```diff
@@ -151,4 +151,153 @@ describe('va-privacy-agreement', () => {
...
+      action: 'click',
+      componentName: 'va-privacy-agreement',
+      details: {
+        checked: false
+      },
+    });
+
+  });
+
+  it.skip('uswds v3 does not fire analytics event when `enableAnalytics` prop is not set', async () => {
```
jamigibbs, line 275: There are a couple of tests added that are skipped?

packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.scss
```diff
@@ -1,9 +1,18 @@
...
 :host {
   display: block;
 }
 
+va-checkbox::part(checkbox) {
+  max-width: 100%;
+}
+
+.privacy-policy {
+  font-size: 1.06 * $v3-font-base-size;
```
jamigibbs, line 13: Can you add a comment here like "Source Sans Pro normalization for v3 - 16.96px" or something?

packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.tsx
```diff
@@ -64,27 +67,59 @@ export class VaPrivacyAgreement {
...
-        </va-checkbox>
-      </Host>
-    );
-  }
+    const { uswds } = this;
 
+    if (uswds) {
+      return (
+        <Host>
+          {/* <div style={{borderTop: '1px solid red'}}></div> */}
```
jamigibbs, line 75: I think we can remove any commented out code like this.

packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.tsx
```diff
@@ -64,27 +67,59 @@ export class VaPrivacyAgreement {
...
+            <span class="privacy-policy" slot="description">
+              Please read and accept the&nbsp;
+              <a href="/privacy-policy/" target="_blank">
+                privacy policy
+                <i
+                  class="fa-arrow-up-right-from-square"
+                  aria-hidden="true"
+                  role="img"
+                ></i>
+                <span class="sr-only">opens in a new window</span>
```
jamigibbs, line 94: There is a UWDS specific screen reader only class you can use here. It imports as:

`@use 'uswds-helpers/src/styles/usa-sr-only';`

packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.tsx
```diff
@@ -64,27 +67,59 @@ export class VaPrivacyAgreement {
...
-    );
-  }
+    const { uswds } = this;
 
+    if (uswds) {
+      return (
+        <Host>
+          {/* <div style={{borderTop: '1px solid red'}}></div> */}
+          <va-checkbox
+            uswds={true}
```
jamigibbs, line 77: You can just make this `uswds`. Just the presence of the prop makes it truthy.

jamigibbs:
@LillyBoot  Do we want the label of the checkbox bold when there's an [error for v3](https://1678-privacy-agreement-v3--60f9b557105290003b387cd5.chromatic.com/?path=/docs/uswds-va-privacy-agreement--default)? I don't think we have an error state in the sketch file but I'm pretty sure that the label is bold for the other v3 components when there's an error:

Screenshot 2023-04-20 at 10 30 14 AM

LillyBoot:
@jamigibbs I looked at what we had in sketch for the USWDS error states and it seems that there is a bit of inconsistency on whether or not we bold the label. I am going to bring it up during our design meeting today. My current feeling is that if we bold some labels we should bold all of them, but I will update you.

packages/web-components/src/components/va-privacy-agreement/test/va-privacy-agreement.e2e.ts
harshil1793: A couple of tests, like V1, were initially skipped. However, upon re-evaluation, they appear to be passing now, so I have removed the skip.

packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.tsx
```diff
@@ -64,27 +67,59 @@ export class VaPrivacyAgreement {
...
+                  role="img"
+                ></i>
+                <span class="sr-only">opens in a new window</span>
+              </a>
+              .
+            </span>
+          </va-checkbox>
+        </Host>
+      );
+    } else {
```
jamigibbs, line 101: I know that we have done an `else` in previous v3 variation updates but we don't really need it because the `return` statement will not be called twice. If we remove the `else`, we can also maintain a clean version history (the carriage returns make it appear that all of this code has changed when it has not).

packages/web-components/src/components/va-checkbox/va-checkbox.scss
harshil1793: You raise a good point. The selector :`host(:not([uswds])) input:checked+label::before` was used in the v1 version and seems to have carried over to v3. However, I agree that it may not be necessary for v3 and I have also noticed that some CSS properties are being applied to V3 that shouldn't be. FYI, [V3 checkbox](https://design.va.gov/storybook/?path=/docs/uswds-va-checkbox--default)  in production have small check icon. Let me review it again.

jamigibbs: That is interesting. I'm not sure when the check got so small like that. 🤔 

Screenshot 2023-04-20 at 4 51 30 PM

It would be great if you were able to figure out how to fix that in this PR. But it is also reasonable if you just create a separate ticket for it. It does not seem to be an issue when it's loaded in the v3 `va-privacy-agreement`.

harshil1793: I was able to fix it by adding `reflect` to `uswds` under checkbox component.

packages/web-components/src/components/va-checkbox/va-checkbox.scss
```diff
@@ -24,6 +24,10 @@
   outline-offset: 4px;
 }
 
+.usa-checkbox__input:checked + [class*="__label"]::before {
+  background-size: 1.3rem;
+}
```
jamigibbs, line 29: Why do we need to add a USWDS override here in the web component? Wouldn't any necessary styling for the v3 checkbox just come directly from the USWDS module?

harshil1793: This was the fix for the check icon size. This is not within the scope of the PR. If overriding is something we want to avoid I can undo this change and create a ticket.

jamigibbs: Ah, I see. So what I think needs to happen here is that you need to add the USWDS style to the formations overrides stylesheet to convert the rem to px.

This should be converted to pixels because the base font in Formation is different than the base font for USWDS:

Screenshot 2023-04-21 at 11 27 09 AM

Try adding this to the `global/_formation_overrides.scss` stylesheet (and remove your override in the component):

```css
.usa-checkbox__input:checked + [class*="__label"]::before {
    background-size: rem-override(0.75rem);
}
```

packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.scss
jamigibbs: This class needs to be made specific to v3 now that the if/else has been removed (`:host[uswds]`). It's affecting both v1 and v3 at the moment. And I also would like to see the class name changed to `.description`.

packages/web-components/src/components/va-checkbox/va-checkbox.scss
harshil1793: Sure, and can I update the `0.75rem` to `1.3rem` ?

jamigibbs: You should leave it at 0.75rem and let the function `rem-override` convert it to the correct size pixel in `_formation_overrides.scss`

jamigibbs:
Here is an example of the label bold during the error state in the checkbox v3 variation for comparison: https://1678-privacy-agreement-v3--60f9b557105290003b387cd5.chromatic.com/?path=/docs/uswds-va-checkbox--error

harshil1793:
I've added the logic to make first label bold when in the error state. Also, I've added test case as well.

jamigibbs (approved):
Thanks, Harshil! 👍  

Please keep the `core` version at 15.2.0 after you sync with `main`. 15.2.0 will be our next release version.

PR #684, opened 2023-04-25
(bug) fixed styling issue for sub header

mdewey:
## Chromatic

https://va-accordion-item&#x2F;alignment-patch--60f9b557105290003b387cd5.chromatic.com

## Description

- Fixing a bug that was introduced by a CSS Change. 

### Currently
image

### After 
image

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@department-of-veterans-affairs/component-library",
   "description": "VA.gov component library. Includes React and web components.",
-  "version": "15.2.0",
+  "version": "15.2.1",
   "license": "MIT",
   "scripts": {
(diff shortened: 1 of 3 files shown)


packages/core/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@department-of-veterans-affairs/component-library",
   "description": "VA.gov component library. Includes React and web components.",
-  "version": "15.2.0",
+  "version": "15.2.1",
```
jamigibbs, line 4: We probably won't be creating a release right after this is merged (we have a few other PRs coming that we can include in the next release) so do you mind reverting this? The web-components version bump is good though.

jamigibbs (approved):

PR #690, opened 2023-04-28
Converting selected components to use px instead of rem - 906

Andrew565:
## Chromatic

https://asteele-uxpin-updates-906--60f9b557105290003b387cd5.chromatic.com

---

## Description
Relates to [#906](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/906).

- Modifying styles on va-button to make better use of style-dictionary-generated SCSS tokens, as well as outputting rem values as px to improve compatibility with UXPin.
- Added v1-font-base-size
- Added v1_rem_to_px function
- Renamed rem_override to v3_rem_to_px function to reduce confusion

## Testing done
Tested in UXPin app and storybook using Brave/Chrome.

## Acceptance criteria
- [X] Still looks the same as it did.
- [X] Renders in UXPin correctly.

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-button/va-button.scss
+++ b/packages/web-components/src/components/va-button/va-button.scss
@@ -1,7 +1,5 @@
 @forward 'settings';
-
 @use 'usa-button/src/styles/usa-button';
-
 @import '../../global/formation_overrides';
 
 :host {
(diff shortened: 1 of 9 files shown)


packages/web-components/src/components/va-button/va-button.scss
```diff
@@ -17,65 +15,78 @@
...
+    border: 0;
+    color: $color-white;
+    cursor: pointer;
+    display: flex;
+    font-family: $font-family-sans;
+    font-size: v1_rem_to_px(1.6rem);
+    font-weight: 700;
+    justify-content: center;
+    line-height: 1;
+    margin: 0.5em 0.5em 0.5em 0;
```
jamigibbs, line 36: `em` doesn't need to be adjusted?

packages/web-components/src/mixins/uswds-error-border.scss
```diff
@@ -3,7 +3,7 @@
 
 :host([error][uswds]:not([error=''])) {
     border-left: 0.4rem solid $v3-color-error-dark; /* USWDS red-60v, $theme-color-error-dark */
```
jamigibbs, line 5: Is it supposed to be all `rem`? If so, this was missed: `border-left: 0.4rem`

Andrew565: That rem is a value which should be addressed with a formation_override entry, as it's specific to the v3 version.

packages/web-components/src/components/va-button/va-button.scss
jamigibbs: Had to google it because I wasn't sure why:

> The `rem` unit is only relative to the document's root element, while the `em` unit is only relative to the immediate parent of the targeted element.

packages/web-components/src/mixins/uswds-error-border.scss
jamigibbs: This is a style override of USWDS so that we can display our own error style. So what is `0.4rem` being based off of? The Formation base font or the USWDS base font? As it is now, it's not being addressed by either a formation override or a v3 override. 

If we're not sure which conversion function to pass it into (v1 or v3), then why don't we just directly change it to whatever pixel value our [Sketch file](https://www.sketch.com/s/610156b6-f281-4497-81f3-64454fc72156/p/447FB6DA-215E-4F15-A71B-94B7686DA614/canvas) says the error border width should be? (I'm seeing it as 4px):

Screenshot 2023-04-28 at 11 50 57 AM

jamigibbs: With that said, I think we need to also change `padding-left: v3_rem_to_px(1rem);` to just `16px` (like the Sketch file also says).

Andrew565: I'm good with updating to what the sketch file says.

Andrew565:
No longer needed.

PR #695, opened 2023-05-08
Add va-card component

allisonlu:
## Chromatic

https://va-card--60f9b557105290003b387cd5.chromatic.com

## Description
Consider this a `va-card` lite web component. The reason why this was developed so quickly is for `va-notification` scaffolding (branch and PR coming soon). Structure of the web component and its story is based off of `va-alert`.

This PR does not include variations of Card, like [USWDS's Card component](https://designsystem.digital.gov/components/card/). I'm not sure if the VA design system team had a vision for `va-card`, but I thought keeping it "lite" will keep it flexible for collaboration and changes. 

See [Sketch link](https://www.sketch.com/s/be292806-8dbf-4201-bfa7-0edbba103347/a/bgmjKmz) for default and box-shadow mockups.

- Completes a task on https://github.com/department-of-veterans-affairs/va.gov-team/issues/57714

## Testing done
- locally, visually on Storybook
- e2e tests pass

## Screenshots
localhost_6006__path=_docs_components-va-card--default

## Acceptance criteria
- [ ]

## Definition of done


packages/web-components/src/components/va-card/va-card.tsx
```diff
@@ -0,0 +1,41 @@
...
+  render() {
+    const {
+      showShadow
+    } = this;
+
+    const classes = classnames('va-card',  {'show-shadow': showShadow});
+
+    return (
+      <Host>
+        <div class={classes}>
```
jamigibbs, line 35: Can you leverage the `<Host>` element for applying the classes? In the dom, this would just be the `<va-card>` element so you should be able to add style directly to it (that would save you from having the extra div).

In CSS, style is applied to the web component top level element like:
 
`:host { }`

Example: https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/src/components/va-omb-info/va-omb-info.css#L1-L3

packages/web-components/src/components/va-card/va-card.tsx
```diff
@@ -0,0 +1,41 @@
...
+          <slot name="content"></slot>
```
jamigibbs, line 36: I wonder if we need to name this slot at all. 🤔  If you just used `<slot></slot>`, then the component could be used without needing any other markup around the content.

```
  <va-card show-shadow={showShadow}>
     <p>Example card content</p>
  </va-card>
```

packages/web-components/src/components/va-card/va-card.tsx
```diff
@@ -0,0 +1,41 @@
...
+  Host, 
+  Prop, 
+  h 
+} from '@stencil/core';
+import classnames from 'classnames';
+
+/**
+ * @componentName Card
+ * @maturityCategory dont_use
+ * @maturityLevel proposed
```
jamigibbs, line 12: I think that we can set this to "Use with Caution: Candidate" [like this other new component](https://design.va.gov/storybook/?path=/docs/components-va-need-help--default) we recently added:

```
 @maturityCategory caution
 @maturityLevel candidate
 ```

File: packages/web-components/src/components/va-card/va-card.scss [hunk 4]
```diff
@@ -0,0 +1,9 @@
+.va-card {
+  border: 1px solid var(--color-gray-medium);
+  background-color: var(--color-white);
+  padding: 1.6rem;
```
jamigibbs, line 4: There is no way you would have been aware of this because we have not documented it but can you convert any `rem` to `px`? 

We are managing two different library versions of USWDS right now and the body font size is set different in each of them so rem's have been really difficult for us to manage during the transition.

packages/web-components/src/components/va-card/va-card.tsx
allisonlu: you're right, we don't need to name it. i figured this out as i was working on `va-notification`. it was extremely difficult to target content within `<slot name={content}></slot>` rather than just `<slot></slot>`

PR #713, opened 2023-06-07
add/convert maintenance banner to a web component

powellkerry:
## Chromatic

https://1508-maintenance-banner--60f9b557105290003b387cd5.chromatic.com

## Description
Convert "Banner - Maintenance" react component to a web component. 
Sketch Design: https://www.sketch.com/s/2476e4ee-fa82-4a19-bf11-dcb9b790acb7/p/D413251A-9268-4E48-9E4C-8E8A3A6B31F9/canvas

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1508

## Testing done
- Browser testing: Chrome, Firefox, Safari, MS Edge
- Mac Voice Over

## Screenshots
Screenshot 2023-06-07 at 3 53 30 PM
Screenshot 2023-06-07 at 3 53 52 PM

## Acceptance criteria
- [ ] Warning shows if current date is before startsAt and after warnStartsAt
- [ ] Alert shows if current date is after startsAt and before expiresAt
- [ ] Maintenance banner does not show if previous 2 conditions are not met
- [ ] Maintenance banner does not show if the user has dismissed the banner

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-maintenance-banner/va-maintenance-banner.scss
+++ b/packages/web-components/src/components/va-maintenance-banner/va-maintenance-banner.scss
@@ -6,6 +6,8 @@
   border-top: 8px solid;
   background-color: var(--color-gray-lightest);
   position: relative;
+  max-width: 1000px;
+  margin: 0px auto;
 }
 
 .maintenance-banner--warning {
@@ -19,6 +21,7 @@
 .maintenance-banner__title {
   font-family: var(--font-serif);
   margin: 0 0 8px;
+  font-size: 20px;
 }
 
 .maintenance-banner__body {
@@ -36,7 +39,7 @@
   position: absolute;
   height: 32px;
   width: 32px;
-  top: 12px;
+  top: 18px;
   left: 18px;
 }
 
--- a/packages/web-components/src/components/va-maintenance-banner/va-maintenance-banner.scss
+++ b/packages/web-components/src/components/va-maintenance-banner/va-maintenance-banner.scss
@@ -1,0 +1,86 @@
+@forward 'settings';
+@use 'uswds-helpers/src/styles/usa-sr-only';
+@use 'usa-button/src/styles/usa-button';
+@use 'usa-icon/src/styles/usa-icon';
+
+@import '../../global/formation_overrides';
+@import '../../mixins/buttons.css';
+
+
+
+.usa-maintenance-banner {
+  border-top: 8px solid;
+  background-color: var(--color-gray-lightest);
+  position: relative;
+}
+
+.usa-maintenance-banner--warning {
+  border-color: var(--color-gold);
+}
+
+.usa-maintenance-banner--error {
+  border-color: var(--color-secondary-dark);
+}
+
+.usa-maintenance-banner__title {
+  font-family: var(--font-serif);
+  margin: 0 0 8px;
+}
+
+.usa-maintenance-banner__body {
+  padding: 16px 56px;
+  position: relative;
+
+}
+
+.usa-maintenance-banner__body::before {
+  font-family: "Font Awesome 5 Free";
+  font-size: 26px;
+  font-style: normal;
+  font-weight: 900;
+  line-height: 1;
+  position: absolute;
+  height: 32px;
+  width: 32px;
+  top: 12px;
+  left: 18px;
+}
+
+.usa-maintenance-banner--warning .usa-maintenance-banner__body::before {
+  content: '\F071'; /* fa-exclamation-triangle */
+}
+
+.usa-maintenance-banner--error .usa-maintenance-banner__body::before {
+  content: '\F06A'; /* fa-exclamation-circle */
+  color: var(--color-secondary-dark);
+}
+
+.usa-maintenance-banner__close {
+  background-color: transparent;
+  color: var(--color-gray-dark);
+  font-size: 2.25rem;
+  padding: 0;
+  position: absolute;
+  margin: 16px;
+  right: 0;
+  top: 0;
+  width: auto;
+  z-index: 9;
+  white-space: nowrap;
+}
+
+.usa-maintenance-banner__close:hover {
+  color: var(--color-base);
+}
+
+.usa-maintenance-banner i::before {
+  -webkit-font-smoothing: antialiased;
+  content: '\F057'; /* fa-times-circle*/
+  display: inline-block;
+  font-family: 'Font Awesome 5 Free';
+  font-style: normal;
+  font-variant: normal;
+  font-weight: 900;
+  line-height: 1;
+  text-rendering: auto;
+}
--- a/packages/web-components/src/components/va-maintenance-banner/va-maintenance-banner.scss
+++ b/packages/web-components/src/components/va-maintenance-banner/va-maintenance-banner.scss
@@ -1,39 +1,33 @@
-@forward 'settings';
-@use 'uswds-helpers/src/styles/usa-sr-only';
-@use 'usa-button/src/styles/usa-button';
-@use 'usa-icon/src/styles/usa-icon';
-
-@import '../../global/formation_overrides';
 @import '../../mixins/buttons.css';
 
 
 
-.usa-maintenance-banner {
+.maintenance-banner {
   border-top: 8px solid;
   background-color: var(--color-gray-lightest);
   position: relative;
 }
 
-.usa-maintenance-banner--warning {
+.maintenance-banner--warning {
   border-color: var(--color-gold);
 }
 
-.usa-maintenance-banner--error {
+.maintenance-banner--error {
   border-color: var(--color-secondary-dark);
 }
 
-.usa-maintenance-banner__title {
+.maintenance-banner__title {
   font-family: var(--font-serif);
   margin: 0 0 8px;
 }
 
-.usa-maintenance-banner__body {
+.maintenance-banner__body {
   padding: 16px 56px;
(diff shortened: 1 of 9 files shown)


packages/web-components/src/components/va-maintenance-banner/va-maintenance-banner.tsx
```diff
@@ -0,0 +1,150 @@
...
+  formatDate,
+  isDateAfter,
+  isDateBefore,
+  isDateSameDay,
+} from '../../utils/date-utils';
+
+/**
+ * @componentName Maintenance Banner
+ * @maturityCategory use
+ * @maturityLevel deployed
```
jamigibbs, line 13: I am not certain of this maturity level. One of the criteria I see for [USE:DEPLOYED](https://design.va.gov/about/maturity-scale#use-deployed) says:
- Is deployed in production for more than 3 months

I wonder if this should be [USE WITH CAUTION: AVAILABLE](https://design.va.gov/about/maturity-scale#use-with-caution-available) for now?

```
 * @maturityCategory caution
 * @maturityLevel proposed
```

packages/web-components/src/components/va-maintenance-banner/va-maintenance-banner.scss
```diff
@@ -0,0 +1,86 @@
+@forward 'settings';
+@use 'uswds-helpers/src/styles/usa-sr-only';
+@use 'usa-button/src/styles/usa-button';
+@use 'usa-icon/src/styles/usa-icon';
+
+@import '../../global/formation_overrides';
```
jamigibbs, line 6: Are we adding a USWDS v3 variation at the same time? These modules would be specific to that but I don't see a `uswds` prop.

```
@forward 'settings';
@use 'uswds-helpers/src/styles/usa-sr-only';
@use 'usa-button/src/styles/usa-button';
@use 'usa-icon/src/styles/usa-icon';

```

packages/web-components/src/components/va-maintenance-banner/va-maintenance-banner.tsx
```diff
@@ -0,0 +1,150 @@
...
+  styleUrl: 'va-maintenance-banner.scss',
+  shadow: true,
+})
+export class VaMaintenanceBanner {
+
+  maintenanceBannerEl: HTMLDivElement;
+  /**
+   * Whether or not an analytics event will be fired.
+   */
+  @Prop() enableAnalytics?: boolean = false;
```
jamigibbs, line 28: I think if we are not wiring up analytics yet, we probably shouldn't advertise that there is analytics being fired. Should we just create a new ticket for completing that task after this component is done? Here is an example issue for adding analytics to va-notification: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1803

packages/web-components/src/components/va-maintenance-banner/va-maintenance-banner.scss
```diff
@@ -0,0 +1,86 @@
...
+  width: auto;
+  z-index: 9;
+  white-space: nowrap;
+}
+
+.usa-maintenance-banner__close:hover {
+  color: var(--color-base);
+}
+
+.usa-maintenance-banner i::before {
```
jamigibbs, line 76: I am not sure about using the `.usa-` namespace for these css classes. I don't think the other components use that (non-v3). Is this because we are taking the styles from Formation?

jamigibbs:
I think something isn't quite right with the Storybook examples. The code example is showing `forwardRef` instead of the component name:

Screenshot 2023-06-08 at 9 34 35 AM

packages/web-components/src/components/va-maintenance-banner/va-maintenance-banner.tsx
```diff
@@ -0,0 +1,150 @@
...
+  isDateAfter,
+  isDateBefore,
+  isDateSameDay,
+} from '../../utils/date-utils';
+
+/**
+ * @componentName Maintenance Banner
+ * @maturityCategory use
+ * @maturityLevel deployed
+ * @guidanceHref form/maintenance-banner
```
jamigibbs, line 14: Should this have a form href reference? I think this is for linking it on the design.va.gov documentation site under the [forms dropdown](https://design.va.gov/components/form/) but I'm not sure if this component will be there.

jamigibbs:
Is it possible that the maintenance content would need to contain links or something more than just plain text? If so, should we consider changing this from a prop to just passing the content into a slot? With the React version, I think it could display HTML so they'd be losing that flexibility with a web component prop.

Screenshot 2023-06-08 at 10 23 43 AM

packages/web-components/src/components/va-maintenance-banner/va-maintenance-banner.tsx
powellkerry: Good question. I am unsure what needs analytics on this component. Should there be an event for whenever a maintenance banner renders and maybe when a user dismisses it? There is not anything else that is actionable on this component.

jamigibbs: I'm guessing we'll want something similar to va-promo-banner or any of the components that capture a close event: https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/src/components/va-promo-banner/va-promo-banner.tsx#L77-L104

packages/web-components/src/components/va-maintenance-banner/va-maintenance-banner.tsx
powellkerry: Updated

packages/web-components/src/components/va-maintenance-banner/va-maintenance-banner.tsx
powellkerry: I added the events, please check my work though :)

packages/web-components/src/components/va-maintenance-banner/va-maintenance-banner.scss
powellkerry: removed the .usa-

packages/web-components/src/components/va-maintenance-banner/va-maintenance-banner.tsx
powellkerry: removed guidanceHref

packages/web-components/src/components/va-maintenance-banner/va-maintenance-banner.tsx
jamigibbs: @powellkerry eeep, I think I pasted the wrong suggestion. The "available" reference should be this:

```
@maturityLevel available
```

jamigibbs:
> @jamigibbs The code preview in storybook should reflect the actual component now.

@powellkerry It does look better now! 

One other thing I noticed with Storybook though is that all of the other banner components are grouped tidy but this new one is alphabetizing lower. I realize the React version is already taking up this component's name "Banner - Maintenance" so I'm not exactly sure the answer here. I kind of think we should also give this web component the title "Banner - Maintenance" to keep it organized but some thoughts on that:

- Can we move the React component to the deprecated section? Not sure if there's some other steps we need to do first before that happens.
- Should we add the word "React" to the react component's title to distinguish it from the web component?
- Maybe we are ok with two components being displayed as "Banner - Maintenance" on Storybook for now?

Screenshot 2023-06-09 at 1 15 28 PM

packages/web-components/src/components/va-maintenance-banner/va-maintenance-banner.tsx
jamigibbs: I think the other component was using `innerContent` because everything was being passed through a slot. We have some props here though so usually we just do something like this to fine tune it:

```
        details: {
          header: this.maintenanceTitle,
          starts: this.startsAt,
          expires: this.expiresAt
        },
```

I don't know, we (DST) can capture whatever we think is valuable. For a maintenance banner, those bits seems useful-ish?

jamigibbs: We will also need to add the component to vets-website here: https://github.com/department-of-veterans-affairs/vets-website/blob/main/src/platform/site-wide/component-library-analytics-setup.js

powellkerry:
@jamigibbs that is a great suggestion. I have updated the stories to be organized better. 

After further investigation I don't know how to allow HTML without loosing the ability to have a different warning message than active maintenance message. We usually allow HTML using "slot", but we can't do that if there are 2 messages that could contain HTML. I'll investigate this a bit more on Monday.

babsdenney (approved):
Design approved!

packages/web-components/src/components/va-maintenance-banner/va-maintenance-banner.tsx
powellkerry: Updated for the component, will work on vets-website PR

packages/web-components/src/components/va-maintenance-banner/va-maintenance-banner.scss
powellkerry: Removed unused CSS imports. Not USWDS yet

packages/web-components/src/components/va-maintenance-banner/va-maintenance-banner.tsx
powellkerry: Updated

jamigibbs (approved):

PR #716, opened 2023-06-12
60122 va select inert

rhasselle-oddball:
## Chromatic

https://60122-va-select-inert--60f9b557105290003b387cd5.chromatic.com

## Description
- department-of-veterans-affairs/va.gov-team#60122
- additional context in this slack thread: https://dsva.slack.com/archives/C01DBGX4P45/p1686238079010729

## Testing done
- Tested in chrome with usdws true and false (will apply to both)
- yarn tests
- Tested in IE11 mode in Edge - did not work perfectly (but I think it's because the options weren't showing up at all, which may be unrelated to inert). Do we need full backward compatibility here?

Questions: 
- Do we need a new storybook example for inert? or is it OK to just include the CSS?
- Do we need a new property for inert? or simply having the CSS is fine when the user passes the native `inert` property?

```
  <va-select
    aria-live-region-text="You selected"
    hint="This is a hint"
    label="Branch of Service"
    name="branch"
    value="navy"
    inert
    usdws
  >
    <option value="navy">
      Navy
    </option>
    <option value="army">
      Army
    </option>
    <option value="marines">
      Marines
    </option>
    <option value="air-force">
      Air Force
    </option>
    <option value="coastguard">
      Coastguard
    </option>
  </va-select>
```

## Screenshots
without inert, usdws
image

with inert, usdws
image

## Acceptance criteria
- [x] added new css
- [ ] do i need to add a new storybook example or use case?

## Definition of done
- [ ] Documentation has been updated, if applicable - do I need to update any documentation?

### DIFF
--- a/packages/storybook/stories/va-select-uswds.stories.jsx
+++ b/packages/storybook/stories/va-select-uswds.stories.jsx
@@ -92,6 +92,56 @@ const Template = ({
   );
 };
 
+const InertTemplate = ({
+  label,
+  name,
+  value,
+  required,
+  error,
+  uswds,
+  hint,
+  'aria-live-region-text': ariaLiveRegionText,
+  options,
+  'use-add-button': useAddButton,
+}) => {
+  const [modifiedOptions, setModifiedOptions] = useState(options);
+
+  return (
+    <>
+      <p>To make this read only add <code>inert</code> to the element.</p>
+      {useAddButton && (
+        <button
+          style={{ fontSize: '16px' }}
+          onClick={() => {
+            setModifiedOptions([
+              ...modifiedOptions,
+              <option key="6" value="new">
+                Something new
+              </option>,
+            ]);
+          }}
+        >
+          Add &quot;Something new&quot;
+        </button>
+      )}
+      <va-select
+        uswds={uswds}
+        label={label}
+        name={name}
+        value={'navy'}
+        required={required}
+        error={error}
+        hint={hint}
+        inert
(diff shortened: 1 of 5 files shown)


Andrew565:
I think adding a storybook story would be a really good idea, there’s no better way for other people to be made aware that this option exists otherwise. Because it’s a native attribute, I don’t think you need to spell it out as a separate property.

rhasselle-oddball:
one discovery is that when using inert, the label also becomes unclickable/selectable. not sure if that will be a deal breaker or not.

packages/storybook/stories/va-select-uswds.stories.jsx
```diff
@@ -92,6 +92,59 @@ const Template = ({
...
+      )}
+      <va-select
+        uswds={uswds}
+        label={label}
+        name={name}
+        value={'navy'}
+        required={required}
+        error={error}
+        hint={hint}
+        inert
```
rhasselle-oddball, line 138: this is same as Template above, but just added this property.

packages/web-components/src/index.html
```diff
@@ -18,6 +18,29 @@
 
   </head>
   <body>
-    <my-component first="Stencil" last="'Don't call me a framework' JS"></my-component>
+    <va-select
```
Andrew565, line 21: Can you undo this change to the index.html file please? Just confusing.

Andrew565 (changes requested):

Andrew565 (approved):

PR #717, opened 2023-06-12
Update va-table to support pagination

jamigibbs:
## Chromatic

https://1821-va-table-data-render--60f9b557105290003b387cd5.chromatic.com

### The Problem
The va-table component did not entirely support pagination or dynamic content being passed into its slots. The component was adding CSS classes and attributes to table cells in the `componentDidLoad` lifecycle. This lifecycle is called only once after the component has loaded. That means that any later updates to the table cells (new data) would not have those classes and attributes applied.

That's why the table style was not consistent when data was being dynamically added to the table. Here is what it looked like before when paginating:

243804061-ad34e3e7-30be-45d6-9c11-5d16d0df0817

### The Solution

❌ **First Approach**

I attempted a couple of different ways to make the data display consistently. First, I tried watching the slot content with `slotchange` as was suggested in the [original draft PR](https://github.com/department-of-veterans-affairs/component-library/pull/408). Unfortunately, the `slotchange` event was not firing consistently. There were still occasions when the cells were not getting the classes or attributes they needed. Possibly because:

> the slotchange event doesn't fire if the children of a slotted node change — only if you change (e.g. add or delete) the actual nodes themselves. [Reference](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/slotchange_event)

I think because we had children in the slotted node (the cells of rows) that weren't always being fully added or deleted, it was not always firing.

❌ **Second Approach**

I also tried using other [component lifecycles](https://stenciljs.com/docs/component-lifecycle) that might trigger when there was a content change (`componentDidUpdate`, `componentWillRender`, `componentDidRender`, etc) but there was not one that was always consistent.

✅ **Third Approach**

This PR is adding a `MutationObserver` to the table body slot to watch for any changes there and process cells with classes or attributes that they need. I found this approach to be the only one that was consistent for watching for all elements within the slot (parent, children, etc) every time there was a data update.

**Demo with MutationObserver**

https://github.com/department-of-veterans-affairs/component-library/assets/872479/5a09a8c4-7395-456b-8f9f-aeda7d4835f9

**Mobile**

Screenshot 2023-06-15 at 12 42 59 PM

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1821

## Testing done
Storybook

## Acceptance criteria
- [ ] The va-table component displays table cell styles consistently when used with pagination.

## Definition of done

### DIFF
--- a/packages/storybook/stories/va-table.stories.jsx
+++ b/packages/storybook/stories/va-table.stories.jsx
@@ -228,14 +228,14 @@ const Pagination = args => {
       >
         <va-table-row slot="headers">
           {columns.map((col, index) => (
-            <span role="cell" key={`table-example-${index}`}>{col}</span>
+            <span key={`table-example-${index}`}>{col}</span>
           ))}
         </va-table-row>
 
         {currentData.map((row, i) => (
-          <va-table-row id={`table-example-${i}`} key={`table-example-${i}`}>
+          <va-table-row key={`table-example-${i}`}>
             {row.map(item => (
-              <span role="cell" key={`${item}-${i}`}>{item}</span>
+              <span key={`${item}-${i}`}>{item}</span>
             ))}
           </va-table-row>
         ))}
--- a/packages/storybook/stories/va-table.stories.jsx
+++ b/packages/storybook/stories/va-table.stories.jsx
@@ -87,7 +87,7 @@ const paginationData = [
     '**3456',
   ],
   [
-    '04/05/2019',
+    '03/05/2019',
     '$3,261.10',
     'Compensation & Pension - Recurring',
     'Direct Deposit',
@@ -95,15 +95,7 @@ const paginationData = [
     '**3456',
   ],
   [
-    '04/05/2019',
-    '$3,261.10',
-    'Compensation & Pension - Recurring',
-    'Direct Deposit',
-    'BANK OF AMERICA, N.A.',
-    '**3456',,
-  ],
-  [
-    '04/05/2019',
+    '02/05/2019',
     '$3,261.10',
     'Compensation & Pension - Recurring',
     'Direct Deposit',
@@ -111,7 +103,7 @@ const paginationData = [
     '**3456',
   ],
   [
-    '04/05/2019',
+    '01/05/2019',
     '$3,261.10',
     'Compensation & Pension - Recurring',
     'Direct Deposit',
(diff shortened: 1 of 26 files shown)


packages/web-components/src/components/va-table/test/va-table.e2e.ts
```diff
@@ -121,7 +121,7 @@ describe('va-table', () => {
   });
 
   it('right aligns columns with numeric data', async () => {
-    const rightAlignClass = 'medium-screen:vads-u-text-align--right';
+    const rightAlignClass = 'text-align-right';
```
jamigibbs, line 124: I am replacing the original Formation styling class with one that is contained within the component itself. This will guard us for when Formation is eventually removed.

packages/storybook/stories/va-table.stories.jsx
```diff
@@ -52,15 +53,15 @@ const Template = args => {
...
         </va-table-row>
 
-        {rows.map(row => (
-          <va-table-row>
-            {row.map(item => (
-              <span>{item}</span>
+        {rows.map((row, index) => (
+          <va-table-row key={`row-default-${index}`}>
+            {row.map((item, index2) => (
+              <span key={`row-item-default-${index2}`}>{item}</span>
```
jamigibbs, line 64: I added keys to mapping elements because we were getting React errors in the console in Storybook without them.

packages/storybook/stories/va-table.stories.jsx
```diff
@@ -75,6 +76,180 @@ const defaultArgs = {
...
+  ],
+  [
+    '03/05/2018',
+    '$3,261.10',
+    'Compensation & Pension - Recurring',
+    'Direct Deposit',
+    'BANK OF AMERICA, N.A.',
+    '**3456',
+  ],
+];
```
jamigibbs, line 192: I chose this demo content because it's the same mock data that was used when the original issue was reported.

packages/web-components/src/components/va-table/va-table-slot.css
```diff
@@ -13,3 +13,9 @@ va-table > va-table-row[slot] button:hover {
 va-table > va-table-row[slot] button svg {
   width: 2.4rem;
 }
+
+@media screen and (min-width: 768px) {
+  va-table-row span.text-align-right {
+    text-align: right!important;
+  }
+}
```
jamigibbs, line 21: This is exactly the same styles as what the previous `medium-screen:vads-u-text-align--right` class was appyling.

packages/web-components/src/components/va-table/test/va-table.e2e.ts
ataker: Love the proactive change!

ataker (approved):

PR #728, opened 2023-06-26
Add va-accordion USWDS v3 variation

powellkerry:
## Chromatic

https://1719-v3-accordion--60f9b557105290003b387cd5.chromatic.com

## Description
Convert existing va-accordion web-component to use USWDS styles and architecture. Maintain backwards compatibility with the previous version.

Sketch Design: https://www.sketch.com/s/a52734dd-00d0-44f1-9c9e-ff4016130e5c/a/OeoJpYv#Inspect
USWDS Design: https://designsystem.digital.gov/components/accordion/

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1719

## Testing done
- Browser testing in Chrome, Firefox, Safari, Edge
- Axe Dev Tools accessibility 
- VoiceOver testing

## Screenshots
Screenshot 2023-06-26 at 11 38 37 AM
Screenshot 2023-06-26 at 11 39 09 AM

## Acceptance criteria
- [x] Maintain backwards compatibility
- [x] Pass accessibility tests

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-accordion/va-accordion-item.scss
+++ b/packages/web-components/src/components/va-accordion/va-accordion-item.scss
@@ -1,4 +1,30 @@
+@forward 'settings';
+
+@use 'usa-accordion/src/styles/usa-accordion';
+
+@import '../../global/formation_overrides';
+
 @import '../../mixins/focusable.css';
+
+
+
+.usa-accordion__header, .usa-accordion__subheader {
+  display: block;
+}
+
+.usa-accordion__header ::slotted([slot=header-icon]), .usa-accordion__subheader ::slotted([slot=subheader-icon]){
+  margin-right: 15px;
+  width: inherit;
+}
+
+.usa-accordion__subheader {
+  font-weight: 400;
+  font-size: 16px;
+  line-height: 26px;
+  margin-top: 2px;
+}
+
+
 
 :host {
   display: block;
--- a/packages/web-components/src/components/va-accordion/va-accordion-item.scss
+++ b/packages/web-components/src/components/va-accordion/va-accordion-item.scss
@@ -7,27 +7,28 @@
 @import '../../mixins/focusable.css';
 
 
-
-.usa-accordion__header, .usa-accordion__subheader {
+/**
+  Custom classes and overrides to add in subheader and icon functionality
+*/
+.va-accordion__header, .va-accordion__subheader {
   display: block;
 }
 
-.usa-accordion__header ::slotted([slot=icon]), .usa-accordion__subheader ::slotted([slot=subheader-icon]){
+.va-accordion__header ::slotted([slot=icon]), .va-accordion__subheader ::slotted([slot=subheader-icon]){
   margin-right: 15px;
   width: inherit;
 }
 
-.usa-accordion__subheader {
+.va-accordion__subheader {
   font-weight: 400;
   font-size: 16px;
   line-height: 26px;
   margin-top: 2px;
 }
 
-.usa-accordion__button {
-  position: relative;
-}
-
+/**
+  Override to use font awesome icons
+*/
 .usa-accordion__button[aria-expanded=false] {
   background-image: url("data:image/svg+xml,%3C%3Fxml version%3D%221.0%22 encoding%3D%22UTF-8%22%3F%3E%3Csvg xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22 width%3D%2264%22 height%3D%2264%22 viewBox%3D%220 0 64 64%22%3E%3Ctitle%3Eplus%3C%2Ftitle%3E%3Cpath fill%3D%22%23171717%22 fill-rule%3D%22evenodd%22 d%3D%22M36 0c1.111 0 2.056.425 2.833 1.272.778.849 1.166 1.88 1.166 3.091V24h19.636c1.125 0 2.095.335 2.907 1.006l.184.16c.849.778 1.272 1.723 1.272 2.834v8c0 1.112-.425 2.057-1.272 2.834-.849.778-1.88 1.166-3.091 1.166H39.999v19.636c0 1.125-.335 2.094-1.006 2.906l-.16.185c-.778.848-1.722 1.272-2.833 1.272h-8.001c-1.111 0-2.056-.424-2.834-1.272-.778-.849-1.166-1.88-1.166-3.09V39.998H4.363c-1.125 0-2.094-.335-2.906-1.006l-.185-.16C.424 38.054 0 37.11 0 36v-8.001c0-1.111.424-2.056 1.272-2.834C2.121 24.387 3.152 24 4.363 24H24V4.363c0-1.125.335-2.094 1.006-2.906l.16-.185C25.944.423 26.889 0 28 0z%22%2F%3E%3C%2Fsvg%3E");
   background-size: 1.5rem;
--- a/packages/web-components/src/components/va-accordion/va-accordion-item.scss
+++ b/packages/web-components/src/components/va-accordion/va-accordion-item.scss
@@ -27,6 +27,14 @@
 }
 
 /**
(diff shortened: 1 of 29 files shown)


packages/web-components/src/components/va-accordion/va-accordion-item.tsx
```diff
@@ -49,6 +50,16 @@ export class VaAccordionItem {
    */
   @Prop() level?: number = 2;
 
+  /**
+   * The id of the accordion item, required for accessibility. (USWDS only)
+   */
+  @Prop() bordered?: boolean = false;
```
jamigibbs, line 56: Is the documentation text correct for this prop?

```
  /**
   * The id of the accordion item, required for accessibility. (USWDS only)
   */
  @Prop() bordered?: boolean = false;
```

packages/web-components/src/components/va-accordion/va-accordion.tsx
```diff
@@ -70,55 +97,71 @@ export class VaAccordion {
...
+        let items = getSlottedNodes(this.el, 'va-accordion-item');
+        items.forEach((item) => {
+          let el = item as Element;
+          let button = el.shadowRoot.querySelector('button.usa-accordion__button') as HTMLButtonElement;
+          if (button.getAttribute('aria-expanded') === 'true') {
+            button.click()
+          }
+        })
+      }
+    } else {*/
```
jamigibbs, line 113: Do we need all of this commented out code?

packages/web-components/src/components/va-accordion/va-accordion-item.scss
```diff
@@ -1,5 +1,56 @@
...
+.usa-accordion__subheader {
+  font-weight: 400;
+  font-size: 16px;
+  line-height: 26px;
+  margin-top: 2px;
+}
+
+.usa-accordion__button {
+  position: relative;
+}
```
jamigibbs, line 29: Are these USWDS style overrides?

```
.usa-accordion__header, .usa-accordion__subheader {
  display: block;
}

.usa-accordion__header ::slotted([slot=icon]), .usa-accordion__subheader ::slotted([slot=subheader-icon]){
  margin-right: 15px;
  width: inherit;
}

.usa-accordion__subheader {
  font-weight: 400;
  font-size: 16px;
  line-height: 26px;
  margin-top: 2px;
}

.usa-accordion__button {
  position: relative;
}
```

packages/web-components/src/components/va-accordion/va-accordion.tsx
```diff
@@ -48,15 +49,41 @@ if (Build.isTesting) {
...
+  /**
+   * If `true`, doesn't fire the CustomEvent which can be used for analytics tracking.
+   */
+  @Prop() disableAnalytics?: boolean = false;
+
+  /**
+   * Optional accordion section heading text. Only used in analytics event. Default is null.
+   */
+  @Prop() sectionHeading?: string = null;
+
```
jamigibbs, line 86: The diff makes it seems like all of these props are new to this file. I can understand the `uswds` one but is that also right for the others?

packages/web-components/src/components/va-accordion/va-accordion.tsx
```diff
@@ -188,30 +222,70 @@ export class VaAccordion {
...
-      <Host>
-        {!this.openSingle && (
-          <button
-            ref={el => (this.expandCollapseBtn = el as HTMLButtonElement)}
-            onClick={() => this.expandCollapseAll(!this.expanded)}
-            aria-label={
-              this.expanded
-                ? i18next.t('collapse-all-aria-label')
-                : i18next.t('expand-all-aria-label')
+    const {uswds = false, openSingle} = this;
```
jamigibbs, line 225: I'm curious why `uswds` has to be set to false here.

jamigibbs:
I think something is a little off with the Font Awesome icon override. It looks thinner/smaller:

v1:
Screenshot 2023-06-27 at 12 42 39 PM

v3:
Screenshot 2023-06-27 at 12 42 43 PM

jamigibbs:
A small PR administration request; can we name this PR the way we have for all of the other USWDS v3 variations for our release notes? ie. Add va-accordion USWDS v3 variation

jamigibbs:
I was expecting to find `.usa-accordion--bordered .usa-accordion__content` in the formation override stylesheet (because it's full of rem's) but I don't see it there. Do we know where this style is being applied to the border? It looks thinner than what's on the USWDS site:

Screenshot 2023-06-27 at 12 51 26 PM

packages/web-components/src/components/va-accordion/va-accordion-item.scss
powellkerry: The uswds styles do not include a subheader and icons for the header and subheader lines. These are the styles to add them.

packages/web-components/src/components/va-accordion/va-accordion.tsx
powellkerry: This was an artifact from development that I forgot to remove.

packages/web-components/src/components/va-accordion/va-accordion.tsx
powellkerry: This is correct-ish. The only new prop is uswds, but I moved the props to the top of the file instead of having them buried beneath the logic at line 160.

packages/web-components/src/components/va-accordion/va-accordion-item.scss
jamigibbs: ah! Ok so I guess what is confusing is that it's using the `.usa-accordion` namespace which makes me think it's actually from the USWDS module. But it's really just custom classes we added for styling features that are VA specific.

This might be the first time we've done something like this (add a new feature to the USWDS variation - subheader and icons in this case) so it's an interesting scenario we will probably come across again in other components.

Will mimicking a USWDS class for custom styling make it more or less confusing when we eventually need to reconcile a component for the USWDS team? @Andrew565 Considering you've done some preliminary work for "giving back" these components, what are your thoughts on how we should do this? Would it be easier to identify our custom features by using generic classes, or, create best-guess names that are similar to what's in the modules (but not actually in the modules)? Do we even know if that team wants any of our "enhancements"?

packages/web-components/src/components/va-accordion/va-accordion.tsx
jamigibbs: Oh yes! They really were buried down there. 😂 Good idea to group them together.

powellkerry:
@jamigibbs Good catch on the content rems. I didn't see that, probably because the v1 styles were still applied. I have fixed it so that the v1 styles are not applied and added the formation overrides.

packages/web-components/src/components/va-accordion/va-accordion.tsx
powellkerry: Removed

packages/web-components/src/components/va-accordion/va-accordion-item.tsx
powellkerry: This was incorrect and I updated it, I also realized that I needed to clean up the unused bordered prop on the va-accordion.tsx file. Great find!

babsdenney (approved):
The design looks great!

babsdenney:
@powellkerry I was looking at this again and noticed that the USWDS sets a max width for the content inside the accordion. Looking at our previous accordion we didn't have a max-width set. This might be a big visual shift in content for teams using the accordion. 

@humancompanion-usds should we be setting a max width for the content within the accordion?

humancompanion-usds:
@babsdenney - Great question. They set it to 64ex to keep it to a reasonable reading length. I think we need a11y to weigh in here. If they would back this change then I'd be all for it despite the repercussions.

jamigibbs (approved):
Great job @powellkerry!

I think we might just want to do a minor version update and then 🚀

Andrew565 (approved):
LGTM, great job.

PR #731, opened 2023-06-28
Prevent screen reader from reading modal twice

ataker:
## Chromatic

https://prevent-double-screen-reader-modal--60f9b557105290003b387cd5.chromatic.com

## Description
Closes [Screen readers are reading modal content out twice #1849](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1849)

## Testing done
Tested with Mac Voiceover in chrome and firefox

## Acceptance criteria
- [ ] Screen reader doesn't repeat twice when opening a modal.

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-modal/va-modal.tsx
+++ b/packages/web-components/src/components/va-modal/va-modal.tsx
@@ -399,18 +399,17 @@ export class VaModal {
       return (
         <Host>
           <div class={wrapperClass} 
-            role={status === 'warning' || status === 'error' ? 'alertdialog' : 'dialog' }
+            role='dialog'
             aria-label={ariaLabel}
             aria-describedby="description"
             aria-modal="true"
-            tabIndex={-1}
           >
             <div class={contentClass}>
               {closingButton}
               <div class={bodyClass}>
                 <div role="document">
                   {modalTitle && (
-                    <h2 class={titleClass} tabIndex={-1} id="heading">
+                    <h2 class={titleClass} tabindex={-1} id="heading">
                       {modalTitle}
                     </h2>
                   )}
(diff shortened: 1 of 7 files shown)


ataker:
I can't get this to consistently work with the proper accessibility tags with the `alertdialog` role. Removing it and all modals will be considered `dialog`.

packages/web-components/src/components/va-modal/va-modal.tsx
```diff
@@ -443,7 +437,7 @@ export class VaModal {
                       <li class="usa-button-group__item">
                         {!unstyled && (
                           <va-button
-                            onClick={e => this.handlePrimaryButtonClick(e)}
+                            onClick={e => this.handleSecondaryButtonClick(e)}
```
ataker, line 440: Not related to the issue, but fixing bug in USWDS modal where both secondary button was performing the action of the primary button.

Andrew565 (approved):
LGTM, as long as it's actually working. Didn't test with a screenreader.

packages/web-components/src/components/va-modal/va-modal.tsx
```diff
@@ -475,8 +469,13 @@ export class VaModal {
       const bodyClass = status ? 'va-modal-alert-body' : 'va-modal-body';
       const titleClass = status ? 'va-modal-alert-title' : 'va-modal-title';
       return (
-        <Host aria-label={ariaLabel} aria-modal="true" role={ariaRole(status)}>
-          <div class={wrapperClass} tabIndex={-1}>
+        <Host>
+          <div class={wrapperClass}
+            role='dialog'
```
jamigibbs, line 474: If we are not toggling between "dialog" and "alertdialog" any longer, is there a utility method we can remove as well?

jamigibbs: Oh my bad! It is removed. :D

ataker:
Feedback I've gotten is that they would prefer the `alertdialog` role with the chance of double reading rather than have everything be `dialog`, so reverting that. This isn't completely fixed, but it does seem isolated to Mac Voiceover and never happens on the first open, so hopefully this will be a rare incident.

PR #733, opened 2023-06-29
va-alert: Update background only functionality and examples

jamigibbs:
## Chromatic

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/908

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/storybook/stories/va-alert-uswds.stories.jsx
+++ b/packages/storybook/stories/va-alert-uswds.stories.jsx
@@ -111,7 +111,6 @@ const Template = ({
 
 const BackgroundOnlyTemplate = ({
   'background-only': backgroundOnly,
-  'show-icon': showIcon,
   'close-btn-aria-label': closeBtnAriaLabel,
   closeable,
   headline,
@@ -123,7 +122,6 @@ const BackgroundOnlyTemplate = ({
         uswds={uswds}
         status="info"
         background-only={backgroundOnly}
-        show-icon={showIcon}
         disable-analytics="false"
         visible="true"
         close-btn-aria-label={closeBtnAriaLabel}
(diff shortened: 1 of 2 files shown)


packages/web-components/src/global/_formation_overrides.scss
```diff
@@ -349,6 +349,9 @@
   .usa-alert--slim .usa-alert__body {
     padding-left: rem-override(3.5rem);;
   }
+  .usa-alert--no-icon .usa-alert__body {
```
jamigibbs, line 352: We don't actually have a "no icon" example even though it's one of the [USWDS examples](https://designsystem.digital.gov/components/alert/) but if we do want to add that for the USWDS variation, I've gone ahead and added the class so that the rem's are converted correctly if we ever do want to add it.

Andrew565 (approved):
LGTM

packages/web-components/src/global/_formation_overrides.scss
Andrew565: I feel like this might be an example of "no longer used code", which we should proactively remove if we no longer need it. I'd rather add styles back in if needed later, than have unused code lingering around 'just in case'.

jamigibbs: @Andrew565  Yeah I mean it wasn't there to begin with so I was proactively _adding_ it. Happy to take it out if we are confident that we won't be using the "Alert with no icon" USWDS variation.

humancompanion-usds: We just discussed in DSC and we are going to suggest to USWDS that they also drop the no icon variation.

PR #751, opened 2023-07-13
va-date & va-memorable-date: Switch to zero pad function

Mottie:
## Chromatic

https://1834-zero-pad--60f9b557105290003b387cd5.chromatic.com

## Description

While monitoring Sentry errors, we noticed quite a large number of [`undefined is not a constructor (evaluating 'new Intl.NumberFormat(j,p)')` errors](http://sentry.vfs.va.gov/organizations/vsp/issues/?project=4&query=is%3Aunresolved+Intl.NumberFormat&statsPeriod=90d) (~3k events in the past 90 days). See [#1834](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1834).

This happens because the [`Intl.NumberFormat` requires modern up-to-date browsers](https://caniuse.com/?search=Intl.NumberFormat), and we have Veterans still using Safari v9. We can't use [`padStart` since it also requires Safari v10+](https://caniuse.com/?search=padStart)

This PR switches from using `intl.NumberFormat` to `slice` - since it works in legacy browsers

## Testing done

Added tests for new `zeroPadStart` date utility function

## Screenshots

N/A

## Acceptance criteria
- [ ] Numbers < 10 will get zero-padded
- [ ] Use method that will work in Safari v9

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-date/va-date.tsx
+++ b/packages/web-components/src/components/va-date/va-date.tsx
@@ -16,6 +16,7 @@ import {
   validate,
   getErrorParameters,
   checkIsNaN,
+  zeroPadStart,
 } from '../../utils/date-utils';
 
 /**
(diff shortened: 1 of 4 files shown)


Andrew565 (approved):
LGTM

PR #754, opened 2023-07-18
Added the mobile team to codeowners and tried to plan directories

timwright12:
The VA Mobile team is running a test to discover what it would take to embed our design system and components inside this repo. For this PR I'm adding the initial files and directories where I think we'll be primarily working and updating CODEOWNERS to reflect that so our PRs won't bother the main design system team.

This would also be replicative of the end result workflow.


Andrew565 (approved):
LGTM

PR #774, opened 2023-07-28
va-memorable-date: Fix hint text not reading for screen readers (uswds variation)

jamigibbs:
## Chromatic

https://1936-memorable-date-hint-a11y--60f9b557105290003b387cd5.chromatic.com

## Description
The PR updates the uswds `hint-text` so that it's placed within the `legend`. Otherwise, the hint text was not getting read out by screen readers.

This also adds an "Extra hint text" Storybook example which was also tested to be reading with screen readers. 

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1936

## Testing done
Storybook
Voice Over Chrome (v1 and v3)

## Screenshots

Screenshot 2023-07-28 at 12 22 30 PM

Screenshot 2023-07-28 at 12 22 44 PM

## Acceptance criteria
- [ ] Hint text is read by screen readers for va-memorable-date

## Definition of done


packages/web-components/src/components/va-memorable-date/va-memorable-date.scss
```diff
@@ -25,6 +25,10 @@
   margin-bottom: 0;
 }
 
+.usa-hint {
+  display: block;
+}
```
jamigibbs, line 30: The [USWDS markup for memorable date](https://designsystem.digital.gov/components/memorable-date/) uses a `span` which naturally displays inline. But our variation has added a `legend` element which the span is now added to. This means that it was displaying in line with the label text.

I added this overriding style to fix that alignment issue but we could also just change the `span` to a `div`. If we feel like that's a better approach instead of style overrides, I can remove this and update the markup.

it-harrison (approved):
looks good!

PR #786, opened 2023-08-04
va-radio-option: Set tab index to -1 for radio button input

ataker:
## Chromatic

https://1961-multiple-tab-presses--60f9b557105290003b387cd5.chromatic.com

## Description
Closes [v3 va-radio button uses more than one tab to get to next focusable item #1961](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1961)

## Testing done
Tested locally in Storybook in Chrome, Firefox, Edge, Safari

## Acceptance criteria
- [ ] Only on tab press to exit focus on radio buttons

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-radio-option/va-radio-option.tsx
+++ b/packages/web-components/src/components/va-radio-option/va-radio-option.tsx
@@ -104,6 +104,7 @@ export class VaRadioOption {
             checked={checked}
             disabled={disabled}
             onClick={() => this.handleChange()}
+            tabIndex={-1}
           />
           <label htmlFor={id} id="option-label" class="usa-radio__label">
             {label}
--- a/packages/web-components/package.json
+++ b/packages/web-components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/web-components",
-  "version": "4.45.3",
+  "version": "4.45.4",
   "description": "Stencil Component Starter",
   "main": "dist/index.cjs.js",
   "module": "dist/index.js",
--- a/packages/web-components/src/components/va-radio-option/test/va-radio-option.e2e.ts
+++ b/packages/web-components/src/components/va-radio-option/test/va-radio-option.e2e.ts
@@ -89,7 +89,7 @@ describe('va-radio-option', () => {
     <va-radio-option uswds="" id="yes2" label="Yes - Any Veteran" name="yes" value="2" class="hydrated">
       <mock:shadow-root>
         <div class="usa-radio">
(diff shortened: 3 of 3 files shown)


packages/web-components/src/components/va-radio-option/va-radio-option.tsx
```diff
@@ -104,6 +104,7 @@ export class VaRadioOption {
             checked={checked}
             disabled={disabled}
             onClick={() => this.handleChange()}
+            tabIndex={-1}
```
jamigibbs, line 107: I'm not super clear about how we manage the tab indexes for va-radio and va-radio-option.

There are a few different functions in `va-radio` that seems to be managing tabindex. I think it's `getSlottedNodes` that is dealing with the va-radio-option. Is there anything we can do with the existing tabindex management in the parent component instead of making all of the `input` elements in va-radio-option set to tab-index="-1"? 

https://github.com/department-of-veterans-affairs/component-library/blob/77ab7cf9553d338b85919bdf2b8ae39c11c8d217/packages/web-components/src/components/va-radio/va-radio.tsx#L185-L211

humancompanion-usds: Also, how does this work in a page where either no tabindex is set on any element or tabindex is set on some other component? Or are we protected because this is in the Shadow DOM? Do we know that this is the root cause of the double-tab bug? Or is this a hunch that setting the tabindex might fix it?

ataker: I'm confident this is the right solution. This updates the `input` inside the `va-radio-option`, while the code in `va-radio.tsx` is updating the `va-radio-option` itself. The code is correctly mimicking the behavior of a vanilla radio button group. You tab into the entire group and use the arrow keys to only select one option, then tab out. This is opposed to most inputs, for example checkbox group, where you tab through each option, able to select them as you go, and need to tab through all before leaving the inputs.

The updating tabindex in this code isn't affecting the tab presses while using the components, it's setting the tabindex so it's correct when you tab into it again.

The bug is because of the way `va-radio` handles the focus. The `va-radio-option` holds the focus, not the `input` itself, so when tab is pressed to leave the radio button group, it finds the `input` in the `va-radio-option` and focuses that, making it seem like nothing is changing in focus and needing the second tab press to leave. 

And yes, the tabindex is encapsulated by the shadow DOM, so this code setting the `va-radio-options` tabindex won't affect tab order on the entire page.

jamigibbs (approved):

PR #790, opened 2023-08-08
va-pagination: Add optional chaining to activeElement blur

jamigibbs:
## Chromatic

https://va-pagination-active-blur--60f9b557105290003b387cd5.chromatic.com

## Description
No ticket - This is from a support request in Slack: https://dsva.slack.com/archives/C01DBGX4P45/p1691419861861739

The issue reported was that when using a filtering mechanism with the va-pagination component, an error started to appear in the console:

image

From that Slack message:

> To reproduce nav to https://staging.va.gov/my-health/secure-messages/inbox/, enter `test` in filter box, click Filter button and then use pagination

> vets.gov.user+41@gmail.com
> 109SsNrLgPv5

I was not able to reproduce this locally but I think the problem might be that when we are removing the focus state with `.blur`, there could be scenarios where an `.activeElement` element does not exist yet. So this adds optional chaining to account for that.

## Testing done
vets website staging
Storybook

## Acceptance criteria
- [ ] The issue that reports blur not being able to initiate with a null element has been addressed.

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-pagination/va-pagination.tsx
+++ b/packages/web-components/src/components/va-pagination/va-pagination.tsx
@@ -74,7 +74,7 @@ export class VaPagination {
     this.pageSelect.emit({ page });
 
     // Reset focus to the active page button.
-    (this.el.shadowRoot.activeElement as HTMLElement).blur();
+    (this.el?.shadowRoot?.activeElement as HTMLElement)?.blur();
 
     if (this.enableAnalytics) {
       const detail = {
--- a/packages/web-components/package.json
+++ b/packages/web-components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/web-components",
-  "version": "4.45.4",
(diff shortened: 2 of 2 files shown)


packages/web-components/src/components/va-pagination/va-pagination.tsx
```diff
@@ -74,7 +74,7 @@ export class VaPagination {
     this.pageSelect.emit({ page });
 
     // Reset focus to the active page button.
-    (this.el.shadowRoot.activeElement as HTMLElement).blur();
+    (this.el?.shadowRoot?.activeElement as HTMLElement)?.blur();
```
jamigibbs, line 77: This line was [originally added](https://github.com/department-of-veterans-affairs/component-library/pull/602) because we had a defect where the selected page was not losing its active style (making it appears like two pages were selected at the same time).

Screenshot 2023-08-08 at 2 27 24 PM

powellkerry (approved):

PR #791, opened 2023-08-08
memorable-date - Fix date validation messages, focus invalid fields (NVDA/screen reader bug)

powellkerry:
## Chromatic

https://1593-memdt-validate-bug--60f9b557105290003b387cd5.chromatic.com

## Description
- Re-order NaN/null date validation so that correct validation messages show when the user enters text instead of a number in fields
- Focus invalid fields after failed validation in the priority month, day, year

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1983

## Testing done
- Mac VoiceOver 
- Local testing done in Chrome, Safari, Edge, and Firefox

## Screenshots
Screenshot 2023-08-08 at 2 23 55 PM

## Acceptance criteria
- [x] Screen reader announces correct error message
- [x] Invalid field is focussed

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-memorable-date/va-memorable-date.tsx
+++ b/packages/web-components/src/components/va-memorable-date/va-memorable-date.tsx
@@ -48,6 +48,11 @@ if (Build.isTesting) {
   shadow: true,
 })
 export class VaMemorableDate {
+
+  monthfield: HTMLElement;
+  dayfield: HTMLElement;
+  yearfield: HTMLElement;
+
   @Element() el: HTMLElement;
   /**
    * Render marker indicating field is required.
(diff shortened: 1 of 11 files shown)


packages/web-components/src/components/va-memorable-date/va-memorable-date.tsx
```diff
@@ -243,6 +262,7 @@ export class VaMemorableDate {
             class='usa-form-group--month-select'
             reflectInputError={error ? true : false}
             value={month ? String(parseInt(month)) : month}
+            ref={(field) => this.monthfield = field}
```
jamigibbs, line 265: I'm not sure that I understand why we need `ref` on the fields. What does this do different than just querySelect'ing the field directly with a class or id? (ie. `this.el.shadowRoot.querySelector`)

jamigibbs: I think I understand why now. Is this so that you can apply focus?

powellkerry: My logic around this is that the reference to the fields is stored once on render, whereas the querySelector would need to query the DOM each time validation is ran. 

I don't know if `ref` is a stencil/react thing and not a thing with native web components? If that is the case maybe we should change it to use querySelector.

powellkerry: I agree. That change has been made.

jamigibbs (approved):

PR #804, opened 2023-08-18
fix scroll bug in va-number-input

it-harrison:
## Chromatic

---

## Description
This PR prevents `va-number-input` values from being influenced by scrolling when under focus. The fix implements a change recommended [here](https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/).  This is a major release because the inputMode prop is removed as being unnecessary. 

Closes [1989](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1989)

## Testing done
local testing in chrome, firefox, edge, safari

## Screenshots

https://github.com/department-of-veterans-affairs/component-library/assets/8867779/cd7d0333-d69c-4d05-a0ef-8126cc396d60

## Acceptance criteria
- [ ] `va-number-input` scroll value does not increment on scroll when in focus

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-number-input/va-number-input.tsx
+++ b/packages/web-components/src/components/va-number-input/va-number-input.tsx
@@ -47,11 +47,6 @@ export class VaNumberInput {
   @Prop() required?: boolean = false;
 
   /**
-   * The inputmode attribute.
-   */
-  @Prop() inputmode?: 'decimal' | 'numeric';
-
-  /**
    * Emit component-library-analytics events on the blur event.
    */
   @Prop() enableAnalytics?: boolean = false;
@@ -148,7 +143,6 @@ export class VaNumberInput {
       label,
       required,
       error,
-      inputmode,
       name,
       max,
       min,
@@ -202,8 +196,9 @@ export class VaNumberInput {
             aria-describedby={ariaDescribedbyIds}
             aria-invalid={error ? 'true' : 'false'}
             id="inputField"
-            type="number"
-            inputmode={inputmode ? inputmode : null}
+            type="text"
+            inputmode="numeric"
+            pattern="[0-9]+(\.[0-9]{1,})?"
             name={name}
             max={max}
             min={min}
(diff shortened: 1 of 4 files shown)


packages/web-components/src/components/va-number-input/va-number-input.tsx
```diff
@@ -46,11 +46,6 @@ export class VaNumberInput {
    */
   @Prop() required?: boolean = false;
 
-  /**
-   * The inputmode attribute.
-   */
-  @Prop() inputmode?: 'decimal' | 'numeric';
```
ataker, line 52: My only concern was that removing the decimal/numeric would remove some sort of validation we were relying on, but `inputmode` only affects what [keyboard is displayed on digital](https://css-tricks.com/everything-you-ever-wanted-to-know-about-inputmode/). So 👍

ataker (approved):

packages/web-components/src/components/va-number-input/va-number-input.tsx
jamigibbs: I think that's a good callout because I see a lot of instances of this prop being used in vets-website. Besides leaving behind some cruft, do we think this should be considered a breaking `major` change?

Screenshot 2023-08-18 at 2 25 51 PM

jamigibbs:
@it-harrison Just a gentle nudge here to see if you had a chance to look at this a little more about removing the `inputmode` prop. And if there is a follow up ticket, can you link to it on this PR?

https://dsva.slack.com/archives/C04PAU4GTRP/p1692724455104579

it-harrison:
> @it-harrison Just a gentle nudge here to see if you had a chance to look at this a little more about removing the `inputmode` prop. And if there is a follow up ticket, can you link to it on this PR?
> 
> https://dsva.slack.com/archives/C04PAU4GTRP/p1692724455104579

Hi @jamigibbs I think the right answer is to add the prop back. It's purpose is to hint to the browser which virtual keyboard to show on mobile. The only difference between "numeric" and "decimal" is that the "." is an option in "decimal" which is important. From my experiments allowing this prop doesn't alter the fix that was the purpose of this ticket. And by keeping the prop we won't need to alter any `vets-website` code. I will open a separate PR to add it back since this PR is closed.

## with `inputmode` = `numeric`

Simulator Screenshot - iPhone SE (3rd generation) - 2023-08-23 at 13 36 52

---

## with `inputmode` = `decimal`


Simulator Screenshot - iPhone SE (3rd generation) - 2023-08-23 at 13 37 07

jamigibbs:
@it-harrison Very nice investigation! I can see how that without having the decimal option on keyboards, it would be a pretty big blocker for users in some situations. I agree if it doesn't add revert the scrolling issue you were fixing here, it sounds like we're good to add it back in. 👍

PR #806, opened 2023-08-18
va-memorable-date & va-text-input: Remove unnecessary input message 

jamigibbs:
## Chromatic

https://1965-mem-date-hint-text--60f9b557105290003b387cd5.chromatic.com

## Description
This PR will remove the max length and char count message that displays below the va-text-input field when that component is being used in va-memorable date.

It was discussed during sprint planning that this text was never intended to be displayed in va-memorable-date because it is redundant. va-memorable-date includes the same input guidance in the existing hint text.

Screenshot 2023-08-18 at 11 57 48 AM

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1965

## Testing done
Storybook

## Screenshots

**Before**

Screenshot 2023-08-18 at 12 02 05 PM

**After**

Screenshot 2023-08-18 at 12 04 28 PM

## Acceptance criteria
- [ ]

## Definition of done


packages/web-components/src/components/va-text-input/va-text-input.tsx
```diff
@@ -243,14 +243,15 @@ export class VaTextInput {
       charcount
     } = this;
 
+    const isMessageDisplayed = !this.el.classList.contains('memorable-date-input') as boolean;
```
jamigibbs, line 246: I experimented with a few different ways to figure out if the component was being used as a child to va-memorable-date.

- `this.el.closest('va-memorable-date');`
  - this doesn't guarantee we aren't going to find another memorable date component somewhere else on the page which would give a false positive.
- `this.el.parentElement`
  - the direct parent element isn't the va-memorable-date component so we'd either have to keep chaining the parent element (this.el.parentElement.parentElement) until we found it, or, hook into another distinctly memorable date class (which could eventually be changed)

Both of these methods felt brittle. Explicitly putting a class on the text input component in memorable date guaranteed we'd find it.

But if anyone knows of a way to do this without having to add the new class to va-memorable-date, I'd love to try it out.

packages/web-components/src/components/va-text-input/va-text-input.tsx
```diff
@@ -312,12 +313,12 @@ export class VaTextInput {
               {messageAriaDescribedby}
             </span>
           )}
-          {!charcount && maxlength && value?.length >= maxlength && (
-              <span class={messageClass} aria-live="polite">
+          {isMessageDisplayed && !charcount && maxlength && value?.length >= maxlength && (
```
jamigibbs, line 316: Originally I was thinking we could just do this with css `display: none` but that doesn't remove the element from the DOM. Adding this to the display condition removed it completely.

packages/web-components/src/components/va-text-input/va-text-input.tsx
```diff
@@ -370,13 +371,13 @@ export class VaTextInput {
               {messageAriaDescribedby}
             </span>
           )}
-          {maxlength && value?.length >= maxlength && (
-            <small part="validation" aria-live="polite">
+          {isMessageDisplayed && maxlength && value?.length >= maxlength && (
+            <small id="maxlength-message"  part="validation" aria-live="polite">
```
jamigibbs, line 375: Adding id's for testing.

packages/web-components/src/components/va-text-input/va-text-input.tsx
ataker: My only other thought would be to pass it as a prop rather than have it check for a class. But then we would be officially supporting the feature which I'm not sure is something we want to do.

My other thought is if we keep it as a class check, is `memorable-date-input` descriptive of what it's doing? It feels like we're trying to hide a small hack in here rather than have a descriptive class name.

jamigibbs: > My only other thought would be to pass it as a prop rather than have it check for a class. But then we would be officially supporting the feature which I'm not sure is something we want to do.

Yes, I was having the same conflict with that but this is such a specific use case (a case that only us in the design system would be dealing with) that making an entire prop for it seemed too public.

> is memorable-date-input descriptive of what it's doing?

I guess I was thinking this class name would be very general just to "identify this as an input in memorable date" so `memorable-date-input` seemed ok. I didn't want to get too specific in case we needed to use it for other reasons. I'm not sure what that reason would ever be though.

But I did spend way too much time thinking about how to name _this_ variable though: `isMessageDisplayed`

I don't love that it's super generic because it's being assigned based on a specific class in a totally different component but I also didn't want to make it just about va-memorable-date (inside of va-text-input) in case we wanted to toggle the messages for other things? Naming is hard. 

```js
const isMessageDisplayed = !this.el.classList.contains('memorable-date-input') as boolean;
```

harshil1793 (approved):

PR #817, opened 2023-08-25
va-checkbox | Add description id to aria-describedby

Mottie:
## Chromatic

https://56080-checkbox-description--60f9b557105290003b387cd5.chromatic.com

## Description
See [#56080](https://github.com/department-of-veterans-affairs/va.gov-team/issues/56080)

Our Supplemental Claim form has a large amount of legal text (8 paragraphs plus a bullet list) which would not work because the `message-aria-describedby` only accepts strings. So we moved the wall of content inside the description slot, but the screen reader still won't read out the content, so this PR adds a `description` ID to the input's `aria-describedby` inside the shadow DOM. It was tested and the screen reader does read out the slotted content (see screenshot)

Additionally, the `error-message` ID on the span was changed to `checkbox-error-message` so that the USWDS and original checkbox code better matches. This doesn't look like it will make any tests in vets-website fail because React-Testing Library doesn't support shadow DOM and appears to only check the outer `error` attribute, and looking through Cypress E2E tests reveals label checks more than anything else.

## Testing done

Updated tests to check `aria-describedby` IDs

## Screenshots

va-checkbox with Voiceover window showing the checkbox label and description being read out

## Acceptance criteria
- [x] Descriptions read out by screen reader
- [x] Error message ID consistency
- [x] All tests passing 

## Definition of done


packages/web-components/src/components/va-checkbox/va-checkbox.tsx
```diff
@@ -188,11 +191,15 @@ export class VaCheckbox {
         'usa-legend': true,
         'usa-label--error': error
       });
-      const ariaDescribedbyIds = `${messageAriaDescribedby ? 'input-message' : ''} ${error ? 'checkbox-error-message' : ''}`
-      .trim() || null; // Null so we don't add the attribute if we have an empty string
+      const ariaDescribedbyIds = [
+        messageAriaDescribedby ? 'input-message' : '',
+        error ? 'checkbox-error-message' : '',
+        hasDescription ? 'description' : '',
```
jamigibbs, line 197: Would there ever be a scenario when we wouldn't want the description content associated with aria-describedby?

Mottie: I think if no description is provided?

Mottie: Oh, I see what you're saying... Monday was really cray-cray for me. I can't think of any scenario like that. If it ever came up, we could modify the code to accommodate the need.

jamigibbs (approved):

PR #848, opened 2023-09-14
1884 continue back button bug fix

harshil1793:
## Chromatic

---

## Description
Closes [#1884](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1884)

## Screenshots
V1:
Screenshot 2023-09-14 at 11 12 42 AM
V3:
Screenshot 2023-09-14 at 11 22 34 AM

## Acceptance criteria
- [ ]

## Definition of done


harshil1793:
I just realized I could rename class name from `button-group` to `continue-button-group` to differentiate it from other versions.

powellkerry (approved):
This solution looks good. Are we sure this is needed? @rsmithadhoc is the element ordering an issue for assistive technologies/is there a way to tell them what order to announce the elements like `tabindex`?

rsmithadhoc:
> This solution looks good. Are we sure this is needed? @rsmithadhoc is the element ordering an issue for assistive technologies/is there a way to tell them what order to announce the elements like `tabindex`?

Yeah, you would want the visual presentation to match the DOM. Any assistive tools that pull the DOM and give you a list of elements would be out of order. `tabindex="1"` (or any number greater than 1) should only be avoided for that same reason, it creates a tab order that is different from the visual or screen reader order. Typically, you want to have the elements in order in the DOM and visually as normal, then let the browser handle the tab order and everything works as expected.

That being said, I question why we are putting the Continue button on top of the Back button in the mobile view. I'd expect a left to right control to be converted to a top to bottom control and the browser handles that. Switching these around in a mobile view creates issues (as we are seeing here). I get that Continue is the primary action, but forcing it to be the first item in the mobile view opens the door to these ordering problems.

rsmithadhoc (approved):
The solution looks good, I think my previous comment shouldn't hold this fix up, but something for us to consider in the future.

humancompanion-usds:
To Ryan's point, the US System does keep the same visual order and DOM order. This is a deviation we should eliminate. That's what we decided in the bug.

harshil1793:
@humancompanion-usds After analyzing the comments on the issue, I've come up with this solution. My aim was to take into account both your suggestions regarding the "Back" button and any future considerations, while also addressing the specific issue raised. As of now, we are not aligning the button order with USWDS standards. Should we aim to do that as part of this fix?

humancompanion-usds:
> As of now, we are not aligning the button order with USWDS standards. Should we aim to do that as part of this fix?

Yes, we should align with the US System unless we have a good reason to not do so. If there is a good reason, let me know what that is. I'm not aware of one. Otherwise, we should match their design. Thanks.

harshil1793:
@humancompanion-usds  Before making that change, I have a question about the other two button pairs, 'Yes' and 'Update.' Would it make sense to standardize all three pairs for consistency, or should we keep them as they are?

PR #851, opened 2023-09-14
va-telephone: Change extension prop from number to string type

jamigibbs:
## Chromatic

https://va-telephone-ext-type--60f9b557105290003b387cd5.chromatic.com

## Description
When deprecating the React Telephone component for va-telephone, I was getting a failed test on vets-website because it was checking for an extension of `0000` but the web component was not displaying it.

The `extension` prop was being set correctly but the web component wasn't able to accept the value `0000`. In VS Code, a linting error was showing as:

> Octal literals are not allowed. Use the syntax '0o0'.ts(1121)

Screenshot 2023-09-14 at 4 11 30 PM

I was able to verify this issue in Storybook locally.

I tried using `Number` and `parseInt` to make sure it was getting passed as a number type but it still didn't work _specifically_ for the value `0000`. Other numbers and numeric strings worked fine.

What seems to fix this is changing the `extension` prop type from number to string. This is the same as what the `contact` prop is as well.

related PR https://github.com/department-of-veterans-affairs/vets-website/pull/25632

## Testing done
Storybook

## Screenshots

Screenshot 2023-09-14 at 4 49 40 PM

## Acceptance criteria
- [ ] The extension prop works when passing it `0000` as well as any other extension variation

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-telephone/va-telephone.tsx
+++ b/packages/web-components/src/components/va-telephone/va-telephone.tsx
@@ -25,9 +25,9 @@ export class VaTelephone {
   @Prop() contact!: string;
 
   /**
-   * Optional phone number extension
+   * Optional numeric string phone number extension
    */
-  @Prop() extension?: number;
+  @Prop() extension?: string;
 
   /**
    * Indicates if the phone number can be clicked or not
@@ -95,7 +95,7 @@ export class VaTelephone {
    */
   static formatPhoneNumber(
     num: string,
-    extension: number,
+    extension: string,
     international: boolean = false,
     vanity: string,
     tty: boolean = false,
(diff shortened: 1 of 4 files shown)


packages/web-components/src/components/va-telephone/va-telephone.tsx
```diff
@@ -25,9 +25,9 @@ export class VaTelephone {
   @Prop() contact!: string;
 
   /**
-   * Optional phone number extension
+   * Optional numeric string phone number extension
    */
-  @Prop() extension?: number;
+  @Prop() extension?: string;
```
jamigibbs, line 30: The `contact` prop (for the main phone number) is also type `string`: 

https://github.com/department-of-veterans-affairs/component-library/blob/12df9ae4d584652cec9dfe3ef3cbcc0540831395/packages/web-components/src/components/va-telephone/va-telephone.tsx#L22-L25

micahchiang (commented):
Not that anyone would intentionally do this, but should we be worried at all about people inadvertently entering letters for their extensions?

Screenshot from 2023-09-14 15-10-32

jamigibbs:
> We could add some kind of extra check to make sure whatever is being passed into the prop evaluates to a number before we display it?

This `0000` value is super strange. If add a validation like this:

```css
      if (extension) {
        const ext = Number(extension);
        if (!isNaN(ext)) formattedNum = `${formattedNum}, ext. ${ext}`;
      }
```
It works fine for disallowing words but passing in "0000", displays like this.

Screenshot 2023-09-15 at 9 06 30 AM

JavaScript (or TypeScript?) is not interpreting that value in a way I would expect. 🤔

micahchiang:
@jamigibbs - Hah I suppose extensions could be letters. I'm thinking of those old 1-800-XXXX numbers from infomercials etc. I guess words are easier to remember than numbers :shrug: 

The `Number('0000')` thing is interesting. I did some digging, and `Number()` attempts to [parse a string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number#number_coercion) as if the string contains a [number literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#numeric_literals). I guess, because the string '0000' technically contains all numbers smaller than 8, it gets interpreted as an [octal number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Numbers_and_dates#octal_numbers). (I didn't know all this, I just read up on it lol)

micahchiang (approved):

PR #855, opened 2023-09-19
bump core version number for out of band release

micahchiang:
## Description
Updates core version for release.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@department-of-veterans-affairs/component-library",
   "description": "VA.gov component library. Includes React and web components.",
-  "version": "21.0.0",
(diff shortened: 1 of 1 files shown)


packages/core/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@department-of-veterans-affairs/component-library",
   "description": "VA.gov component library. Includes React and web components.",
-  "version": "21.0.0",
+  "version": "21.1.0",
```
jamigibbs, line 4: Should this be a patch? `21.0.1` I think the two PRs this includes are just patches.

micahchiang: Updated!

jamigibbs (approved):
🚀

PR #856, opened 2023-09-19
va-number-input: 2052 default number validation message

powellkerry:
## Chromatic

https://2052-number-validation--60f9b557105290003b387cd5.chromatic.com

## Description
Add default validation to number field. Validation message should only show when the `error` prop is not defined and when the entered value is not valid.

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2052

## Testing done
Local testing in Chrome, Edge, Safari, and Firefox

## Screenshots
Screenshot 2023-09-19 at 11 08 13 AM

## Acceptance criteria
- [ ] Default error message should show when field is invalid and no error has been supplied to the component

## Definition of done


jamigibbs:
Do we visually want the validation error to look like what it does when `error` is true (with the left and input border)?

Screenshot 2023-09-19 at 3 01 04 PM

Screenshot 2023-09-19 at 3 01 10 PM

packages/core/src/i18n/translations/en.js
```diff
@@ -31,6 +31,7 @@ export default {
   'date-hint': 'Please enter 2 digits for the month and day and 4 digits for the year',
   'date-hint-with-select': 'For example: January 19 2000',
   'date-error': 'Please enter a complete date',
+  'number-error': 'Please enter a valid number',
```
jamigibbs, line 34: Can we add the new translation string(s) to our translations spreadsheet? https://docs.google.com/spreadsheets/d/1iIg9pP5GG-QFk7DrsyoKJv9GebtYmDuzKf_heqxWks4/edit#gid=0

I'm also not completely sure if the Spanish and Tagalog are correct.

packages/web-components/src/components/va-number-input/test/va-number-input.e2e.ts
```diff
@@ -378,5 +378,35 @@ describe('va-number-input', () => {
...
+    await page.setContent('<div><va-number-input uswds label="test input" /></div>');
+    const inputEl = await page.find('va-number-input >>> input');
+
+    await inputEl.press('a');
+    await inputEl.press('b');
+    await inputEl.press('c');
+    await inputEl.press('Tab');
+
+    const label = await page.find('va-number-input >>> label');
+    expect(label.getAttribute('class')).toContain('usa-label--error')
```
jamigibbs, line 396: You should be able to check that a specific translation text is loaded by matching the translation key. Something like this:

```js
expect(label).toEqualText('number-error');
```

packages/web-components/src/components/va-number-input/test/va-number-input.e2e.ts
```diff
@@ -378,5 +378,35 @@ describe('va-number-input', () => {
...
+  })
+
+  it.skip('should not show default validation message when error prop is defined', async () => {
```
jamigibbs, line 399: If these are uswds specific tests, can we add "uswds" to the name of the tests like we have the others?

jamigibbs: And to that point, should we also add identical tests for when the `uswds` prop is not set?

packages/core/src/i18n/translations/en.js
powellkerry: I added that to the spreadsheet, I don't really know what all the fields mean so please double check my work. I used google translate for the Spanish and Tagalog translations, so they are probably wrong. Is there a process for translations to follow/translators to verify strings?

packages/web-components/src/components/va-number-input/test/va-number-input.e2e.ts
powellkerry: Done

packages/web-components/src/components/va-number-input/test/va-number-input.e2e.ts
powellkerry: updated

powellkerry:
@jamigibbs I think I have addressed all the feedback. Can you re-review?

jamigibbs (approved):
I think this looks good to me but until we're certain that the Spanish and Tagalog text is correct, I'd probably just remove it for now and create a follow up ticket to get that content.

PR #859, opened 2023-09-20
Va-pagination: add handlers to uswds pagination component

it-harrison:
## Chromatic

https://add-handlers-uswds-pagination--60f9b557105290003b387cd5.chromatic.com

---

## Description
This PR adds event handlers to the anchor tags in the uswds version of the pagination component so that the component can function.
Closes [65926](https://github.com/department-of-veterans-affairs/va.gov-team/issues/65926)

## Testing done
local testing

## Acceptance criteria
- [ ]

## Definition of done


packages/web-components/src/components/va-pagination/va-pagination.scss
```diff
@@ -5,6 +5,10 @@
 @import '../../mixins/focusable.css';
 @import '../../global/formation_overrides';
 
+:host([uswds]) a.usa-pagination__next-page > div, a.usa-pagination__previous-page > div {
+  display: flex;
```
jamigibbs, line 9: This seems like it might be a divergence from the USWDS styles. Can you just mention a little bit about why we need to do something different from the prev/next styles that what's in the USWDS example? 

https://designsystem.digital.gov/components/pagination/

it-harrison: @jamigibbs The markup is a little different in this component than in USWDS due to the issues with directly including SVG in the render method -- in this component the SVG sits inside a container div. Without this css tweak the container div has a height that is bigger than the SVG which results in the SVG and the text not being vertically centered. Previously I got rid of the container div but after adding handlers I realized doing this causes problems with rendering so this seemed the simplest fix.

jamigibbs (approved):

PR #866, opened 2023-09-25
va-radio: Add aria-describedby to header

jamigibbs:
## Chromatic

https://2061-va-radio-aria--60f9b557105290003b387cd5.chromatic.com

## Description

This PR adds a new prop `header-aria-describedby` so that additional context from outside of the web component can be read by screen readers from inside the web component when the header element is put into focus. This is a specific ask by the forms library team who are building a form using "one thing per page" and the header of the component is put into focus like this:

270711114-f9dace19-7408-4d6c-8778-32c34bbde9d6

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2061

## Testing done
Storybook; VoiceOver

## Screenshots

**Default/v1 variation**

Screenshot 2023-09-26 at 12 03 45 PM

**USWDS variation**

Screenshot 2023-09-26 at 12 22 05 PM

## Acceptance criteria
- [ ] additional context can be applied to the focused header for screen readers

## Definition of done


packages/web-components/src/components/va-radio/va-radio.tsx
```diff
@@ -74,6 +74,13 @@ export class VaRadio {
    * Insert a header with defined level inside the label (legend)
    */
   @Prop() labelHeaderLevel?: string;
+
+  /**
+   * An optional message that will be read by screen readers when the header is focused. The label-header-level
+   * prop must be set for this to be active.
+   */
+  @Prop() headerAriaDescribedby?: string;
```
jamigibbs, line 82: I've made this specific to the (optional) header but I had been thinking about **what if** this is wanted or useful when that isn't set. Just the default `legend` text. 

That opened up a can of worms for me because it seemed like a header tag within a legend behaved in different ways when trying to connect it with additional context (using aria-describedby). It seemed like I would instead need to use `aria-labelledby` on the `fieldset` (this doesn't work with the header tag though).

I also didn't want to try and solve for something that wasn't actually needed or wanted so I'm adding this as something that is _only_ available with `label-header-level` (which was the original ask).

**Markup With header level prop**

```html
<fieldset>
  <legend>
    <h1>My label text</h1>
  </legend>
</fieldset>
```

**Markup Without header prop**

```html
<fieldset>
  <legend>
    My label text
  </legend>
</fieldset>
```

jamigibbs: If we do want that additional screen reader text for both header and non-header text, an alternate approach could be to do something like append the header text and the legend text with the additional string (setting `sr-only` on it) instead of using any `aria-` variation. Something like:

```js
{HeaderLevel ? (
  <HeaderLevel part="header">
    {label} <span class="sr-only">{headerAriaDescribedby}</span>
  </HeaderLevel>
) : (
  {label} <span class="sr-only">{headerAriaDescribedby}</span>
)}
```

That's what this article is saying was their solution:

https://blog.tenon.io/accessible-validation-of-checkbox-and-radiobutton-groups/

rsmithadhoc (approved):
I tested this out with NVDA and I think the implementation looks good. One question I have is if they are going to be able to apply the focus to the heading like their example, I think it's possible but maybe not as trivial as a non-shadow DOM element? Maybe something for them to figure out. I've given this my approval. 

However, I will say I don't love the idea of having a heading inside the radio component to begin with. It seems like added complexity/maintenance to our component when this pattern should be achievable without our intervention. But that was pre-existing, so I'll start a separate discussion as to not hold this up.

jamigibbs:
Thanks for your feedback @rsmithadhoc.

> However, I will say I don't love the idea of having a heading inside the radio component to begin with. It seems like added complexity/maintenance to our component when this pattern should be achievable without our intervention. But that was pre-existing, so I'll start a separate discussion as to not hold this up.

There is an issue in our backlog about adding this header option to other components too (currently on pause) so that might be a good place to continue your feedback about it: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2078

PR #871, opened 2023-09-26
va-memorable-date, va-select - Accessibility updates from staging review

powellkerry:
## Chromatic

## Description
- va-memorable-date: move error text inside of legend to ensure that it is read by screen readers
- va-select: Fix bug with aria-describedby where the element id was incorrect for the USWDS version, added a fallback to the help text if it is defined

Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/64284

## Testing done
- Local testing via chrome and Mac VoiceOver

## Screenshots
- No visual changes

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-memorable-date/va-memorable-date.tsx
+++ b/packages/web-components/src/components/va-memorable-date/va-memorable-date.tsx
@@ -284,16 +284,17 @@ export class VaMemorableDate {
               {required && <span class="usa-label--required"> {i18next.t('required')}</span>}
               {hint && <div class="usa-hint" id="hint">{hint}</div>}
               <span class="usa-hint" id="dateHint">{hintText}</span>
+              <span id="error-message" role="alert">
+                {error && (
+                  <Fragment>
+                    <span class="usa-sr-only">{i18next.t('error')}</span>
+                    <span class="usa-error-message">{i18next.t(error, errorParameters(error))}</span>
+                  </Fragment>
+                )}
+              </span>
             </legend>
             <slot />
-            <span id="error-message" role="alert">
-              {error && (
-                <Fragment>
-                  <span class="usa-sr-only">{i18next.t('error')}</span>
-                  <span class="usa-error-message">{i18next.t(error, errorParameters(error))}</span>
-                </Fragment>
-              )}
-            </span>
+            
             <div class="usa-memorable-date">
               {monthDisplay}
               <div class="usa-form-group usa-form-group--day">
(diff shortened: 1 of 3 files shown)


jamigibbs (approved):

briandeconinck:
@powellkerry Testing with NVDA, there's something that's preventing the error message from being read when tabbing into the memorable date component. Looking at it with @jeana-adhoc it looks like it might be something connected to the `role="alert"` on the error message, which is a little concerning. I'm going to try to dig into this a bit next week to see if we can figure out the right way forward here.

cc @rsmithadhoc

rsmithadhoc:
I looked into getting the error to read when re-tabbing to the field. I wasn't able to get it to work with the current markup.

One possible idea is utilizing the input field's error message, possibly making it sr-only so it doesn't misalign the fields. I don't know if that is a great solution.
image

I'll try some more things today and let you know if I come up with anything better.

rsmithadhoc:
@powellkerry The other way I was able to get this to read out was by setting `shadow: false` on `va-text-input` and putting `aria-describedby="error-message"` on the inner `<input>` tag itself. It seems like other methods are not able to see outside of the `va-text-input` shadow DOM and read out that error message. Using `aria-describedby` with an ID of the error message is the way I've seen it done before, and that works perfectly in plain HTML, but our setup complicates this.

image
image

However, I'm not sure if it is feasible to disable the shadow DOM for the input (or if it is even possible to do it as a one-off for memorable date). It definitely had some styling issues. Because of that, I think my first comment still may be the better choice, unless you know of any workaround.

powellkerry:
@rsmithadhoc @jamigibbs I have updated memorable date to pass validation errors to the individual va-text-input error props. It seems like a bit of hack, but I also added the `sr-only` class on each of the va-text-input fields' errors when the memorable date component loads. In my testing the error is now read on each invalid field when focussed.

packages/web-components/src/components/va-memorable-date/va-memorable-date.tsx
```diff
@@ -192,6 +192,14 @@ export class VaMemorableDate {
   })
   componentLibraryAnalytics: EventEmitter;
 
+  componentDidLoad() {
+    // We are setting the error on each va-text-input for screen readers, but do not want to show it visually. 
+    const textInputs = this.el.shadowRoot.querySelectorAll('va-text-input');
+    textInputs.forEach((input) => {
+      input.shadowRoot.querySelector('#input-error-message').classList.add('sr-only');
+    });
```
jamigibbs, line 200: I'm ok with this approach. I think trying to add a conditional `sr-only` directly in the child components would make things more complicated than they need to be.

I just want to double check that we don't need to do this also with va-select when the month dropdown is used?

rsmithadhoc:
@powellkerry Definitely seems hacky, but might be the only way due to the shadow DOM. 

I tested with NVDA, it worked perfectly. But with VoiceOver, it was also reading the error on the month field (for an error with day). It is picking that up from inside the `<legend>`. So I think the error display under the hint text may need to be outside of the `<legend>`.
image

image

Other than that, I think the solution looks good.

powellkerry:
@rsmithadhoc I moved the error span back outside of the legend element. Let me know if that is better or worse.

rsmithadhoc (approved):
Looks good, thanks!

packages/web-components/src/components/va-memorable-date/va-memorable-date.tsx
powellkerry: I have added this solution to the month drop down, it also seems that validation was failing on the month drop down. I have added a quick fix for that so that the value of "0" will not pass validation.

PR #880, opened 2023-09-29
va-modal: require modalTitle

powellkerry:
## Chromatic

https://2087-require-modal-title--60f9b557105290003b387cd5.chromatic.com

## Description
This change is to make the modal title required. Not having a title was causing the word 'undefined' used as part of ids, and was causing issues with accessibility

- Require modalTitle
- Update storybook to remove example of modal without a title
- 
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2087

## Testing done
Local testing in storybook

## Screenshots
Prop updated to be required:

Screenshot 2023-09-29 at 2 03 34 PM

No other visual changes

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/storybook/stories/va-modal-uswds.stories.jsx
+++ b/packages/storybook/stories/va-modal-uswds.stories.jsx
@@ -169,9 +169,6 @@ WithoutButtons.args = {
   'secondary-button-text': undefined,
 };
 
-export const WithoutTitle = Template.bind(null);
-WithoutTitle.args = { ...defaultArgs, 'modal-title': undefined };
-
 export const WithNestedWebComponents = ({
   'click-to-close': clickToClose,
   'disable-analytics': disableAnalytics,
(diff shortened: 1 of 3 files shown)


jamigibbs (approved):
Would this cause any issues when we deploy to vets-website? Like do we know if there are many va-modal's out there that don't have a title set?

powellkerry:
Looks like there are 3 instances where a modal does not have a title:
src/applications/claims-status/components/AddFilesForm.jsx
src/applications/static-pages/dependency-verification/components/dependencyVerificationModal.jsx
src/platform/site-wide/announcements/components/HomepageRedesignModal.jsx

I will contact those code owners. From my testing the component does break in storybook when the title isn't set, but it may cause build errors.

jamigibbs:
> Looks like there are 3 instances where a modal does not have a title

Maybe for our own administrative house keeping stuff we should label this a `major` change for our release notes because I guess technically this is an API change for the component.

powellkerry:
With further feedback I am closing this PR and will create another for:
- Add a condition that if modalTitle is not set or is set to an empty string, do not include aria-label="undefined modal" or aria-label=" modal" at all.

PR #886, opened 2023-10-05
2076 Fix V3 and non-V3 CSS Selectors (Part 1)

Andrew565:
## Chromatic

https://2076-fix-non-uswds-styles--60f9b557105290003b387cd5.chromatic.com

---

## Description
Relates to #[2076](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2076)

Converts styles for V1 and V3 to make use of more specific selectors. Previously, non-V3 styles were 'selected' in CSS by checking for the presence or absence of the `uswds` attribute, looking like:

```
:host(:not([uswds]))
```

However, this was not specific enough, because if the `uswds` attribute was present but set to `false`, the CSS engine would still ignore the V1 style. This PR replaces the above with this:

```
:host(:not([uswds='true'], [uswds='']))
```

This will check for both the presence of the attribute (`=''`) and/or if it is present and true (`='true'`). In other places, where we *want* the attribute to be true, I've replaced this:

```
:host([uswds])
```

with this:

```
:host([uswds]:not([uswds='false']))
```

which will match if the `uswds` attribute is present but not set to `false`.

This PR covers the first batch of components:
- Accordion
- Alert
- Breadcrumb
- Button
- Button Pair
- Checkbox

## Testing done
Tested locally using Storybook

## Screenshots
No screenshots as the before and after are identical.

## Acceptance criteria
- [X] Successfully converts non-V3 styles

## Definition of done


powellkerry (approved):
LGTM

jamigibbs (approved):
Looks great! Thanks for adding the additional scenarios. The description is really clear too. 👍

harshil1793 (approved):

PR #887, opened 2023-10-05
va-modal: Fix logic for aria-label, don't set aria-label if title is not provided

powellkerry:
## Chromatic

https://2087-modal-undefined-label--60f9b557105290003b387cd5.chromatic.com

## Description
- Don't set aria-label if title is undefined
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2087

## Testing done
- Local testing through storybook
- E2E tests are passing

## Screenshots
- No visual changes

## Acceptance criteria
- [ ] aria-label is not set when title is undefined

## Definition of done

### DIFF
--- a/packages/web-components/package.json
+++ b/packages/web-components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/web-components",
-  "version": "4.45.26",
+  "version": "4.45.27",
   "description": "Stencil Component Starter",
   "main": "dist/index.cjs.js",
   "module": "dist/index.js",
--- a/packages/web-components/src/components/va-modal/va-modal.tsx
+++ b/packages/web-components/src/components/va-modal/va-modal.tsx
@@ -402,7 +402,7 @@ export class VaModal {
 
     if (!visible) return null;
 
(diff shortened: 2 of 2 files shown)


powellkerry:
@jamigibbs I have not written an issue to add a custom aria-describedby prop to modal. Is this something we want to do? (I think this is what you were asking from Laura's comment)

cc @micahchiang

jamigibbs:
> @jamigibbs I have not written an issue to add a custom aria-describedby prop to modal. Is this something we want to do? (I think this is what you were asking from Laura's comment)

@powellkerry  Oh sorry I should have been specific. It was @rsmithadhoc that mentioned it so I just wanted to make sure if it was still something we wanted to do at some other time. This is what is in that comment from Ryan:

> Their team is setting a title through aria-labeledby, which is the proper way to do it and the way USWDS recommends. **We should update our component to match, but that's outside the scope of this, so I will create a separate issue for updates to bring this in line with USWDS.**

powellkerry:
@jamigibbs I didn't find an issue from @rsmithadhoc. I have created a new one here: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2189

PR #888, opened 2023-10-06
2076 Fix V3 and non-V3 CSS Selectors (Part 2)

Andrew565:
## Chromatic

---

## Description
Relates to #[2076](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2076)

Converts styles for V1 and V3 to make use of more specific selectors. Previously, non-V3 styles were 'selected' in CSS by checking for the presence or absence of the `uswds` attribute, looking like:

```
:host(:not([uswds]))
```

However, this was not specific enough, because if the `uswds` attribute was present but set to `false`, the CSS engine would still ignore the V1 style. This PR replaces the above with this:

```
:host(:not([uswds='true'], [uswds='']))
```

This will check for both the presence of the attribute (`=''`) and/or if it is present and true (`='true'`). In other places, where we *want* the attribute to be true, I've replaced this:

```
:host([uswds])
```

with this:

```
:host([uswds]:not([uswds='false']))
```

which will match if the `uswds` attribute is present but not set to `false`.

This PR covers the second batch of components:
- Checkbox Group
- Date
- Memorable Date
- Modal
- Number Input
- Pagination

## Testing done
Tested locally using Storybook

## Screenshots
No screenshots as the before and after are identical.

## Acceptance criteria
- [X] Successfully converts non-V3 styles

## Definition of done


packages/web-components/src/components/va-checkbox-group/va-checkbox-group.scss
```diff
@@ -24,29 +24,31 @@
...
-}
-
-:host(:not([uswds])) legend {
-  display: block;
-  max-width: 46rem;
-  font-weight: inherit;
-  font-size: inherit;
-  line-height: inherit;
-  padding-inline: 0;
+:host(:not([uswds='true'], [uswds=''])) {
```
micahchiang, line 27: I haven't worked too much with the `:not()` selector, so I was surprised to discover the selector list passed in isn't a "cover all of these scenarios" sort of thing. It only uses whichever selector is the most specific. 

This led me down a deeper rabbit hole specificity weight and `:is()`. 

Could this selector also be written as `:host(:not(:is([uswds='false']))`? 

I don't think that's any clearer necessarily, just curious if it's the same thing.

micahchiang (approved):
Looks like Chromatic didn't load properly, but part 1 of these series of PRs has already been approved and merged, so I think it may be safe to do the same here.

ataker (approved):
I'm wondering is it possible to create a [scss variable](https://css-tricks.com/snippets/sass/use-sass-variable-selector/) to use as our `uswds` selector. This is probably outside the scope of this ticket, but the fact that we split this into 4 tickets is good evidence that we should find a way to refactor this.

PR #897, opened 2023-10-10
Va-pagination: add aria-labels to page numbers (uswds) for screen readers

it-harrison:
## Chromatic

---

## Description
Closes [2115](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2115)

## Testing done
local

## Screenshots
n/a

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-pagination/va-pagination.tsx
+++ b/packages/web-components/src/components/va-pagination/va-pagination.tsx
@@ -317,7 +317,9 @@ export class VaPagination {
                 onClick={() => this.handlePageSelect(1, 'nav-paginate-number')}
                 onKeyDown={e => this.handleKeyDown(e, 1)}
                 href="javascript:void(0)"
-                class="usa-pagination__button">1</a>
+                class="usa-pagination__button"
+                aria-label="page 1, first page"
+                >1</a>
             </li>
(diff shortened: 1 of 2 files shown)


packages/web-components/src/components/va-pagination/va-pagination.tsx
```diff
@@ -317,7 +317,9 @@ export class VaPagination {
                 onClick={() => this.handlePageSelect(1, 'nav-paginate-number')}
                 onKeyDown={e => this.handleKeyDown(e, 1)}
                 href="javascript:void(0)"
-                class="usa-pagination__button">1</a>
+                class="usa-pagination__button"
+                aria-label="page 1, first page"
```
ataker, line 321: This is only showing up if:
1. There is a previous button and
2. The first page is not in`pageNumbersToRender`

So if you're on the first page, or you're not far enough into the results for there to be an ellipsis separating page 1 from the other page numbers being rendered this will read as "page 1" instead of "page 1, first page".

ataker (changes requested):
Also should we be updating v1 component as well in this update?

ataker (approved):

PR #910, opened 2023-10-13
update search script to support uswds query

micahchiang:
## Chromatic

https://mc-2517-uswds-components--60f9b557105290003b387cd5.chromatic.com

---
## Description
This pull request adds support to the design system dashboard cli to search for all components using the `uswds` prop. 
Running `yarn search uswds` will use a regex that matches any web component or web component react binding where the `uswds` prop is being used. 
Related to https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2157

## Testing done
Locally tested the regex in both the cli tool and vets-website to ensure results matched up

## Screenshots
Screenshot from 2023-10-13 11-43-43
Screenshot from 2023-10-13 11-43-56

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/design-system-dashboard-cli/src/find-ds-components.js
+++ b/packages/design-system-dashboard-cli/src/find-ds-components.js
@@ -71,7 +71,6 @@ function tallyResults(vwComponents, contentBuildWC) {
   vwComponents.forEach(i => {
     const manifest = getManifest(i.path);
     const app = manifest ? manifest.appName : 'Platform(ish)';
-
     i.matches.forEach(m => {
       const componentName = m[1];
       if (componentName) {
@@ -143,6 +142,9 @@ function findUsedReactComponents(vwModules, regexPattern) {
  * Search vets-website and the content build for design system components.
  *
  * @param searchStrings {string[]} - The components to display usage for
+ *
+ * Passing in the string 'uswds' will output a list of all web components and
+ * web component react bindings where the uswds prop is used
  */
 function findComponents(searchStrings) {
   const vwModules = readAllModules(`${repos['vets-website']}/src`);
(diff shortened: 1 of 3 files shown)


packages/design-system-dashboard-cli/src/find-ds-components.js
```diff
@@ -163,18 +166,30 @@ function findComponents(searchStrings) {
   );
 
   const wcTagRegex = /<(va-[^\s>]+)/gms;
-  const vwWebComponents = search(vwModules, wcTagRegex);
-  const contentBuildWC = search(contentTemplates, wcTagRegex);
+  const wcUswds3Regex = /<(Va[^\s]+|va-[^\s]+)(\s|\n)[^>]*?uswds/gms;
```
micahchiang, line 169: This regex matches either react binding usage by looking for `<Va` or regular web component usage by looking for `<va`. Then, it looks for the presence of `uswds` either on the same line or accounting for multiline prop declarations.

packages/design-system-dashboard-cli/src/find-ds-components.js
```diff
@@ -163,18 +166,30 @@ function findComponents(searchStrings) {
...
-  ];
+  let vwWebComponents;
+  let contentBuildWC;
+  let vwComponents;
 
-  const data = tallyResults(vwComponents, contentBuildWC);
+  if (searchStrings?.includes('uswds')) {
+    vwWebComponents = search(vwModules, wcUswds3Regex);
+    contentBuildWC = search(contentTemplates, wcUswds3Regex);
+    vwComponents = [...vwWebComponents];
```
micahchiang, line 178: when using `yarn search uswds`, we only care about web components and react bindings that are using the uswds prop, so it's the only list we'll use when we call `tallyResults`. `usedReactBindings` isn't used because that list is only looking at import statements.

packages/design-system-dashboard-cli/src/find-ds-components.js
```diff
@@ -163,18 +166,30 @@ function findComponents(searchStrings) {
...
+    vwComponents = [
+      ...usedReactComponents,
+      ...vwWebComponents,
+      ...usedReactBindings,
+    ];
+  }
 
-  return filterSearchedComponents(data, searchStrings);
+  const data = tallyResults(vwComponents, contentBuildWC);
+  return searchStrings?.includes('uswds')
```
micahchiang, line 190: `filterSearchedComponents` expects the name of a component, not the name of a prop. Passing `uswds` to it returns nothing because there is no uswds component. Instead, the data is just returned directly.

packages/design-system-dashboard-cli/src/find-ds-components.js
micahchiang: I'm a fan of https://regex101.com/. I like how the site explains each segment of a regex and has a quick reference window for tokens.

micahchiang:
Requested re-review because I added ds playground apps to the ignore list

jamigibbs (approved):
Works for me! 

Screenshot 2023-10-18 at 2 40 28 PM

powellkerry (approved):

Issue #356, opened 2022-04-08
Export type definitions for typescript projects


bkjohnson:
## Description

The component library is [being used in a React typescript project](https://dsva.slack.com/archives/C01DBGX4P45/p1649446167005429) and typescript is throwing an error:

```
Could not find a declaration file for module
```

We should provide a type declaration file for the library.

### Details

The `web-components` sub-package is currently building with `declaration: false`:

https://github.com/department-of-veterans-affairs/component-library/blob/96692e5cd93ad0bb3140342d551b092f9c5d2e83/packages/web-components/tsconfig.json#L5

According to [the tsconfig documentation](https://www.typescriptlang.org/tsconfig#declaration) we may need that to be `true`.

We also declare the location of a type definition file in our `package.json`:

https://github.com/department-of-veterans-affairs/component-library/blob/96692e5cd93ad0bb3140342d551b092f9c5d2e83/packages/web-components/package.json#L9

---

A build in `web-components` currently will create a `dist/types/index.d.ts` file:

![image](https://user-images.githubusercontent.com/2008881/162516751-330f21d2-469a-4a1c-a83d-010afc9b87b9.png)

However, there is no type definition file in the `core` build:

![image](https://user-images.githubusercontent.com/2008881/162517028-d9d58c32-ed09-40ee-83ce-bb9431c26fd9.png)

We also may need to make sure that the `core` package has these type definition files and/or has a `types` property in the `package.json`.

## Acceptance criteria
- [ ] Type definition file is provided for the component library

mjknight50:
Hi @k80bowman   Circling back this open issue.  Is there a plan or roadmap for this from the platform team?

mjknight50:
@humancompanion-usds Our team is interested in moving to typescript and having the declarations for the components library would be very helpful.  Is this something you could help us move along?

humancompanion-usds:
@micahchiang - Do you want to help me prioritize this?

micahchiang:
This is a great idea that we should definitely do. We can add it as a stretch item for the time being, as it feels less urgent than the current things we have slated for sprint 38.

caw310:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Platform-Design-System-5f8de67192551b0012ebb802/issues/department-of-veterans-affairs/component-library/356?planning-poker) with Zenhub @Andrew565 @ataker @harshil1793 @micahchiang @nickjg231 @powellkerry @rmessina1010

PR #921, opened 2023-10-18
Build index.d.ts files

powellkerry:
## Chromatic

https://356-build-types-file--60f9b557105290003b387cd5.chromatic.com

## Description
Set `"declaration": true` to begin generating the index.d.ts file for components
Add section to core/webpack.config.js to copy the types directory to dist
Closes https://github.com/department-of-veterans-affairs/component-library/issues/356

## Testing done
local testing using `yarn build` in the web-components directory and core

## Screenshots
types directory built in web-components via `yarn build`
Screenshot 2023-10-18 at 2 29 14 PM

types directory copied to dist/components folder of /core during `yarn build`
Screenshot 2023-10-18 at 2 26 15 PM

## Acceptance criteria
- [ ]

## Definition of done


packages/core/webpack.config.js
```diff
@@ -62,6 +62,11 @@ module.exports = {
           to: './',
           context: path.resolve(__dirname, '../', 'web-components/dist'),
         },
+        {
+          from: '*',
+          to: './components/types',
+          context: path.resolve(__dirname, '../', 'web-components/dist/types'),
+        },
```
jamigibbs, line 69: I don't fully understand how the team that is needing this will be using it so did we need to double check that this is a good location for them?

I also noticed this comment in the original ticket description that I'm not sure is relevant or not:

> We also may need to make sure that the core package has these type definition files and/or has a types property in the package.json.

Andrew565 (commented):
Should we generate the index.d.ts file(s) and check them into the repo?

powellkerry:
@Andrew565 The files are generated as part of the build scripts. I don't think that we should check them in because that would cause issues with merge conflicts and automated builds on deploys.

packages/core/webpack.config.js
powellkerry: I added that to the package.json file.

mjknight50:
> @mjknight50 Can you tell us if this is a solution that will work for providing types for the components in the component library?

I think it should be fine with the `"declaration": true` but I'll have to test to be sure.

Andrew565 (approved):

mjknight50:
Circling back to this ticket, sorry...  
Would it be possible to get the @types published here?

`By default all visible ”@types” packages are included in your compilation. Packages in node_modules/@types of any enclosing folder are considered visible. For example, that means packages within ./node_modules/@types/, ../node_modules/@types/, ../../node_modules/@types/, and so on.`

https://www.typescriptlang.org/tsconfig#types

PR #925, opened 2023-10-25
2220 - Minimal header and crisis line modal accessibility improvements

Andrew565:
## Chromatic

https://2220-minimal-header-a11y--60f9b557105290003b387cd5.chromatic.com

---

## Description
Closes [#2220](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2220)
* Adding focus outline and aria-label to VA logo in Minimal Header
* Improving 'focus trap' of the crisis line modal
* Attempting to make CLM get announced by screen readers

## Testing done
Tested locally using Brave and VoiceOver

## Screenshots
N/A

## Acceptance criteria
- [ ] Meets accessibility requirements

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-crisis-line-modal/va-crisis-line-modal.tsx
+++ b/packages/web-components/src/components/va-crisis-line-modal/va-crisis-line-modal.tsx
@@ -15,7 +15,6 @@ import { CONTACTS } from '../../contacts';
 })
 export class VACrisisLineModal {
   @State() isOpen: boolean = false;
-  parentNode: HTMLElement;
 
   setVisible() {
     this.isOpen = true;
@@ -25,14 +24,7 @@ export class VACrisisLineModal {
     this.isOpen = false;
   }
 
-  private getParentNode() {
-    this.parentNode = document.querySelector('div.va-crisis-line-container');
-  }
-
-  connectedCallback() {
-    this.getParentNode();
-  }
-
+  /* eslint-disable i18next/no-literal-string */
   render() {
     return (
       <Host>
@@ -57,14 +49,12 @@ export class VACrisisLineModal {
           </button>
         </div>
         <va-modal
-          clickToClose
           modalTitle="We’re here anytime, day or night – 24/7"
           onPrimaryButtonClick={() => this.setNotVisible()}
           onCloseEvent={() => this.setNotVisible()}
           visible={this.isOpen}
           large={true}
           uswds={true}
-          ariaHiddenNodeExceptions={[this.parentNode]}
         >
           <p>
             If you are a Veteran in crisis or concerned about one, connect with
--- a/packages/web-components/src/components/va-crisis-line-modal/va-crisis-line-modal.tsx
+++ b/packages/web-components/src/components/va-crisis-line-modal/va-crisis-line-modal.tsx
@@ -15,6 +15,7 @@ import { CONTACTS } from '../../contacts';
 })
 export class VACrisisLineModal {
   @State() isOpen: boolean = false;
+  parentNode: HTMLElement;
 
   setVisible() {
     this.isOpen = true;
@@ -22,6 +23,14 @@ export class VACrisisLineModal {
 
   setNotVisible() {
     this.isOpen = false;
+  }
+
+  private getParentNode() {
+    this.parentNode = document.querySelector('div.va-crisis-line-container');
+  }
+
+  connectedCallback() {
+    this.getParentNode();
   }
 
   render() {
@@ -54,18 +63,13 @@ export class VACrisisLineModal {
           onCloseEvent={() => this.setNotVisible()}
           visible={this.isOpen}
           large={true}
+          uswds={true}
+          ariaHiddenNodeExceptions={[this.parentNode]}
         >
-          <div
-            id="modal-crisisline"
-            class="va-overlay va-modal va-modal-large"
-            role="alertdialog"
-          >
-            <div class="va-crisis-panel va-modal-inner" role="dialog">
-              <div class="va-overlay-body va-crisis-panel-body">
-                <p>
-                  If you are a Veteran in crisis or concerned about one, connect
-                  with our caring, qualified responders for confidential help.
-                  Many of them are Veterans themselves.
+          <p>
+            If you are a Veteran in crisis or concerned about one, connect with
+            our caring, qualified responders for confidential help. Many of them
+            are Veterans themselves.
                 </p>
                 <ul class="va-crisis-panel-list">
                   <li>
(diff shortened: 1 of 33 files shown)


packages/web-components/src/components/va-minimal-header/va-minimal-header.scss
```diff
@@ -19,6 +19,10 @@
   min-height: 30px;
   min-width: 48px;
 
+  &:focus {
```
powellkerry, line 22: This needs to be added to the `a` tag instead of the image. When tabbing through the gold outline doesn't show because the image is not the focusable element.

packages/web-components/src/components/va-crisis-line-modal/va-crisis-line-modal.tsx
```diff
@@ -48,6 +48,7 @@ export class VACrisisLineModal {
           </button>
         </div>
         <va-modal
```
powellkerry, line 50: For some reason you cannot navigate the modal with the keyboard. I was testing with voice over on and was trying to navigate by using ctrl option space and also by just using the tab key. I could never get the focus to move from the close button.

packages/web-components/src/components/va-crisis-line-modal/va-crisis-line-modal.tsx
```diff
@@ -59,7 +60,7 @@ export class VACrisisLineModal {
             class="va-overlay va-modal va-modal-large"
             role="alertdialog"
           >
-            <div class="va-crisis-panel va-modal-inner">
+            <div class="va-crisis-panel va-modal-inner" role="dialog">
```
harshil1793, line 63: The screen reader doesn't seem to pick up on the dialog role.

jamigibbs:
Are we ok with this PR title being descriptive enough for our release notes? "2220 - Accessibility improvements #925"

Andrew565:
> Are we ok with this PR title being descriptive enough for our release notes? "2220 - Accessibility improvements"

Updated!

packages/web-components/src/components/va-crisis-line-modal/va-crisis-line-modal.tsx
```diff
@@ -48,23 +57,19 @@ export class VACrisisLineModal {
           </button>
         </div>
         <va-modal
+          clickToClose
```
micahchiang, line 60: Just tested in Chromatic, and I'm still seeing that focus isn't trapped to the modal. Notice how focus continues down into the tabs in storybook:

[Screencast from 11-02-2023 01:46:55 PM.webm](https://github.com/department-of-veterans-affairs/component-library/assets/15097156/4048e91b-d700-4958-a0d6-f442ef7da2b7)

I think we may have to hand roll our own custom solution in order to keep focus in our crisis line modal. I did some digging, and it looks like the modal in vets-website could be using [this custom addFocusBehaviorToCrisisLineModal](https://github.com/department-of-veterans-affairs/vets-website/blob/80c7239aef28a99b99a2cbc0fd84c7baeb36741b/src/platform/site-wide/accessible-VCL-modal.js#L13) handler to keep track of the first and last focusable elements in the modal. 

We may be able to crib the helper used in there to get those elements, using the `getParentNode` function you added above. Something like: 

```javascript
function getTabbableElements(node) {
  return node.querySelectorAll(
    'input,select,a[href],textarea,button,[tabindex]',
  );
}

const elements = getTabbableElements(this.parentNode);
const firstEl = elements[0];
const lastEl = elements[elements.length - 1]

// and then some listeners to handle tab and/or shift do something like

// handle shift + tab
if document.activeElement === firstEl { 
  lastEl.focus()
  e.preventDefault()
} 

// handle last element tab
if document.activeElement === lastEl {
  firstEl.focus()
  event.preventDefault()
}
```
And then probably something else to handle a key press of the first element, which would be the close button.

micahchiang (commented):
Left a comment that proposes an idea for how to handle focus trap in the modal.

packages/web-components/src/components/va-modal/va-modal.tsx
```diff
@@ -354,8 +358,11 @@ export class VaModal {
     // Prevents scrolling outside modal
     disableBodyScroll(this.el);
 
+    // Get the docs-root element (from Storybook) if present, so that it can be excluded
+    const docsRoot = document.querySelector('#docs-root');
```
jamigibbs, line 362: When I remove this `docsRoot` (including from the `hideExceptions` assignment), the analytics tests pass so my guess is that it's causing a null error within the testing environment and stopping the event code block below from firing.

But maybe a better question is to ask if we should be putting code into components that are specifically for Storybook? This seems like it might be an anti-pattern.

Andrew565: I couldn't find a way to mitigate this issue in the storybook story itself, but did find a way to get this same effect without breaking the automated tests.

jamigibbs: @Andrew565 I trust what you're saying I'm just still a little bit hung up that we have to add Storybook specific logic to the component. The prop `ariaHiddenNodeExceptions` was supposed to allow for managing conditions like this.

It should be used with the React binding version of the web component (`VaModal`) though. Maybe that's why you said you were just getting a string and not the node?

So in `va-crisis-line-modal` it would be:

```js

const docsRoot = document.querySelector('#docs-root');

  <VaModal
    ariaHiddenNodeExceptions={[docsRoot]}
    modalTitle="We’re here anytime, day or night – 24/7"
    onPrimaryButtonClick={() => this.setNotVisible()}
    onCloseEvent={() => this.setNotVisible()}
    visible={this.isOpen}
    large={true}
    uswds={true}
  >
```

I realize this kind of just moves Storybook specific stuff into another component but at least we aren't changing logic in the modal component itself and the prop is being used as it's intended.

rsmithadhoc:
Unfortunately, I'm running into a few issues with the latest round of updates:
- VoiceOver is no longer announcing the dialog and reading the title as it was previously. Tabbing to the individual links inside the modal does not read them out as well.
- Now the focus jumps to VA logo, then to the close button when tabbing past the last element in the modal. So technically focus lands in the modal, but that may be confusing because a screen reader would start to read the VA logo label then "Close" (but due to the previous issue mentioned, it isn't reading anything at the moment).

I'm wondering if we need to look at modal as part of a separate update. The minimal header itself looks good, but the modal seems to be causing the issues and might need a bit of a deeper look.

Andrew565:
Closing in favor of splitting out the focus trap and screen reader issues, smaller PR for remainder of work can be found here: https://github.com/department-of-veterans-affairs/component-library/pull/940

PR #927, opened 2023-10-27
Mc 2517 sql query update

micahchiang:
## Chromatic

https://mc-2517-sql-query-update--60f9b557105290003b387cd5.chromatic.com

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2157

This pull request addresses two needs in our design system dashboard cli:
1. It adds a new column to our SQL query, `uswds`, which represents the total count where a component is using the uswds prop. 
2. It fixes a gotcha where the cli was only accounting for single line, single use imports, not multiline imports. The implication here is that react binding usages were being under counted. 

For a little more context, checkout the slack conversation [here](https://dsva.slack.com/archives/G01BJ3ESXL4/p1698249090101329)

## Testing done
Locally against search results in vets-website.

Using `VaTextInput` as an example:

<details><summary>New regex properly matching search result count in vets-website</summary>

image

image

</details>

<details><summary>Old regex on main branch, only matching single line imports</summary>

image

Screenshot from 2023-10-27 10-00-49

</details>

## Screenshots
<details><summary>Example of old regex not matching multiline imports</summary>

Screenshot from 2023-10-25 12-36-18

</details>

<details><summary>Example of new regex matching multiline imports properly</summary>

Screenshot from 2023-10-25 12-46-55

</details>

## Acceptance criteria
- [x] Bugs have been fixed

## Definition of done


packages/design-system-dashboard-cli/src/find-ds-components.js
```diff
@@ -56,17 +56,25 @@ function filterSearchedComponents(data, searchStrings = []) {
...
     data[componentName][appName]++;
     data[componentName].total++;
   }
 
+  function incrementUswdsCount(componentName) {
+    if (!data[componentName].uswds) {
+      data[componentName].uswds = 0;
+    }
+    data[componentName].uswds++;
+  }
```
micahchiang, line 76: Does the same thing as incrementCount but specifically for the uswds property only

packages/design-system-dashboard-cli/src/find-ds-components.js
```diff
@@ -106,6 +114,23 @@ function tallyResults(vwComponents, contentBuildWC) {
...
+        // If the component is not yet in the collection, initialize it as an empty array
+        if (!instancesByComponent[componentName])
+          instancesByComponent[componentName] = [];
+
+        // Only include the app path if it hasn't already been included previously
+        if (!instancesByComponent[componentName].includes(i.path))
+          instancesByComponent[componentName].push(i.path);
+      }
+    });
+  });
```
micahchiang, line 132: Does a count of the queried uswds components. I couldn't think of an easy way to modify the first query that returns vwComponents, so the query just runs again.

packages/design-system-dashboard-cli/src/find-ds-components.js
```diff
@@ -115,25 +140,44 @@ function tallyResults(vwComponents, contentBuildWC) {
...
-      // First item will be the import match, second will be the use
-      const componentName = m[1];
-      const componentRegex = new RegExp(`<(${componentName})`, 'g');
-      // Search each file that imports a component for each use of that component
-      // and add the search results to the return array
-      return singleFileUses.concat(
-        search(readFile(match.path), componentRegex)
-      );
+      // First item will be the import match, second will be a string of component names
+      const componentNames = m[1];
```
micahchiang, line 167: Originally `m[1]` was assuming a single string. With the new regex, even if it is just a single string like `VaTextInput`, we're transforming it to an array to handle the cases where there are more than one import. See the jsdoc comment above the function for a broader explanation.

packages/design-system-dashboard-cli/src/find-ds-components.js
```diff
@@ -171,6 +215,7 @@ function findComponents(searchStrings) {
   let vwWebComponents;
   let contentBuildWC;
   let vwComponents;
+  const uswdsV3Components = [...search(vwModules, wcUswds3Regex)];
```
micahchiang, line 218: I couldn't think of a good way to search for vw components and keep track of the uswds prop "in place", so just decided to search explicitly for them separately.

powellkerry (approved):

PR #945, opened 2023-11-13
update web-components to use variables from css-library

it-harrison:
## Chromatic

https://2236-remove-variables-file--60f9b557105290003b387cd5.chromatic.com

---

## Description
This PR makes changes to allow for the removal of the [variables.scss file](https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/src/global/_variables.scss) from the web-component package, which is a duplicate of [this file](https://github.com/department-of-veterans-affairs/component-library/blob/bef641ef2c85077d4a013d5fe32030fd176912bc/packages/css-library/dist/tokens/scss/variables.scss#L5) in css-library.
A [companion PR](https://github.com/department-of-veterans-affairs/component-library/pull/946) for the css-library package _**must be merged and published before**_ this PR is merged.

Closes [2236](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2236)

## Testing done
local testing done by using verdaccio to publish css-library and then consume in web-components:
**css-library**:

Screenshot 2023-11-13 at 2 00 13 PM

**verdaccio**:
Screenshot 2023-11-13 at 1 28 03 PM

**web-components - build**:

Screenshot 2023-11-13 at 1 48 29 PM

**screenshots of some components that use variables from css-library** 

`va-statement-of-truth`
Screenshot 2023-11-13 at 2 12 12 PM

`va-accordion`
Screenshot 2023-11-13 at 2 10 30 PM

`va-button`
Screenshot 2023-11-13 at 2 14 07 PM

`<va-segmented-progress-bar`

Screenshot 2023-11-13 at 2 20 51 PM

## Acceptance criteria
- [ ]

## Definition of done
- [ ] Documentation has been updated, if applicable
- [ ] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)

PR #946, opened 2023-11-13
update package.json for publishing

it-harrison:
## Chromatic

https://make-css-library-publishable--60f9b557105290003b387cd5.chromatic.com

---

## Description
This PR makes the css-library package able to be published to npm so that it can be used as a dependency in the web-component package. There is a [companion PR here](https://github.com/department-of-veterans-affairs/component-library/pull/945), which updates web-components to use css-library to import the [variables](https://github.com/department-of-veterans-affairs/component-library/blob/bef641ef2c85077d4a013d5fe32030fd176912bc/packages/css-library/dist/tokens/scss/variables.scss#L5) that are generated in css-library (and therefore ends the need to copy the file manually to web-components).

## Testing done
local testing

## Acceptance criteria
- [ ]

## Definition of done


rmessina1010 (approved):
LGTM

jamigibbs:
Do we want this added to the [component library release notes](https://github.com/department-of-veterans-affairs/component-library/releases)? If not then this PR should be tagged as `ignore-for-release`

jamigibbs:
@it-harrison I saw this was merged but I'm still not clear if we want this PR to be a part of the release notes for the component-library https://github.com/department-of-veterans-affairs/component-library/pull/946#issuecomment-1808788895

I know that we intend for this package to have its own independent release notes but that has not been set up yet. Won't labeling this PR as `major` lump it into the web component releases?

packages/css-library/package.json
```diff
@@ -1,9 +1,10 @@
 {
   "name": "@department-of-veterans-affairs/css-library",
-  "private": true,
+  "version": "0.0.1",
```
jamigibbs, line 3: Would it be an issue to start at this version number because version `0.1.0` already exists and if/when we ever make a `minor` version update to the package, we will not be able to publish it as `0.1.0`.

https://www.npmjs.com/package/@department-of-veterans-affairs/css-library?activeTab=versions

cc @micahchiang

micahchiang: hmm that's a good point. Tbh, I was the one that told @it-harrison it'd probably be okay to start at `0.0.1`, so that's my bad. I think given the life cycle stage css library is in, it's probably not too big of a deal to eventually just skip over `0.1.0`, unless folks feel strongly about that.

jamigibbs: No big deal. Someone might just be a little confused for minute. If we're following [Semantic Versioning](https://semver.org/#spec-item-4), the most important release will be `1.0.0` anyway which should signal it's ready for the public.

- > (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.
- > Version 1.0.0 defines the public API.

PR #947, opened 2023-11-13
va-search-input: add USWDS v3 variation

jamigibbs:
## Chromatic

https://1862-va-search-uswds-v3--60f9b557105290003b387cd5.chromatic.com

## Description
Sketch reference: https://www.sketch.com/s/610156b6-f281-4497-81f3-64454fc72156/p/447FB6DA-215E-4F15-A71B-94B7686DA614/canvas
USWDS reference: https://designsystem.digital.gov/components/search/

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1862

## Testing done
Storybook
Browserstack

## Screenshots
Screenshot 2023-11-13 at 3 17 48 PM
Screenshot 2023-11-13 at 3 17 55 PM

Screenshot 2023-11-13 at 4 29 25 PM

## Acceptance criteria
- [ ]

## Definition of done


packages/web-components/src/components/va-search-input/va-search-input.tsx
```diff
@@ -120,7 +136,14 @@ export class VaSearchInput {
    * Fires a submit and analytics events.
    */
   private handleSubmit = () => {
-    this.inputRef.dispatchEvent(
+    const form = this.el.shadowRoot.querySelector('form');
+    if (form) {
+      form.onsubmit = (event) => {
+        event.preventDefault();
```
jamigibbs, line 142: The `form` element will submit (reloading the page) unless we prevent the default behavior.

packages/web-components/src/components/va-search-input/va-search-input.tsx
```diff
@@ -120,7 +136,14 @@ export class VaSearchInput {
...
+      }
+    }
+
+    this.el.dispatchEvent(
       new SubmitEvent('submit', {
```
jamigibbs, line 147: Originally the component was dispatching the submit event using`this.inputRef` (the `input` element) but because the input element is now a child of `form`, the event would bubble up and be prevented by the parent element `form`  (at line 142). So the submit event would never actually fire. 

To fix that, I changed the element we are dispatching off of to the component itself `this.el`.

packages/web-components/src/components/va-search-input/va-search-input.tsx
```diff
@@ -402,6 +428,77 @@ export class VaSearchInput {
...
+              {small &&
+                <va-icon
+                  icon="search"
+                  size={3}
+                  srtext="Search"
+                />
+              }
+            </button>
+            {isListboxOpen && (
+              <ul
```
jamigibbs, line 474: The "type ahead" feature was part of the original component so I've integrated into the v3 variation too. This is what it looks like for each of the USWDS v3 display types:

**default**

typeahead-default

**big**


typeahead-big

**small**

typeahead-small

packages/web-components/src/components/va-search-input/va-search-input.tsx
```diff
@@ -402,6 +428,77 @@ export class VaSearchInput {
...
+              value={value}
+            />
+            <button 
+              class="usa-button" 
+              type="submit"
+              onClick={handleButtonClick}
+            >
+              {!small && <span class="usa-search__submit-text">Search</span>}
+              {small &&
+                <va-icon
```
jamigibbs, line 466: The USWDS markup has this being an `img` element but now that we have  the `va-icon` component, I thought it would be better to start using it for component composition.

packages/web-components/src/components/va-search-input/va-search-input.scss
```diff
@@ -1,7 +1,49 @@
...
+.usa-search [type=submit] {
+  width: auto;
+}
+
+.usa-button {
+  line-height: 0;
+}
+
+.usa-input {
+  max-width: 100%;
```
jamigibbs, line 38: The USWDS example forces a set width for the input length. I thought that we would want this to be more flexible and allow for the wrapping container to manage the width.

packages/web-components/src/components/va-search-input/va-search-input.scss
```diff
@@ -1,7 +1,49 @@
...
+:host([uswds]:not([uswds='false'])[big="true"]) #va-search-listbox  {
+  top: 52px;
+}
+
+.usa-search [type=submit] {
+  width: auto;
+}
+
+.usa-button {
+  line-height: 0;
```
jamigibbs, line 34: Because I am using `va-icon` instead of an `img` element, I needed to add this line-height reset otherwise the icon was sitting a pixel or two off center.

packages/web-components/src/components/va-search-input/va-search-input.scss
```diff
@@ -11,11 +53,11 @@
...
   border-color: var(--color-gray);
   display: inline-block;
-  height: 4.8rem;
+  height: 48px;
   margin: 0;
   max-width: none;
-  min-width: 24.8rem;
-  padding: 1.2rem 0.8rem;
+  min-width: 248px;
+  padding: 12px 8px;
```
jamigibbs, line 60: I went ahead and adjusted all of the `rem` to `px` for the v1 version because of the anticipated issues with the USWDS v3 root font-size but if we'd prefer to keep these as rem for now, I'm happy to revert.

packages/web-components/src/components/va-search-input/va-search-input.tsx
```diff
@@ -402,6 +428,75 @@ export class VaSearchInput {
...
+              role={role}
+              value={value}
+            />
+            <button 
+              class="usa-button" 
+              type="submit"
+              onClick={handleButtonClick}
+            >
+              {!small && <span class="usa-search__submit-text">Search</span>}
+              <va-icon
```
jamigibbs, line 465: The USWDS markup has this being an `img` element to display their search icon. But now that we have the `va-icon` component, I thought it would be better to start using it like this.

packages/web-components/src/components/va-search-input/va-search-input.scss
```diff
@@ -0,0 +1,145 @@
...
+}
+
+.usa-input {
+  max-width: 100%;
+}
+
+/* if small prop is active, or, if mobile size is active, display va-icon search button */
+:host([small]:not([small='false'])) va-icon {
+  display: block;
+}
```
jamigibbs, line 40: A side-effect of using `va-icon` meant that I added the CSS display logic so that the search icon would always display on small viewports as well as when the small prop is set.

jamigibbs:
I have not added i18next support to the component yet even though the USWDS site provides the Spanish word for "search" because:

1) We have a prop that dynamically sets the button text `button-text` so it didn't makes sense to manage that translation within the component itself if teams would be making it something different.
2) We also need the ["Search Suggestions" translation](https://github.com/department-of-veterans-affairs/component-library/blob/4f093a3fd851d4470f5e2fff0f4810ac9806a637/packages/web-components/src/components/va-search-input/va-search-input.tsx#L474) for the type ahead feature and historically, we do not start integrating i18next until we have all available strings translated.

Screenshot 2023-11-13 at 4 46 01 PM

I have updated our [translations spreadsheet](https://docs.google.com/spreadsheets/d/1iIg9pP5GG-QFk7DrsyoKJv9GebtYmDuzKf_heqxWks4/edit#gid=0) though with "Buscar" because we didn't have that yet.

packages/web-components/src/components/va-search-input/va-search-input.tsx
jamigibbs: I'm removing `va-icon` because I think it's just making the CSS more complicated. But I needed to add the icon to our assets folder in order to keep it as an `img` element.

https://github.com/department-of-veterans-affairs/component-library/pull/947/commits/b35efcec38777f577019f2e4d7c443101fc7e68a

micahchiang (commented):
Love how thorough this PR is. Not sure if it's a storybook thing, or something else, but when I toggle some of the props in Canvas (big, small, and uswds) the button text disappears and never re-renders: 

[Screencast from 11-14-2023 07:58:31 AM.webm](https://github.com/department-of-veterans-affairs/component-library/assets/15097156/ac52f020-752e-4790-b0e6-0251ef8eab94)

jamigibbs:
> when I toggle some of the props in Canvas (big, small, and uswds) the button text disappears and never re-renders:

@micahchiang I think that's happening because the button text needs to be explicitly set on the `button-text` prop when the props refresh like that in Storybook. I added the value "Search" as a [default to the prop](https://github.com/department-of-veterans-affairs/component-library/blob/bdfa0f67b48814a0ab30055af92b17355dfb2d85/packages/web-components/src/components/va-search-input/va-search-input.tsx#L56) which is what displays when the component loads (unless it's the `small` variation) but that doesn't force a value if the component is toggled in Storybook. 

I'm not sure if we should always require a value there though because sometimes it's just the search icon. I'll think about this a little more but I'm leaning towards this just being a Storybook thing at the moment

Screenshot 2023-11-14 at 10 40 54 AM

But having said that, it does reveal another issue that I'm looking into. Setting the button text to "Search VA.gov" doesn't display correctly 🤔 

Screenshot 2023-11-14 at 11 05 51 AM

micahchiang (approved):
lgtm :rocket:

jamigibbs:
@danbrady @LWWright7 I am not sure if there is a Sketch file yet for the v3 variation of va-search-input but I wanted to make sure you were aware of the type ahead feature that we have added to keep parity with the v1 version: https://github.com/department-of-veterans-affairs/component-library/pull/947#discussion_r1391709361

Andrew565 (approved):

PR #951, opened 2023-11-16
va-omb-info, va-crisis-line-modal, va-minimal-header: Fix for focus trapping

Andrew565:
## Chromatic

https://2207-omb-focus-issues--60f9b557105290003b387cd5.chromatic.com

---

## Description
Closes [#2207](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2207)
Relates to [#2255](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2255)

Detects if focus is 'focused' on the modal-triggering buttons within the components
Redirects focus back to the Modal if it is visible

## Testing done
Tested locally using Brave and Storybook

## Screenshots
N/A

## Acceptance criteria
- [X] Focus is 'trapped' so that it stays in the modal, if it is visible

## Definition of done


packages/web-components/src/components/va-crisis-line-modal/va-crisis-line-modal.tsx
```diff
@@ -24,12 +26,21 @@ export class VACrisisLineModal {
     this.isOpen = false;
   }
 
+  // Redirects focus back to the modal, if the modal is open/visible
+  private trapFocus() {
+    const modalCloseButton = this.el.shadowRoot
+      .querySelector('va-modal')
+      .shadowRoot.querySelector('button.va-modal-close') as HTMLElement;
+    modalCloseButton?.focus();
```
jamigibbs, line 34: So fresh and so clean. ✨

packages/web-components/src/components/va-omb-info/va-omb-info.tsx
```diff
@@ -50,6 +50,12 @@ export class VaOmbInfo {
     this.modalContents = null;
   };
 
+  // Redirects focus back to the modal, if the modal is open/visible
+  private trapFocus = () => {
+    const modalCloseButton = this.el.shadowRoot.querySelector('va-modal').shadowRoot.querySelector('button.va-modal-close') as HTMLElement;
```
jamigibbs, line 55: This is kooky nit feedback but could this be formatted the same as what you've done for the `va-crisis-line-modal` version of the same method?

packages/web-components/src/components/va-modal/va-modal.tsx
```diff
@@ -178,7 +178,7 @@ export class VaModal {
   /**
    * Save focusable children within the modal. Populated on setup
    */
-  @State() focusableChildren: HTMLElement[] = null;
+  focusableChildren: HTMLElement[] = null;
```
jamigibbs, line 181: I'm just curious why `@State()` needed to be removed here?

Andrew565: It was causing an error/warning about the state updating during the wrong part of the lifecycle which could trigger unnecessary re-renders. All of the logic still works just fine without it being a 'state' variable, and it only ever gets updated once.

packages/web-components/src/components/va-crisis-line-modal/test/va-crisis-line-modal.e2e.ts
```diff
@@ -86,7 +86,10 @@ describe('va-crisis-line-modal', () => {
     const triggerButton = await page.find(
       'va-crisis-line-modal >>> button.va-overlay-trigger',
     );
+    console.log("found trigger button");
     await triggerButton.click();
+    console.log("trigger button clicked");
```
jamigibbs, line 91: A couple of rogue console logs in the test.

jamigibbs (approved):
I think this is a reasonable approach. The tabbing stays in the modal when I try it and when the modal opens, it's on the modal close button and VoiceOver reads the heading of the modal.

I'm not sure if focus should return back to the crisis line button when the modal is closed though? It doesn't seem to do that for me. I don't have all of the context of previous feedback though. Maybe Ryan can confirm if/when he tests as well.

packages/web-components/src/components/va-crisis-line-modal/test/va-crisis-line-modal.e2e.ts
```diff
@@ -87,6 +87,7 @@ describe('va-crisis-line-modal', () => {
       'va-crisis-line-modal >>> button.va-overlay-trigger',
     );
     await triggerButton.click();
+    await page.waitForChanges();
```
jamigibbs, line 90: @Andrew565 Sorry, I missed this in my last review. Do you need this still?

rsmithadhoc:
I'm still running into issues with this. I'm testing via the Chromatic link and using both Firefox and Chrome. Let me know if it is something with that Chromatic link.

- The focus can leave the modal and go to browser controls.
    - For crisis line modal, it still hops to the VA logo outside the modal.
        image
- Shift + Tab reverses focus correctly, but stops at the first element. The expected behavior would be to go in reverse and cycle back to the last element in the modal, then continue.
- When opening the modal, I think focus needs to be on the element with `dialog`, so a screen reader reads out that a dialog has opened.
    Examples:
    image
    image
- In Firefox with VoiceOver, it doesn't read anything in the modal. However, it does with Chrome.

Example of what should happen [here](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/) and [here](https://getbootstrap.com/docs/5.0/components/modal/). With these modals, the focus never leaves. Holding Tab or Shift + Tab will just cycle through the modal elements and never leave. Then, upon closing it, focus is returned to the button that triggered the modal.

packages/web-components/src/components/va-crisis-line-modal/test/va-crisis-line-modal.e2e.ts
Andrew565: Probably not, thanks for catching.

rsmithadhoc (approved):
Looks good to me for the focus trapping portion, thanks!

PR #956, opened 2023-11-21
Add v3 va-featured-content component

ataker:
## Chromatic

https://dst1858-v3-featuredcontent--60f9b557105290003b387cd5.chromatic.com

## Description
Closes [Conversion to USWDS v3 Summary Box / VA Featured Content](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1858)

## Testing done
Tested locally in storybook in chrome, firefox, safari, edge.
Added unit tests

## Screenshots
Screenshot 2023-11-21 at 11 24 04

## Acceptance criteria
- [ ] V3 version of Featured Content displays as expected
- [ ] Added to storybook
- [ ] Tests added

## Definition of done


jamigibbs:
In the [USWDS example](https://designsystem.digital.gov/components/summary-box/), there are a lot of other `usa-summary-box` classes. How significantly different are these styles with the Formation styles currently displaying? Is it possible to use them? Like if we removed Formation today, what would this component look like?

Screenshot 2023-11-21 at 10 41 05 AM

rmessina1010 (approved):
LGTM

packages/web-components/src/components/va-featured-content/va-featured-content.tsx
```diff
@@ -7,18 +7,35 @@ import { Component, Host, h } from '@stencil/core';
...
-          <slot name="headline"></slot>
-          <slot />
-        </div>
-      </Host>
-    );
+    const { uswds } = this;
+    if (uswds) {
+      return (
+        <Host>
+          <div class="usa-summary-box">
```
jamigibbs, line 24: The [USWDS example](https://designsystem.digital.gov/components/summary-box/) has all of these on the wrapping element (which you might able to simplify by putting them on the `<Host>` element):

```
  class="usa-summary-box"
  role="region"
  aria-labelledby="summary-box-key-information"
```

And is there any way to assign the other USWDS classes? Maybe programmatically on the assigned nodes?

- usa-summary-box__body
- usa-summary-box__heading
- usa-summary-box__text
- usa-summary-box__link

```js
    this.el.shadowRoot
      .querySelector('slot')
      .assignedNodes();
```

jamigibbs: I was also thinking that maybe, even if you were able to assign the classes to matching elements, the styles would be blocked by the shadow dom so this might be a good candidate for `shadow: false` on the component.

ataker: I was looking for a way to add these, thanks! I can't find a way to access the actual slot content to add `usa-summary-box__link` to the links, but was able to add the others.

packages/web-components/src/components/va-featured-content/va-featured-content.tsx
```diff
@@ -7,18 +7,53 @@ import { Component, Host, h } from '@stencil/core';
...
-          <slot name="headline"></slot>
-          <slot />
-        </div>
-      </Host>
-    );
+    const { uswds } = this;
+    if (uswds) {
+      return (
+        <Host>
+          <div class="usa-summary-box" role="region" aria-labelledby="summary-box-key-information">
```
jamigibbs, line 40: The `aria-labelledby` attribute references an `id` on another element. I don't see it connected to anything here.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby

packages/web-components/src/components/va-featured-content/va-featured-content.tsx
```diff
@@ -7,18 +7,53 @@ import { Component, Host, h } from '@stencil/core';
...
+      return
+    }
+    // add uswds classes
+    const nodes = this.el.shadowRoot
+      .querySelectorAll('slot')
+    const headline = nodes[0];
+    const content = nodes[1];
+    
+    headline.classList.add('usa-summary-box__heading');
+    content.classList.add('usa-summary-box__text');
```
jamigibbs, line 32: I'm not sure if adding this class to a slot element does anything. Because when I try and fiddle with the CSS in the browser, nothing happens.

Screenshot 2023-11-21 at 2 54 05 PM

I wonder if it needs to be on an actual element like `div` before it applies to child elements. 🤔

ataker:
Rolling over to next sprint to give more time to respond to PRs

packages/web-components/src/components/va-featured-content/va-featured-content.tsx
ataker: Yeah, I'm conflicted. I don't see the point of these classes when they're not applying styles. I'm ok with leaving them since this is sorta its own component, and we can more strictly enforce a structure if we want to make an actual `summary-box` component

jamigibbs: Does it work if we turn off the component's shadow dom and some how programmatically apply the classes to an actual element (not a `slot`)? The purpose of having a USWDS v3 variation is to use their styling.

packages/web-components/src/components/va-featured-content/va-featured-content.tsx
jamigibbs: @ataker Were you able to look into this issue?

packages/web-components/src/components/va-featured-content/va-featured-content.tsx
jamigibbs: We can at least add the `usa-summary-box__text` class though, right?

```
        <Host>
          <div class="usa-summary-box" role="region">
            <div class="usa-summary-box__body">
              <slot name="headline"></slot>
               <div class="usa-summary-box__text"> <------
                 <slot />
               </div>
            </div>
          </div>
        </Host>
```

ataker: That class is already being added to the slot element dynamically.

ataker:
Wrote a follow-up ticket to determine how we should proceed with styling slot content. Going to merge this in for now and we can revisit it down the line.

packages/web-components/src/components/va-featured-content/va-featured-content.tsx
ataker: This was removed

jamigibbs: This is a feature from the USWDS example that provides an accessible name to the featured content so we might want to have that integrated. Maybe when @rsmithadhoc is able to do an a11y review, he can provide some guidance for that.

Screenshot 2023-12-06 at 8 53 39 AM

PR #958, opened 2023-11-23
set up temp demo package

rmessina1010:
Experiment branch.
Script to deploy via  Verdaccio


micahchiang:

PR #970, opened 2023-12-01
va-telephone: add prop for passing a string to aria-describedby

jamigibbs:
## Chromatic

https://2254-telephone-aria-describedby--60f9b557105290003b387cd5.chromatic.com/?path=/docs/components-va-telephone--aria-described-by

## Background Info

When the `va-telephone` [web component was created](https://github.com/department-of-veterans-affairs/component-library/pull/233), it was essentially a direct port of the original React `Telephone` component (now deprecated).

The React component was able to use a standard `aria-describedby` attribute and the web component continued to demonstrate that usage in Storybook under the assumption that it would continue to work. Unfortunately, it does not because the ID can not be referenced through the shadow dom. 

**Does not work**

Screenshot 2023-12-01 at 2 17 46 PM

## Description

This PR adds the prop `messageAriaDescribedby` which will be used to communicate additional context to the phone number within the component's shadow dom. It also updates the Storybook example to use this prop.

This is a similar approach that we have also done for:
- va-checkbox
- va-number-input
- va-statement-of-truth
- va-text-input
- va-textarea

Now when the phone number is in focus, there will be additional context available to the screen reader if a string is passed into the prop.

```html
  <va-telephone
    contact="8773459876"
    message-aria-describedby="Main number to facility"
  />
```

**Number Focused**

Screenshot 2023-12-01 at 2 32 17 PM

**More Content Available Message**

Screenshot 2023-12-01 at 2 32 30 PM

**The aria-describedby string is available**

Screenshot 2023-12-01 at 2 37 23 PM

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2254

## Testing done
Chrome, Voiceover

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/packages/web-components/package.json
+++ b/packages/web-components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/web-components",
-  "version": "4.49.6",
+  "version": "4.50.0",
   "description": "Stencil Component Starter",
   "main": "dist/index.cjs.js",
   "module": "dist/index.js",


laflannery (commented):
I just want to clarify, make sure I understand - There could potentially be 2 labels now, a visual label outside the component/shadow dom and a sr label, which is the new `message-aria-describedby` prop. Is that correct?

jamigibbs:
> I just want to clarify, make sure I understand - There could potentially be 2 labels now, a visual label not part of the component/shadow dom at all and a sr label, which is the new message-aria-describedby prop. Is that correct?

@laflannery I think technically the answer to this is yes. The prop gives the flexibility to have sr-only text that could be anything (the same as the visual label or something more/less descriptive). Usually it's just the same text so I'm not sure I would describe it as two different labels. But it could be different if the situation needs it to be.

This would be an example usage (I've also updated the [Storybook example](https://2254-telephone-aria-describedby--60f9b557105290003b387cd5.chromatic.com/?path=/docs/components-va-telephone--aria-described-by) to demonstrate this):

```html
 <span>Main number to facility</span>
  <va-telephone
    contact="8773459876"
    message-aria-describedby="Main number" <--- linked to an sr-only element internally
  />
```

Andrew565 (approved):

PR #976, opened 2023-12-06
va-breadcrumbs: make uswds version support React Router links

it-harrison:
## Chromatic

https://2188-breadcrumb-with-link--60f9b557105290003b387cd5.chromatic.com

---
## Description
This PR makes the uswds version of va-breadcrumbs compatible with React Router. The original issue mentioned the `<Link` component. Making the uswds version play nicely with `<Link>` would require a total rewrite of va-breadcrumbs since currently the uswds version does not use slots. However, this PR makes the uswds version work with React Router (or, most likely, other Routers too) by manipulating the History api. So in this PR the component API for the uswds version does **not** change.

Note: in this PR the component directly manipulates the History API. Another approach is to emit an event and let the VFS team handle the route change programmatically in a handler that makes use of utilities that React Router provides. This approach imposes a small burden on the VFS team but may provide more future proof solution (in case, for example, directly manipulating the History API does not work well with a future version of React Router).

Closes [2188](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2188)

## QA Checklist

## Screenshots
In `vets-website` I modified [this file ](https://github.com/department-of-veterans-affairs/vets-website/blob/main/src/applications/simple-forms/mock-simple-forms-patterns/containers/IntroductionPage.jsx) because it is part of an app that uses React Router. The app runs locally at http://localhost:3001/mock-simple-forms-patterns/introduction:

Screenshot 2023-12-06 at 9 58 13 AM

The `href` values in the `breadcrumbList` prop correspond to pages in the app.

Here is a video of navigation:

https://github.com/department-of-veterans-affairs/component-library/assets/8867779/f1c41f9a-3998-4a52-92e6-c252bc4a3bfe

## Acceptance criteria

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
+++ b/packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
@@ -75,6 +75,13 @@ export class VaBreadcrumbs {
   componentLibraryAnalytics: EventEmitter;
 
   /**
+   * Are the hrefs provided in the objects of the breadcrumbList to be used 
+   * in an app with a Router, e.g. React Router?
+   * Has no effect if uswds is false
+   */
+  @Prop() useRouter?: boolean = false;
+
+  /**
    * This method is used to update the formattedBreadcrumbs state.
    * It is only invoked when the uswds attribute is set to true on the component.
    * It either parses a JSON string into an array of breadcrumb objects or uses the passed array as is.
@@ -239,6 +246,29 @@ export class VaBreadcrumbs {
     });
   }
 
+  /**
+   * Changes the route when uswds and useRoute props are true.
+   * Route will be set to the href value in the breadcrumb.
+   */
+  handleRouteChange(e: MouseEvent, href: string): void {
+    e.preventDefault();
+    window.history.pushState(null, '', href);
+    // create event to notify Router
+    const popStateEvent = new PopStateEvent('popstate');
+    // notify Router
+    window.dispatchEvent(popStateEvent);
+  }
+
+  /**
+   * Handle click event on breadcrumb when uswds is true
+   */
+  uswdsHandler(e: MouseEvent, href: string) {
+    if (this.useRouter) {
+      this.handleRouteChange(e, href);
+    }
+    this.fireAnalyticsEvent(e);
+  }
+
   render() {
     const { label, uswds, wrapping } = this;
     const wrapClass = classnames({
@@ -260,7 +290,7 @@ export class VaBreadcrumbs {
                       <span>{item.label}</span>
                     </a>
                   ) : (
-                    <a class="usa-breadcrumb__link" href={item.href} onClick={e => this.fireAnalyticsEvent(e)}>
+                    <a class="usa-breadcrumb__link" href={item.href} onClick={e => this.uswdsHandler(e, item.href)}>
                       <span>{item.label}</span>
                     </a>
                   )}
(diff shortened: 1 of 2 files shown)


packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
```diff
@@ -239,6 +246,29 @@ export class VaBreadcrumbs {
...
+    // notify Router
+    window.dispatchEvent(popStateEvent);
+  }
+
+  /**
+   * Handle click event on breadcrumb when uswds is true
+   */
+  uswdsHandler(e: MouseEvent, href: string) {
+    if (this.useRouter) {
+      this.handleRouteChange(e, href);
```
jamigibbs, line 267: What about situations where teams are needing both `Link` and `a` tags like here?

https://github.com/department-of-veterans-affairs/vets-website/blob/main/src/applications/combined-debt-portal/debt-letters/containers/DebtDetails.jsx#L57-L67

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
```diff
@@ -239,6 +246,29 @@ export class VaBreadcrumbs {
...
+   * Changes the route when uswds and useRoute props are true.
+   * Route will be set to the href value in the breadcrumb.
+   */
+  handleRouteChange(e: MouseEvent, href: string): void {
+    e.preventDefault();
+    window.history.pushState(null, '', href);
+    // create event to notify Router
+    const popStateEvent = new PopStateEvent('popstate');
+    // notify Router
+    window.dispatchEvent(popStateEvent);
```
jamigibbs, line 259: I'm not sure what you mean by "notify Router". Is this specifically for [React Router](https://reactrouter.com/en/main)?

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
it-harrison: @jamigibbs  This situation would definitely be hard to handle with the current approach without, say, adding complexity to the breadcrumbList prop, ie for each breadcrumb another property like `isLink`.

it-harrison:
> Do we see this approach limiting teams so that they couldn't use other `<Link>` props that would normally be available to them? Like could this snow ball into us having to also implement some other options listed here on the Link documentation? `relative`, `preventScrollReset`, `replace`, `state`, etc...
> 
> https://reactrouter.com/en/main/components/link

@jamigibbs good point on this - assuming we weren't doing a rewrite that would allow for slotted content, having the component emit an event that is handled in the React app would avoid us having to build more features into the component.

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
it-harrison: @jamigibbs Dispatching the event here isn't required to make the navigation work (with v 5). But from my understanding dispatching the event might more closely mimic what React Router is doing so I included it in case it increases compatibility with other versions.

jamigibbs:
>  Another approach is to emit an event and let the VFS team handle the route change programmatically in a handler that makes use of utilities that React Router provides. This approach imposes a small burden on the VFS team but may provide more future proof solution (in case, for example, directly manipulating the History API does not work well with a future version of React Router).

@it-harrison Would it be too much effort to make another POC of how this would work for teams? I'm curious how much effort this would be for them. And for the component, would we just need to suppress the native click (so maybe a new prop is still needed to tell it to do that) and then emit a custom event?

PR #979, opened 2023-12-08
va-breadcrumbs: add support for React Router via custom event

it-harrison:
## Chromatic

https://2188-poc-ii--60f9b557105290003b387cd5.chromatic.com

## Description
This PR makes the uswds version of va-breadcrumbs compatible with React Router. The original issue mentioned the <Link component. Making the uswds version play nicely with <Link> would require a total rewrite of va-breadcrumbs since currently the uswds version does not use slots. 

This PR makes the uswds version work with React Router by emitting an event the VFS team can handle in its app with the utilities that React Router provides. 

Closes [2188](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2188)

## QA Checklist

## Screenshots
In vets-website I modified [this file ](https://github.com/department-of-veterans-affairs/vets-website/blob/main/src/applications/simple-forms/mock-simple-forms-patterns/containers/IntroductionPage.jsx)because it is part of an app that uses React Router. The app runs locally at http://localhost:3001/mock-simple-forms-patterns/introduction:

Screenshot 2023-12-11 at 11 59 05 AM

A video of the page in action:

https://github.com/department-of-veterans-affairs/component-library/assets/8867779/26f9de7f-da2e-41e9-9196-ac77f9da53ff

## Acceptance criteria

## Definition of done


it-harrison:
@jamigibbs event-based POC here

it-harrison:
> Thanks @it-harrison! I am curious to hear what the other engineers think but to me this feels like a cleaner and more flexible solution for everyone.
> 
> I suspect in situations where a [router and native anchor links](https://github.com/department-of-veterans-affairs/vets-website/blob/main/src/applications/combined-debt-portal/debt-letters/containers/DebtDetails.jsx#L57-L67) are mixed, the standard link behavior could also be managed in the callback if needed?

@jamigibbs yes, I agree, the standard behavior could be managed in a callback/handler. I feel this is a cleaner, easier to maintain solution, too.

packages/storybook/stories/va-breadcrumbs-uswds.stories.jsx
```diff
@@ -108,6 +108,49 @@ const WrappingCrumbsTemplate = ({
...
+    console.log(`the href is: ${href}`)
+  }
+
+  return (
+    <div>
+      <p>Some of the breadcrumbs in this example have a <code>isRouterLink:true</code> property.
+        When the corresponding anchor tag is clicked these links emit a <code>route-change</code> event.
+        This event can be handled in a React component where utilities provided 
+        by React Router can be used to change the page under view.
+      </p>
```
jamigibbs, line 140: I wonder if this example could be clearer with a small code block of what the callback could look like. Maybe similar to how we're explaining how to handle custom validation on memorable date?

https://2188-poc-ii--60f9b557105290003b387cd5.chromatic.com/?path=/docs/components-va-memorable-date--custom-validation

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
```diff
@@ -239,6 +250,26 @@ export class VaBreadcrumbs {
...
+   */
+   handleRouteChange(e: MouseEvent, href: string): void {
+    e.preventDefault();
+    this.routeChange.emit({href});
+  }
+
+  /**
+   * Handle click event on breadcrumb when uswds is true
+   */
+  uswdsHandler(e: MouseEvent, breadcrumb: { href: string, isRouterLink?: boolean}): void {
```
jamigibbs, line 265: I think naming this `uswdsHandler` doesn't really explain what the method does in a future proof way. Like at some point there will be no "uswds" and "non-uswds" variations. Maybe just `handleClick`?

packages/storybook/stories/va-breadcrumbs-uswds.stories.jsx
it-harrison: @jami - good call, done.

Andrew565 (approved):

jamigibbs (approved):
Great! Let's ship it! 🚀

micahchiang (commented):
I noticed something when testing this in Chromatic while having the dev console open. The console.log for the last breadcrumb in the list never fires. It's because the href is hardcoded to `#content`. 

Wouldn't this mean that if a user were to be on the last page, use the breadcrumb to go to the second or third link, and then attempt to return to the last page using the breadcrumb, they would be unable to do so? 

Here's a screen cap where the console logs fire for breadcrumbs 2 and 3, but not for 4: 
[Screencast from 12-13-2023 12:59:36 PM.webm](https://github.com/department-of-veterans-affairs/component-library/assets/15097156/725f0c66-ffaf-4fcb-a2f4-719dd76606c3)

it-harrison:
> I noticed something when testing this in Chromatic while having the dev console open. The console.log for the last breadcrumb in the list never fires. It's because the href is hardcoded to `#content`.
> 
> Wouldn't this mean that if a user were to be on the last page, use the breadcrumb to go to the second or third link, and then attempt to return to the last page using the breadcrumb, they would be unable to do so?
> 
> Here's a screen cap where the console logs fire for breadcrumbs 2 and 3, but not for 4: [Screencast from 12-13-2023 12:59:36 PM.webm](https://github.com/department-of-veterans-affairs/component-library/assets/15097156/725f0c66-ffaf-4fcb-a2f4-719dd76606c3)

@micahchiang yes, it would...I assume normally the breadcrumbs are re-rendered after navigation so the last breadcrumb is always the current page, meaning you would have to use the back button to return to the previous page if the previous page were "deeper" in the hierarchy.

 So then if they don't re-render the breadcrumbs there will be the issue you describe...and in that case the breadcrumbs are not acting like normal breadcrumbs (where the last breadcrumb doesn't correspond to the current page)
 
 I guess it matters how the component is used? I'm not sure if we need to provide guidance or change something about the component.

jamigibbs:
> So then if they don't re-render the breadcrumbs there will be the issue you describe.

I might not entirely understand the issue but we do have an example in Storybook already to show that the [breadcrumbs can re-render](https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/storybook/stories/va-breadcrumbs-uswds.stories.jsx#L49-L62). Would it make the router example less confusing if we also added that functionality to that example too? 

Here's how the re-render is doing it with just some basic `useState` functionality:

https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/storybook/stories/va-breadcrumbs-uswds.stories.jsx#L49-L62

PR #981, opened 2023-12-11
Add custom CSS type selectors from Formation

jamigibbs:
## Chromatic

https://2043-move-formation-base-styles--60f9b557105290003b387cd5.chromatic.com

## Description
This is adding styles from Formation that have been identified as:

1. being a parent class > child element/tag style (a parent custom class styling na HTML element), and, 
2. still being used in vets-website or content-build

I have left a comment on each style block. **It might be possible not to have to add any of them with a little extra work** like replacing instances of them with a web component or putting the styles directly in the app using them.

This is the Formation PR with related updates there: 
- https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/pull/1001

related issue:
- https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2043

## QA Checklist
- [x] Component maintains 1:1 parity with design mocks
  - Not applicable
- [x] Text is consistent with what's been provided in the mocks
  - Not applicable
- [x] Component behaves as expected across breakpoints
  - Not applicable
- [x] Accessibility expert has signed off on code changes (if applicable. If not applicable provide reason why)
  - Not applicable
- [X] Designer has signed off on changes (if applicable. If not applicable provide reason why)
  - Not applicable
- [X] Tab order and focus state work as expected
  - Not applicable
- [x] Changes have been tested against screen readers (if applicable. If not applicable provide reason why)
  - Not applicable
- [x] New components are covered by e2e tests; updates to existing components are covered by existing test suite
  - Not applicable
- [x] Changes have been tested in vets-website using Verdaccio (if applicable. If not applicable provide reason why)
  - I have tested Formation using Verdaccio with those related updates. I have not tested CSS Library there though because it has not been connected yet. We have a separate ticket for that here: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2287
 
## Screenshots
N/A

## Acceptance criteria
- [ ] QA checklist has been completed
- [ ] Screenshots have been attached that cover desktop and mobile screens

## Definition of done
- [ ] Documentation has been updated, if applicable
- [ ] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)

PR #982, opened 2023-12-12
Add CSS type selectors from Formation to CSS Library

jamigibbs:
## Chromatic

https://2042-css-library-formation-selectors--60f9b557105290003b387cd5.chromatic.com

## Description
This PR is adding CSS type seletions from [Formation](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/master/packages/formation/sass/base/_va.scss) to the CSS Library.

- https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2042

## QA Checklist
  - Not applicable
  - Not applicable
  - Not applicable
  - Not applicable
  - Not applicable
  - Not applicable
  - Not applicable
  - Not applicable
  - I have tested Formation using Verdaccio with those related updates. I have not tested CSS Library there though because it has not been connected yet. We have a separate ticket for that here: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2287
## Acceptance criteria

## Definition of done


packages/css-library/src/stylesheets/print.scss
```diff
@@ -0,0 +1,29 @@
+.print-only {
+  display: none;
+}
+
+@media print {
+  header, footer, nav {
```
jamigibbs, line 6: A print stylesheet! How do we feel about adding this? There was a mix of CSS elements and classes and instead of splitting them up, I figured they could all just be together. 

Here is the original in Formation: https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/master/packages/formation/sass/modules/_m-print.scss

packages/css-library/src/stylesheets/elements.scss
```diff
@@ -5,3 +5,158 @@
 // p {
 //    @include @usa-paragraph
 // }
+
+@use "../../dist/tokens/scss/variables" as *;
+
+$color-link-default-hover: rgba($color-black, 0.05);
```
jamigibbs, line 11: I mentioned this in the [other PR](https://github.com/department-of-veterans-affairs/component-library/pull/981/files#r1423104505) but here it is again. How do add this new color token to the variables build?

packages/css-library/src/stylesheets/elements.scss
```diff
@@ -5,3 +5,158 @@
...
+$small-screen:  481px;  // TODO: USWDSv3 breakpoints?
+$large-screen:  1201px; // TODO: USWDSv3 breakpoints?
```
jamigibbs, line 13: What is our plan with breakpoints? Are we adopting the USWDS v3 breakpoints? These variables are form Formation here:

https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/master/packages/formation/sass/base/_b-breakpoints.scss

packages/css-library/src/stylesheets/elements.scss
```diff
@@ -5,3 +5,158 @@
...
+  color: $color-base;
+}
+
+article > h1 {
+  margin-top: 0;
+}
+
+hr {
+  margin: 2.5em 0;
+  margin: 48px 0 40px; // converted to 16px base font: 3rem 0 2.5rem
```
jamigibbs, line 36: I'm converting `rem` to pixels so that it's compatible with both 10px and 16px base fonts.

packages/css-library/src/stylesheets/elements.scss
```diff
@@ -5,3 +5,158 @@
...
+
+dd + dt {
+  margin-top: 1.5em;
+}
+
+// Figure / Caption
+figure {
+  font-size: 0.85em;
+  margin-left: -15px; // converted to 16px base font: -1.875rem/2
+  margin-right: -15px; // converted to 16px base font: -1.875rem/2
```
jamigibbs, line 118: Same here. Converted rem to pixels so that it's compatible with both 10px and 16px base fonts.

powellkerry (approved):

packages/css-library/src/stylesheets/elements.scss
micahchiang: yeah the opacity thing is tricky. I wonder if there's a `uswds-system-color-grey--` token out there that has a rgba value already converted to hex. Or maybe there's a way to update our build script to support something like that.

packages/css-library/src/stylesheets/elements.scss
micahchiang: As far as I know, we are replacing one breakpoint but keeping the others - https://vfs.atlassian.net/wiki/spaces/DST/pages/2731245815/DR001+VA+Design+System+v3+upgrade+rollout+steps

768px will be replaced with 640px.

packages/css-library/src/stylesheets/print.scss
micahchiang: I'll have to read up on `print` lol. I'm inclined to keep it separate because it feels like a pretty specific use case and does some minor things to the `.schemaform-` classes, but it's a weak opinion.

jamigibbs: @micahchiang Now that you mention `.schemaform-`, should that even be in Formation or CSS Library? Wouldn't that be better in the forms library itself?

```css
  .schemaform-title,
  .schemaform-subtitle,
  .screen-only {
    display: none;
  }
```

packages/css-library/src/stylesheets/elements.scss
jamigibbs: @micahchiang Do we have a ticket yet for adding breakpoints to the CSS Library? I guess at a minimum we need [these](https://design.va.gov/foundation/breakpoints#names-and-values) but I see [other stuff](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/master/packages/formation/sass/base/_b-breakpoints.scss#L21) in Formation too that might need to be audited. And possibly some hidden things in these [imports](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/master/packages/formation/sass/shared-variables.scss)?

packages/css-library/src/stylesheets/print.scss
jamigibbs: Nevermind! It seems like it's in many different apps so probably a legit global need.

Screenshot 2023-12-13 at 10 44 15 AM

packages/css-library/src/stylesheets/elements.scss
micahchiang: We might be able to use some sort of style dictionary transform - https://github.com/amzn/style-dictionary/blob/main/examples/advanced/transitive-transforms/tokens/color/overlay.json5

https://amzn.github.io/style-dictionary/#/transforms?id=transitive-transforms

jamigibbs: @micahchiang I just found that too! 😂 Looks promising!

jamigibbs: Looking at the transform approach, I think that might complicate things beyond what we need. If we were needing to transform the composed rgba into any other color representation, then that sounds like the right approach.

But just composing it like this seems to be working fine so I think this might be the way to go for now:

https://github.com/department-of-veterans-affairs/component-library/blob/b27e5104f18d7f4eea8aba7d10f495ba0ed2db57/packages/css-library/tokens/color.json#L7-L11

packages/css-library/src/stylesheets/elements.scss
jamigibbs: I'm just going to add the breakpoint tokens that I need for this PR for now: https://github.com/department-of-veterans-affairs/component-library/pull/982/commits/15b2b06fa98bdb3e6d27a44a9d8d941656570679

micahchiang: Sorry, we don't have a ticket yet, I'll write one to audit and port the remaining stuff we need. I think I saw you added two in a recent commit.

packages/css-library/src/stylesheets/print.scss
micahchiang: yeah all of those schemaform- classes are gonna be an interesting thing to deal with. Their use is very pervasive to say the least.

packages/css-library/src/stylesheets/elements.scss
micahchiang: Very nice! Definitely the way to go here :+1: out of curiosity, could we also do something like 

```css
"link-default-hover": {
  "alpha": "0.05"
  "value": "rgba({uswds-system-color-gray-100}, {color.link-defaul-hover.alpha})"
}
```
or, could we do something like this where this is the converted hex equivalent for `rgba(0, 0, 0, 0.05)`

```css
"link-default-hover": {
  "value": "{uswds-system-color-gray-100}0d"
}
```

packages/css-library/src/stylesheets/elements.scss
micahchiang: This is interesting - https://github.com/uswds/uswds/blob/62d297a7f09230e5fc4c8b9d9aaae099a4ebda2b/packages/uswds-core/src/styles/settings/_settings-utilities.scss#L34-L56 

I think we might be able to get away with just configuring what breakpoints we want to use in our [settings file](https://github.com/department-of-veterans-affairs/component-library/blob/8e80295be8a003232b1bae837f9e9fefb7f2ae31/packages/css-library/src/stylesheets/utilities.scss#L12). Question would remain about overriding some of the default pixel values though.

micahchiang: eh, now that I think about it. We probably need to bring the breakpoints over anyway. There are plenty of places in vets-website where applications are using them directly in custom style sheets =\

micahchiang: Ticket to add breakpoints to css library - https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2334

packages/css-library/src/stylesheets/elements.scss
jamigibbs: >  out of curiosity, could we also do something like
> 
> ```css
> "link-default-hover": {
>   "alpha": "0.05"
>   "value": "rgba({uswds-system-color-gray-100}, {color.link-defaul-hover.alpha})"
> }
> ```

Yes! Good call. I like using the USWDS black instead. 


> or, could we do something like this where this is the converted hex equivalent for `rgba(0, 0, 0, 0.05)`
> 
> ```css
> "link-default-hover": {
>   "value": "{uswds-system-color-gray-100}0d"
> }
> ```

No, I don't believe this is the same thing. Assigning this as just a hex value means that a `transition-property` will not work. It needs the `alpha` value to display the fade effect going from black to 0.05 of black.

jamigibbs: As a sanity check, I tested using a hex value instead of `0,0,0` for rgba in vets-website because the[ rgba function specs don't show hex as an example](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/rgb):

Screenshot 2023-12-14 at 8 54 20 AM

Screenshot 2023-12-14 at 9 19 33 AM

It looks like `rgba` automatically converts the hex to a red/green/blue value so I think using `uswds-system-color-gray-100` instead of 0,0,0 will be fine. 

And the support for hex color notation looks good (except for IE which we no longer support anyway): https://caniuse.com/css-rrggbbaa

**Chrome**
Screenshot 2023-12-14 at 9 18 18 AM

micahchiang (approved):

PR #1009, opened 2024-01-05
va-radio: add new form pattern support

harshil1793:
## Chromatic

https://2269-update-radio--60f9b557105290003b387cd5.chromatic.com

Radio Input component: https://2269-update-radio--60f9b557105290003b387cd5.chromatic.com/?path=/docs/uswds-va-radio--default

---

## Description

This PR adds support for the new forms pattern to the `va-radio`. It includes new props for each component:

- `useFormsPattern` : String "single" or "multiple"
- `formHeadingLevel` : Number
- `formHeading` : String

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2269

## QA Checklist

## Screenshots
Screenshot 2024-01-05 at 3 55 40 PM

Screenshot 2024-01-05 at 3 55 24 PM

Screenshot 2024-01-05 at 3 55 16 PM

## Acceptance criteria

## Definition of done


danbrady:
> @danbrady or @LWWright7 Do you have a real world example of this from Figma or somewhere in the forms library (or just from your expertise) of what type of content we should use in our example? There's 3 different sections:

@jamigibbs I circled back through this issue chain and found the following screenshot of content (from [#2269](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2269)). Does this work?

image

jamigibbs:
> > @danbrady or @LWWright7 Do you have a real world example of this from Figma or somewhere in the forms library (or just from your expertise) of what type of content we should use in our example? There's 3 different sections:
> 
> @jamigibbs I circled back through this issue chain and found the following screenshot of content (from [#2269](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2269)). Does this work?
> 

Oh perfect. @harshil1793 How do you feel about updating the Storybook example(s) to that it's like what's in the original ticket like Dan is showing above?

packages/web-components/src/components/va-radio/va-radio.tsx
```diff
@@ -220,48 +236,77 @@ export class VaRadio {
...
+        <Host aria-invalid={error ? 'true' : 'false'} aria-label={ariaLabel} aria-labelledby={ariaLabeledByIds}>
+          {formsHeading}
+          <div class="input-wrap">
+            <fieldset class="usa-fieldset">
+              <legend class={legendClass} part="legend">
+                {HeaderLevel ? (
+                  <HeaderLevel part="header" aria-describedby={headerAriaDescribedbyId}>{label}</HeaderLevel>
+                ) : (
+                  label
+                )}
```
micahchiang, line 286: Just want to check my understanding here. If a user sets the props `labelHeaderLevel`, `useFormsPattern`, and `formHeadingLevel`, then this component would potentially render two separate headers: One right before `div.input-wrap` and one inside the `legend`. Is this something we want, or, is it something we need to handle?

harshil1793: Good point! @danbrady, could you please confirm whether this was discussed during the design process?

danbrady: I haven't been involved in this effort, so I don't know how much help I can be here. Looking back at the [initial ticket](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2269) it says:

> useFormsPattern?: string. Accepts either "single" or "multiple" to indicate if the form is a single input or will have multiple inputs. Enabling this will add a heading and description for integrating into the forms pattern.
**Enabling this will remove headings/descriptions from fieldset, legend, these will be placed outside.**

I'm not sure if that answer your question though. @rsmithadhoc or @jamigibbs can you shed some light here?

harshil1793: The ticket description answers the question. Thanks Dan

rsmithadhoc:
Hi @harshil1793,

Currently, the `form-title` and `form-description` are not read out when tabbing to the field. I think there are a few reasons for this:
- The `label` prop gets set as the `aria-label`, so that may override `aria-labeledby`. I think if we are using the forms pattern, the `label` prop can be ignored.
  - This also read out "Select a historical figure", even though it was the burn pit example question. This may just be a re-used variable for the Storybook example.
- The `aria-labelledby` would have to be attached to `fieldset` tag (see the screenshot on the issue). That way, when tabbing to the radio, it knows the form title and description are connected to that radio group and will read those out.

Thanks!

harshil1793:
@rsmithadhoc, I've resolved the issue; now the heading and description are being read out upon tabbing. However, I noticed that the field titles are not being announced, which appears to be a pre-existing bug. I'll go ahead and open an issue to address this.

rsmithadhoc (changes requested):
@harshil1793 I was not able to get it working as expected, I think a few more changes are needed.

- I don't believe this should have the `tabindex`. We don't want the heading to be a tabbable element.
    - The `aria-labelledby="form-title form-description"` will handle the title being read out when tabbing to the field, so we don't need an explicit tab stop on the heading.
- The `<fieldset>` needs the `aria-labelledby`, that is currently still on the host element.

This example HTML functions as expected, as long as we can replicate this in our component, it should work as intended. To test this, you can save the code below to an HTML file, open VoiceOver (Command + F5), then Tab. It should read out everything, "Yes" (the currently selected option), the title, and the description (see screenshot).
image

```
<div id="form-title">
    Burn pit S2.1, did you serve in any of there locations?
</div>

<div id="form-description">
    <p>HTML passed into the form-description slot:</p>
    <ul>
        <li>Bahrain</li>
        <li>Iraq</li>
        <li>Kuwait</li>
        <li>Oman</li>
        <li>Qatar</li>
        <li>Saudi Arabia</li>
        <li>Somalia</li>
        <li>The United Arab Emirates(UAE)</li>
        <li>The airspace above any of these locations</li>
    </ul>
</div>

<fieldset aria-labelledby="form-title form-description">
    <input type="radio" name="yes" id="yes" value="yes" checked />
    <label for="yes">Yes</label>

    <input type="radio" name="no" id="no" value="no" />
    <label for="no">No</label>
</fieldset>

```

rsmithadhoc (changes requested):
@harshil1793 I think it is mostly looking good, nice work here, there are a couple of small things that may just be issues with the Storybook examples.

- For **Forms Pattern Single Error**, it doesn't announce the radio option.
    - I think this is related to **Forms Pattern Single** and **Forms Pattern Single Error**, saying "Radio button _ of 8" even though there are only 3 options - this might just be an issue with the Storybook example having the same IDs. There are multiples of radio group 5 on the page (one example `id="group51input"`), but the **Forms Pattern Single Error** (radio group 6) options still say `group5`.
    image
- For **Forms Pattern Multiple**, it is not announcing "Select one historical figure (*Required)". I think the `form-description` div is not populated on this example and would need `sr-only`.
    image

rsmithadhoc (approved):
@harshil1793 Looks good, thanks!

jamigibbs (approved):
Great job, Harshil. Thank you!

PR #1015, opened 2024-01-11
adds util function to  wrap slot assignedElementts

rmessina1010:
A test solution to:
When creating v3/USWDS components, we import USWDS styles and apply them by marking up the component with USWDS classes (starting with usa-). However, when we accept slot content it is isolated from outside styles, including USWDS, and so the content remains unstyled.

In this attempt, I assign a new blank div element to assigned content, then attach the classes to it, and NOT the slot.

[https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2298](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2298)


micahchiang (commented):
I just realized I never finished submitting this review, ugh. Sorry!

PR #1017, opened 2024-01-11
va-file-input: Adding a V3-compatible (USWDS) version

Andrew565:
## Chromatic

https://1500-file-input-v3--60f9b557105290003b387cd5.chromatic.com/

---

## Description
Closes [#1500](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1500)

## QA Checklist
  - Not applicable, this is a new component

### Desktop
Screenshot 2024-01-08 at 4 46 11 PM
### Mobile
Screenshot 2024-01-08 at 4 46 27 PM

## Acceptance criteria

## Definition of done


LWWright7 (approved):

packages/web-components/src/components/va-file-input/va-file-input-upgrader.ts
```diff
@@ -0,0 +1,609 @@
+/* eslint-disable i18next/no-literal-string */
+import { Sanitizer, selectOrMatches } from '../../utils/utils';
+
+const PREFIX = 'usa';
```
jamigibbs, line 4: Can you mention what (if any) changes we needed to make from the original USWDS script in order to get this working for the web component?

https://github.com/uswds/uswds/blob/202448875393d39d663614d2d8f9a12c91fed75e/packages/usa-file-input/src/index.js

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -61,10 +63,25 @@ export class VaFileInput {
...
   /**
    * Emit component-library-analytics events on the file input change event.
    */
-   @Prop() enableAnalytics?: boolean = false;
+  @Prop() enableAnalytics?: boolean = false;
+
+  /**
+   * Whether or not the component will use USWDS v3 styling.
+   */
+  @Prop({ reflect: true }) uswds?: boolean = false;
```
jamigibbs, line 79: Why does this need to be made reflective?

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -61,10 +63,25 @@ export class VaFileInput {
...
+
+  /**
+   * A random id used for listening to changes
+   */
+  changeListenerId: number;
```
jamigibbs, line 84: Where is this used?

packages/web-components/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/web-components",
-  "version": "4.55.0",
+  "version": "5.0.0",
```
jamigibbs, line 3: When adding components, this can just be `minor` (removing/breaking changes are major).

https://github.com/department-of-veterans-affairs/component-library?tab=readme-ov-file#minor

Screenshot 2024-01-18 at 10 16 34 AM

packages/web-components/src/components/va-file-input/va-file-input.scss
```diff
@@ -0,0 +1,51 @@
...
+
+/* V3/USWDS Styles */
+.usa-label--required,
+.usa-error-message {
+  color: var(--v3-color-error-dark);
+}
+
+:host([error]:not([error=''][uswds]:not([uswds='']))) {
+  border-left-color: var(--v3-color-error-dark);
+}
```
jamigibbs, line 39: There's a USWDS specific mixin for the error display. Can that be used here too?

```
@import '../../mixins/uswds-error-border.scss';
```

packages/web-components/src/components/va-file-input/va-file-input.scss
```diff
@@ -0,0 +1,51 @@
...
+
+.usa-file-input .usa-file-input__input[type] {
+  padding: 0;
+}
+
+.usa-file-input__choose {
+  color: var(--color-link-default);
+}
```
jamigibbs, line 47: Why do we need to override the USWDS color here?

packages/web-components/src/components/va-file-input/va-file-input.scss
```diff
@@ -0,0 +1,51 @@
...
+.usa-error-message {
+  color: var(--v3-color-error-dark);
+}
+
+:host([error]:not([error=''][uswds]:not([uswds='']))) {
+  border-left-color: var(--v3-color-error-dark);
+}
+
+.usa-file-input .usa-file-input__input[type] {
+  padding: 0;
```
jamigibbs, line 42: I'm surprised we need to override the padding for this USWDS style. Is there something in the VA system that is adding padding that we're trying to fight against?

packages/web-components/src/components/va-file-input/test/va-file-input.e2e.ts
```diff
@@ -113,4 +115,123 @@ describe('va-file-input', () => {
...
+      'va-file-input >>> .usa-file-input__instructions',
+    );
+
+    expect(instructions).not.toBeNull();
+
+    const input = await page.$('pierce/.usa-file-input__input');
+    expect(input).not.toBeNull();
+
+    await input
+      .uploadFile(filePath)
```
jamigibbs, line 222: Is it possible to add tests for some of the drag and drop functionality? There are some methods available on the `Element` type that we might be able to use. Does something like this work?

```
await input.drop(filePath)
```

Screenshot 2024-01-18 at 10 24 37 AM

packages/web-components/src/components/va-file-input/va-file-input.scss
```diff
@@ -0,0 +1,51 @@
+@forward 'settings';
+
+@use 'usa-file-input/src/styles/usa-file-input';
+@use 'usa-label/src/styles/usa-label';
+@use 'usa-hint/src/styles/usa-hint';
+@import '../../global/formation_overrides';
+
+/* Original Styles */
```
jamigibbs, line 8: This is a nitpick but with all of the other USWDS variations, we have added the USWDS classes to the top of the stylesheet above the `/* Original Styles */` comment.

I'm not sure why we're diverging in this component. For example: https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/src/components/va-radio/va-radio.scss

jamigibbs:
Doing a visual check against the USWDS version, I see a few discrepancies.

The SVG looks smaller in Storybook and after drop, there is no focus.

Screenshot 2024-01-18 at 10 29 42 AM

Screenshot 2024-01-18 at 10 29 48 AM

And does the font for the file name look the same? It seems like the font family is different for some reason. 🤔 

**USWDS**
Screenshot 2024-01-18 at 10 37 27 AM

**Storybook**
Screenshot 2024-01-18 at 10 37 18 AM

packages/web-components/src/components/va-file-input/va-file-input-upgrader.ts
Andrew565: I didn't keep track, but there are a lot of changes.

packages/web-components/src/components/va-file-input/va-file-input.tsx
Andrew565: I just assumed we always made it reflective, because it's reflected everywhere else.

packages/web-components/src/components/va-file-input/va-file-input.tsx
Andrew565: Removed, no longer needed

packages/web-components/src/components/va-file-input/va-file-input.scss
Andrew565: It was the wrong color blue when I tested it.

packages/web-components/src/components/va-file-input/va-file-input.scss
Andrew565: Yes, there was a pre-existing style that was pushing the target down by 5-10px, so it was causing a glitch where the cursor would change at the wrong position.

packages/web-components/src/components/va-file-input/test/va-file-input.e2e.ts
Andrew565: Tried this, doesn't work. A quick review of Puppeteer's docs suggest this is meant for dragging html elements from one place to another, not for file uploads.

packages/web-components/src/components/va-file-input/va-file-input.tsx
jamigibbs: When `{ reflect: true }` is set, it renders the prop in the DOM. Where it's in one of the existing components, it's probably because the component is being composed as part of a parent component and we need access to that prop setting in the DOM as part of that integration.

So unless there's a valid reason to add it for va-file-input, we don't need it set.

https://stenciljs.com/docs/properties#reflect-properties-values-to-attributes-reflect

Most components don't need it.

Screenshot 2024-01-18 at 1 34 19 PM

packages/web-components/src/components/va-file-input/va-file-input.scss
jamigibbs: This is a different color?

--color-link-default is #005ea2 which is rgb(0, 94, 162)

Screenshot 2024-01-18 at 1 48 29 PM

Andrew565:
Fixed the file preview image size.

packages/web-components/src/components/va-file-input/va-file-input-upgrader.ts
```diff
@@ -0,0 +1,609 @@
...
+const DRAG_CLASS = `${PREFIX}-file-input--drag`;
+const LOADING_CLASS = 'is-loading';
+const INVALID_FILE_CLASS = 'has-invalid-file';
+const GENERIC_PREVIEW_CLASS_NAME = `${PREFIX}-file-input__preview-image`;
+const GENERIC_PREVIEW_CLASS = `${GENERIC_PREVIEW_CLASS_NAME}--generic`;
+const PDF_PREVIEW_CLASS = `${GENERIC_PREVIEW_CLASS_NAME}--pdf`;
+const WORD_PREVIEW_CLASS = `${GENERIC_PREVIEW_CLASS_NAME}--word`;
+const VIDEO_PREVIEW_CLASS = `${GENERIC_PREVIEW_CLASS_NAME}--video`;
+const EXCEL_PREVIEW_CLASS = `${GENERIC_PREVIEW_CLASS_NAME}--excel`;
+const SR_ONLY_CLASS = `sr-only`;
```
jamigibbs, line 27: This was changed from:

```
const SR_ONLY_CLASS = `${PREFIX}-sr-only`;
```

To:
```
const SR_ONLY_CLASS = `sr-only`;
```
If you import the usa-sr-only module into the stylesheet, I don't think you need to make that change.

```
@use 'uswds-helpers/src/styles/usa-sr-only';
```

For example: https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/src/components/va-text-input/va-text-input.scss#L3

packages/web-components/src/components/va-file-input/va-file-input-upgrader.ts
jamigibbs: I was able to create a basic diff here: https://www.diffchecker.com/D8NTH2nR/

packages/web-components/src/components/va-file-input/va-file-input.scss
jamigibbs: Sorry, I still don't understand what's going on here. The padding does look a little different between them.

**USWDS**

Screenshot 2024-01-18 at 2 06 42 PM

**Storybook**
Screenshot 2024-01-18 at 2 06 39 PM

There is a `rem` value in the USWDS class. Does that just need to be added to the formation override stylesheet?

Screenshot 2024-01-18 at 2 09 10 PM

packages/web-components/src/components/va-file-input/va-file-input-upgrader.ts
Andrew565: Thanks!

packages/web-components/src/components/va-file-input/va-file-input.scss
Andrew565: Agreed, no longer appear to need this style.

Andrew565:
@jamigibbs I think I've addressed all of the issues you've brought up so far, but please let me know if I've missed anything. The "v3 emits vachange event once" test is passing locally but not on CI, so that's annoying. And of course it's failing without an error message saying why. 🙄 I'm very tempted to just skip this test for now, do you have an opinion on that?

jamigibbs:
> @jamigibbs I think I've addressed all of the issues you've brought up so far, but please let me know if I've missed anything. The "v3 emits vachange event once" test is passing locally but not on CI, so that's annoying. And of course it's failing without an error message saying why. 🙄 I'm very tempted to just skip this test for now, do you have an opinion on that?

It does seem flaky. It's passing now but I was able to get it to fail locally too after a few tries. We can either merge it now as-is if it's passing but if you do end up needing to skip it, just create a follow-up ticket for us to fix it and link to it here.

jamigibbs:
@Andrew565 I think the only couple of things I see left is that there are some USWDS `rem` values not adjusting for Formation:

Screenshot 2024-01-19 at 11 26 05 AM

Screenshot 2024-01-19 at 11 25 28 AM


And were you able to figure out what's up with the font-family difference on the file name text? 

Screenshot 2024-01-19 at 11 24 32 AM

Andrew565:
@jamigibbs I don't know what's up with the font-family difference, no.

jamigibbs:
> @jamigibbs I don't know what's up with the font-family difference, no.

@Andrew565 ok. Well I spent a few minutes on it and I wonder if this is a bug with the USWDS example? The font-family is pulling in from some accordion class.

Screenshot 2024-01-23 at 8 47 05 AM

The reason I think it's important for us to document this is because we often get issues submitted about font family being incorrect and I want to makes sure we either a) fix it now so we don't have those issues logged or b) understand why USWDS looks one way and the VA variation looks another way so we can explain it. This is an fyi to @danbrady as well in case you get feedback about Figma discrepancy too.

danbrady:
It looks like the `.site-accordion-code` class that's changing the font family is from the *site's* CSS (more specifically `_uswds-theme-custom-styles.scss` line #2461), not USWDS.  I loaded USWDS 3.7 in isolation in this [Codepen](https://codepen.io/danbrady/pen/jOddKwW) and dropped in the same File Input code block, dropped in a file and inspected, and I do *not* see Public Sans, but Source Sans Pro Web.

jamigibbs:
> It looks like the `.site-accordion-code` class that's changing the font family is from the _site's_ CSS (more specifically `_uswds-theme-custom-styles.scss` line #2461), not USWDS. I loaded USWDS 3.7 in isolation in this [Codepen](https://codepen.io/danbrady/pen/jOddKwW) and dropped in the same File Input code block and inspected, and I do _not_ see Public Sans, but Source Sans Pro Web.

👍  As long as we're all on the same page about which font should be displaying. Thanks for checking @danbrady!

packages/web-components/src/components/va-file-input/va-file-input-upgrader.ts
jamigibbs: @Andrew565 I think you missed updating this `sr-only` Formation class to the USWDS `usa-sr-only`

https://github.com/department-of-veterans-affairs/component-library/blob/549701650399c9496c72af0399db41fa3ea4b428/packages/web-components/src/components/va-file-input/va-file-input.tsx#L188C17-L188C66

```html
  <span id="input-error-message" role="alert">
    {error && (
      <Fragment>
        <span class="sr-only">{i18next.t('error')}</span> <-------
        <span class="usa-error-message">{error}</span>
      </Fragment>
    )}
  </span>
```

jamigibbs (approved):
I see one thing here that you'll want to double check before merging: https://github.com/department-of-veterans-affairs/component-library/pull/1017#discussion_r1463490175

PR #1030, opened 2024-01-19
Parse slotted content and add USWDS classes

rmessina1010:
Chromatic: https://60f9b557105290003b387cd5-cvdxmgtcos.chromatic.com/?path=/docs/components-va-featured-content--default

A test solution to:
When creating v3/USWDS components, we import USWDS styles and apply them by marking up the component with USWDS classes (starting with usa-). However, when we accept slot content it is isolated from outside styles, including USWDS, and so the content remains unstyled.

In this attempt, I have reinterpreted the original assignment. This PR intends to:

- [x] Add classes and ID as prescribed in the corresponding  USWDS documentation; 
- [x] The CSS styles of slotted content should then  adhere to the styles imported into the global stylesheet (the light DOM). 
- [x] If a wrapper is needed, I am altering the component  to hard code the wrapper with the appropriate classes and attributes applied.
- [x] But options are open so this can be done  conditionally and programmatically

https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2298


packages/web-components/src/utils/slot-child-class-append.ts
```diff
@@ -0,0 +1,106 @@
...
+                el.classList.add(theClass);
+            }
+            if (!shallow){
+                const subEls:NodeListOf<HTMLElement>= el.querySelectorAll(selector);
+                subEls.forEach((subEl:HTMLElement)=>subEl.classList.add(theClass))
+            }
+       }); 
+   }
+   return;
+}
```
micahchiang, line 66:
```suggestion
       els.forEach((el:HTMLElement) => {
            if (el.matches(selector)) { 
                el.classList.add(theClass);
            }
            if (!shallow) {
                const subEls:NodeListOf<HTMLElement> = el.querySelectorAll(selector);
                subEls.forEach((subEl:HTMLElement) => subEl.classList.add(theClass))
            }
       }); 
   }
   return;
}
```

packages/web-components/src/utils/slot-child-class-append.ts
```diff
@@ -0,0 +1,106 @@
...
+
+// Applies IDs as described in iterateAddAttrs
+export function addIdToSlottedElements(slotEl:HTMLSlotElement, selector:string, theId:string, shallow:boolean) : null {
+   const els:Element[]= slotEl.assignedElements();
```
micahchiang, line 70:
```suggestion
   const els:Element[] = slotEl.assignedElements();
```

packages/web-components/src/utils/slot-child-class-append.ts
```diff
@@ -0,0 +1,106 @@
...
+   if (els && els.length) {
+       for ( let i=0; i<els.length; i++){
```
micahchiang, line 72:
```suggestion
       for (let i=0; i < els.length; i++) {
```

packages/web-components/src/utils/slot-child-class-append.ts
```diff
@@ -0,0 +1,106 @@
...
+           const el:Element= els[i];
```
micahchiang, line 73:
```suggestion
           const el:Element = els[i];
```

packages/web-components/src/utils/slot-child-class-append.ts
```diff
@@ -0,0 +1,106 @@
...
+               break; 
+           }
+           if (!shallow){
+                const subEl:HTMLElement = el.querySelector(selector) ;
+                if (subEl){
+                    subEl.id=theId;
+                    break;
+                }
+            }
+       }
```
micahchiang, line 85:
```suggestion
           if (el.matches(selector)) { 
               el.id = theId;
               break; 
           }
           if (!shallow) {
                const subEl:HTMLElement = el.querySelector(selector) ;
                if (subEl) {
                    subEl.id=theId;
                    break;
                }
            }
       }
```

packages/web-components/src/utils/slot-child-class-append.ts
```diff
@@ -0,0 +1,106 @@
...
+   }
+   return;
+}
+
+// creates a new div element, adds the class 
+// and coreesponding slot.name  to it,
```
micahchiang, line 91:
```suggestion
// and corresponding slot.name to it, 
```

packages/web-components/src/utils/slot-child-class-append.ts
```diff
@@ -0,0 +1,106 @@
...
+   }
+   return;
+}
+
+// creates a new div element, adds the class 
+// and coreesponding slot.name  to it, 
+// appends it to the list returned by slot.assignedElements()
+// removes the slot attribute from all siblings 
+// and moves them into  the newly created div
+export function wrapSlotElements(slot:HTMLSlotElement, clss:string ) : null{
```
micahchiang, line 95:
```suggestion
export function wrapSlotElements(slot:HTMLSlotElement, class:string) : null {
```

packages/web-components/src/utils/slot-child-class-append.ts
```diff
@@ -0,0 +1,106 @@
...
+    const children:Element[] = slot.assignedElements();
+    const wrapper:HTMLElement = document.createElement('div');
+    if (slot.name) { wrapper.slot = slot.name};
```
micahchiang, line 98:
```suggestion
    if (slot.name) { wrapper.slot = slot.name };
```

packages/web-components/src/utils/slot-child-class-append.ts
```diff
@@ -0,0 +1,106 @@
...
+    children.at(-1).after(wrapper);
+    wrapper.classList.add(clss);
```
micahchiang, line 100:
```suggestion
    wrapper.classList.add(class);
```

packages/web-components/src/utils/slot-child-class-append.ts
```diff
@@ -0,0 +1,106 @@
...
+    for (let i=0; i<children.length; i++   ){
```
micahchiang, line 101:
```suggestion
    for (let i = 0; i < children.length; i++) { 
```

micahchiang (commented):
A lot of these are stylistic changes that likely would have been caught by a linter. Additionally though, the impression I get is that embedding `iterateAddAttrs` in the component itself is making an assumption that these are things we want every single time we use this. Not sure if we do or not, or if that's even correct, but that's what it feels like on first pass.

packages/web-components/src/components/va-featured-content/va-featured-content.tsx
```diff
@@ -1,4 +1,15 @@
 import { Component, Host, h, Prop, Element, State } from '@stencil/core';
+import { iterateAddAttrs, SlotTarget } from '../../utils/slot-child-class-append';
+
+const targetingArray:SlotTarget[] =[
+  { slotName: 'headline',
+    selectorClasses: [["h3","usa-summary-box__heading",true]],
+    selectorId: [['h3', 'summary-box-key-information']]
+  },
+  { slotName: '',
+    selectorClasses: [['a', 'usa-summary-box__link'], ['ul', 'usa-list']],
```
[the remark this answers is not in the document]
rmessina1010: That's the ambiguity I was facing. A VFS team could or could not pass any of the above,  so this parses _any_ possible composition of the `summary-box` component fits  the [example](https://designsystem.digital.gov/components/summary-box/#component-code). This configuration would vary component by component. 

The function of the utils is to parse through the slotted elements and add the classes ( that the content team do not have to... so I thought it became our responsibility.  I made the assumption  from my interpretation of the sample content, provided above , if there are more specific conditions that can be adjusted...and on a slot by slot basis :)

packages/web-components/src/utils/slot-child-class-append.ts
rmessina1010: it's my understanding that 'class' is a keyword in javascript. would 'aClass'  be more readable?

packages/web-components/src/components/va-featured-content/va-featured-content.tsx
```diff
@@ -1,4 +1,15 @@
 import { Component, Host, h, Prop, Element, State } from '@stencil/core';
+import { iterateAddAttrs, SlotTarget } from '../../utils/slot-child-class-append';
+
+const targetingArray:SlotTarget[] =[
+  { slotName: 'headline',
+    selectorClasses: [["h3","usa-summary-box__heading",true]],
+    selectorId: [['h3', 'summary-box-key-information']]
```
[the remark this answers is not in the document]
rmessina1010: This variable is storing a config for the utility.  I noticed in the example given to me, that some elements, were assigned id's as well as classes. Since I was told the VFS teams would not be responsible for assigning class attributes, I assumed they would also not assign corresponding ID's. This value can be configured  on a component slot by component slot basis. What you see here is what my interpretation of the needed IDs specific to the [featured-content example](https://designsystem.digital.gov/components/summary-box/#component-code). Also, since IDs are meant to be unique, it targets only the first selector that matches...as a safety. The selectors can be customized to fit the needs, and if the VFS/content teams give the id's to the slotted content themselves this step can/is omitted. Classes are different as they can be repeated, each matched selector ( in this case h3) would get assigned `usa-summary-box__heading`

This is trying to handle assignments from the premise that:
1) VFS/content team  will slot the elements providing no guidance other than the slot-name
2) They could compose (pass any element or grouping of element)
3) My interpretation of [summary-box guidance](https://designsystem.digital.gov/components/summary-box/#guidance), which leaves a lot open possibilities

packages/web-components/src/utils/slot-child-class-append.ts
micahchiang: lol you're right, woops. Maybe `classname` would be better here?

micahchiang:

PR #1059, opened 2024-02-16
va-file-input: Add focus border, fix input alignment, and update error state

jamigibbs:
## Chromatic

https://2441-file-input--60f9b557105290003b387cd5.chromatic.com

## Description

Figma: https://www.figma.com/file/afurtw4iqQe6y4gXfNfkkk/VADS-Component-Library?node-id=199%3A1212&mode=dev
USWDS Example: https://designsystem.digital.gov/components/file-input/

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2441

## QA Checklist

## Screenshots

**Focus Before**

Screenshot 2024-02-16 at 12 47 23 PM

**Focus After**

Screenshot 2024-02-16 at 12 49 15 PM

**Error Before**

Screenshot 2024-02-16 at 2 36 50 PM

**Error After**

Screenshot 2024-02-16 at 2 37 33 PM

## Acceptance criteria

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-file-input/va-file-input.scss
+++ b/packages/web-components/src/components/va-file-input/va-file-input.scss
@@ -4,15 +4,21 @@
 @use 'usa-label/src/styles/usa-label';
 @use 'usa-hint/src/styles/usa-hint';
 @use 'uswds-helpers/src/styles/usa-sr-only';
+
 @import '../../mixins/uswds-error-border.scss';
 @import '../../global/formation_overrides';
+@import '../../mixins/focusable.css';
 
-/* V3/USWDS Styles */
 [hidden] {
   display: none;
 }
 
-/* Original Styles */
+.usa-file-input__input {
+  box-sizing: border-box; // Formation override.
+  font-family: var(--font-source-sans);
+}
+
+/** Original Component Style **/
 @import '../../mixins/accessibility.css';
 @import '../../mixins/form-field-error.css';
 @import '../../mixins/hint-text.css';
--- a/packages/web-components/src/components/va-file-input/va-file-input.scss
+++ b/packages/web-components/src/components/va-file-input/va-file-input.scss
@@ -18,6 +18,11 @@
   font-family: var(--font-source-sans);
 }
 
+:host([error]:not([error=''])) .usa-file-input__target {
+  border-color: var(--color-secondary-dark);
+  border-width: 2px;
+}
+
 /** Original Component Style **/
 @import '../../mixins/accessibility.css';
 @import '../../mixins/form-field-error.css';
--- a/packages/web-components/src/components/va-file-input/va-file-input.scss
+++ b/packages/web-components/src/components/va-file-input/va-file-input.scss
@@ -14,7 +14,7 @@
 }
 
 .usa-file-input__input {
-  box-sizing: border-box; // Formation override.
+  box-sizing: inherit;
   font-family: var(--font-source-sans);
 }
 
--- a/packages/web-components/src/components/va-file-input/va-file-input.scss
+++ b/packages/web-components/src/components/va-file-input/va-file-input.scss
@@ -14,7 +14,7 @@
 }
 
 .usa-file-input__input {
-  box-sizing: inherit;
+  box-sizing: border-box;
   font-family: var(--font-source-sans);
 }
 
--- a/packages/web-components/package.json
+++ b/packages/web-components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/web-components",
-  "version": "5.0.3",
+  "version": "5.0.4",
   "description": "Stencil Component Starter",
   "main": "dist/index.cjs.js",
   "module": "dist/index.js",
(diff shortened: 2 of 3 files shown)


packages/web-components/src/components/va-file-input/va-file-input.scss
```diff
@@ -4,15 +4,26 @@
...
+@import '../../mixins/focusable.css';
 
-/* V3/USWDS Styles */
 [hidden] {
   display: none;
 }
 
-/* Original Styles */
+.usa-file-input__input {
+  box-sizing: border-box; // Formation override.
```
jamigibbs, line 17: This fixes the misalignment/overflow of the input but it's unusual that we need to add this at all because I would expect the USWDS module to do this automatically for us. Why would we need to make an adjustment here? 🤔

Screenshot 2024-02-16 at 12 47 23 PM

When I inspected the computed styles on the [USWDS site,](https://designsystem.digital.gov/components/file-input/) their component had `box-sizing: border-box` set where ours did not. 

It loads for them from [`uswds/elements/src/styles/_sizing.scss`](https://github.com/uswds/uswds/blob/develop/packages/uswds-elements/src/styles/_sizing.scss):

Screenshot 2024-02-16 at 2 41 07 PM

And according to the [General Settings](https://designsystem.digital.gov/documentation/settings/#general-settings) reference for `$theme-global-border-box-sizing`, this should be applying for us too. But it isn't! 

I tried to add it to our USWDS settings, but that also did not change how it was loading:

Screenshot 2024-02-16 at 3 12 14 PM

My thought is that there is something with Formation that is making the `box-sizing: inherit` come from something other than the `html` element (which has the box-sizing set correctly to `border-box`).

So until Formation is gone, we may need to explicitly set the box sizing like this.

packages/web-components/src/components/va-file-input/va-file-input.scss
```diff
@@ -4,15 +4,26 @@
...
+  font-family: var(--font-source-sans);
```
jamigibbs, line 18: The font was loading as Arial so I'm making it Source Sans here. We believe that it might be a bug on the USWDS site which is loading it as Public Sans. 

Original convo here:

https://github.com/department-of-veterans-affairs/component-library/pull/1017#issuecomment-1906220668

packages/web-components/src/components/va-file-input/va-file-input.scss
```diff
@@ -4,15 +4,26 @@
...
+}
+
+:host([error]:not([error=''])) .usa-file-input__target {
+  border-color: var(--color-secondary-dark);
+  border-width: 2px;
```
jamigibbs, line 23: This adds the error state to the drop zone dashed border.

Screenshot 2024-02-16 at 2 37 33 PM

jamigibbs:
Something I noticed when comparing the computed styles between the USWDS and VA version is that the `rem` values converted as part of the `_formation-override` stylesheet are not exactly matching. So there are some slight visual differences between the two. 

A couple reasons that this might be happening:

1. We need to update to the latest release of USWDS (currently they are on [version 3.7.1](https://github.com/uswds/uswds/releases/tag/v3.7.1) and we are using 3.4.0).
2. The [`rem-ovveride`](https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/src/global/_functions.scss) function is not working as expected.

All of this won't matter though once we get the Formation base font updated.

Screenshot 2024-02-16 at 10 26 15 AM

danbrady (commented):
This looks good to me. Nice work!

I did notice one minor interaction difference between this and the USWDS. If you attach a file, then click to change that attach file, but hit Cancel instead, in VADS it negates that second click so the first selected file remains. However, in the USWDS, after clicking Cancel, it removes that first file, essentially reinitializing the whole component.

jamigibbs:
> I did notice one minor interaction difference between this and the USWDS. If you attach a file, then click to change that attach file, but hit Cancel instead, in VADS it negates that second click so the first selected file remains. However, in the USWDS, after clicking Cancel, it removes that first file, essentially reinitializing the whole component.

Thanks, Dan! Great catch on that. I will create a follow-up ticket for us to look into it!

jamigibbs:
@danbrady Here is the follow-up ticket about the "cancel" behavior. Feel free to update if you have any other details or clarifying information: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2511

danbrady (approved):

rsmithadhoc (approved):

jamigibbs:
> A couple reasons that this might be happening:

> We need to update to the latest release of USWDS (currently they are on [version 3.7.1](https://github.com/uswds/uswds/releases/tag/v3.7.1) and we are using 3.4.0).
The [rem-overide](https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/src/global/_functions.scss) function is not working as expected.

We have an issue created here for updating the USWDS package version: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2512

PR #1066, opened 2024-02-22
va-breadcrumbs: first breadcrumb label defaults to "VA.gov home"

jamigibbs:
## Chromatic

https://638-rename-breadcrumb-home--60f9b557105290003b387cd5.chromatic.com

## Description

This will add logic to `va-breadcrumbs` so that the first label will default to "VA.gov home" with the option to override that by setting a new prop `home-veterans-affairs` to `false`.

Figma reference: https://www.figma.com/file/afurtw4iqQe6y4gXfNfkkk/VADS-Component-Library?node-id=798%3A931&mode=dev
related issue https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/638

## QA Checklist

## Screenshots

Screenshot 2024-02-22 at 9 50 33 AM

## Acceptance criteria

## Definition of done


powellkerry (approved):

packages/web-components/src/components/va-breadcrumbs/va-breadcrumbs.tsx
```diff
@@ -40,6 +40,12 @@ export class VaBreadcrumbs {
    *  Represents a list of breadcrumbs. Use an array of objects with label and href properties, and then use JSON.stringify() to convert to a string. This prop is available when `uswds` is set to `true`.
    */
   @Prop() breadcrumbList?: any;
+
+  /**
+   *  When true, the first breadcrumb label will be "VA.gov home". 
+   */
+  @Prop() homeVeteransAffairs?: boolean = true;
```
jamigibbs, line 47: Naming is hard. I originally had this as `homeVA` but with the capitalization, the prop would need to be set like `home-v-a` on the vanilla web component which is ick.

Then I thought about just doing `homeVa` with the lowercase "a" but I didn't want it to look like we were not using the correct VA branding.

I landed on this `homeVeteransAffairs` because it never hurts to be more explicit about what this prop does and we can avoid the capitalization confusion.

powellkerry:
The code looks good, the only thing I noticed is that in storybook when you toggle the new prop it doesn't update the rendered demo component at the top of the page. I don't know if that is a feature or a bug :)

jamigibbs:
> The code looks good, the only thing I noticed is that in storybook when you toggle the new prop it doesn't update the rendered demo component at the top of the page. I don't know if that is a feature or a bug :)

@powellkerry Oh good catch! I didn't add that new prop to the Storybook pages but also I made all of the examples use "VA.gov home" in case teams copy/pasted the example. They'd already be off to a good start using the correct home label then.

But I wonder if we should add an example at least showing that it can be changed? 🤔  That might cause less confusion.

powellkerry (approved):

PR #1073, opened 2024-02-29
Ignore false UI diffs in  va-banner and va-loading-indicator

ataker:
## Chromatic

https://dst2547-add-chromatic-ignores--65a6e2ed2314f7b8f98609d8.chromatic.com

### va-maintenance-banner
Diffs in va-maintenance-banner were caused by the spinning loading icon. Solution was to add a `isChromatic` check and if that is true add a class to the icon that prevents the animation from playing.

### va-loading-indicator
Diffs in va-loading-indicator were caused by the the dates changing in the banner. Solution was to hard code the dates. The side effect of this is that the error maintenance banner is only supposed to show if the maintenance is currently active, so I added an `isError` prop to force the error banner in storybook and chromatic. Test added for `isError` prop.

Closes [Add an ignore or threshold to false UI diffs in va-banner and va-loading-indicator #2547](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2547)

## QA Checklist
- [ ] (N/A no accessibility changes made) Accessibility expert has signed off on code changes (if applicable. If not applicable provide reason why)
- [ ] (N/A no design changes made)  Designer has signed off on changes (if applicable. If not applicable provide reason why)
- [ ]  (N/A no accessibility changes made) Changes have been tested against screen readers (if applicable. If not applicable provide reason why)

## Screenshots

Screenshot 2024-03-07 at 17 07 46
Screenshot 2024-03-07 at 17 07 34

## Acceptance criteria

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-checkbox/va-checkbox.scss
+++ b/packages/web-components/src/components/va-checkbox/va-checkbox.scss
@@ -18,6 +18,7 @@
 .usa-checkbox {
   max-width: 320px;
   background: transparent;
+  padding-bottom: 1px;
 }
 
 .usa-checkbox__input:focus + [class*='__label']::before {
--- a/packages/web-components/src/components/va-checkbox/va-checkbox.scss
+++ b/packages/web-components/src/components/va-checkbox/va-checkbox.scss
@@ -18,7 +18,6 @@
 .usa-checkbox {
   max-width: 320px;
   background: transparent;
-  padding-bottom: 1px;
 }
 
 .usa-checkbox__input:focus + [class*='__label']::before {
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@department-of-veterans-affairs/component-library",
   "description": "VA.gov component library. Includes React and web components.",
-  "version": "37.0.0",
+  "version": "37.1.0",
   "license": "MIT",
   "scripts": {
     "build": "webpack"
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@department-of-veterans-affairs/component-library",
   "description": "VA.gov component library. Includes React and web components.",
-  "version": "37.1.0",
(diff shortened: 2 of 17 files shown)


ataker:
It wasn't until writing up the PR that I realized we need to manually update the maintenance banner date in the future. The banners only show if the end time is in the future so having a hard coded date will break this once the date passes. Not sure if we want to push the date further out, create a reminder to update it, or find another solution. 

Seems like we would need to set the `diffThreshold` too high for the test to be useful at all and we can't ignore elements in the shadow DOM so I don't have another solution currently.

Andrew565 (approved):

Andrew565:
I approved this, but I'd be more than happy to see the hardcoded date for the maintenance banner just pushed out to some absurd date like 2099 just so we don't have to worry about it in our life times.

jamigibbs:
> I approved this, but I'd be more than happy to see the hardcoded date for the maintenance banner just pushed out to some absurd date like 2099 just so we don't have to worry about it in our life times.

I plan to live until 150! 😂

Does something like this work?

```
const year = new Date().getFullYear() + 50;
const maintenanceTime = new Date(year, 5, 21);
```

packages/web-components/src/components/va-checkbox/va-checkbox.scss
```diff
@@ -18,6 +18,7 @@
 .usa-checkbox {
   max-width: 320px;
   background: transparent;
+  padding-bottom: 1px;
```
jamigibbs, line 21: Can you add this to the Storybook example instead? Adding it here means we're diverging from USWDS styles.

ataker: This was just a test, and then it passed after removing this, so we should just keep an eye on [this test](https://www.chromatic.com/test?appId=65a6e2ed2314f7b8f98609d8&id=65ea3bc591ce6dd1397ab13f) in the future. Updating the storybook makes a lot more sense.

PR #1075, opened 2024-03-06
va-radio-option: revert z-index and add Formation style overrides for USWDS

jamigibbs:
## Chromatic

https://2491-va-radio-input--65a6e2ed2314f7b8f98609d8.chromatic.com

## Description
This will resolve the issue of Formation input styles overriding the USWDS style for the va-radio-option `input` element. The previous solution that added a negative `z-index` caused issue with Cypress testing and it also doesn't follow the USWDS original styles.

USWDS reference: https://designsystem.digital.gov/components/radio-buttons/

related issue https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2491
related pr https://github.com/department-of-veterans-affairs/component-library/pull/1053

**before**

before

**after**

after

## QA Checklist

## Acceptance criteria

## Definition of done

### DIFF
--- a/packages/web-components/package.json
+++ b/packages/web-components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/web-components",
-  "version": "5.1.0",
+  "version": "5.1.1",
   "description": "Stencil Component Starter",
   "main": "dist/index.cjs.js",
   "module": "dist/index.js",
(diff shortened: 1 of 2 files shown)


Andrew565 (approved):

Mottie:
Cypress tests work if you add `forced: true` when clicking - it then ignores the -999em repositioning.

jamigibbs:
> Cypress tests work if you add `forced: true` when clicking - it then ignores the -999em repositioning.

@Mottie You're right! This is what we started doing but we didn't want to have to do this ad nauseam. More importantly though, we want to say as close to the USWDS styles as possible. The `z-index` diverges from that. When Formation is removed, the goal is to not have any overrides.

PR #1094, opened 2024-03-21
va-checkbox: Add part attribute to label and description

agravell047:
## Chromatic

https://Add-label-prop-to-va-checkbox--65a6e2ed2314f7b8f98609d8.chromatic.com

---
closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2606

Added label prop to allow bold label for va-checkbox 

## Screenshots
There is no UI change for this, a "label" prop was added to the component. We can then reference this prop in a CSS file to make a custom bold label.

### DIFF
--- a/packages/web-components/src/components/va-checkbox/va-checkbox.tsx
+++ b/packages/web-components/src/components/va-checkbox/va-checkbox.tsx
@@ -242,7 +242,7 @@ export class VaCheckbox {
             <label htmlFor="checkbox-element" id="option-label" class="usa-checkbox__label" part="label">
               {label}&nbsp;
               {required && <span class="usa-label--required">{i18next.t('required')}</span>}
-              {checkboxDescription && <span class="usa-checkbox__label-description" aria-describedby="option-label">{checkboxDescription}</span>}
+              {checkboxDescription && <span class="usa-checkbox__label-description" aria-describedby="option-label" part="description">{checkboxDescription}</span>}
             </label>
             {messageAriaDescribedby && (
               <span id='input-message' class="sr-only dd-privacy-hidden">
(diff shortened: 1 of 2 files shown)


packages/web-components/src/components/va-checkbox/va-checkbox.tsx
```diff
@@ -239,7 +239,7 @@ export class VaCheckbox {
               disabled={disabled}
               onChange={this.handleChange}
             />
-            <label htmlFor="checkbox-element" id="option-label" class="usa-checkbox__label">
+            <label htmlFor="checkbox-element" id="option-label" class="usa-checkbox__label" part="label">
               {label}&nbsp;
               {required && <span class="usa-label--required">{i18next.t('required')}</span>}
               {checkboxDescription && <span class="usa-checkbox__label-description" aria-describedby="option-label">{checkboxDescription}</span>}
```
jamigibbs, line 245: I think that you will also want to add `part` to the description element too. Otherwise all of the content wrapped by the label will be bold like this:

Screenshot 2024-03-21 at 12 19 29 PM

Once that's done, then there will be more flexibility with the weights:

```css
va-checkbox::part(label) {
  font-weight: bold;
}

va-checkbox::part(description) {
  font-weight: normal;
}
```

Screenshot 2024-03-21 at 12 21 30 PM

agravell047: @jamigibbs just updated!

jamigibbs (approved):
This looks good @agravell047 thank you again for your contribution! 

Before merging, can you update[ this package.json](https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/package.json#L3) file with a `patch` version update to 6.0.4?

PR #1115, opened 2024-04-11
va-table: add uswds borderless table variant

it-harrison:
## Chromatic

https://1860-borderless-table--65a6e2ed2314f7b8f98609d8.chromatic.com

---

## Description
This PR adds the uswds borderless table to the `<va-table>` component.

1. Note: the api supports the inclusion of links, including React Router links, but does not allow for the use of arbitrary elements or styles. Including styles (by means of classes) is not tenable due to the shadow dom. Arbitrary elements cannot be included in the table data through a straightforward process. I did not any instances in `vets-website` where these things would be needed so I opted for a simpler api for the first iteration.

Closes [1860](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1860)

## QA Checklist

## Screenshots
Screenshot 2024-04-11 at 3 12 40 PM

## Acceptance criteria

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-table/va-table.scss
+++ b/packages/web-components/src/components/va-table/va-table.scss
@@ -1,3 +1,4 @@
+@forward 'settings';
 @use 'usa-table/src/styles/usa-table';
 @import '~@department-of-veterans-affairs/css-library/dist/tokens/scss/variables.scss';
 
--- a/packages/web-components/src/components/va-table/va-table.scss
+++ b/packages/web-components/src/components/va-table/va-table.scss
@@ -1,0 +1,44 @@
+@use 'usa-table/src/styles/usa-table';
+@import '~@department-of-veterans-affairs/css-library/dist/tokens/scss/variables.scss';
+
+
+:host([uswds='true']) table, 
+:host([uswds='true']) caption {
+  font-size: 1.06 * $v3-font-base-size;
+}
+
+:host(:not([uswds='true'])) {
+  display: table;
+  font-family: var(--font-source-sans);
+  border-collapse: collapse;
+  margin: 0;
+  padding: 0;
+  width: 100%;
+}
+
+::slotted([slot='headers']) {
+  background: var(--vads-color-base-lightest);
+  font-weight: 700;
+}
+
+:host(:not([uswds='true'])) caption {
+  text-align: left;
+  padding: 0 0 0.5rem;
+  font-weight: 700;
+  font-size: 2rem;
+  font-family: var(--font-serif);
+  margin-bottom: 1.2rem;
+}
+
+@media screen and (max-width: 768px) {
+  :host(:not([uswds='true'])) thead {
+    display: none;
+  }
+
+  ::slotted(va-table-row) {
+    border-bottom: 1px solid var(--vads-color-gray-medium);
+  }
+  ::slotted(va-table-row):first-child {
+    border-top: 1px solid var(--vads-color-gray-medium);
+  }
+}
--- a/packages/web-components/src/components/va-table/va-table.tsx
+++ b/packages/web-components/src/components/va-table/va-table.tsx
@@ -243,7 +243,7 @@ export class VaTable {
     if (typeof cell === 'string') {
       content = cell;
     } else {
-      const { text, href, isRouterLink } = cell;
+      const { text, href, isRouterLink, testId } = cell;
       if (href) {
         if (isRouterLink) {
           content =
@@ -256,6 +256,9 @@ export class VaTable {
         }
       } else {
         content = text;
+      }
+      if (testId) {
+        content = <span data-testid={testId}>{content}</span>
       }
     }
     return content;
--- a/packages/web-components/src/components/va-table/va-table.tsx
+++ b/packages/web-components/src/components/va-table/va-table.tsx
@@ -1,4 +1,5 @@
-import { Component, Element, Host, Prop, State, h } from '@stencil/core';
+import { Component, Element, Host, Prop, State,  h } from '@stencil/core';
+import classnames from 'classnames';
 import { isNumeric } from '../../utils/utils';
 import ascendingIcon from '../../assets/sort-arrow-up.svg?format=text';
 import descendingIcon from '../../assets/sort-arrow-down.svg?format=text';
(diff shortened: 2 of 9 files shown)


powellkerry (approved):
Looks good to me.
I also learned from testing with voiceover that if a column only has numbers, voiceover will announce that column as the first column. Seems odd, but that is what happened when testing. When I added a character in the year column it announced the column order correctly.

jamigibbs:
The [v1 version has a few different Storybook stories](https://design.va.gov/storybook/?path=/docs/components-va-table--custom-components) that I don't see yet in the USWDS story. Is it possible to add those?

I'm also seeing the USWDS version showing up in the Chromatic link on the v1 Storybook page.

packages/web-components/src/components/va-table/va-table.tsx
```diff
@@ -41,110 +44,157 @@
...
+        );
+        button.classList.add(
+          'vads-u-padding--0',
+          'vads-u-margin--0',
+          'vads-u-text-decoration--none',
+          'vads-u-font-weight--bold',
+          'vads-u-background-color--gray-lightest',
+          'vads-u-color--base',
+        );
+        button.innerHTML = `${columns[this.sortColumn]}${icon}`;
```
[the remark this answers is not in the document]
jamigibbs: Is it possible to add this button element to the rendered markup instead of dynamically? It would be nice to avoid having to use `innerHTML` if we could..

If not possible, then we should probably sanitize the content being passed into it using `Sanitizer.escapeHTML` or similar to address this warning.

USWDS provided us with some utility functions that we have adopted for va-file-input that might be useful:

https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/src/utils/utils.ts#L128-L165

packages/web-components/src/components/va-table/va-table.scss
```diff
@@ -0,0 +1,44 @@
+@use 'usa-table/src/styles/usa-table';
```
jamigibbs, line 1: Do you need to also @forward the USWDS settings?

packages/web-components/src/components/va-table/va-table.scss
```diff
@@ -0,0 +1,44 @@
...
+@media screen and (max-width: 768px) {
+  :host(:not([uswds='true'])) thead {
+    display: none;
+  }
+
+  ::slotted(va-table-row) {
+    border-bottom: 1px solid var(--vads-color-gray-medium);
+  }
+  ::slotted(va-table-row):first-child {
+    border-top: 1px solid var(--vads-color-gray-medium);
```
jamigibbs, line 42: I am curious why we need to apply a different color here than what the USWDS `usa-table` package is providing?

packages/web-components/src/components/va-table/va-table.tsx
```diff
@@ -41,110 +44,157 @@ export class VaTable {
    */
   @Prop() descending?: boolean = false;
 
+  /**
+   * Whether this component will use the uswds styling.
+   */
+  @Prop() uswds?: boolean = false;
+
+  /**
+   * If uswds is true, shouldl the table be borderless or not.
```
jamigibbs, line 53: typo `shouldl`

packages/web-components/src/components/va-table/va-table.tsx
```diff
@@ -177,20 +229,89 @@ export class VaTable {
...
+            <tr>{header.map(cell => <th scope="col">{cell}</th>)}</tr>
+          </thead>
+          <tbody>
+            {rows.map(row => this.makeTableRow(row))}
+          </tbody>
+      </table>
+      )
+    } else {
+      return (
+        <Host role="table">
```
jamigibbs, line 304: Do we need to add `<Host role="table">` to the uswds variation too?

jamigibbs:
This is just one example but anywhere that the USWDS table style modules is using `rem` will need to be added to the [formation-override stylesheet](https://github.com/department-of-veterans-affairs/component-library/blob/228e9f47c4a15d1212a3ef45ec4487f324f7aa62/packages/web-components/src/global/_formation_overrides.scss). @powellkerry flagging you about this because we have the Formation PR that removes that still in flight.

Screenshot 2024-04-12 at 1 32 23 PM

jamigibbs:
I don't think the responsive view is displaying correctly (I just QA'd that really quick in the browser though). Do you mind checking that? This is what the USWDS site says it should be: https://designsystem.digital.gov/components/table/#responsive-stacked-table

Screenshot 2024-04-12 at 1 38 35 PM

danbrady:
Tagging @LWWright7 to help review this next week. Thanks!

packages/web-components/src/components/va-table/va-table.scss
it-harrison: @jamigibbs good catch - done!

packages/web-components/src/components/va-table/va-table.scss
it-harrison: @jamigibbs - I'm not sure about those colors; I didn't touch the colors for the v1 version - this file was touched only to import the uswds styles. But soon those v1 styles will be removed so it probably doesn't matte.

LWWright7:
@danbrady - LGTM!! I don't see any issues!

packages/web-components/src/components/va-table/va-table.tsx
it-harrison: @jamigibbs - fixed!

packages/web-components/src/components/va-table/va-table.tsx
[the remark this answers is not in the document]
it-harrison: @jamigibbs - the only affects the V1 component. It was flagged in this PR because I had to enclose the code in an `if` statement, so I think it is out of scope for this ticket?

packages/web-components/src/components/va-table/va-table.tsx
```diff
@@ -41,110 +44,157 @@ export class VaTable {
...
 
+  /**
+   * Whether this component will use the uswds styling.
+   */
+  @Prop() uswds?: boolean = false;
+
+  /**
+   * If uswds is true, should the table be borderless or not.
+   */
+  @Prop() borderless?: boolean = true;
```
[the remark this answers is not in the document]
it-harrison: @jamigibbs I agree - that would decrease number of props we need say the prop is type and can be one of `standard`, `borderless`, or `striped`

PR #1124, opened 2024-04-18
CSS Library: Migrate VA styles from Formation

jamigibbs:
## Chromatic

https://2537-formation-styles-to-css-library--65a6e2ed2314f7b8f98609d8.chromatic.com

### What is this PR doing?

- Moves necessary Formation files into CSS Library. Reference this [import mapping spreadsheet](https://docs.google.com/spreadsheets/d/10j_mjiEMASj0I5gzdZHIlmD2gxLtIHNF3Z_MMQZr7Ig/edit#gid=0).
  - **Many of these files and code blocks should be replaced by USWDS style imports or web components.**
- Updated anywhere that `node-sass` functionality was being used for `dart-sass` compatibility
- Updated tokens in those stylesheets to semantic tokens (if available)
- Added fonts and image assets
  - Some font files may still be missing but those files should just need to be dropped into the fonts folder.
- The build script is updated to generate assets
 
### What has not been added from Formation?

The CSS Library already offers utilities and tokens that we will want to use instead of porting over the old stuff:

- grid
- utilities
- breakpoints
- font awesome

### What's next after this PR is merged?

- Swapping Formation links for CSS Library so that we can migrate off of the Formation package.
- Replace styles with USWDS imports or web components and start deleting some of this stuff.
- Organization! This is just a mapping from Formation but as we remove a lot of these files with the USWDS v3 migration, we should re-organize the CSS Library

closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2537

## Acceptance criteria

## Definition of done


packages/css-library/src/assets/fonts/bitter-bold.ttf
jamigibbs, line 1: I've added the Bitter font files (from Formation) here which will copy over into `dist/fonts`.

I have also loaded USWDS v3 `@font-face` font imports which loads the font face for and references the font files in the USWDS package for these:

  - Roboto Mono Web
  - Source Sans Pro Web
  - Merriweather Web

https://github.com/department-of-veterans-affairs/component-library/blob/9b902d53335676bd8ed2a96f94b8640d131e3aa6/packages/css-library/src/stylesheets/base/_fonts.scss#L7

**The problem though** is that they have added "Web" to all of these font names so at some point we will need to update wherever there is a `font-family` assignment for "Source Sans Pro" to "Source Sans Pro Web".

packages/css-library/src/assets/img/arrow-both.png
jamigibbs, line 1: I have added any images from Formation that were being referenced in the stylesheets that were moved over. But I would anticipate that we wouldn't actually need any images in the CSS Library like this eventually. All necessary images would instead be coming from component-library or USWDS directly.

packages/css-library/src/stylesheets/_mixins.scss
```diff
@@ -0,0 +1,434 @@
+//============ PORTED MIXINS FROM FORMATION ==============//
```
jamigibbs, line 1: There is zero organization here because I don't expect any of these to survive the Formation culling. I just moved over whatever mixins were necessary to keep the ported stylesheets working.

packages/css-library/src/stylesheets/_override-function.scss
```diff
@@ -0,0 +1,159 @@
+@use 'sass:math';
```
jamigibbs, line 1: I had to make some modifications to this file in order for it to work with `dart-sass` but I expect that this will also eventually go away as we remove stylesheets.

packages/css-library/src/stylesheets/_shame.scss
```diff
@@ -0,0 +1,189 @@
+// This file is a place where you put all the code you're not proud of,
```
jamigibbs, line 1: Screenshot 2024-04-23 at 2 49 51 PM

packages/css-library/src/stylesheets/base/_fonts.scss
```diff
@@ -0,0 +1,36 @@
+/*
+The uswds-fonts package loads @font-face for:
+  - Roboto Mono Web
+  - Source Sans Pro Web
+  - Merriweather Web
+*/
+@use "uswds-fonts";
```
jamigibbs, line 7: This will add `@font-face` imports for those fonts listed above from the USWDS v3 package. They all include the name "Web" though which is different from v1.

packages/css-library/src/stylesheets/base/_va.scss
```diff
@@ -0,0 +1,374 @@
+@import '../override-function';
+
+@import 'uswds-core/src/styles/mixins/typography/usa-content-styles';
```
jamigibbs, line 3: This stylesheet is essentially [the same as `elements.scss`](https://github.com/department-of-veterans-affairs/component-library/blob/9b902d53335676bd8ed2a96f94b8640d131e3aa6/packages/css-library/src/stylesheets/elements.scss) but I could not move that file because it's being used in the component library and has `rem` values.

This file is the "16px root font" version and `elements.scss` is the "10px root font" version.

packages/css-library/src/stylesheets/fonts.scss
jamigibbs, line 1: Moved this fonts stylesheet to `/stylesheets/base/`

packages/css-library/src/stylesheets/formation-overrides/_variables.scss
```diff
@@ -0,0 +1,113 @@
...
+@import '../override-function';
+
+$font-path: "./fonts";  // TODO: Verify asset path
+$image-path: "./img"; // TODO: Verify asset path
+
+$label-border-radius: 2px;
+$h5-font-size: scale-rem(1.5rem);
+
+$color-black-light: #212121;
+$color-base: $color-black-light;
```
jamigibbs, line 15: These are all "imposter tokens". They are only here so that the ported stylesheets continue to work. These should all go away as we remove things.

packages/css-library/src/stylesheets/_core.scss
```diff
@@ -0,0 +1,48 @@
+@use "../../dist/tokens/scss/variables" as *;
+
+@use 'uswds-core' with (
+  $theme-show-notifications: false,
+  // TODO: verify USWDS font path
+  $theme-font-path: '../node_modules/@uswds/uswds/dist/fonts',
+);
+
+$font-path: "./fonts";  // TODO: Verify asset path
+$image-path: "./img"; // TODO: Verify asset path
```
jamigibbs, line 10: These asset paths point to where the assets will be in `dist/` but when we start connecting stylesheets, this will need to be verified that it's correct.

packages/css-library/src/stylesheets/lib/_normalize.scss
```diff
@@ -0,0 +1,13 @@
...
+    @forward "uswds-elements/lib/normalize";
+    @forward "uswds-core";
+    @forward "uswds-fonts";
+    @forward "uswds-elements";
+    @forward "uswds-helpers";
+
+  See: https://github.com/uswds/uswds/blob/98566ec5dc12807a2b0bc36699a033fe3a711d44/packages/uswds-global/_index.scss
+*/
+
+@use "uswds-elements/lib/normalize";
```
jamigibbs, line 13: I thought we could just start using the USWDS normalize import instead of housing it ourself. 

In Formation, it was imported from USWDS v1 which was using v3.0.3. USWDS v3 is using [v8.0.1](https://github.com/uswds/uswds/blob/98566ec5dc12807a2b0bc36699a033fe3a711d44/packages/uswds-elements/lib/_normalize.scss#L4)

Here is a diff of those two versions:

https://www.diffchecker.com/XH2TFwMU/

Andrew565: I'm in favor of moving to V3's normalize.

Andrew565 (approved):

micahchiang (commented):
I think this looks good for getting us to a spot where we can cut over from Formation to CSS-Library and begin swapping imports. Exciting! 

We'll have to do some digging around the uswds repo as a group, but my hope is that we'll be able to replace a lot of this stuff with direct imports from `uswds-core`

packages/css-library/src/stylesheets/lib/_normalize.scss
micahchiang: I agree with Andrew. Where impact is minimal, we should try and get onto V3 stuff as we go.

PR #1127, opened 2024-04-21
Fix va-icon svg path to work in prod and dev environs, take 2

rmessina1010:
This PR attempts to reconcile the asset path structure of va-icon with vets-website and other va-components
 
### Actions
* Removes Stencil/core getAssetPath(). _Note: this had no functional effect on the src string, we could revert this_
* Changes the path directory for `sprite.svg` from `/assets/sprite.svg`to `/img/sprite.svg`
* Creates a new directory, `img`, in `web-components` and a copy of` sprite.svg`.  To Do: Test if this is necessary, or if we can rely on webpack copy alone
* Adjusted copy plugin in webpack
* Adjusted copy plugin in Stencil config
* Creates a new directory, `img`, in `storybook` and a copy of` sprite.svg`-- storybook would not work without this

* Check vets-website `img` build
* May need to update/upload to proper s3 bucket; still unclear as to how that would go

## Chromatic

https://rm-2680-rm-icon-fix-v2--65a6e2ed2314f7b8f98609d8.chromatic.com


rmessina1010:
The icons in va-icon components confirmed to work, in dev and staging, see attached. 
The current working solution was based on the following:

- the image resource paths for our web components must be set to` ../img` ( or `/img`, which am currently testing)
- as such I added an  `~/img` directory to `web-components`. and adjusted webpack to in core to copy the `~/img` directory into `dist/img`  
- I added `~/img` to storybook
- I did a similar adjustment to `vets-website`, to allow for local builds, and copied sprite.svg` into `src/site/assets/img/`
- finally, I also needed to add `sprite.svg` into `content-build ` at `src/site/assets/img/`; worthy of note content build does not use `component-library` as a dependency, they seem to have `web-components` only

After closing, I will create a new issues:
1. Adjust resource paths
2. Research how to manage and synchronize `sprite.svg` across repos


Screenshot 2024-04-24 at 1 28 00 AM
Screenshot 2024-04-24 at 1 27 33 AM

Andrew565 (approved):

Andrew565 (approved):

PR #1128, opened 2024-04-21
va-table: adding USWDS borderless variation

it-harrison:
## Chromatic

https://1860-borderless-ii--65a6e2ed2314f7b8f98609d8.chromatic.com

---

## Description
This PR adds a `bordeless` table uswds variant to the `<va-table>` web-component. Due to limitations with slots and tables in HTML and the need to maintain V1 functionality (including the inclusion of arbitrary markup in table cells), a new approach was needed. 

The API for `va-table` has not changed but under the hood `va-table` is now a wrapper component and the new `va-table-inner` component is the component that handles the actual rendering.

In the case of the V1 component the `va-table` wrapper simply forwards slot content to `va-table-inner`. 

In the case of V3, `va-table` extracts the `span` elements from within the `va-table-row` components, adds a slot name, and then forwards them to `va-table-inner`, which generates a corresponding slot for each cell, thereby allowing for arbitrary markup (like `va-link` or `va-telephone` components) to be rendered inside the `td` and `th` elements in the table that `va-table-inner` renders.

Closes [1860](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1860)

## QA Checklist

## Acceptance criteria

## Definition of done


it-harrison:
> Should we make this follow the same standard as the other components and assume V3 styling as the default?

@Andrew565 I was wondering that but left it V1 default just because right now table seems to be an exception to the rule

packages/web-components/src/components/va-table/va-table.tsx
```diff
@@ -1,37 +1,47 @@
...
 @Component({
   tag: 'va-table',
-  styleUrl: 'va-table.css',
+  styleUrl: 'va-table.scss',
   shadow: true,
 })
-export class VaTable {
-  headers: HTMLElement[] = null;
-
+export class VaTableWrapper {
```
jamigibbs, line 21: The class name is `VaTableWrapper` in `va-table.tsx` and in `va-table-inner.txs` the class name is `VaTable`. Is that right?

packages/web-components/src/components/va-table/va-table.tsx
```diff
@@ -40,157 +50,168 @@ export class VaTable {
...
+    this.cells = cells;
+  }
 
-    handleCellAdjustments();
+  /**
+   * Generate a DocumentFragment that holds the cells 
+   * to be slotted into a va-table-inner component
+   */
+  makeFragment(): DocumentFragment {
+    const frag = document.createDocumentFragment();
```
jamigibbs, line 121: This is really interesting. Is the purpose of using `DocumentFragment` instead of direct dom manipulation because of cost savings? (speed) Or is there a specific web component reason to use it?

packages/web-components/src/index.html
```diff
@@ -18,6 +18,18 @@
 
   </head>
   <body>
-    <my-component first="Stencil" last="'Don't call me a framework' JS"></my-component>
+    <va-table uswds table-title="a cool table">
```
jamigibbs, line 21: Was this just for testing?

packages/web-components/src/components/va-table/va-table-inner/test/va-table.e2e.ts
jamigibbs, line 1: Did this test need to be moved under the `va-table-inner` folder? Or can it be put back under `va-table`?  Should there be separate test files for the wrapper component and the original component?

packages/web-components/src/components/va-table/va-table-inner/va-table-inner.scss
```diff
@@ -1,3 +1,22 @@
+@forward 'settings';
+@use 'usa-table/src/styles/usa-table';
+@import '~@department-of-veterans-affairs/css-library/dist/tokens/scss/variables.scss';
+
+
+:host([uswds='true']) table, 
+:host([uswds='true']) caption {
+  font-size: 1.06 * $v3-font-base-size;
```
jamigibbs, line 8: Is this font size adjustment to avoid having to update the `formation-overides.scss` file?

packages/web-components/src/components/va-table/va-table-inner/va-table-inner.scss
```diff
@@ -1,3 +1,22 @@
...
+}
+
+:host([uswds='true']){
+  td slot::slotted(span:empty)::before,
+  th slot::slotted(span:empty)::before {
+   content: "---";
```
jamigibbs, line 14: Nice! 👍

jamigibbs (commented):
This is a very creative solution! I have just a few comments/questions but thank you for working so hard on this one.

packages/web-components/src/components.d.ts
```diff
@@ -1194,10 +1194,47 @@ export namespace Components {
...
         "sortColumn"?: number;
+        /**
+          * The title of the table
+         */
+        "tableTitle"?: string;
+        /**
+          * If uswds is true, the type of table
+         */
+        "tableType"?: | 'borderless'
+    | 'striped';
```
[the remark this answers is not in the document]
it-harrison: @jamigibbs - I was wondering that too - I will comment it out for now

packages/web-components/src/components/va-table/va-table.tsx
it-harrison: @jamigibbs - good catch - fixed!

packages/web-components/src/components/va-table/va-table.tsx
it-harrison: @jamigibbs - I didn't run any tests as far as speed but it seemed the easiest way to add the cells and from the research I did seemed the preferred way (but I could be wrong).

packages/web-components/src/index.html
it-harrison: @jamigibbs - it slipped through - I always try to check that I don't commit that file!

rsmithadhoc (changes requested):
This looks good in my testing. Everything is read out as expected in VoiceOver, JAWS, and NVDA.

Axe scan brings up a code hierarchy issue, but I think the fix is simple. Inside `va-table-row.tsx`, there is a leftover `role="row"` from v1 that I believe can be removed. In v1, the table had this structure which resolved any Axe hierarchy complaints:
```
<va-table role="table">
  <va-table-row role="row">
  ...
```
But it won't have that with v3 (there is no `role="table"` on v3, which is fine), so the `role="row"` is may no longer be relevant. Unless you feel otherwise, I think removing that is the way to go and it will resolve any Axe issues. I re-tested in all screen readers with that change and it still read everything out as expected.

packages/web-components/src/components/va-table/va-table.tsx
```diff
@@ -1,37 +1,48 @@
...
-  @Prop() tableTitle: string;
+  @Prop() tableTitle?: string;
 
   /**
+   * UNCOMMENT when we add more variations
+   * If uswds is true, the type of table
+   */
+  // @Prop() tableType?:
+  //   | 'borderless'
+  //   | 'striped' = 'borderless';
```
jamigibbs, line 40: Instead of commenting out code blocks, could you just remove `striped` and have  `borderless` as the only option on the prop for now? Or does Stencil complain?

```
@Prop() tableType?:  | 'borderless';
```

jamigibbs:
@it-harrison I think this was ok before so maybe just something minor changed in the last round of updates but the [v1 version](https://1860-borderless-ii--65a6e2ed2314f7b8f98609d8.chromatic.com/?path=/docs/components-va-table--default) doesn't look like it's showing the table headers anymore. Can you check that out?

Screenshot 2024-04-24 at 9 57 59 AM

Screenshot 2024-04-24 at 9 58 03 AM

packages/web-components/src/components/va-table/va-table.tsx
it-harrison: @jamigibbs - no problem, done!

packages/web-components/src/components/va-table/va-table-inner/va-table-inner.scss
it-harrison: @jamigibbs - I should have added to the override file - fixed.

packages/storybook/stories/va-table-uswds.stories.jsx
```diff
@@ -0,0 +1,328 @@
...
+            {row.map((item, index2) => (
+              <span key={`cell-default-${index2}`}>{item}</span>
+            ))}
+          </va-table-row>
+        ))}
+      
+      <va-table-row>
+        <span><va-link text="Social Security" href="https://www.ssa.gov/history/35act.html"/></span>
+        <span>
+          <div className='vads-u-background-color--gray-lightest vads-u-padding--2'>
```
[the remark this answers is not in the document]
it-harrison: good call - done!

it-harrison:
> @it-harrison I think this was ok before so maybe just something minor changed in the last round of updates but the [v1 version](https://1860-borderless-ii--65a6e2ed2314f7b8f98609d8.chromatic.com/?path=/docs/components-va-table--default) doesn't look like it's showing the table headers anymore. Can you check that out?
> 
> Screenshot 2024-04-24 at 9 57 59 AM
> 
> Screenshot 2024-04-24 at 9 58 03 AM

@jamigibbs - fixed!

jamigibbs (approved):
I think this looks good! We just need Ryan to verify the a11y issue he mentioned is resolved.

@it-harrison Can you also spin up a [separate ticket for adding the responsive stacking](https://designsystem.digital.gov/components/table/#responsive-stacked-table)? We can do that as a follow-up.

packages/web-components/src/components/va-table/va-table-inner/test/va-table.e2e.ts
it-harrison: @jamigibbs I renamed the test spec inside `va-table-inner` and added another spec at the `va-table` level

it-harrison:
> I think this looks good! We just need Ryan to verify the a11y issue he mentioned is resolved.
> 
> @it-harrison Can you also spin up a [separate ticket for adding the responsive stacking](https://designsystem.digital.gov/components/table/#responsive-stacked-table)? We can do that as a follow-up.

@jamigibbs sounds good. 
@rsmithadhoc can you verify the a11y issue re the "role" attr on the va-table-row?

rsmithadhoc (approved):
Looks good, thanks!

PR #1142, opened 2024-05-06
2642 file input update

nickjg231:
## Chromatic

https://2642-file-input-update--65a6e2ed2314f7b8f98609d8.chromatic.com

## Description
Updates the va-file-input V3 component with new designs and features. These include:

 

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2642

## QA Checklist

## Screenshots
Base component:
image

File uploaded:
image

Error message:
image

Different size header:
image

Additional Info:
image

## Acceptance criteria

## Definition of done


packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -26,7 +26,11 @@ import { fileInput } from './va-file-input-upgrader';
   shadow: true,
 })
 export class VaFileInput {
+  private fileInputRef!: HTMLInputElement;
+
   @Element() el: HTMLElement;
+  @State() file?: File;
+  @State() uploadStatus: 'idle' | 'uploading' | 'success' | 'failure' = 'idle';
```
jamigibbs, line 33: Does this need the `@State()` decorator? I only ask because it will cause a re-render whenever this object value is updated so I just want to make sure that it's necessary. Does it still work if you remove that decorator?

https://stenciljs.com/docs/state#when-to-use-state

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -114,6 +127,29 @@ export class VaFileInput {
     this.el.shadowRoot.getElementById('fileInputField').click();
   };
 
+  private removeFile = () => {
+    this.file = undefined;
```
jamigibbs, line 131: Would it be better to assign this as `null` instead of `undefined`?

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -122,6 +158,44 @@ export class VaFileInput {
     return this.buttonText ? this.buttonText : 'Upload file';
   };
 
+  private formatFileSize = (filesSize): string => {
```
jamigibbs, line 161: I'm not sure I understand what's happening with this function or why it's needed. Do you mind adding clarifying it here in the PR or as a comment in the file?

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -139,60 +213,82 @@ export class VaFileInput {
...
+            <div class="input-hint-message">{hint}</div>
           )}
-          <slot></slot>
-          <span id="input-error-message" role="alert">
+          <div id="input-error-message" role="alert">
             {error && (
               <Fragment>
-                <span class="usa-sr-only">{i18next.t('error')}</span>
-                <span class="usa-error-message">{error}</span>
+                <span class="sr-only">{i18next.t('error')}</span>
```
jamigibbs, line 236: I think that we can still use the USWDS module for `usa-sr-only`. That is not specific to file input.

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -139,60 +213,82 @@ export class VaFileInput {
...
+                <span>{error}</span>
```
jamigibbs, line 237: `usa-error-message` also is not specific to file input. Do the VA designs diverge from the USWDS error styles? If not, can we continue to use the module that contains that class?

LWWright7:
@nickjg231 - just a few minor things..... there is a black border line missing around the blue background that appears in the design.  The only other thing I'm seeing is the difference in the icon.  For the design, I'm not sure where @danbrady got that icon, as I'm not seeing it in our icon library in Figma under Font Awesome or USWDS icons, so it may be a custom .svg icon.

jamigibbs:
Are there examples in Storybook to preview for the **upload indicator** or **additional info slot** like shown in the [ticket's screenshots](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2642)? I wasn't sure how to reproduce that to test it.

Screenshot 2024-05-08 at 1 46 00 PM

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -26,7 +26,11 @@ import { fileInput } from './va-file-input-upgrader';
   shadow: true,
 })
 export class VaFileInput {
+  private fileInputRef!: HTMLInputElement;
+
   @Element() el: HTMLElement;
+  @State() file?: File;
```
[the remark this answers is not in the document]
nickjg231: Good catch, this can be removed now. I initially had this to help with the next task of allowing multiple files within this component (just change to a `File[]`), however a new component for the multi-variant is being created. We don't need to manage this in here anymore because of that.

packages/web-components/src/components/va-file-input/va-file-input.tsx
nickjg231: This is a good question. I do believe it's still useful to have this as a state variable based on that link, as a lot of the render logic is conditional on the value of `uploadStatus`:
image

packages/web-components/src/components/va-file-input/va-file-input.tsx
nickjg231: Adding the following comment:
`/**
   * Converts the size of a file from bytes to a more human-readable format for
   * rendering the file size label. This function calculates the file size in
   * appropriate units (B, KB, MB, GB, TB) based on the size provided. It uses
   * logarithmic scaling to determine the unit, then formats the size to one
   * decimal place for units KB and above.
   *
   * @param {number} filesSize - The size of the file in bytes.
   * @returns {string} The file size formatted as a string with the appropriate unit.
   */
`

LWWright7 (approved):

packages/web-components/src/components/va-file-input/va-file-input.tsx
nickjg231: I am using the USWDS styles for hint and error again

nickjg231:
> Are there examples in Storybook to preview for the **upload indicator** or **additional info slot** like shown in the [ticket's screenshots](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2642)? I wasn't sure how to reproduce that to test it.
> 
> Screenshot 2024-05-08 at 1 46 00 PM

In the engineering channel the other day, we came to the consensus that we actually should not use the progress bar in this component. The upload in our component doesn't upload when the user selects a file, rather when the user submits the form. 

When the user selects a file on the page, the input acts as a pointer to the user's file system and has a little bit of metadata. If we wanted to show the progress bar here, it wouldn't really be in an uploading state long enough to even see the progress bar. Then the user generally moves on in the form and the actual file upload happens at the end. This may mean they don't even see the progress bar upload. In addition, within the va-file-input, we have no knowledge of the form we're in or its events, like if it has been submitted. We would need to add extra props to manage the state, then work with the backend to actually have some sort of value returned from the server. At this point, it seems like it won't behave in a way that's intuitive to the user and would be confusing to handle. We can have a further discussion in the engineers channel if you'd like!

jamigibbs:
A couple other UX things:

1. This icon doesn't look correct compared to [Figma](https://www.figma.com/design/UZys4hvw2Vwuo7U0pwbKbq/File-Input-v3?node-id=71-25203&t=HmWqGnAphmWQmDVE-0):

Screenshot 2024-05-08 at 2 46 46 PM


2. The padding in the header and hint text is not matching Figma

**Figma**
Screenshot 2024-05-08 at 2 49 56 PM

**Storybook**
Screenshot 2024-05-08 at 2 49 52 PM

3. I think this text color for the file size might be too dark?

**Figma**

Screenshot 2024-05-08 at 2 52 18 PM


**Storybook**

Screenshot 2024-05-08 at 2 52 12 PM

jamigibbs:
> In the engineering channel the other day, we came to the consensus that we actually should not use the progress bar in this component. 

@nickjg231 Do we need to update the designs then? Or is it possible then for teams to add the progress component into the va-file-input slot in order to match the designs?

packages/web-components/src/components/va-file-input/va-file-input.tsx
jamigibbs: Hi @nickjg231 Please import the USWDS module for `usa-sr-only`. The `sr-only` class will be going away with Formation. 

Here is an example: 
https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/src/components/va-checkbox/va-checkbox.scss#L3

nickjg231:
> A couple other UX things:
> 
> 1. This icon doesn't look correct compared to [Figma](https://www.figma.com/design/UZys4hvw2Vwuo7U0pwbKbq/File-Input-v3?node-id=71-25203&t=HmWqGnAphmWQmDVE-0):
> 
> Screenshot 2024-05-08 at 2 46 46 PM
> 
> 2. The padding in the header and hint text is not matching Figma
> 
> **Figma** Screenshot 2024-05-08 at 2 49 56 PM
> 
> **Storybook** Screenshot 2024-05-08 at 2 49 52 PM
> 
> 3. I think this text color for the file size might be too dark?
> 
> **Figma**
> 
> Screenshot 2024-05-08 at 2 52 18 PM
> 
> **Storybook**
> 
> Screenshot 2024-05-08 at 2 52 12 PM

@jamigibbs 

1. I had Lucas look into the icon. He says it's a custom file that Dan used on Figma, and isn't something that exists in our project. The one I chose is the closest icon we have to the one in the mockups, and Lucas seems okay with that.
2. This padding above the `va-card` is being updated now
3. This color difference was a typo, this is fixed now

packages/web-components/src/components/va-file-input/va-file-input.tsx
jamigibbs: Thanks for digging into it! I would lean towards removing it if `uploadStatus` still works for all of the different rendered logic. Any chance that we can get to make the component faster (avoiding unnecessary renders) is a good thing. But if the logic needs the re-render to function, it would make sense to keep it.

nickjg231:
> > In the engineering channel the other day, we came to the consensus that we actually should not use the progress bar in this component.
> 
> @nickjg231 Do we need to update the designs then? Or is it possible then for teams to add the progress component into the va-file-input slot in order to match the designs?

@jamigibbs This may require some more discussion with the team, as I didn't think about allowing users to do this with the slot. The only thing I worry is that teams may use slightly different styles across forms so the progress bar could be slightly inconsistent, but otherwise it would take the burden off of us. 

For now, I think this should be merged without references to the progress bar in the code, then we can discuss with designers. @LWWright7 what do you think?

jamigibbs:
> I had Lucas look into the icon. He says it's a custom file that Dan used on Figma, and isn't something that exists in our project. The one I chose is the closest icon we have to the one in the mockups, and Lucas seems okay with that.

@LWWright7 Should we create a follow-up ticket to update the icon then? cc @danbrady

micahchiang (changes requested):
@nickjg231 @jamigibbs - I haven't finished my review yet, but there are couple things out there that I think need to be looked at before merging, so I'm requesting changes until those things are resolved. I'll follow up with a longer comment here in a second.

danbrady:
After a file is attached, the "icon" is actually a preview of that file (like a thumbnail). This functionality exists is in the USWDS File Input as well as the current v3 version that Andrew worked one. I'm not sure of the exact logic behind it, but my understanding if the file *can* have a preview (because it is an image of some type) it will. However, it shows alternate previews if the attachment is a different "non-previewable" file type.

micahchiang:
@nickjg231 @jamigibbs - I haven't finished my review yet, but I'd like to chime in on some of the thoughts already out there.

#### Regarding the loading indicator
> In the engineering channel the other day, we came to the consensus that we actually should not use the progress bar in this component. The upload in our component doesn't upload when the user selects a file, rather when the user submits the form.

Maybe this was part of the discussion after I had to leave the mob session on Monday, so I'm not sure if I'm missing some additional context here. The loading indicator is definitely part of the [file input designs](https://www.figma.com/design/UZys4hvw2Vwuo7U0pwbKbq/File-Input-v3?node-id=71-25203&t=Br4VTfWcxHdhFtYp-0) in Figma. It's also functionality that is current on staging. The following screen cap is from the [10-10EZ Form](https://staging.va.gov/health-care/apply-for-health-care-form-10-10ez/military-service/documents): 

[Screencast from 05-08-2024 01:36:28 PM.webm](https://github.com/department-of-veterans-affairs/component-library/assets/15097156/cb4c371c-e3b1-4d74-bb2d-bef512fa9b30)

Because it already exists like this in staging, regardless of how brief it renders, we'll want to add it here. 

#### Regarding the file icon
I'm fine with @LWWright7's sign off here since he's a designer, but I'd like to make sure things are consistent between what's in here and what exists in Figma. So, if we're going with the icon shown in the screenshots here, then I'd appreciate it if the designs in Figma were updated to avoid confusion. If it turns out for some reason that the custom icon @danbrady used was intentional, then we need to add it to `va-icon` and update the icon used in this component. I'm fine with whatever.

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -114,6 +127,29 @@ export class VaFileInput {
...
+    event.stopPropagation();
+
+    const files = event.dataTransfer.files;
+    if (files.length > 0) {
+      this.file = files[0];
+      this.vaChange.emit({ files: [this.file] });
+      this.uploadStatus = 'success';
+    }
+  }
+
```
micahchiang, line 152: I'm not sure if our web component has always been this way so forgive me if I'm missing something, but with file upload, I'd expect to see files stored in a `FileList` on the input. For some reason it's returning an empty list after upload: 

image

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -139,60 +223,79 @@ export class VaFileInput {
...
+            {uploadStatus !== 'idle' && (
+              <div class="selected-files-wrapper">
+                <div class="selected-files-label">Selected files</div>
+                <va-card class="va-card">
+                  <div class="file-info-section">
+                    <va-icon icon="file_present" size={5} class="file-icon"></va-icon>
+                    <span class="file-label">{file.name}</span>
+                    <span class="file-size-label">{this.formatFileSize(file.size)}</span>
+                  </div>
+                  {(uploadStatus === 'success' || uploadStatus === 'failure') && file && (
```
micahchiang, line 284: Do we need the `uploadStatus` check here? On first pass it seems like we could just check `file` and that would suffice.

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -122,6 +158,54 @@ export class VaFileInput {
...
+   * rendering the file size label. This function calculates the file size in
+   * appropriate units (B, KB, MB, GB, TB) based on the size provided. It uses
+   * logarithmic scaling to determine the unit, then formats the size to one
+   * decimal place for units KB and above.
+   *
+   * @param {number} filesSize - The size of the file in bytes.
+   * @returns {string} The file size formatted as a string with the appropriate unit.
+   */
+  private formatFileSize = (filesSize): string => {
+    const units = ['B', 'KB', 'MB', 'GB', 'TB'];
```
micahchiang, line 172: I really hope no one tries to upload a file that is x GBs or TBs large lol

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -139,60 +223,79 @@ export class VaFileInput {
...
-            aria-describedby={ariaDescribedbyIds}
-            onChange={this.handleChange}
-          />
+            {uploadStatus !== 'idle' && (
+              <div class="selected-files-wrapper">
+                <div class="selected-files-label">Selected files</div>
+                <va-card class="va-card">
+                  <div class="file-info-section">
+                    <va-icon icon="file_present" size={5} class="file-icon"></va-icon>
+                    <span class="file-label">{file.name}</span>
```
micahchiang, line 281: @danbrady @LWWright7 - We need to decide on how to approach name truncation for files with long names. Would appreciate some design input here, and potentially an update to the figma specs to reflect name truncation. 

Currently long names extend past the card: 
Screenshot from 2024-05-08 13-15-52

micahchiang (changes requested):
Requesting changes for the following reasons: 
- we should use our design tokens where we can. I left comments where they can be used.
- we should be supporting loading indicator since it's part of the Figma designs unless @danbrady and @LWWright7 say otherwise.
- I think we're meant to be displaying preview thumbnails instead of static icons
- we need to decide how to truncate long file names. i.e. if a file name is longer than 25 characters should we display a truncated version of it?

packages/web-components/src/components/va-file-input/va-file-input.tsx
harshil1793: Indeed, that would be quite an edge case! However, it might be wise to consider implementing a limit on file size or handling such scenarios gracefully to prevent potential issues. What are your thoughts?

packages/web-components/src/components/va-file-input/va-file-input.tsx
danbrady: Ok, I think we can handle long file names by splitting the file name and the uploading label / file size to be on different lines. In other words, the file name will be on top line and then either the "Uploading..." label or the file weight on the next line. If the file name is long, it should wrap, and not truncate it, so it does not break outside the boundaries of the card. (This stays inline with what USWDS does.) Some designs below for reference:

### **Previous design**: 
(This didn't factor long file names)
image



---

### Updated design:
image

---

### Updated design with long name and in uploading state:
image

---

### Updated design with long name attached:
image

---

For Reference:
**USWDS with Long File Name**:
Screenshot 2024-05-13 at 1 56 32 PM


**NOTE**: The default body text has a line-height that's too much for this component. So instead I used [line-height-2](https://design.va.gov/foundation/utilities/line-height) (1.15) for the file name and size text, which renders 16.96px text to have 19.504 line-height. 

I've updated the design files with this. Hit my up with any questions!

micahchiang (approved):
Changes look good. Thanks for your work on this @nickjg231 and @danbrady

PR #1154, opened 2024-05-20
storybook 7 upgrade

it-harrison:
## Chromatic

https://storybook-7-upgrade--65a6e2ed2314f7b8f98609d8.chromatic.com

---

## Description
This PR upgrades component-library's Storybook from 6 to 7. **Note** there are some styling issues which this PR does not address but which are slated for work in [this ticket](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2923).

Closes [2634](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2634)

## Acceptance criteria

## Definition of done


packages/web-components/stencil.config.ts
```diff
@@ -35,7 +35,7 @@ export const config: Config = {
   },
   outputTargets: [
     reactOutputTarget({
-      componentCorePackage: '@department-of-veterans-affairs/web-components/dist/types',
+      componentCorePackage: '../dist/types',
```
jamigibbs, line 38: This should probably be reverted. 

Is there anything else that we were throwing against the wall when trying to get this to work that can or should be reverted too?

.github/workflows/node.js.yml
```diff
@@ -13,15 +13,45 @@ jobs:
...
+
+      - name: Install deps in packages
+        run: yarn workspaces foreach --all -v install
+
+      # - name: Test packages
+      #   run: yarn workspaces foreach --all -v run test
+
+      # - name: Install dependencies in react-components
+      #   run: yarn install
+      #   working-directory: packages/react-components
```
jamigibbs, line 37: Do we still need or want these steps combined? If so, can any of this be cleaned up?

packages/storybook/.storybook/style.scss
```diff
@@ -5,6 +5,13 @@ $formation-asset-path: '~@department-of-veterans-affairs/formation/assets';
 @import '@department-of-veterans-affairs/formation/sass/base/_b-breakpoints.scss';
 @import '@department-of-veterans-affairs/formation/sass/site/_m-crisis-line.scss';
 
+@import '@department-of-veterans-affairs/formation/dist/formation.min.css';
+
+// @import '@department-of-veterans-affairs/formation/dist/formation.min.css';
+// @import '@department-of-veterans-affairs/formation/dist/formation';
+// @import '@department-of-veterans-affairs/component-library/dist/main.css';
```
jamigibbs, line 12: Do we need any of these imports still? I'm not sure why th component-library `main.css` needed to be commented out specifically.

packages/storybook/.storybook/style.scss
```diff
@@ -50,4 +57,8 @@ a.sbdocs-a:hover {
 
 button {
   padding: 10px 20px !important;
+}
+
+html {
+  color: orangered;
```
jamigibbs, line 63: Was this just for testing?

packages/storybook/package.json
```diff
@@ -13,22 +14,26 @@
...
+    "@storybook/addon-links": "^7.6.10",
+    "@storybook/addon-mdx-gfm": "7.6.19",
+    "@storybook/addon-storysource": "^7.6.10",
+    "@storybook/react": "^7.6.10",
+    "@storybook/react-webpack5": "7.6.19",
     "babel-loader": "^8.2.2",
+    "chromatic": "^11.5.1",
     "css-loader": "^5.2.4",
-    "node-sass": "^7.0.0",
+    "node-sass": "npm:sass@^1.49.9",
```
jamigibbs, line 34: I did this because my local build was failing for the `web-components` package during debugging. Is it still needed?

packages/storybook/package.json
```diff
@@ -13,22 +14,26 @@
...
     "react-dom": "^17.0.2",
-    "sass-loader": "^12.4.0"
+    "sass-loader": "^12.4.0",
+    "storybook": "7.6.19"
```
jamigibbs, line 37: I'm seeing some differences in the URL for individual Stories. For example, the [single select variation for va-accordion](https://design.va.gov/storybook/?path=/docs/uswds-va-accordion--single-select) is currently:

```
path=/docs/uswds-va-accordion--single-select
```
In the [Chromatic preview](https://storybook-7-upgrade--65a6e2ed2314f7b8f98609d8.chromatic.com/?path=/story/uswds-va-accordion--single-select), it's changed to this:

```
path=/story/uswds-va-accordion--single-select
```

Is there any way to keep the same URLs so they don't break?

packages/storybook/.storybook/style.scss
it-harrison: @jamigibbs - yes, thank you!

.github/workflows/chromatic.yml
```diff
@@ -27,11 +27,14 @@ jobs:
...
-          workingDir: packages/storybook
-          skip: "dependabot/**"
+        run: yarn chromatic storybook build --project-token=${{ secrets.CHROMATIC_PROJECT_TOKEN }} --only-changed=true
+        working-directory: packages/storybook
+        env:
+          STORYBOOK_BUILD_TIMEOUT: 10000000
+          # projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+      - name: Output build log details
+        if: always() 
+        run: cat /home/runner/work/component-library/component-library/packages/storybook/build-storybook.log
```
micahchiang, line 37: Do we know if switching back to the chromatic github action works? If the file name change really does resolve the issue, we probably want to switch back to the action, since it'll ensure we're using the latest chromatic cli

packages/storybook/package.json
it-harrison: @jamigibbs - although the generated urls are different, the old ones are not broken. The one qualification is that in storybook 6 the story `https://design.va.gov/storybook/?path=/docs/uswds-va-accordion--default` would take you to the summary page for the component, but in storybook 7 it will take you to the actual default component - I tried but don't think there is a way around this since in storybook 7 each component has an extra story -- `Docs` -- which is analogous to the `--default` in storybook 6.

.github/workflows/chromatic.yml
it-harrison: @micahchiang switched back and all is good

packages/storybook/package.json
it-harrison: @jamigibbs switched back and everything still works

.github/workflows/node.js.yml
it-harrison: @jamigibbs I cleaned this action up (the `foreach` was still failing on install and test so I broke it up into three steps)

jamigibbs (approved):

micahchiang (approved):

PR #1160, opened 2024-05-23
Fixes line-height on va-file-input

rmessina1010:
## Chromatic

https://rm-lineheight-hotfix--65a6e2ed2314f7b8f98609d8.chromatic.com

 This is a quick fix,  adding vads-u-line-height--2 class to va-file-input to tighten the line-height  (so as to be in step with figma). It seems the was reverted  before merging, so I am adding it again.


it-harrison (approved):
lgtm

PR #1209, opened 2024-06-27
Add css and tests to FA icon search

Andrew565:
## Chromatic

https://9999-add-css-to-icon-search--65a6e2ed2314f7b8f98609d8.chromatic.com

---

## Description
Adds looking at css/scss and unit/spec tests to Font Awesome icon searches

## Definition of done

### DIFF
--- a/packages/design-system-dashboard-cli/src/search-files.js
+++ b/packages/design-system-dashboard-cli/src/search-files.js
@@ -38,6 +38,27 @@ function readAllModules(rootDir) {
 }
 
 /**
+ * Find all .css and .scss files.
+ *
+ * @param rootDir {string} - The path to the root directory to search
+ * @return {Module[]}
+ */
+function readAllStylesheets(rootDir) {
+  const cssFiles = glob.sync(`${rootDir}/**/*.@(css|scss)`, {
+    ignore: [
+      `${rootDir}/**/tests/**`,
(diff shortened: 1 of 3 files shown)


packages/design-system-dashboard-cli/src/search-files.js
```diff
@@ -37,6 +37,47 @@ function readAllModules(rootDir) {
...
+ * Find all .unit and .spec files.
+ *
+ * @param rootDir {string} - The path to the root directory to search
+ * @return {Module[]}
+ */
+function readAllTests(rootDir) {
+  const cssFiles = glob.sync(`${rootDir}/**/*.@(unit|spec).*`, {
+    ignore: [
+      // This mock-form directory exists in vets-website and we don't want to include it
+      `${rootDir}/**/_mock-form/**`,
```
micahchiang, line 71: This is great and makes me think: There are several other applications we could probably ignore if we wanted to: 
1. The ds playground apps (though at this point we may want to straight up remove them)
2. The burial v6 poc: this was a holdover from the VA spike team contract and will likely never be published. 

As an extension, it'd be cool to read the `registry.json` file in content-build to determine which applications have `vagovprod: false`. We'd have to conduct an eye to test to figure out which of those are truly just dead artifacts we could delete or ignore, but it could also help us get a better picture of what applications are actually maintained.

micahchiang (approved):

PR #1217, opened 2024-07-02
va-text-input: New prefix and suffix variant

harshil1793:
## Chromatic

https://2882-va-text-input-prefix-suffix--65a6e2ed2314f7b8f98609d8.chromatic.com

---

## Description
This pull request introduces new variants for handling prefixes and suffixes in the input field component. The changes include the addition of new properties `inputPrefix` and `inputSuffix` with support for dynamic content and character limits.

**Changes:**

**Prefix Variant:**

- New Prop `inputPrefix`
- Supports either an object (for icon support) or a string.
- Limits the number of characters to 25.
- Implements dynamic left padding to the input field to accommodate the prefix.

**Suffix Variant:**

- New Prop `inputSuffix`
- Supports only string values.
- Limits the number of characters to 25.
- Implements dynamic right padding to the input field to accommodate the suffix.

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2882

## QA Checklist

## Acceptance criteria

## Definition of done


harshil1793:
The introduction of a character limit for the inputPrefix and inputSuffix properties is intended to ensure that the input field maintains proper alignment and spacing. Longer prefixes and suffixes can cause UI inconsistencies, such as overflowing content and misaligned input fields. By limiting the character length to 25, we can prevent these issues while still providing enough flexibility for meaningful prefixes and suffixes.

powellkerry (approved):
LGTM

packages/web-components/src/components/va-text-input/va-text-input.tsx
```diff
@@ -176,6 +177,16 @@ export class VaTextInput {
    */
   @Prop() currency?: boolean = false;
 
+   /**
+   * This property displays a prefix that accepts either a string or an object with an icon string.
+   */
+   @Prop() inputPrefix?: string;
```
jamigibbs, line 183: > an object with an icon string
What does this mean? Is there a certain format that needs to be followed in order to use a custom icon?

packages/web-components/src/components/va-text-input/va-text-input.tsx
```diff
@@ -199,6 +213,40 @@ export class VaTextInput {
...
+    this.updatePaddingLeft();
+    this.updatePaddingRight();
+  }
+
+  private updatePaddingLeft = () => {
+    let parsedPrefix: string | { icon: string };
+    try {
+      parsedPrefix = JSON.parse(this.inputPrefix);
+    } catch (e) {
+      parsedPrefix = String(this.inputPrefix);
```
jamigibbs, line 231: Is this necessary because we are handling two different Types (string & object) in the same prop? If so, has it been considered to just make two separate props (one for custom prefix letters and one for displaying an icon).

packages/storybook/stories/va-text-input-uswds.stories.jsx
```diff
@@ -392,7 +384,27 @@ WithCharacterCount.args = { ...defaultArgs, maxlength: '10', charcount: true };
 export const WithCurrency = Template.bind(null);
 WithCurrency.args = {
   ...defaultArgs,
-  currency: true
+  currency: true,
+};
+
+export const WithCreditcard = Template.bind(null);
```
jamigibbs, line 390: Instead of naming this story "With Credit Card", can we make it more general like "With Icon"?

Screenshot 2024-07-03 at 10 22 46 AM

packages/web-components/src/components/va-text-input/va-text-input.tsx
```diff
@@ -340,9 +390,39 @@ export class VaTextInput {
     const pattern = this.getPattern();
 
     const currencyWrapper = classnames({
-      'currency-wrapper': currency
+      'currency-wrapper': currency,
+      'prefix-wrapper ': inputPrefix,
+      'suffix-wrapper ': inputSuffix
```
jamigibbs, line 395: There are USWDS classes for prefix and suffix. Are you able to use any of these? You will probably need to import the `usa-input-prefix-suffix` package.

https://designsystem.digital.gov/components/input-prefix-suffix/

packages/storybook/stories/va-text-input-uswds.stories.jsx
```diff
@@ -60,7 +65,9 @@ const defaultArgs = {
   'charcount': false,
   'max': undefined,
   'min': undefined,
-  'currency': undefined
+  'currency': undefined,
```
jamigibbs, line 68: Since we will have prefix with icons supported, the old `currency` prop seems redundant now. I wonder if it would be good to create a follow-up ticket to deprecate this. 🤔

harshil1793: Yes, I had that in mind. I can create a follow-up ticket to deprecate the `currency` prop.

packages/web-components/src/components/va-text-input/va-text-input.tsx
harshil1793: I did not realize that, thanks for pointing it out. I can push an update for it.

packages/web-components/src/components/va-text-input/va-text-input.tsx
harshil1793: I thought about this but wanted to keep it generic, and I was unsure which path to choose. Now that you mention it, I agree that adding a new prop will make it cleaner compared to the current implementation.

harshil1793: I've create a separate prop for icon and for custom prefix.

packages/web-components/src/components/va-text-input/va-text-input.tsx
harshil1793: This won't be an issue any more as I created a new prop.

packages/storybook/stories/va-text-input-uswds.stories.jsx
harshil1793: I created a new ticket.

danbrady (approved):

packages/storybook/stories/va-text-input-uswds.stories.jsx
jamigibbs: @harshil1793 Thanks Harshil! Can you link to that ticket?

packages/web-components/src/components/va-text-input/va-text-input.tsx
```diff
@@ -419,6 +484,7 @@ export class VaTextInput {
                 onInput={handleInput}
                 onBlur={handleBlur}
                 aria-describedby={ariaDescribedbyIds}
+                style={style}
```
jamigibbs, line 487: Now that we have the USWDS `.usa-input-group` class working, is it necessary for us to add custom inline style? When I toggle off this style, it seems to be working with just the USWDS styles:

Screenshot 2024-07-09 at 9 31 50 AM

packages/web-components/src/components/va-text-input/va-text-input.tsx
```diff
@@ -433,6 +499,7 @@ export class VaTextInput {
                 min={min}
                 max={max}
               />
+              {inputSuffix && <div id="text" class="usa-input-suffix" part="suffix" aria-hidden="true">{inputSuffix.substring(0, 25)}</div>}
```
jamigibbs, line 502: Is this `id="text"` necessary still?

jamigibbs (commented):
Those updates are looking great! I have just a few follow-up questions about things that we might be able to cleanup.

packages/web-components/src/components/va-text-input/va-text-input.tsx
harshil1793: No its not, I've removed it.

packages/web-components/src/components/va-text-input/va-text-input.tsx
harshil1793: While testing, I found that the padding for the USA input is not dynamic. By "dynamic," I mean that the padding does not adjust based on the length of the prefix or suffix. Hence, we need to implement inline styling that dynamically adjusts the padding.

jamigibbs (approved):
Approving but I think @danbrady has a good question about if we need icons for both prefix and suffix. The [USWDS guidance](https://designsystem.digital.gov/components/input-prefix-suffix/#using-the-input-prefix-suffix-component) says that it can be accepted for either but I'm not sure if application teams have a need for it.

> Use text or icons. Input prefixes and suffixes can accept either plain text or [icons](https://designsystem.digital.gov/components/icon/).

danbrady:
@jamigibbs We decided to discuss this with Matt for his thoughts upon his return. We can add a new ticket for this enhancement if we think we should have this feature. This way we don't hold up the release.
(BTW, I couldn't think of a use case for an icon suffix to an input, but might need to do some research. One concern I have is having an suffix icon look a button/action on that input.)

jamigibbs:
> (BTW, I couldn't think of a use case for an icon suffix to an input, but might need to do some research. One concern I have is having a suffix icon might make it look like a button/action on that input.)

@danbrady I couldn't think of a situation either except maybe to accommodate a RTL language (ie. Arabic, Hebrew).

danbrady:
After discussing with Matt, we should stay aligned with the USWDS for this component. So while they don't show examples for suffix icon, it is possible to achieve with their component. They also have the following documentation:

>**Use text or icons**. Input prefixes and suffixes can accept either plain text or icons.

So VADS should have the same features.

@caw310 Do you want to make a new ticket for this or reopen this one?

caw310:
@danbrady I created a ticket after our meeting with Matt. 
https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3063

PR #1221, opened 2024-07-11
Align error alert with other statuses

rsmithadhoc:
## Chromatic

https://alert-error-update--65a6e2ed2314f7b8f98609d8.chromatic.com

---

## Description
Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2910

Removed the special conditions for the error status. Adding a `role="alert"` to the web component itself will turn this into an ARIA live region and announce the alert when it appears, so this code was not needed. This gives some flexibility and allows usage of other `role` (`role="status"`, etc.) attributes as well.

## QA Checklist
    - N/A - no visual change.

## Screenshots

No before screenshot: I was previously able to recreate it not doing that for error when it was brought up at DSC, but for some reason I couldn't get that to happen now. But I was able to confirm that with the changes made, it still works as expected. 

This shows how it functions with `role="alert"` added to the host element. When the alert is shown, it correctly interrupts and announces the alerts content.

alert

## Acceptance criteria

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-alert/va-alert.tsx
+++ b/packages/web-components/src/components/va-alert/va-alert.tsx
@@ -177,8 +177,6 @@ export class VaAlert {
     if (definedStatuses.indexOf(status) === -1) {
       status = 'info';
     }
-    // const role = status === 'error' ? 'alert' : null;
-    // const ariaLive = status === 'error' ? 'assertive' : null;
     /* eslint-enable i18next/no-literal-string */
 
     if (!visible) return <div aria-live="polite" />;
@@ -189,12 +187,10 @@ export class VaAlert {
         'usa-alert--slim': slim,
       });
 
-      {console.log(this.el.getAttribute('data-label'))}
       return (
         <Host>
           <div
             role={this.el.getAttribute('data-role')}
-            // aria-live={ariaLive}
             class={classes}
             aria-label={this.el.getAttribute('data-label')}
           >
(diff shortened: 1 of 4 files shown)


packages/web-components/src/components/va-alert/va-alert.tsx
```diff
@@ -188,13 +186,13 @@ export class VaAlert {
         'usa-alert--success': status === 'continue',
         'usa-alert--slim': slim,
       });
+
       return (
         <Host>
           <div
-            role={this.el.getAttribute('data-role') || role}
```
jamigibbs, line 194: In your screenshot example, you mentioned that it functions with `role="alert"` added to the host element. 

Does that mean that we should let teams know (or updated our guidance) that if they want that functionality, they should do something like: 

```html
<va-alert role="alert/status" ... />
``` 
Is it worthwhile to add a Storybook example?

rsmithadhoc: Yeah, good call, I will add an example to Storybook. I'll do a separate PR for docs since it isn't really clear over there as well.

rsmithadhoc: @jamigibbs I added those examples and also put up a separate PR for some documentation updates https://github.com/department-of-veterans-affairs/vets-design-system-documentation/pull/3007, thanks!

jamigibbs (approved):
Nice update! Thanks Ryan!

PR #1228, opened 2024-07-16
va-file-input: update screen reader message when file is uploaded or removed

ataker:
## Chromatic

https://file-input-sr-alerts--65a6e2ed2314f7b8f98609d8.chromatic.com

## Description
Closes [v3 File Input should have a success alert #2947](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2947)

## QA Checklist

## Screenshots
Screenshot 2024-07-16 at 16 11 42
Screenshot 2024-07-16 at 16 17 07

## Acceptance criteria

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-file-input/va-file-input.tsx
+++ b/packages/web-components/src/components/va-file-input/va-file-input.tsx
@@ -198,7 +198,8 @@ export class VaFileInput {
   private updateStatusMessage(message:string) {
     // Add delay to encourage screen reader readout
     setTimeout(() => {
-      this.el.shadowRoot.querySelector("#statusMessage").textContent = message
+      const statusMessageDiv = this.el.shadowRoot.querySelector("#statusMessage");
+      statusMessageDiv ? statusMessageDiv.textContent = message : "";
     }, 1000);
   }
 
--- a/packages/web-components/src/components/va-file-input/va-file-input.tsx
+++ b/packages/web-components/src/components/va-file-input/va-file-input.tsx
@@ -426,7 +426,7 @@ export class VaFileInput {
                     </Fragment>
                   )}
                 </span>
-                <div class='usa-sr-only' aria-live="polite" id="statusMessage">No files selected.</div>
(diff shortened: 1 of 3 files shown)


packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -168,6 +170,8 @@ export class VaFileInput {
       this.vaChange.emit({ files: [] });
     }
     this.file = null;
+    this.updateStatusMessage(`File removed. No file selected.`);
```
rsmithadhoc, line 173: This was not getting read out in my testing. I think once it gets deleted, it may only be in the DOM briefly, and then it renders the component's initial state with "No files selected."

rsmithadhoc (changes requested):
@ataker This worked as expected in VoiceOver, but JAWS and NVDA had some repetition of alerts after adding the file: "No file selected", "No file selected", "You have selected the file: ..." I think that "No file selected" would need to be hidden from the component's initial state. I believe that SR text can be removed entirely, since we are now reading the label/hint text which says to upload a file, that SR text feels redundant. I think that would solve the redundancy and we can keep the SR messages to only when an action occurs.

Keep:
- "You have selected the file: ..." on add.
- "File removed. No file selected." on delete.

Remove:
- "No file selected." on initial load.

Thanks!

rsmithadhoc (changes requested):
Success message and removal messages are looking good. It is a bit verbose after adding a file because the operating system's file browser starts reading the browser title, but the success message does get read out eventually. I compared this to a standard HTML `<input type="file">` and it seems to do the same thing, so I don't believe we can do anything about that.

That "No file selected." still needs to be removed. Otherwise, when you add a file it will say "No file selected. No file selected. No file selected." for each input on the page. I tested with a branch and it corrected that https://github.com/department-of-veterans-affairs/component-library/pull/1233/commits/b1ca926cff643ccce6fc4835d8baa7fdd9b9e0b8

rsmithadhoc (approved):

PR #1245, opened 2024-07-29
add consistent types to call signatures

micahchiang:
## Chromatic

https://{{head.ref}}--65a6e2ed2314f7b8f98609d8.chromatic.com

---

## Description
Some of our functions contain things like return types and parameter types while others do not. Many functions also include JSDoc annotations that include parameter types and descriptions. While this is helpful, it doesn't necessarily provide the static type checking we may want from Typescript.

For example, the `replaceName` function in our instance of the file input upgrader. While the `s` parameter is given the `String` type in its JSDoc comment, it's never assigned a type in the parameter list and defaults to `any`, which introduces a potential silent failure like this: 

Screenshot from 2024-07-29 12-54-20

Assigning the `s` parameter an explicit type of `string` gets us the intellisense we'd expect from the typescript server: 

Screenshot from 2024-07-29 12-54-42

This pull request adds types to our function type expressions that were missing them.

### DIFF
--- a/packages/web-components/src/components/va-file-input/va-file-input-upgrader.ts
+++ b/packages/web-components/src/components/va-file-input/va-file-input-upgrader.ts
@@ -46,14 +46,14 @@ let DEFAULT_FILE_STATUS_TEXT = '';
  * @param {HTMLElement} el the element within the file input
  * @returns {FileInputContext} elements
  */
-const getFileInputContext = el => {
-  const dropZoneEl = el.closest(DROPZONE);
+const getFileInputContext = (el: HTMLElement): { dropZoneEl: HTMLElement, inputEl: HTMLInputElement } => {
+  const dropZoneEl = el.closest(DROPZONE) as HTMLElement;
 
   if (!dropZoneEl) {
     throw new Error(`Element is missing outer ${DROPZONE}`);
   }
 
-  const inputEl = dropZoneEl.querySelector(INPUT);
+  const inputEl = dropZoneEl.querySelector(INPUT) as HTMLInputElement;
 
   return {
     dropZoneEl,
@@ -66,7 +66,7 @@ const getFileInputContext = el => {
  *
  * @param {HTMLElement} el An element within the file input component
  */
-const disable = el => {
+const disable = (el: HTMLElement): void => {
   const { dropZoneEl, inputEl } = getFileInputContext(el);
 
   inputEl.disabled = true;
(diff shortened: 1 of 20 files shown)


packages/web-components/src/components/va-file-input/va-file-input-upgrader.ts
```diff
@@ -300,17 +301,16 @@ const removeOldPreviews = (dropTarget, instructions) => {
  * @param {Object} fileNames - The selected files found in the fileList object.
  * @param {Array} fileStore - The array of uploaded file names created from the fileNames object.
  */
-const updateStatusMessage = (statusElement, fileNames, fileStore) => {
+const updateStatusMessage = (statusElement: HTMLDivElement | Element, fileNames: FileList | File[], fileStore: string[]): void => {
```
Andrew565, line 304: This can be simplified to `statusElement: Element` as every element will match that type.

packages/web-components/src/components/va-file-input/va-file-input-upgrader.ts
```diff
@@ -364,12 +364,12 @@ const addPreviewHeading = (fileInputEl, fileNames) => {
  * @param {HTMLDivElement} dropTarget - The drag and drop target area.
  */
 
-const handleChange = (e, fileInputEl, instructions, dropTarget) => {
+const handleChange = (e: Event, fileInputEl: HTMLInputElement, instructions: HTMLDivElement, dropTarget: HTMLDivElement) => {
   const multiple = fileInputEl.getAttribute('multiple'); // TODO: Double-check this works correctly when multiple is re-enabled
-  const fileNames = multiple ? e.target.files : [e.target.files[0]];
+  const fileNames = multiple ? (e.target as HTMLInputElement).files : [(e.target as HTMLInputElement).files[0]];
```
Andrew565, line 369: Instead of casting the type twice, you could store `e.target` as a separate variable and cast it once maybe?

packages/web-components/src/utils/date-utils.ts
```diff
@@ -148,12 +149,12 @@ export const days = {
  * Return true if the year is a leap year.
  * (Exported for testing purposes)
  */
-export function checkLeapYear(year: number) {
+export function checkLeapYear(year: number): boolean {
```
Andrew565, line 152: I'd love to rename this to `isLeapYear` but that may be more trouble than it's worth for what's essentially a personal preference.

Andrew565 (commented):
Just a few suggestions!

PR #1253, opened 2024-08-02
Tool for getting codeowner team member names

Andrew565:
## Chromatic

https://9999-codeowner-scanner-tool--65a6e2ed2314f7b8f98609d8.chromatic.com

---

## Description
Tool that takes a codeowner team name and spits out names of that team's members.

### DIFF
--- a/codeowner-scan.sh
+++ b/codeowner-scan.sh
@@ -1,0 +1,13 @@
+#!/bin/sh
+
+RETJSON=$(gh api \
+  -H "Accept: application/vnd.github+json" \
+  /orgs/department-of-veterans-affairs/teams/$1/members)
+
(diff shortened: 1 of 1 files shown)


codeowner-scan.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
```
micahchiang, line 2: This is nice! Would you mind adding a comment in here that provides just a little more guidance? Maybe a brief description of the command line arg it takes, whether or not you need to be authenticated to gh api, etc? Thanks!

micahchiang (approved):
Just left a suggestion to add a comment that provides some additional info on how to use this. Looks good otherwise :+1:

PR #1263, opened 2024-08-07
CSS Library: generate medium-screen breakpoint utility classes

jamigibbs:
## Chromatic

https://3045-generate-medium-screen-bp--65a6e2ed2314f7b8f98609d8.chromatic.com

## Description
We are using a breakpoint called `medium-screen` (768px) which is not available in the USWDS system. The CSS Library is generating utility classes from the USWDS system though and this PR will also add the `medium-screen` breakpoint utility classes to the utilities.css stylesheet.

It will generate those breakpoint utility classes for the [following packages](https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/css-library/src/stylesheets/utilities.scss#L8-L24) _except_ for `u-bg-color` and `u-color`. I didn't see these being generated for other breakpoints so I'm leaving them out:

```scss
  $u-border,
  $u-display,
  $u-flex-grid,
  $u-flexbox,
  $u-font,
  $u-height,
  $u-line-height,
  $u-margin,
  $u-measure,
  $u-padding,
  $u-position,
  $u-text-align,
  $u-text-decoration,
  $u-visibility,
  $u-width
```

This is hopefully going to be temporary because we have an open [PR with the USWDS core team](https://github.com/uswds/uswds/pull/5985) to add custom breakpoints. If/when that is merged and the USWDS package is updated, we will need a follow-up task to remove this code and add the custom breakpoint to the theme settings.

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3045

## Acceptance criteria

## Definition of done


packages/css-library/src/utilities/flex-grid.scss
```diff
@@ -51,4 +51,4 @@ $u-flex-grid-cols: (
   ),
 );
 
-$u-flex-grid: map-collect($u-flex-grid-cols, $u-flex-grid-cols);
+$u-flex-grid: map-collect($u-flex-grid-col, $u-flex-grid-cols);
```
jamigibbs, line 54: Just a small bug that didn't get caught in the initial commit. We should see now that the `vads-l-col` utilities are generating.

packages/css-library/src/utilities/border.scss
```diff
@@ -7,7 +7,7 @@ $u-border-base: (
   border: (
     base: 'vads-u-border',
     modifiers: (
-      noModifier: '-',
+      noModifier: '',
```
jamigibbs, line 10: A small bug fix. When it was set as `noModifier: '-',`, some border utilities were getting generated like this:

```css
  .mobile\:vads-u-border--7px {
    border-: 7px solid;
  }
```
Instead of this:

```css
  .mobile\:vads-u-border--7px {
    border: 7px solid;
  }
```

packages/css-library/src/stylesheets/utilities.scss
```diff
@@ -24,5 +24,134 @@ $utilities-package: map-collect(
...
+
+@media all and (min-width: $medium-screen-breakpoint) {
+  // $u-border
+  @include generate-utilities-from-settings($u-border);
+
+  // $u-display
+  @include generate-utilities-from-settings($u-display);
+
+  // $u-flex-grid
+  @each $property, $property-map in $u-flex-grid {
```
jamigibbs, line 71: The `u-flex-grid` and `u-margin` settings are more complicated packages that the others. So instead of continuing to build out the `generate-utilities-from-settings` logic, I decided to just give them their own sections.

All of this will hopefully be removed anyway once USWDS core is updated to allow for custom breakpoints.

powellkerry (approved):

Andrew565 (approved):

packages/css-library/src/stylesheets/utilities.scss
```diff
@@ -24,5 +24,136 @@ $utilities-package: map-collect(
...
+  @include generate-utilities-from-settings($u-display);
+
+  // $u-flex-grid
+  @each $property, $property-map in $u-flex-grid { 
+    $u-base: map-get($property-map, base);
+    $u-values: map-get($property-map, values);
+    $u-property: map-get($property-map, property);
+
+    @each $name, $value in $u-values {
+      $extend: map-get($value, extend);
```
micahchiang, line 79: I don't understand what `$extend` ends up evaluating to here. Could you explain?

micahchiang (approved):
Really great work! Just left a question about `$extend` because I don't think I get that part. Otherwise, generated utilities look good :+1: Exciting that this gets us moving forward with the rest of the utility swaps. Hopefully that uswds PR gets merged soon :crossed_fingers:

packages/css-library/src/stylesheets/utilities.scss
jamigibbs: The `flex-grid` utilities have a block of CSS styles that are named `extend` in the tokens like this (maybe not the greatest name) :

https://github.com/department-of-veterans-affairs/component-library/blob/3ca3fcc424f932a70feda6f871fa6847b56739cd/packages/css-library/src/tokens/flex-grid.scss#L10-L18

I wasn't able to just dump this into the CSS brackets (maybe my Sass inexperience?) so I'm iterating over them instead like this:

```
  @each $css-property, $css-value in $extend {
    #{$css-property}: $css-value;
  }
```

With the flex property added in the end, it generates like this:

```css
  .medium-screen\:vads-l-col {
    box-sizing: border-box;
    flex-basis: 0;
    max-width: 100%;
    min-height: 1px;
    min-width: 0;
    position: relative;
    width: 100%;
    flex-grow: 1;
  }
```

PR #1305, opened 2024-08-27
fix: focus outline color fix for v3 components

harshil1793:
## Chromatic

https://2936-focus-v3-components-fix--65a6e2ed2314f7b8f98609d8.chromatic.com

---

## Description
This PR fixes focus outline color for v3 components.

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2936

## QA Checklist

## Screenshots
**Breadcrumbs:**
Screenshot 2024-08-27 at 7 04 02 PM

**Banner promo:**
Screenshot 2024-08-27 at 7 03 53 PM

**Statement of truth:**
Screenshot 2024-08-27 at 7 04 32 PM

**Privacy agreement:**
Screenshot 2024-08-27 at 7 04 22 PM

**Back to top:**
Screenshot 2024-08-27 at 7 03 40 PM

## Acceptance criteria

## Definition of done


packages/web-components/src/components/va-privacy-agreement/va-privacy-agreement.scss
```diff
@@ -23,4 +23,8 @@ a {
 .privacy-policy-icon {
   margin-left: 4px;
   vertical-align: text-top;
+}
+
+a:focus {
+  outline: 2px solid var(--vads-color-action-focus-on-light);
```
jamigibbs, line 29: Does importing (modifying it if necessary) the mixin not work?

https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/src/mixins/focusable.css

For example: 
https://github.com/department-of-veterans-affairs/component-library/blob/c021b5f55dd524926eabe7f78d3014022ff386a3/packages/web-components/src/components/va-minimal-footer/va-minimal-footer.scss#L1

harshil1793: You're right, I missed that. I've pushed an update now that imports and modifies the mixin as needed. Thanks for catching that!

jamigibbs:
The [original ticket](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2936) says that the outside yellow outline should be removed from this banner (keep the inside border):

Screenshot 2024-08-28 at 12 05 42 PM

> Currently outlines the banner itself with the yellow outline, then the internal elements with the default browser outline.
Instead, we should remove the outline from the banner itself as it is not an interactive element, then apply the yellow outline to the link and button inside the banner.

harshil1793:
For the banner outline, the issue is coming from the formation because `va-promo-banner` has an href attribute. I tried to override it, but it doesn't seem to work. Do you think I should make a change in formation to fix this?
Screenshot 2024-08-28 at 2 32 25 PM

jamigibbs:
> For the banner outline, the issue is coming from the formation because `va-promo-banner` has an href attribute. I tried to override it, but it doesn't seem to work. Do you think I should make a change in formation to fix this? Screenshot 2024-08-28 at 2 32 25 PM

Can you reset it if you put it in the [web component global stylesheet](https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/src/global/main.css)?

packages/web-components/src/global/main.css
```diff
@@ -9,3 +9,7 @@ va-process-list li :is(h2, h3, h4, h5) {
   clear: none;
   padding-top: 0.188rem;
 }
+
+va-promo-banner {
+  outline: none !important;
```
jamigibbs, line 14: Can you do this without the `!important` keyword (make it more specific)? For example:

```css
va-promo-banner[href]:focus {
  outline: none;
}
```

harshil1793: Actually, it doesn't work without the `!important` keyword.

packages/web-components/src/components/va-back-to-top/va-back-to-top.css
```diff
@@ -43,7 +44,6 @@ a {
   padding: 0;
   pointer-events: none;
   position: absolute;
-  transition: all 0.25s ease-in;
```
jamigibbs, line 46: I don't think this should be removed. We want the component to transition when it comes into view. You can see how it fades in on this page here: https://www.va.gov/burials-memorials/memorial-items/

Andrew565: We could change the `all` to `opacity` so that it only transitions the opacity and not the focus outline.

jamigibbs: @Andrew565 That's a good idea. It might also need some kind of positioning transition too? It kind of moves up when it comes in view:


https://github.com/user-attachments/assets/578fb9b0-4a78-478d-8cfb-37c366b17a94

harshil1793: I was able to fix it with background color transition, and the positioning transition works as well.

danbrady:
@harshil1793 I notice the focus ring disappears on the Back to top if it's hovered over:


https://github.com/user-attachments/assets/adf2e6a4-e3ac-4fc7-86c2-424907da6c84

harshil1793:
@danbrady I believe that is existing behavior, I don't mind fixing that if its a bug.

danbrady:
@harshil1793 Ah, you're right. I see the bug in the existing component too. It would be great if you can fix it now with this update. If the issue starts to take up too much time though, let's create a separate ticket for it. Thanks!

danbrady (approved):

PR #1318, opened 2024-09-09
CSS Library: remove grid classes

micahchiang:
## Chromatic

https://mc-remove-grid-classes--65a6e2ed2314f7b8f98609d8.chromatic.com

---
## Description
We're removing grid classes from css-library for the time being until we have a better way to handle the transition from Formation.


packages/css-library/src/stylesheets/utilities.scss
```diff
@@ -9,7 +9,6 @@ $utilities-package: map-collect(
   $u-border,
   $u-color,
   $u-display,
-  $u-flex-grid,
```
jamigibbs, line 12: Maybe just comment this out like the other grid stuff instead of remove?

jamigibbs (approved):

jamigibbs:
[Slack convo](https://dsva.slack.com/archives/G01BJ3ESXL4/p1725893154851419?thread_ts=1725647550.741689&cid=G01BJ3ESXL4) for reference.

PR #1352, opened 2024-10-04
css-library: update margin auto value

micahchiang:
## Chromatic

https://mc-margin-auto-fix--65a6e2ed2314f7b8f98609d8.chromatic.com

---

## Description
The 404 page on va.gov is mis-aligned, this is because margin-auto values in css-library were configured to use -1px instead of `auto`. This PR updates the map value for margin auto

## QA Checklist
- [ ] Changes have been tested in vets-website using Verdaccio (if applicable. If not applicable provide reason why)

## Screenshots
<details><summary>Current settings in prod x--auto is set to -1px</summary>
 
image

</details>

<details><summary>With margins fixed in css-library</summary>

image

</details>

## Acceptance criteria
- [ ] QA checklist has been completed
- [ ] Screenshots have been attached that cover desktop and mobile screens

## Definition of done
- [ ] Documentation has been updated, if applicable
- [ ] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)

PR #1355, opened 2024-10-07
update version for core and web-components

micahchiang:
## Chromatic

https://mc-comp-lib-47.1.0--65a6e2ed2314f7b8f98609d8.chromatic.com

---
## Description
Updates version for core and web-components. css-library was updated on Friday [here](https://github.com/department-of-veterans-affairs/component-library/pull/1352/files#diff-1c40d3600debc345902064e33f34f49a99e1230a0a9ec56e6aa590bd93188d11)

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@department-of-veterans-affairs/component-library",
   "description": "VA.gov component library. Includes React and web components.",
-  "version": "47.0.0",
+  "version": "47.1.0",
   "license": "MIT",
   "scripts": {
     "build": "webpack"
--- a/packages/web-components/package.json
+++ b/packages/web-components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/web-components",
-  "version": "15.0.0",
+  "version": "15.1.0",
   "description": "Stencil Component Starter",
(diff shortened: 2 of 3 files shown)


packages/core/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@department-of-veterans-affairs/component-library",
   "description": "VA.gov component library. Includes React and web components.",
-  "version": "47.0.0",
+  "version": "47.1.0",
```
jamigibbs, line 4: maybe should be a patch?

packages/web-components/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@department-of-veterans-affairs/web-components",
-  "version": "15.0.0",
+  "version": "15.1.0",
```
jamigibbs, line 3: If there are no web-component changes, I don't think this should change.

jamigibbs (approved):

packages/web-components/package.json
micahchiang: Looking at [this](https://github.com/department-of-veterans-affairs/component-library/compare/v47.0.0...main), it seems like there's one web-components commit that needs to go out?

jamigibbs: @micahchiang  oh good catch! For some reasons I thought it was just CSS Library.

PR #1373, opened 2024-10-24
va-combo-box: add new component

oleksii-morgun:
## Chromatic

https://add-uswds-combo-box--65a6e2ed2314f7b8f98609d8.chromatic.com

## Description

Related to https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3154

Designs https://www.figma.com/design/uddjxccqvUyz5FSg1aD5wi/ZFeat---Triage-Group-Selection-Flows?node-id=1804-73207&node-type=canvas&t=Rhhpivvbmxpw3vw2-0

USWDS combobox: https://designsystem.digital.gov/components/combo-box/

## QA Checklist

## Acceptance criteria

## Definition of done


packages/web-components/src/components/va-combo-box/va-combo-box.tsx
```diff
@@ -0,0 +1,251 @@
...
+          <select
+            class="usa-select"
+            aria-describedby={ariaDescribedbyIds}
+            aria-invalid={invalid || error ? 'true' : 'false'}
+            name={name}
+            id="options"
+            onChange={e => this.handleChange(e)}
+            disabled={disabled}
+          >
+            {this.populateOptions()}
```
jamigibbs, line 245: The spacing for the dropdown options don't match USWDS. Are we able to sync that up with their style module?

It also looks like our input has rounded corners but they are square.

**Storybook**

Screenshot 2024-11-01 at 9 13 22 AM

**USWDS Example***

Screenshot 2024-11-01 at 9 13 06 AM

packages/web-components/src/components/va-combo-box/va-combo-box.scss
```diff
@@ -0,0 +1,59 @@
...
+@use 'uswds-helpers/src/styles/usa-sr-only';
+@use 'usa-icon/src/styles/usa-icon';
+@use 'usa-select/src/styles/usa-select';
+@use 'usa-input/src/styles/usa-input';
+@use 'usa-hint/src/styles/usa-hint';
+
+@import '../../mixins/uswds-input-width.scss';
+@import '../../mixins/uswds-error-border.scss';
+
+
```
jamigibbs, line 15: We diverge a little bit with USWDS and use a yellow border on focus instead of blue so we'll want to do that with the combobox as well.

Here is an example of the yellow focus border on va-input:

Screenshot 2024-11-01 at 9 17 01 AM

We have a mixin that you can import here should make that update for you:

```css
@import '../../mixins/focusable.css';
```

packages/web-components/src/components/va-combo-box/test/va-combo-box.e2e.ts
```diff
@@ -0,0 +1,159 @@
...
+            </span>`);
+    const element = await page.find('va-combo-box');
+
+    // validate that the label is followed by the hint
+    expect(element.shadowRoot.innerHTML).toContain(
+      `<label for="options" class="usa-label" id="options-label">${labelText}</label><span class="usa-hint" id="input-hint">${hintText}</span>`,
+    );
+  });
+
+  it('renders disabled', async () => {
```
jamigibbs, line 146: At a minimum we'd like to see a test for every prop and event for the component. That means I'd expect a minimum of 13 tests here but I only see 7. Do you mind beefing up these tests a little bit more?

packages/web-components/src/components/va-combo-box/va-combo-box.tsx
```diff
@@ -0,0 +1,251 @@
...
+  State,
+  Fragment,
+} from '@stencil/core';
+import classnames from 'classnames';
+import { i18next } from '../..';
+import { comboBox } from './va-combo-box-library.js';
+
+/**
+ * @componentName Combo Box
+ * @maturityCategory dont_use
```
jamigibbs, line 19: I wonder if this should be Use With Caution: Candidate? 🤔  

Screenshot 2024-11-01 at 9 22 40 AM

Here are the [guidelines for the maturity scale](https://design.va.gov/about/maturity-scale#dont-use-proposed).

The Use With Caution: Candidate requirements say:

- Design System Team and/or Design System Council have evaluated the proposal.
- The Design System Team will be creating documentation for the component or pattern.
- The component or pattern may be in limited use (i.e. it may have already been shipped by a team). This could also be known as: “In work”, “Draft”, “Beta”, or “Give it a go! YMMV”.

packages/storybook/stories/va-combo-box-uswds.stories.jsx
```diff
@@ -0,0 +1,115 @@
...
+WithHintText.args = {
+  ...defaultArgs,
+  hint: 'This is example hint text',
+};
+
+export const Disabled = Template.bind({});
+Disabled.args = {
+  ...defaultArgs,
+  value: 'elderberry',
+  disabled: true,
```
jamigibbs, line 114: I don't think we want to advertise that there is a disabled state. There are a11y concerns with this functionality. 

Or maybe we don't want to have this functionality in the component at all? I will let the design and a11y reviewers weigh in on that though.

Screenshot 2024-11-01 at 9 28 50 AM

packages/web-components/src/components/va-combo-box/va-combo-box.scss
```diff
@@ -0,0 +1,59 @@
...
+::slotted(optgroup){
+  display: none;
+}
+
+.usa-combo-box{
+  font-size: 12px;
+  max-width: 100%;
+  width: 100%;
+  box-sizing: border-box;
+}
```
jamigibbs, line 42: I'm curious why we need to override the USWDS styling class here?

packages/web-components/src/components/va-combo-box/va-combo-box.tsx
```diff
@@ -0,0 +1,251 @@
...
+  @Prop() messageAriaDescribedby?: string;
+
+  /**
+   * The event emitted when the selected value changes
+   */
+  @Event() vaSelect: EventEmitter;
+
+  /**
+   * Whether an error message should be shown - set to false when this component is used inside va-date or va-memorable-date
+   * in which the error for the va-select will be rendered outside of va-select
```
jamigibbs, line 101: The documentation comments for this prop is for `va-select` and I don't think this prop is even needed for va-combo-box but just when `va-select` is being used as a child component in those other components.

For `va-combo-box`, we can probably just use the `error` prop as a falsy/truthy value to determine if the error is displayed.

packages/web-components/src/components/va-combo-box/va-combo-box.tsx
```diff
@@ -0,0 +1,251 @@
...
+  /**
+   * The event used to track usage of the component. This is emitted when an
+   * option is selected and enableAnalytics is true.
+   */
+  @Event({
+    eventName: 'component-library-analytics',
+    composed: true,
+    bubbles: true,
+  })
+  componentLibraryAnalytics: EventEmitter;
```
jamigibbs, line 114: I've been advised previously by @jestutt from platform analytics that we should not add analytics code to a component until we have coordinated with her. So I think we should remove any custom analytics code for now and we'll create a follow-up ticket to get it added.

Unless @oleksii-morgun your team needs it for your user testing?

packages/web-components/src/components/va-combo-box/va-combo-box.tsx
```diff
@@ -0,0 +1,251 @@
...
+  componentDidLoad() {
+    const comboBoxElement = this.el.shadowRoot.querySelector('.usa-combo-box');
+    const labelElement = this.el.shadowRoot.querySelector('label');
+    if (comboBoxElement) {
+      comboBox.init(comboBoxElement, labelElement);
+      comboBox.on(comboBoxElement);
+    }
+    const inputElement = this.el.shadowRoot.querySelector('input');
+    if (inputElement && this.error) {
+      inputElement.classList.add('usa-input--error');
```
jamigibbs, line 134: Do you mind explaining a little more why displaying the error needs to be done in a component lifecycle instead of just using conditional markup in the render?

packages/web-components/src/components/va-combo-box/va-combo-box.tsx
```diff
@@ -0,0 +1,251 @@
...
+      throw new Error('va-combo-box: label is a required property');
+    }
+  }
+
+  componentDidLoad() {
+    const comboBoxElement = this.el.shadowRoot.querySelector('.usa-combo-box');
+    const labelElement = this.el.shadowRoot.querySelector('label');
+    if (comboBoxElement) {
+      comboBox.init(comboBoxElement, labelElement);
+      comboBox.on(comboBoxElement);
```
jamigibbs, line 130: I'm not sure I understand what the label needs to be added in this way instead of just in the markup. Do you mind clarifying?

packages/web-components/src/components/va-combo-box/va-combo-box.tsx
```diff
@@ -0,0 +1,251 @@
...
+    bubbles: true,
+  })
+  componentLibraryAnalytics: EventEmitter;
+
+  connectedCallback() {
+    i18next.on('languageChanged', () => {
+      forceUpdate(this.el);
+    });
+    if (!this.label) {
+      throw new Error('va-combo-box: label is a required property');
```
jamigibbs, line 121: I'm not sure about this approach for throwing an Error on required props. 🤔  Generally we will just allow the markup for the prop to display or not display.

packages/web-components/src/components/va-combo-box/va-combo-box.tsx
```diff
@@ -0,0 +1,251 @@
...
+    const target: HTMLSelectElement = e.target as HTMLSelectElement;
+    this.value = target.value;
+
+    if (this.enableAnalytics) {
+      const detail = {
+        componentName: 'va-combo-box',
+        action: 'change',
+        details: {
+          label: this.label,
+          selectLabel: this.value,
```
jamigibbs, line 177: Analytics will need to be coordinated with @jestutt before we can add this.

jamigibbs (commented):
Thank you so much @oleksii-morgun for working on this! This is going to be a really nice addition to the component library. I've just left a few questions and comments for discussion. I've also tagged our designer group to have a look too when they have a chance. 🙇🏼‍♀️

oleksii-morgun:
@jamigibbs thank you for a thorough review. I will have to get back to this the following week as I am out next week. And will look deeper into your suggestions.

rsmithadhoc (changes requested):
Overall, this looks good, thanks for implementing this! There are a few things that came out of [accessibility testing](https://design.va.gov/about/accessibility/accessibility-testing-for-design-system-components#testing-methodology):
- Hiding the disabled state in Storybook, as Jami mentioned. We advise teams to hide controls that are not applicable, rather than disable them.
- Cuts off at 300% zoom and above
    - Set browser window to 1280px wide, zoom in to 300%, 400%.
      zoom
- Keyboard - Because the toggle icon is a button that can be clicked, it should also be accessible by keyboard.
  keyboard-toggle
- Screen readers (All) - **With Error/With Hint Text** variation - Error and hint text should be read by screen readers when initially tabbing to the field. I believe `aria-describedby` IDs need to be applied to the `input` element.
- Safari - X icon button right padding is missing.
  image

Thanks!

packages/web-components/src/components/va-combo-box/va-combo-box.tsx
oleksii-morgun: Yes, I should be able to tweak this slightly

packages/web-components/src/components/va-combo-box/va-combo-box.scss
oleksii-morgun: awesome, will apply that here

packages/web-components/src/components/va-combo-box/test/va-combo-box.e2e.ts
oleksii-morgun: sure, I can work on adding more coverage here

packages/web-components/src/components/va-combo-box/va-combo-box.tsx
oleksii-morgun: If you feel this is a more appropriate status, I can definitely update that. I just did not want to skip over any phase of maturity yet, given there are a11y issues we want to address before this can be used

packages/storybook/stories/va-combo-box-uswds.stories.jsx
oleksii-morgun: understood, will remove from the doc but will keep `disabled` functional as it is now

packages/web-components/src/components/va-combo-box/va-combo-box.tsx
oleksii-morgun: no, not for GA analytics. Our team can still track this with DD RUM

packages/web-components/src/components/va-combo-box/va-combo-box.tsx
oleksii-morgun: so this is where the tricky part with uswds combobox component comes in. I am actually not adding a label here, but rather passing a reference to the `init` function. This is a slight change I hade to make to `init` function. Prior to this, the function was trying to identify a label with `document.querySelector` but it was not able to due to `label` in our case being rendered in `shadow-root`. This is just a workaround. `<label>` is actually added within `<Host>`

packages/web-components/src/components/va-combo-box/va-combo-box.tsx
oleksii-morgun: that is is due to `<input>` being added by the `init` function. Have to wait until `init` is called, `input` rendered and then assign `error` class if that prop is present

packages/web-components/src/components/va-combo-box/va-combo-box.tsx
oleksii-morgun: I was not sure how to approach this. But technically `init` function will fail if `<label>` is not present. I decided to keep `label` prop as required and throw a console error if it is not provided so that it is more explicit to why component may not be behaving as expected.

oleksii-morgun:
@rsmithadhoc thank you for your thorough review and detailed feedback. I will look into making changes accordingly.

packages/web-components/src/components/va-combo-box/va-combo-box.scss
oleksii-morgun: I think I may have had some styling issues when testing with `yarn serve` instead of storybook. I will give it a test

packages/web-components/src/components/va-combo-box/va-combo-box.tsx
oleksii-morgun: yes, this makes sense. I did indeed use `va-select` as a starting for this, and may have left this behind unintentionally. I will clean this 👍🏻

packages/web-components/src/components/va-combo-box/va-combo-box.tsx
jamigibbs: @oleksii-morgun Ideally, we should just be using the USWDS styling classes for visual parity (ie `usa-combo-box` `usa-select`):

Screenshot 2024-11-21 at 9 20 44 AM

Usually the only reasons we diverge (like override the styles from their modules in the component stylesheet) is to fix a bug they haven't yet fixed, or, if we are adding additional functionality.

packages/web-components/src/components/va-combo-box/va-combo-box.tsx
jamigibbs: @oleksii-morgun Sounds good. I think then that you can go ahead and remove the analytics event and other related code and we will create a follow-up ticket to coordinate adding it at a later time.

packages/web-components/src/components/va-combo-box/va-combo-box.tsx
oleksii-morgun: yep, I will revert back my custom styling on padding for this and just keep using uswds.

packages/web-components/src/components/va-combo-box/va-combo-box.tsx
jamigibbs: @oleksii-morgun ah! my bad for not digging far enough into the `va-combo-box-library` script. Most of that script is directly from the USWDS? 

- https://github.com/uswds/uswds/blob/develop/packages/usa-combo-box/src/index.js

packages/web-components/src/components/va-combo-box/va-combo-box.tsx
jamigibbs: @oleksii-morgun  I see what you mean. In the script file, there's also some logic to throw an error for it too. Is that not flagging as I would expect? I just want to make sure we aren't adding duplicate errors.

https://github.com/department-of-veterans-affairs/component-library/blob/1af41e7a68ee8ca524f294356fa6ce4435c31a03/packages/web-components/src/components/va-combo-box/va-combo-box-library.js#L202-L212

packages/web-components/src/components/va-combo-box/va-combo-box-library.js
```diff
@@ -0,0 +1,842 @@
...
+
+  /**
+   * Enhance a select element into a combo box component.
+   *
+   * @param {HTMLElement} _comboBoxEl The initial element of the combo box component
+   */
+  const enhanceComboBox = (_comboBoxEl, _labelEl) => {
+    const comboBoxEl = _comboBoxEl.closest(COMBO_BOX);
+    console.log('comboBoxEl', _comboBoxEl);
+    console.log('_labelEl', _labelEl);
```
jamigibbs, line 167: Do you mind cleaning up these console.logs before merging? They are printing in the test runs `yarn test`.

Screenshot 2024-11-21 at 9 51 41 AM

packages/web-components/src/components/va-combo-box/va-combo-box.tsx
jamigibbs: @oleksii-morgun I brought this to the team to discuss a little more and it sounds like we'd prefer the `Use With Caution: Candidate` designation since the experimental design proposal has already gone to the design system council.

packages/web-components/src/components/va-combo-box/va-combo-box.tsx
oleksii-morgun: correct, I just made couple minor changes to account for that label reference in `init` function. the rest is copy-paste

packages/web-components/src/components/va-combo-box/va-combo-box-library.js
oleksii-morgun: yikes...

danbrady (changes requested):
Thanks for working on this, @oleksii-morgun! I didn't find any additional design issues than what's already been identified by @jamigibbs and @rsmithadhoc. It looks like the spacing/sizing/padding, VADS focus states, and the clipping on browser zoom (or at 320px/"extra small" in storybook) are still open items. I can review again once updated. Thanks, again!

oleksii-morgun:
@danbrady thanks, been actively working on it, should have updates shortly

packages/web-components/src/components/va-combo-box/va-combo-box.tsx
oleksii-morgun: I just re-tested and realized that at some point I added <label> in markup and now this is irrelevant. I will remove this error from here

oleksii-morgun:
I pushed updated, hopefully addressing all the suggestions above. When y'all have a moment, would you please review 🥺

danbrady:
Lookin' good, @oleksii-morgun! Just a few (hopefully) minor visual tweaks:

* The widths of the input box and the popup menu should match. It looks like the pop up menu is just a tad bigger than it should be.
* The focus ring style on the menu items doesn't match our current, yellow version.
* The list item with the focus ring also looks like it has 1px padding around it, whereas the USWDS version does not have that padding. (Although, I'm not sure if that's because VADS has a different focus ring.)


# Reference:
 
##  YOUR PR:
Screenshot 2024-11-25 at 9 14 04 AM


## USWDS:
  Screenshot 2024-11-25 at 9 14 31 AM

oleksii-morgun:
@danbrady just pushed styling updates to address your findings.

danbrady (approved):

oleksii-morgun:
@rsmithadhoc would you please take another look at the changes I made addressing your feedback?

rsmithadhoc:
> @rsmithadhoc would you please take another look at the changes I made addressing your feedback?

I'll take a look when I get a chance, thanks!

rsmithadhoc (approved):
@oleksii-morgun Accessibility changes look great, thank you!

jamigibbs (approved):
Thank you again for your work on this!

PR #1389, opened 2024-11-05
va-telephone: Allow non-US phone numbers

Andrew565:
## Chromatic

https://3472-allow-non-us-numbers--65a6e2ed2314f7b8f98609d8.chromatic.com

---

## Description
Closes [#3472](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3472)

## QA Checklist
- [X] Accessibility expert has signed off on code changes (if applicable. If not applicable provide reason why)
- [X] Designer has signed off on changes (if applicable. If not applicable provide reason why)

## Screenshots
Screenshot 2024-11-05 at 4 28 11 PM

## Acceptance criteria

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-telephone/va-telephone.tsx
+++ b/packages/web-components/src/components/va-telephone/va-telephone.tsx
@@ -105,16 +105,24 @@ export class VaTelephone {
    * `international` and `extension` args only work on 10 digit contacts
    */
   /* eslint-disable i18next/no-literal-string */
-  static formatPhoneNumber(
-    num: string,
-    extension: string,
-    international: boolean = false,
-    countryCode: string,
-    vanity: string,
-    tty: boolean = false,
-  ): string {
+  static formatPhoneNumber({
+    contact: num,
+    extension,
+    international = false,
+    countryCode,
+    vanity,
+    tty = false,
+  }: {
+    contact: string;
+    extension?: string;
+    international?: boolean;
+    countryCode?: string;
+    vanity?: string;
+    tty?: boolean;
+  }): string {
     const splitNumber = countryCode ? [num] : VaTelephone.splitContact(num);
     let formattedNum = splitNumber.join('');
+
     if (formattedNum.length === 10 && !countryCode) {
       const [area, local, last4] = splitNumber;
       formattedNum = `${area}-${local}-${last4}`;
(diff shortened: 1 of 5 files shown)


ataker (approved):

packages/web-components/src/components/va-telephone/va-telephone.tsx
```diff
@@ -99,31 +104,57 @@ export class VaTelephone {
...
+    tty = false,
+  }: {
+    contact: string;
+    extension?: string;
+    international?: boolean;
+    countryCode?: string;
+    vanity?: string;
+    tty?: boolean;
+  }): string {
+    const splitNumber = countryCode ? [num] : VaTelephone.splitContact(num);
```
jamigibbs, line 123: Should we be validating the International phone number with a regex? I believe this is being done already in the forms library and if this component is supposed to replace that, perhaps we should include that functionality in the web component too.

jamigibbs: Here is the original issue and the linked PR can probably add clarity:
- https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2673

Andrew565: It's an interesting question, because I don't think we do validation anywhere that's not an input. The assumption being that developers would only provide a contact number to the component that they knew to be an accurate phone number. Always dangerous to assume end users always know what they're doing, I know, but I'm not sure what the "error" state would look like if a developer input an invalid phone number. Would we render an error message to the screen or to the console, or possibly not render the component at all?

Andrew565: I think you might still be confusing the tickets, @jamigibbs. The original ticket you linked to (#2673) was for work to be done on the va-text-input component to support international numbers. The regex was included by Matt in the ticket, and is being used by the requesting team via the "pattern" prop. There may be some improvements to documentation that could be made to support teams wishing to use the pattern prop of va-text-input, but if so that's out-of-scope of this ticket (#3472) and this PR.

This PR is for updates to the va-telephone component.

jamigibbs: Ah! I think you're right that I'm getting a disconnect. `va-text-input` itself was never updated for "international support" which is what I thought we agreed to do during the DSC and something different from `va-telephone`. But it looks like we just updated the forms-library which I suppose is fine because that's how we've asked teams to validate before. 👍🏼

jamigibbs (approved):

PR #1400, opened 2024-11-14
PATCH: Remove aria-controls and aria-expanded from `va-accordion` expand all button

1Copenut:
## Commit log
* Removed two ARIA attributes from Expand all button.
* Removed ARIA attributes from va-accordion e2e test.

## Chromatic

https://patch-tpierce-accordionExpandAll--65a6e2ed2314f7b8f98609d8.chromatic.com

### Reasoning for this change
I opted to remove the `aria-expanded` attribute from the `<va-accordion>` expand all button because it was announcing as "Collapse all accordions, expanded" in screen reader testing. This felt contradictory and potentially confusing. I felt the change in button text alone made it very clear that I had expanded 1 to N accordions as a group. 

This change triggered an axe-core error for the `aria-controls` attribute. I did some more digging and determined that JAWS is still the only screen reader adopting `aria-controls`. JAWS uses this attribute to set virtual cursor on the "controlled" content, which here would be multiple things and not especially helpful. I used the following resources to validate this decision:

* [ARIA controls vs ARIA owns](https://www.makethingsaccessible.com/guides/aria-controls-vs-aria-owns/j)
* [Support for aria-controls attribute (aria)](https://a11ysupport.io/tech/aria/aria-controls_attribute)

---
- [x] Link to any related issues in the description so they can be cross-referenced.
- [x] Add the appropriate version patch label (`major`, `minor`, `patch`, or `ignore-for-release`).

## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/96354

## QA Checklist
- [ ] ~~Designer has signed off on changes (if applicable. If not applicable provide reason why)~~
- [ ] ~~New components are covered by e2e tests; updates to existing components are covered by existing test suite~~

## Screenshots

Screenshot 2024-11-14 083728

---
Screenshot 2024-11-14 083901

---
Screenshot 2024-11-14 083813

## Acceptance criteria

## Definition of done
- [ ] ~~Documentation has been updated, if applicable~~


packages/web-components/src/components/va-accordion/va-accordion.tsx
```diff
@@ -195,36 +198,33 @@ export class VaAccordion {
       'usa-accordion': true,
     });
 
-    const accordionItemIDs = [...this.el.children]
-      .filter((el) => el.tagName.toLowerCase() === 'va-accordion-item')
-      .map((el) => el.id);
-
```
1Copenut, line 201: This block was no longer needed after the `aria-controls` attribute was removed. TypeScript was complaining so I opted to remove it.

rsmithadhoc:
Hi @1Copenut, 

This was initially developed the way you have it, using the button text to signal the status of the accordions, but `aria-expanded` was added because JAWS works a little differently from other screen readers. VoiceOver and NVDA announce button text changing, but JAWS does not. Here's my results from testing:
- Current iteration - Toggling back and forth between "Expand all/Collapse all" will announce "expanded/collapsed" each time. But does not re-read the button text when going back to "Expand all +".
- Proposed changes - Toggling back and forth between "Expand all/Collapse all" will **not** announce "expanded/collapsed" each time. And it does not re-read the button text when going back to "Expand all +". So you miss out on hearing the status.

So I don't believe we should remove `aria-expanded` from the Expand/Collapse All button because of that missing button text announcement. I didn't find it too confusing on the other screen readers that read out "Expand all, collapsed" and "Collapse all, expanded".

Can you let me know your thoughts?

Thanks!

1Copenut:
Thank you for testing with JAWS @rsmithadhoc. I'm surprised that JAWS isn't announcing the changed text when you toggle back and forth, but maybe I shouldn't be. :)

I still believe that removing the `aria-expanded` is the right call, but wonder if we could add an `aria-live` region on the container DIV and retest. With a "polite" setting it shouldn't double up in NVDA and VoiceOver, and might smooth out the JAWS experience.

packages/web-components/src/components/va-accordion/va-accordion.e2e.ts
```diff
@@ -14,7 +14,7 @@ describe('va-accordion', () => {
       <va-accordion class="hydrated">
         <mock:shadow-root>
           <div class="usa-accordion">
-            <button aria-controls="" aria-expanded="false" aria-label="expand-all-aria-label" class="va-accordion__button">
+            <button aria-label="expand-all-aria-label" class="va-accordion__button">
```
1Copenut, line 17: /cc @1Copenut 

TODO: Add a `data-cytest="NAME"` attribute here and update the Cypress `expandAccordions()` helper. Could add an optional argument to accept a string value, but definitely need to ensure the helper doesn't break or tests start failing.

rsmithadhoc:
> Thank you for testing with JAWS @rsmithadhoc. I'm surprised that JAWS isn't announcing the changed text when you toggle back and forth, but maybe I shouldn't be. :)
> 
> I still believe that removing the `aria-expanded` is the right call, but wonder if we could add an `aria-live` region on the container DIV and retest. With a "polite" setting it shouldn't double up in NVDA and VoiceOver, and might smooth out the JAWS experience.

I believe `aria-live` was also tested, with a similar result: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1663#issuecomment-1502195858

rsmithadhoc:
> > Thank you for testing with JAWS @rsmithadhoc. I'm surprised that JAWS isn't announcing the changed text when you toggle back and forth, but maybe I shouldn't be. :)
> > I still believe that removing the `aria-expanded` is the right call, but wonder if we could add an `aria-live` region on the container DIV and retest. With a "polite" setting it shouldn't double up in NVDA and VoiceOver, and might smooth out the JAWS experience.
> 
> I believe `aria-live` was also tested, with a similar result: [department-of-veterans-affairs/vets-design-system-documentation#1663 (comment)](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/1663#issuecomment-1502195858)

@1Copenut How do you feel about removing the `aria-controls` for now, but keeping `aria-expanded`? It looks like `aria-controls` can't even see the IDs that it's referencing through the shadow DOM, so it will always trigger an Axe error. Thanks!

rsmithadhoc (approved):
Looks good, thanks!

PR #1424, opened 2024-12-03
New va-alert-sign-in component

Andrew565:
## Chromatic

https://3364-sign-in-alerts--65a6e2ed2314f7b8f98609d8.chromatic.com

---

## Description
Closes [#3364](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3364)

## QA Checklist

## Screenshots
Screenshot 2024-12-03 at 11 25 42 AM
Screenshot 2024-12-03 at 11 26 04 AM
Screenshot 2024-12-03 at 11 26 16 AM
Screenshot 2024-12-03 at 11 26 25 AM
Screenshot 2024-12-03 at 11 26 33 AM

## Acceptance criteria

## Definition of done


packages/web-components/src/components/va-alert-sign-in/va-alert-sign-in.tsx
```diff
@@ -0,0 +1,310 @@
...
+        const innerText = target.shadowRoot.querySelector('a').innerText;
+        const detail = {
+          componentName: 'va-alert-sign-in',
+          action: 'linkClick',
+          details: {
+            clickLabel: innerText,
+            variant: this.variant,
+          },
+        };
+        this.componentLibraryAnalytics.emit(detail);
```
jamigibbs, line 88: Analytics will need to be coordinated with that platform team so we might just want to remove that logic for now and create a follow-up ticket to work with them.

packages/web-components/src/components/va-alert-sign-in/va-alert-sign-in.tsx
```diff
@@ -0,0 +1,310 @@
...
+        </p>
+        <p>
+          <strong>If you don't have a Login.gov or ID.me account,</strong>{' '}
+          create one now. We'll help you verify your identity.
+        </p>
+        <p>
+          <slot name="LoginGovSignInButton"></slot>
+        </p>
+        <p>
+          <slot name="IdMeSignInButton"></slot>
```
jamigibbs, line 263: I'm kind of wondering about the slot names and if it can be simplified. Is there a reason that we couldn't just do like "sign-in-button" and "sign-in-button-additional" for all of the variations?

Or is the purpose of naming the slots like "IdMeSignInButton"  and "LoginGovSignInButton" to give a hint about which button type to use?

packages/web-components/src/components/va-alert-sign-in/va-alert-sign-in.tsx
```diff
@@ -0,0 +1,310 @@
...
+        </p>
+        <p>
+          <va-link
+            href=""
```
jamigibbs, line 267: Should the places that va-link has `href=""` be linking somewhere?

packages/web-components/src/components/va-alert-sign-in/va-alert-sign-in.tsx
```diff
@@ -0,0 +1,310 @@
...
+  styleUrl: 'va-alert-sign-in.scss',
+  shadow: true,
+})
+export class VaAlertSignIn {
+  @Element() el!: any;
+
+  /**
+   * Determines the text content and border/background color.
+   */
+  @Prop() variant: VariantNames = VariantNames.signInRequired;
```
jamigibbs, line 29: The Storybook documentation is rendering the description and default strangely for this prop. Isn't this just accepting a `string` for the variant name?

Screenshot 2024-12-04 at 1 42 27 PM

packages/web-components/src/components/va-alert-sign-in/va-alert-sign-in.tsx
```diff
@@ -0,0 +1,310 @@
...
+  @Prop() noSignInLink?: string;
+
+  /**
+   * Fires when the component has successfully finished rendering for the first
+   * time.
+   */
+  @Event({
+    eventName: 'va-component-did-load',
+    composed: true,
+    bubbles: true,
```
jamigibbs, line 58: I'm not sure I understand why this is needed. Is there some authentication logic that is relying on this? 

Also would it be clearer if the custom event name mentioned that this is the va-alert-sign-in component that loaded? ie `va-alert-sign-in-did-load`

nichia (commented):
Thanks @Andrew565 for the PR, we (the Cartographers team) are following it closely and very eager to start using it. I'm reviewing the PR and want to check if we can add additional Props to enable customization of the component:
- the alert Heading defaults to h2 - some of our static pages needs the heading level to be h3. Can we add a heading level props?
- In additional to the standard Heading text (eg. 'Verify your identity' or 'Sign in'), depending on the apps, we need to append additional service description to the heading eg. 'Sign in to send secure messages' or 'Verify your identity to access more [VA.gov](http://va.gov/) tools and features'. Can we add a service description prop. If provided, append it to the heading?

packages/web-components/src/components/va-alert-sign-in/va-alert-sign-in.tsx
Andrew565: I can comment out the analytics logic I think.

packages/web-components/src/components/va-alert-sign-in/va-alert-sign-in.tsx
Andrew565: Yes, there are at least two different React button components with several different "types" of login methods, so I wanted the slot names to be descriptive of what was expected to go in each slot. Especially for the "either" variant which requires two different sign in buttons.

packages/web-components/src/components/va-alert-sign-in/va-alert-sign-in.tsx
Andrew565: Knew I forgot to fill in something! I'll do my best to figure out where these links should point.

packages/web-components/src/components/va-alert-sign-in/va-alert-sign-in.tsx
Andrew565: Yeah, I see your point about how it's rendering. Yes, it accepts a string, but that string is supposed to be one of the five values only. I'll see if I can reconfigure that to render in Storybook better.

packages/web-components/src/components/va-alert-sign-in/va-alert-sign-in.tsx
Andrew565: This was leftover from the original va-alert code, I'm not sure that it's completely necessary either. I'll remove it.

packages/web-components/src/components/va-alert-sign-in/va-alert-sign-in.tsx
jamigibbs: I think I would prefer just removing code instead of commenting it out. There are plenty of other analytics examples we can reference in other components if we want to re-add it.

packages/web-components/src/components/va-alert-sign-in/va-alert-sign-in.tsx
jamigibbs: We've done something similar for the [va-alert `type` prop](https://design.va.gov/storybook/?path=/docs/components-va-banner--docs) which only accepts certain string values. Just throwing that example out there as maybe a similar [simplified pattern](https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/src/components/va-alert/va-alert.tsx#L159-L163) that could be used.

Andrew565:
@nichia The decision was made to support custom heading levels but not custom heading text at this time. A storybook example has been added showing how to set custom heading levels, with "H2" still the default.

packages/web-components/src/components/va-alert-sign-in/va-alert-sign-in.tsx
```diff
@@ -0,0 +1,312 @@
...
+  //         componentName: 'va-alert-sign-in',
+  //         action: 'linkClick',
+  //         details: {
+  //           clickLabel: innerText,
+  //           variant: this.variant,
+  //         },
+  //       };
+  //       this.componentLibraryAnalytics.emit(detail);
+  //     }
+  //   }
```
jamigibbs, line 85: Can we clean up all of the commented out code?

packages/web-components/src/components/va-alert-sign-in/va-alert-sign-in.tsx
```diff
@@ -0,0 +1,312 @@
...
+
+  /**
+   * Header level for button wrapper. Must be between 1 and 6
+   */
+  @Prop() headingLevel?: number = 2;
+
+  /**
+   * If `true`, doesn't fire the CustomEvent which can be used for analytics tracking.
+   */
+  @Prop() disableAnalytics?: boolean = true;
```
jamigibbs, line 40: This can be removed until analytics is coordinated with the analytics team.

packages/web-components/src/components/va-alert-sign-in/va-alert-sign-in.tsx
```diff
@@ -0,0 +1,312 @@
...
+    return (
+      <Host>
+        <div
+          role={this.el.getAttribute('data-role')}
+          class={classes}
+          aria-label={this.el.getAttribute('data-label')}
+        >
+          <div
+            class="usa-alert__body"
+            // onClick={this.handleAlertBodyClick.bind(this)}
```
jamigibbs, line 297: Can we clean this up too?

packages/web-components/src/components/va-alert-sign-in/va-alert-sign-in.tsx
```diff
@@ -0,0 +1,312 @@
+import {
+  Component,
+  Element,
+  // Event,
+  // EventEmitter,
```
jamigibbs, line 5: Cleanup?

packages/web-components/src/components/va-alert-sign-in/va-alert-sign-in.tsx
```diff
@@ -0,0 +1,312 @@
...
+    /* eslint-enable i18next/no-literal-string */
+
+    const BodyVariants = {
+      [VariantNames.signInEither]: SignInEitherVariant,
+      [VariantNames.signInOptional]: OptionalVariant,
+      [VariantNames.signInRequired]: RequiredVariant,
+      [VariantNames.verifyIdMe]: IdMeVariant,
+      [VariantNames.verifyLoginGov]: LoginGovVariant,
+    };
+    const SignInBody = BodyVariants[variant];
```
jamigibbs, line 286: I'm not sure that we need to make this as complicated as it is but maybe creating Types for props that expect specific string values is something we can explore as a team to do going forward. It might be nice for us to standardize an approach either way.

packages/web-components/src/components/va-alert-sign-in/va-alert-sign-in.tsx
Andrew565: Since we'd need all of this code once analytics is worked out in a week or two, it's my preference to just leave it here.

rsmithadhoc:
I'll review for a11y in https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3385

packages/web-components/src/components/va-alert-sign-in/va-alert-sign-in.tsx
Andrew565: I personally felt this approach was less complicated compared to the way that we handle multiple variants in things like the va-link component. There, we have each variant chosen by whether or not a specific prop is present and/or true, like `active` or `calendar`, and then we present the chosen variant using a series of `if` statements. There's technically nothing that prevents a developer from providing more than one variant prop, which could lead to a race condition. For the va-alert-sign-in component, I chose to more thoughtfully restrict how variants were chosen and thus rendered, ensuring no race conditions. By using the VariantNames enum, we ensure consistency, and provide developers with an easy tool to make sure they are getting precisely what they want.

jamigibbs: > For the va-alert-sign-in component, I chose to more thoughtfully restrict how variants were chosen and thus rendered, ensuring no race conditions

That makes sense. I guess I'm just wondering if this pattern is something that we (design system engineers) should adopt as a standard way of architecting `string` props that need a specific value. It seems like we are doing it different ways in different components. But if we feel this approach is more stable and one that we agree doesn't add too much more maintenance effort, then we should be consistent with it.

packages/web-components/src/components/va-alert-sign-in/va-alert-sign-in.tsx
jamigibbs: @Andrew565 Can we link to that analytics ticket here then? Is it in a sprint yet?

babsdenney:
@Andrew565 This looks good! One thing I'm not sure if we change is the titles of the alerts? This alert doesn't haven't a default alert as it all depends on the use case. Could we change the title to match the [documentation](https://design.va.gov/components/alert/alert-sign-in/)? 
- Required sign-in (verified)
- Optional sign-in (verified)

LWWright7 (commented):
From a design perspective everything looks good to me for the component itself.  

The main thing that jumped out to me is exactly what Barb had said about the alert... 
Screen Shot 2024-12-06 at 1 05 02 PM

and after inspecting, it appears the background color should be orange, but appears dark gray for whatever reason:

Screen Shot 2024-12-06 at 1 08 52 PM

Not sure if this is a bug?

Andrew565:
@babsdenney Storybook requires there to be a "default implementation" of every component or else it doesn't render the page, but I was able to add a duplicate entry for Required Sign In Verified. It won't let me add parentheses.

Screenshot 2024-12-10 at 3 31 07 PM

@LWWright7 The gray background color instead of the orange is due to a glitch in the Formation transition. Apparently, Storybook is still setup to use Formation styles instead of CSS-Library styles, so I'll make a ticket for that.

Screenshot 2024-12-10 at 3 35 39 PM

jamigibbs:
> The gray background color instead of the orange is due to a glitch in the Formation transition. Apparently, Storybook is still setup to use Formation styles instead of CSS-Library styles, so I'll make a ticket for that.

@Andrew565 Just FYI, Ian already wrote a ticket to address Storybook style issues when we upgraded to 7. So it might be a combination of Formation and other custom Storybook styles that changed related to that upgrade: 

- https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2923

jamigibbs (approved):

PR #1426, opened 2024-12-05
va-checkbox: add indeterminate prop

jamigibbs:
## Chromatic

https://2652-va-checkbox-indeterminate--65a6e2ed2314f7b8f98609d8.chromatic.com/?path=/docs/uswds-va-checkbox--docs#indeterminate

## Description

This will add an `indeterminate` prop to `va-checkbox`. When active, the checkbox will display the "mixed" icon in the checkbox which indicates that it's neither checked nor unchecked.

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2652

## QA Checklist

## Screenshots

Screenshot 2024-12-05 at 3 08 15 PM

## Acceptance criteria

## Definition of done


packages/web-components/src/components/va-checkbox/va-checkbox.tsx
```diff
@@ -154,6 +159,26 @@ export class VaCheckbox {
     if (this.enableAnalytics) this.fireAnalyticsEvent();
   };
 
+  /**
+   * For a11y, input.indeterminate must be set with JavaScript, there is no HTML attribute for this.
+   */
+  private updateIndeterminateInput() {
+    const input = this.el.shadowRoot.querySelector('input');
+    if (this.indeterminate && !this.checked) {
+      input.indeterminate = true;
```
jamigibbs, line 168: We are adding this indeterminate attribute manually with JavaScript based on [Ryan's a11y findings here](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3365#issuecomment-2426922846). When I tested with VO, I confirmed that this is what announces "mixed" as the checkbox state when it's active.

packages/web-components/src/components/va-checkbox/va-checkbox.tsx
```diff
@@ -230,14 +256,15 @@ export class VaCheckbox {
             aria-describedby={ariaDescribedbyIds}
             aria-invalid={error ? 'true' : 'false'}
             disabled={disabled}
+            data-indeterminate={indeterminate && !checked}
```
jamigibbs, line 259: The USWDS release notes say:

> when you set `input.indeterminate = true` via JavaScript **or** add the `data-indeterminate` attribute

...but I couldn't get everything working between styles and screenreader support without adding both `indeterminate` using JS and also adding this `data-indeterminate` attribute. Maybe this is a shadow dom limitation?

packages/web-components/src/components/va-checkbox/va-checkbox.tsx
```diff
@@ -230,14 +256,15 @@ export class VaCheckbox {
...
             onChange={this.handleChange}
           />
           <label
             htmlFor="checkbox-element"
             class="usa-checkbox__label"
             part="label"
             role="checkbox"
-            aria-checked={ariaChecked}
+            aria-checked={indeterminate && !checked ? 'mixed' : ariaChecked}
```
jamigibbs, line 267: I'm adding this because otherwise the `aria-checked` state would be either `true` or `false`. It looks like it also accepts the "mixed" value [according to MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-checked). 

But setting `indeterminate` on the input element seems to be what is making the screenreader announcement.

packages/web-components/src/components/va-checkbox/test/va-checkbox.e2e.ts
```diff
@@ -257,4 +257,39 @@ describe('va-checkbox', () => {
     await checkboxLabelEl.click();
     expect(await checkboxEl.getProperty('checked')).toBeTruthy();
   });
+
+  it('sets the data-indeterminate attribute on the input element', async () => {
```
jamigibbs, line 261: I tried to also get a test working that checked if the input element was getting the `.indeterminate` property from being set in JS but I could not get it to work. :/ 

This is kind of a [strange property](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/indeterminate) though so either it's some kind of testing limitation with Stencil or I'm just not writing the test in a way that it can pick up that value.

jamigibbs:
For the Storybook example, I took inspiration from [this resource](https://www.sitelint.com/blog/definitive-guide-to-indeterminate-state-of-a-checkbox#a_practical_example_demonstrating_the_indeterminate_state) but there are different ways to compose a scenario that uses this checkbox state. There are also different ways to style it (like indented options). Are design guidelines for indeterminate checkbox groups something we want to put more thought into?

danbrady (approved):
I've determined this is awesome.

PR #1427, opened 2024-12-05
va-process-list-item: add optional slot for header

powellkerry:
## Chromatic

https://{{head.ref}}--65a6e2ed2314f7b8f98609d8.chromatic.com

## Description
There is currently not an option available to adopt header styles (h1-h6) for the header. There is a way to change the header level but the header styles do not change. 
This change is to add a header slot to give teams the flexibility to use the header element of choice along with the default styling for that element. 
By default, if the header prop is set, the header slot will not render. This is to maintain backwards compatibility.

Closes https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2593

## QA Checklist
- [ x] Accessibility expert has signed off on code changes (if applicable. If not applicable provide reason why)

## Acceptance criteria
- [ x] QA checklist has been completed

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-process-list/va-process-list-item.tsx
+++ b/packages/web-components/src/components/va-process-list/va-process-list-item.tsx
@@ -60,7 +60,7 @@ export class VaProcessListItem {
           <div class="usa-process-list__heading-eyebrow">{statusTextMap[status]}</div>
           : null
         }
-        {header ? <HeaderTag class='usa-process-list__heading'>{header}</HeaderTag> : null}
+        {header ? <HeaderTag class='usa-process-list__heading'>{header}</HeaderTag> : <slot name="header"/>}
         <slot/>
       </Host>
     )
--- a/packages/storybook/stories/va-process-list-uswds.stories.jsx
+++ b/packages/storybook/stories/va-process-list-uswds.stories.jsx
@@ -111,6 +111,51 @@ const HeaderSizeTemplate = (defaultArgs) => {
   );
 };
 
+const HeaderSlotTemplate = (defaultArgs) => {
+  return (
+    <va-process-list>
+      <va-process-list-item>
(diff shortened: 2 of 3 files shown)


packages/web-components/src/components/va-process-list/va-process-list-item.tsx
```diff
@@ -60,7 +60,7 @@ export class VaProcessListItem {
           <div class="usa-process-list__heading-eyebrow">{statusTextMap[status]}</div>
           : null
         }
-        {header ? <HeaderTag class='usa-process-list__heading'>{header}</HeaderTag> : null}
+        {header ? <HeaderTag class='usa-process-list__heading'>{header}</HeaderTag> : <slot name="header"/>}
```
jamigibbs, line 63: This is for sure a valid approach but previously we have just add the `part` attribute for teams to style through the shadow dom. Are there any pros or cons for that approach vs a slot?

jamigibbs: For example:

powellkerry: Good find, I think the example you found will be a better solution with far less freedom to stick an element in the header that should not be there. I've updated the PR

packages/web-components/src/components/va-process-list/va-process-list-item.tsx
```diff
@@ -46,7 +47,7 @@ export class VaProcessListItem {
   render() { 
     const {header, level, checkmark, active, pending, statusText} = this;
     // eslint-disable-next-line i18next/no-literal-string
-    const HeaderTag = `h${level}`;
+    const HeaderLevel = getHeaderLevel(level);
```
jamigibbs, line 50: Nice refactor!

jamigibbs (approved):

PR #1428, opened 2024-12-06
va-table: add scrollable prop

janechodance:
## Chromatic

https://patch-bmt2-va-table-focusable-prop--65a6e2ed2314f7b8f98609d8.chromatic.com

---
## Description
The Your VA Payments view has a table of payments that starts to extend outside the right edge of the viewport when I zoom my browser to 200-250%. This could prevent some users from seeing all of the data without zooming in to a higher resolution where the table adjusts to our responsive layout. Screenshot attached below.
image

Closes [<ticket>](https://github.com/department-of-veterans-affairs/va.gov-team/issues/95134)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/95134

## QA Checklist

## Screenshots

https://github.com/user-attachments/assets/f4f59b6a-c41b-4292-a942-d5cf409f01bb

https://github.com/user-attachments/assets/bec6c149-a7d7-45ba-9998-f690882c47d7

## Acceptance criteria

## Definition of done


packages/web-components/src/components/va-table/va-table-inner/va-table-inner.tsx
```diff
@@ -306,14 +313,14 @@ export class VaTableInner {
...
-    const { tableTitle, tableType, stacked } = this;
+    const { tableTitle, tableType, stacked, isFocusable } = this;
     const classes = classnames({
       'usa-table': true,
       'usa-table--stacked': stacked,
       'usa-table--borderless': tableType === 'borderless',
     });
     return (
-      <div>
+      <div tabIndex={isFocusable ? 0 : undefined}>
```
jamigibbs, line 323: Is adding the `tab-index=0` to this `div` essentially what you're wanting to use to for keyboard tabbing into focus?

I'm going to tag in our a11y specialist Ryan on the PR because I sometimes get confused about when this is needed or not.

jamigibbs (commented):
Hi @janechodance! 👋🏼  Do you mind adding a short description in the PR of the problem is and the solution being submitted here? That will help us understand a little faster instead of trying to work backwards through the linked ticket.

packages/web-components/src/components/va-table/va-table-inner/va-table-inner.tsx
```diff
@@ -60,6 +60,11 @@ export class VaTableInner {
    */
   @Prop() sortable?: boolean = false;
 
+  /**
+   * Does this a table need focus
+   */
+  @Prop() isFocusable?: boolean = false;
```
jamigibbs, line 66: Is it necessary to have this new prop on both `va-table-inner` and `va-table`?

packages/web-components/src/components/va-table/va-table-inner/va-table-inner.tsx
```diff
@@ -60,6 +60,11 @@ export class VaTableInner {
    */
   @Prop() sortable?: boolean = false;
 
+  /**
+   * Does this a table need focus
```
jamigibbs, line 64: > Does this a table need focus

Some grammar issues here.

Screenshot 2024-12-09 at 8 49 25 AM

jamigibbs:
> > Hi @janechodance! 👋🏼 Do you mind adding a short description in the PR of the problem is and the solution being submitted here? That will help us understand a little faster instead of trying to work backwards through the linked ticket.
> 
> I added a description. Let me know if it makes sense. Thank you!

Thanks @janechodance! I left a comment over on the [ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/95134#issuecomment-2528652800) about trying to reproduce the original issue.

packages/storybook/stories/va-table-uswds.stories.jsx
```diff
@@ -370,44 +362,53 @@ const sortData = [
...
+  'columns': sortColumns,
+  'sortable': true,
+};
+
+export const Scrollable = Template.bind(null);
+Scrollable.args = {
+  'table-title': 'This is a scrollable table.',
+  'rows': data,
+  'columns': defaultColumns,
+  'scrollable': true,
```
jamigibbs, line 412: It looks like the [WithPagination Storybook variation](https://github.com/department-of-veterans-affairs/component-library/blob/de72c659f98046d6934cc27da43e65df3cada74c/packages/storybook/stories/va-table-uswds.stories.jsx#L332-L336) needs the `scrollable` prop too because it contains more content. I'm unable to scroll it now on [small screens](https://patch-bmt2-va-table-focusable-prop--65a6e2ed2314f7b8f98609d8.chromatic.com/?path=/story/uswds-va-table--with-pagination) (before it stacks).  Can you add that prop to [that story](https://github.com/department-of-veterans-affairs/component-library/blob/de72c659f98046d6934cc27da43e65df3cada74c/packages/storybook/stories/va-table-uswds.stories.jsx#L332-L336) too?

packages/web-components/src/components/va-table/va-table-inner/va-table-inner.tsx
```diff
@@ -306,14 +313,17 @@ export class VaTableInner {
...
     const classes = classnames({
       'usa-table': true,
       'usa-table--stacked': stacked,
       'usa-table--borderless': tableType === 'borderless',
     });
     return (
-      <div>
+      <div
+        tabIndex={scrollable ? 0 : undefined}
+        class={scrollable ? 'usa-table-container--scrollable' : undefined}
```
jamigibbs, line 325: Instead of `undefined`, can we instead us `null`?  Null would represent the intentional absence of a value which I think better describes how we want to assign a value to these attributes.

jamigibbs (commented):
I have just a few small things that I think would improve these updates but overall it looks great!

packages/storybook/stories/va-table-uswds.stories.jsx
jamigibbs: @janechodance I noticed one more thing from the[ USWDS example](https://designsystem.digital.gov/components/table/#scrollable-table); the default behavior for scrollable tables is to also have `stacked` set to `false`. 

Screenshot 2024-12-11 at 11 34 10 AM

Originally we were thinking that we'd let it stack but I'm going to tag @rsmithadhoc to get his thoughts about if we should add the `stacked: false` prop to our scrollable Storybook example so that it matches USWDS? 🤔 I think I'm fine either way but just looking for a sanity check.

rsmithadhoc: > @janechodance I noticed one more thing from the[ USWDS example](https://designsystem.digital.gov/components/table/#scrollable-table); the default behavior for scrollable tables is to also have `stacked` set to `false`.
> 
> Originally we were thinking that we'd let it stack on small screens but I'm going to tag @rsmithadhoc to get his thoughts about if we should add the `stacked: false` prop to our scrollable Storybook example so that it matches USWDS? 🤔 I think I'm fine either way but just looking for a sanity check.
> 
> Screenshot 2024-12-11 at 11 34 10 AM

@jamigibbs I think I initially said make it stacked by default when we first did the table, just to eliminate any work from teams. However, I'm not too tied to that idea, so I could also go either way. I think the prop would give teams all the options at their disposal, so it might be the most flexible option.

jamigibbs: @rsmithadhoc Thanks Ryan! Yes, I agree it would be good to see the flexibility with the options.

jamigibbs (approved):

PR #1433, opened 2024-12-11
Fix va-table screen reader not announcing sort

ataker:
## Chromatic

https://va-table-sort-screen-reader-fix--65a6e2ed2314f7b8f98609d8.chromatic.com

## Description
Closes [Table sorting - screen reader not announcing on sort #3538](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3538)

Solution I ended up on is just delaying the screen reader text update by 100ms.

Solutions that helped but didn't completely fix the problem:

- Moving screen reader text from `va-table-inner` to `va-table`. `aria=live="polite"` only announces on update, but `va-table-inner` is completely replaced every time a sort occurs, so no announcement occurred. This change helped a little, but consistently no announcement occurred in Safari.
- Only replace the body of the table, not the headers. This required a major rewrite of the sort code and helped, but the announcement still seemed to compete with the button announcement and would be missed.
- Updating screen reader text at different points in the lifecycle. None seems to have a noticeable effect.

## QA Checklist

## Screenshots
Screenshot 2024-12-13 at 11 22 32

## Acceptance criteria

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-table/va-table-inner/va-table-inner.tsx
+++ b/packages/web-components/src/components/va-table/va-table-inner/va-table-inner.tsx
@@ -228,23 +228,6 @@ export class VaTableInner {
   }
 
   // only runs if sortable is true
-  // if a sort has occurred update the sr text to reflect this
-  updateSRtext(
-    thSorted: boolean,
-    content: string,
-    currentSortDirection: string,
-  ) {
-    if (thSorted) {
-      const tableInfo = !!this.tableTitle
-        ? `The table named "${this.tableTitle}"`
-        : 'This table';
-      this.el.shadowRoot.querySelector(
-        'table + div',
-      ).innerHTML = `${tableInfo} is now sorted by ${content} in ${currentSortDirection} order`;
-    }
-  }
-
-  // only runs if sortable is true
   // for a given th, get the current and the next sort directions
   getSortDirections(): {
     currentSortDirection: string;
@@ -298,9 +281,6 @@ export class VaTableInner {
 
           // update title info of span inside th to reflect sort
           this.updateSpan(th, thSorted, nextSortDirection, content);
-
-          // update sr text to reflect sort
-          this.updateSRtext(thSorted, content, currentSortDirection);
         });
     }
   }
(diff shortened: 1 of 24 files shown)


Andrew565 (approved):
Clever solution!

rsmithadhoc:
I'm currently doing testing on this. I think it may work, but I was getting some inconsistency with getting it to read out on Safari. I believe it is due to Safari doing things a bit slower than Chrome. I'll try a few things on my end and test it on Windows as well.

packages/web-components/src/components/va-table/va-table-inner/va-table-inner.tsx
```diff
@@ -319,7 +319,9 @@ export class VaTableInner {
           this.updateSpan(th, thSorted, nextSortDirection, content);
 
           // update sr text to reflect sort
-          this.updateSRtext(thSorted, content, currentSortDirection);
+          setTimeout(()=>{
+            this.updateSRtext(thSorted, content, currentSortDirection);
+          },100)
```
rsmithadhoc, line 324: Update to 500

rsmithadhoc (changes requested):
@ataker The solution looks good, thanks for digging into this one. I'm proposing a slight tweak, but let me know if you see any issues with that.

I tested this with slow/fast voice in VoiceOver and a larger table.
- Voice speed didn't seem to affect anything.
- I found that as the table got larger, the timeout needed to be increased. The original 100ms with the current Storybook example didn't always work. But once I increased that, it worked 100% of the time.
    - 500ms supports 100 rows, so I think that is sufficient for now. It shouldn't impact the user experience, since there are things that need to be read out prior to reading out the SR text, which 500ms should be plenty. I think supporting 100 rows is sufficient, but if not we may need to look at only running `updateSRtext` after sorting has fully completed in a future update.
- I also tested in Chrome/JAWS and Firefox/NVDA and it announced as expected there.

Thanks!

packages/web-components/src/components/va-table/va-table-inner/va-table-inner.tsx
ataker: Thanks Ryan! Updated.

rsmithadhoc (approved):

PR #1437, opened 2024-12-13
va-process-list-item: move status eyebrow into header so that it gets read by screen readers

powellkerry:
## Chromatic

https://98739-proc-list-eyebrow-acc--65a6e2ed2314f7b8f98609d8.chromatic.com

## Description
This PR moves the status eyebrow text inside the header so that it doesn't get skipped by screen readers.

\* moving the eyebrow into the header caused the color of the eyebrow in the current step to be blue. I overrode this so that they are #3d4551

Closes [98739](https://github.com/department-of-veterans-affairs/va.gov-team/issues/98739)

## QA Checklist
- [ x] Designer has signed off on changes (if applicable. If not applicable provide reason why)

## Screenshots
No visual differences:

Before:
Screenshot 2024-12-13 at 10 46 06 AM

After:
Screenshot 2024-12-16 at 10 14 02 AM

## Acceptance criteria

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-process-list/va-process-list-item.tsx
+++ b/packages/web-components/src/components/va-process-list/va-process-list-item.tsx
@@ -59,11 +59,9 @@ export class VaProcessListItem {
       <Host role="listitem" class='usa-process-list__item'>
         {header || status ? (
           // aria-label hack to avoid header being read twice in voiceOver
-          <HeaderLevel part="header" class='usa-process-list__heading' aria-label={`${status ? `${statusTextMap[status]}: ` : ''}${header}`}>
-            <span aria-hidden="true">
-              {status ? <div class="usa-process-list__heading-eyebrow">{statusTextMap[status]}</div> : null}
-              {header ? <div>{header}</div> : null}
-            </span>
+          <HeaderLevel part="header" class='usa-process-list__heading'>
+            {status ? <div class="usa-process-list__heading-eyebrow">{statusTextMap[status]}</div> : null}
+            {header ? <div>{header}</div> : null}
           </HeaderLevel> 
         ): null}
         <slot/>
--- a/packages/web-components/src/components/va-process-list/va-process-list-item.tsx
+++ b/packages/web-components/src/components/va-process-list/va-process-list-item.tsx
@@ -57,11 +57,15 @@ export class VaProcessListItem {
     
     return (
       <Host role="listitem" class='usa-process-list__item'>
-        { status ?
-          <div class="usa-process-list__heading-eyebrow">{statusTextMap[status]}</div>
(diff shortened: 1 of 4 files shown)


packages/web-components/src/components/va-process-list/va-process-list-item.tsx
```diff
@@ -57,11 +57,15 @@ export class VaProcessListItem {
...
     return (
       <Host role="listitem" class='usa-process-list__item'>
-        { status ?
-          <div class="usa-process-list__heading-eyebrow">{statusTextMap[status]}</div>
-          : null
-        }
-        {header ? <HeaderLevel part="header" class='usa-process-list__heading'>{header}</HeaderLevel> : null}
+        {header || status ? (
+          // aria-label hack to avoid header being read twice in voiceOver
+          <HeaderLevel part="header" class='usa-process-list__heading' aria-label={`${status ? `${statusTextMap[status]}: ` : ''}${header}`}>
```
powellkerry, line 62: In VoiceOver when there are multiple elements in the header it will read it as "Header level 3; two items;" and repeat the heading text multiple times. The `aria-label` in the header and `aria-hidden`  on the container fixes this issue, but it seems like there should be a better solution? @rsmithadhoc do you have a suggestion?

Andrew565 (approved):
Development looks good to me!

packages/web-components/src/components/va-process-list/va-process-list-item.tsx
rsmithadhoc: I tested in other screen readers, JAWS didn't recognize the headings at all. I'll take a look and see if there's a better solution.

rsmithadhoc: @powellkerry 
I think the original HTML may be the way to go here.
```
    <h2>
        <span></span>
        <span></span>
    </h2>
```

When reading the page, I see that VoiceOver reads the heading, then repeats the 2 items within the heading, like you said. But when using the command to read the next heading it only reads it once and using the headings menu it only reads it once, as expected.

While not ideal for VoiceOver, we may have to accept it for now. It may be confusing, but I don't think it would hinder the user in navigating the page. JAWS and NVDA only read it once when reading the page. I also tested them reading next heading or using their headings menu and it also functioned as intended with the original HTML.

I think the HTML is fine in the original way, so VoiceOver _should_ be able to handle it. I think it also keeps things much simpler and should work in a wider range of places, as `aria-hidden` can cause some issues, like I saw with JAWS.

Thanks!

danbrady (approved):

rsmithadhoc (approved):

PR #1442, opened 2024-12-19
va-checkbox and va-radio-option: Improved a11y 

Andrew565:
## Chromatic

https://3363-redo-radios-and-checks--65a6e2ed2314f7b8f98609d8.chromatic.com

---

## Description
Closes [#3363](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3363)

Reworks some of the HTML for the va-checkbox and va-radio-option components to be more accessible.

## QA Checklist

## Screenshots
Before:
Screenshot 2024-12-18 at 10 51 13 AM
After:
Screenshot 2024-12-18 at 10 51 25 AM

## Acceptance criteria

## Definition of done


babsdenney:
@Andrew565 Could we fix the box size and radius to align with the USWDS? Attached shows the USWDS with a 20x20 square size and border radius of 2px. 
image

The focus state in the design has a gap between the focus state and the gold outline. Could we follow that design for the focus state. The USWDS also shows a gap. 

**Checkbox in this PR**
image
**VADS**
image
**USWDS**
image

rsmithadhoc (approved):
@Andrew565 This looked good to me. I was able to confirm that this resolves the Axe issues. I also gave it a quick test in VoiceOver and everything sounded as expected.

One minor thing I saw was when initially tabbing to the radio group, there's an additional outline next to the first item. My guess is that a stray pseudo or hidden element might be getting styled.
image

babsdenney:
Similar to the checkboxes, could we increase the size of the radio button to match the USWDS? 

- Increase the size from 16x16 to 20x20
- Create a gap between the focus ring and the radio circle

VADS
image

USWDS
image

Andrew565:
@rsmithadhoc Fixed the ghost outline, thanks for pointing that out.

@babsdenney I fixed the outline gaps and made sure the sizing was consistent for both radios and checkboxes. If you inspect a non-checked checkbox or radio, it will appear in the sizing model as though it's 16x16, but this isn't including the 2px border on all sides, which raises the 'total size' to 20x20. Screenshots attached.

Screenshot 2024-12-20 at 4 18 56 PM
Screenshot 2024-12-20 at 4 25 21 PM
Screenshot 2024-12-20 at 4 25 34 PM

babsdenney:
@Andrew565 That makes sense! I don't know where the sizing strays, I overlayed the USWDS elements with the VADS and you can still see that the VADS radio buttons and checkboxes are slightly smaller. It looks like the selected radio button is accurate and matches the USWDS but the selected checkbox state and unselected radio and checkbox states are smaller than USWDS.  

checkbox and radio

I also am not sure if this makes a difference but the click area for the USWDS checkboxes and radio buttons is larger than the VADS checkboxes. You can click farther to the right on the UWDS elements. 

- [USWDS Checkboxes](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/components-form-inputs-checkbox--default)
- [USWDS Radio buttons](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/components-form-inputs-radio--default)

This doesn't make a difference, so it is not necessary to change, but the example we use for the name is misspelled in the storybook for the [radio buttons](https://3363-redo-radios-and-checks--65a6e2ed2314f7b8f98609d8.chromatic.com/?path=/docs/uswds-va-radio--docs).
image

Andrew565:
@babsdenney Figured out that the USWDS components were 20x20 PLUS a 2px box-shadow, making them effectively 24x24, which is why the size discrepancy. Fixed this for the checkboxes and radio-options, and also fixed the "Soujourner" typo. Should be good for re-review again now.

packages/web-components/src/components/va-radio-option/va-radio-option.tsx
```diff
@@ -72,30 +72,22 @@ export class VaRadioOption {
...
-    const ariaChecked = checked ? 'true' : 'false';
 
-    const inputClass = classnames({
-      'usa-radio__input': true,
-      'usa-radio__input--tile': tile,
+    const containerClass = classnames('usa-radio', {
+      'va-radio-option__container--tile': tile,
     });
     return (
-      <div class="usa-radio">
```
jamigibbs, line 82: It looks like this update will require is to diverge almost completely from the USWDS packages with these classes being removed:
- `usa-radio__input`
- `usa-radio` 

There are similar USWDS classes for `va-checkbox` too. 

Is it possible to keep _some_ USWDS style like padding/margin, spacing, font-size, border-radius and just override those classes in the component stylesheet?

Andrew565: FWIW, the `usa-radio` class is still there, it's just moving to the `containerClass` variable.

As for the rest of the classes that are changing, yes, they are diverging quite a bit from the original USWDS classes, which is why I felt it made sense to replace those classes wholesale rather than try and override them.

For the `usa-radio__input` vs. `va-radio-option__input`, as an example, in our old stylesheet we were overriding just the margin and position, the USWDS doesn't provide any other styles beyond those. In the new stylesheet, we're overriding the margin, position, height, width, border, and box-shadow properties.

In the old styles and layout, the radio button and checkbox "icons" were being generated on the labels using a `::before` pseudo-element. In the new styles and layout, we're generating those "icons" on the inputs themselves.

Getting the 'tile' variant to display correctly likewise required moving  the styles to the `div` container, rather than having them live with the `label` element, due to the `label` not encompassing the `input` element.

So we aren't really getting anything useful from USWDS for these elements any more, so I don't see us needing to receive any upstream updates, unless USWDS completely restructures these components in a way similar to how we've done it.

Andrew565: @jamigibbs Following up on this, did you have any further questions or concerns?

jamigibbs: Sorry I lost track of this one @Andrew565! We were discussing if these changes should be put on hold for the time being for a few different reasons and then I lost track of things. Here is where we're at I believe:

1. The work you did before by adding `role="radio"` and `role="checkbox"` fixed the original Voice Control issue.
2. The ticket [you created for USWDS](https://github.com/uswds/uswds/issues/6068) about the original Voice Control issue has movement and it's possible they could be providing a fix that we could just consume.
3. The ticket this PR is addressing is not a P0 a11y issue (it's not blocking) but a fix requires us to diverge from USWDS. Diverging causes a concerns about not getting any more updates but it also it means we'd have to come back and revert these changes to sync back up with them later.

I think there was a lot of effort put into this work so I definitely don't want to minimize the accomplishment but I'm wondering if it's a good use of our time making these changes now just to come back to sync the components back up with USWDS especially if your original solution allowed Voice Control to work for the time being.

Correct me if I haven't summarized this right though @babsdenney @micahchiang

Andrew565: Thanks Jami. For Micah and Barb, even though this issue isn't a P0, it is causing issues with Axe and causing screen readers to report twice as many options as there are supposed to be (see attached screenshot from Laura). I agree that making a change now and then making a change again once USWDS fixes the issue is less than ideal, but I must disagree that it's not still worth doing. There's no guarantees as to how quickly USWDS will be able to fix this and release it.
Screenshot 2025-01-13 at 10 50 12 AM

jamigibbs: Thanks @Andrew565! I didn't notice that comment before. That problem does seem to have a higher priority than what was mentioned in the ticket description. And it looks like they were able to verify it [in Storybook too with MacOS VoiceOver.](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3363#issuecomment-2587737753) 

So I think at this point I agree with you about the update resolving the miscounting.

jamigibbs: @Andrew565 I had one more hail mary thought; is the original a11y issue a P0 issue? The one that we added the `role="radio"` and `role="checkbox"` attributes for? Could we possibly remove those attributes and then wait for USWDS to make their fix if that problem wasn't a blocker?

Andrew565: The original issue was an accessibility issue, yes. USWDS's original styles take the checkbox and radio inputs and move them `-999px` to the left (i.e. off-screen), which made them unclickable.

babsdenney:
This is looking good Andrew! Thank you! We found some design differences between our radio buttons/checkboxes and the USWDS. 

### Checkboxes:
- The border-radius seems off. It looks like we are using the same border-radius, but the border for our checkboxes looks less rounded. Possibly something to do with the border positioning. 
- The hover-over cursor stays an arrow vs. changing to a hand
- The click area of the USWDS stretches the whole width of the row vs the VADS click area is inline. 
- The focus border width is smaller than USWDS, but this might apply to all of our focus outlines, so this might be a nonissue. 
- The spacing between the checkboxes and radio buttons has more condensed spacing between the rows. I think the USWDS might be using 12px and we use 8px but maybe that is because we want to align with 8px sizing. 

image

### Radio buttons:
These look good! Similar to the checkboxes the spacing between the rows is slightly smaller and the click area is smaller. I think the radio button circles are slightly smaller but not that bad. 

image

Andrew565:
@micahchiang Tested with Verdaccio, everything worked well. The only test that will need to be updated is `src/applications/verify-your-enrollment/tests/e2e/mailingAddress.cypress.spec.js`, and it's a one-liner change from `class="usa-checkbox__label"` to `class="va-checkbox__label"`.

babsdenney:
@Andrew565 This is looking great! Thank you so much for making those design changes. The default options look good now. No issues were found now but I am looking at the tile options more closely, and there are a few things to fix here. 

Thank you for looking into the focus. No worries about that; it will be exciting to get DK's new styles implemented eventually. 

**USWDS Tiles**
image
image

**VADS Tiles**
image
image

- It looks like the tiles have 12px of space around the form element inside the tile, but between the tiles, there are 8px of space. Could we change the gap between the tiles to 8px?
- When the radio or checkbox is selected, the tile's background and border change to a different color. Could we add that?
- The checkbox tile has a place near the border of the tile where the hover state changes, but the tile is not selected when you click on it.

Andrew565:
@babsdenney Improvements made to checkbox and radio option "tile" variant styles.
Screenshot 2025-01-22 at 11 31 37 AM

babsdenney:
@Andrew565 My apologies for going back and forth with this. We were discussing this at the design sync this morning and thought maybe we could take a second look at the checkboxes and styles.  I was able to tweak the styles a bit in devtools to get closer to how the USWDS looks by copying their styles. It's not clear to me where all the existing CSS are coming from and I didn't want to affect anything else, so I mostly overrode existing ones for this component. Here is the checkbox group but if this is helpful I can do this for the radio component as well. Overlaying this checkbox with these styles was almost a complete match the USWDS styles you can see in the image below.

USWDS styles

For the checkbox label: 
I removed all the styles and kept the left padding but moved it to .6rem instead
`.va-checkbox__label { padding-left: 0.6rem; }`

For the checkbox input:
`va-checkbox__input: {
margin-left: 4px;
height: 1.25rem;
width: 1.25rem;
border-radius: 2px;
min-width: auto;
box-shadow: 0 0 0 2px rgb(27, 27, 27);
border: none;
}`

For the focus style you can move the outline-offset to 4px instead.

Andrew565:
@babsdenney Ready for re-review. Updated both the checkboxes and radio-options best I could, let me know if there are any more tweaks you see as necessary.

babsdenney:
@Andrew565 Thank you for making those changes. This is looking really good! A few final things...

Default Checkboxes - 
1. It looks like the focus ring is missing now. 
2. There is a black border when the checkbox is checked and that should change to the blue color.

Radio buttons - 
1. Similar to the checkbox, would you be able to switch to the box-shadow styling for the radio button to align with the USWDS styles? That should make the radio circles look bigger. 

Checkbox Tiles and Radio Button Tiles
1. For the border, it looks like USWDS uses a solid border with a #c9c9c9 color. 
2. For the padding. They are using  1rem on the right and .75rem on the top and bottom.  - padding: 0.75rem 1rem .75rem .5rem;
3. For the checkbox tiles, the focus ring is missing.

Andrew565:
@babsdenney Ready for re-review.

babsdenney:
@Andrew565 Thank you for those updates, the tile options are looking great! 

For the radio buttons, they are still a smidge small. I think if you remove the border and switch to a box shadow instead, that should make them bigger. The red circles are the VADS radio buttons and the black circles are USWDS which shows the size difference. 

image

It looks like there is an error message in the radio button group. Some padding has been added to the top and bottom of the radio button error message, creating extra space between the legend label and the radio group. This is unique to the radio buttons and isn't being applied to the checkboxes. 

<span class="usa-error-message" role="alert"></span>

`.usa-error-message {
    padding-bottom: 0.25rem;
    padding-top: 0.25rem;
    color: rgb(181, 9, 9);
    display: block;
    font-weight: 700;`

babsdenney:
@Andrew565 Just took a look and the radio buttons and checkboxes look almost identical to the USWDS now.  So that is a huge accomplishment I think. Thank you very much for all your work on this! Very much appreciated. This looks good!

PR #1185, opened 2024-06-10
m-action-link: remove typeset ui from icon style, update color-green, fix base font-size

powellkerry:
## Chromatic

https://2538-updates-for-vw-imports--65a6e2ed2314f7b8f98609d8.chromatic.com

## Description
Bug fixes found during QA of vets-website import swap including:
- Changing base font-size from 10px to 16px
- Remove a font declaration from m-action-link that was overriding the icon font
- Updating $color-green to the new approved value

Related to: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2911

## QA Checklist
- [x ] Component maintains 1:1 parity with design mocks
- [x ] Text is consistent with what's been provided in the mocks
- [x ] Component behaves as expected across breakpoints
- [x ] Accessibility expert has signed off on code changes (if applicable. If not applicable provide reason why)
- [x ] Designer has signed off on changes (if applicable. If not applicable provide reason why)
- [x ] Tab order and focus state work as expected
- [x ] Changes have been tested against screen readers (if applicable. If not applicable provide reason why)
- [x ] New components are covered by e2e tests; updates to existing components are covered by existing test suite
- [x ] Changes have been tested in vets-website using Verdaccio (if applicable. If not applicable provide reason why)

## Acceptance criteria
- [ ] QA checklist has been completed

## Definition of done
- [ ] Documentation has been updated, if applicable
- [ ] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)

PR #1444, opened 2024-12-24
css-library: Fixing the html element's font-size so it will respect browser settings

Andrew565:
## Chromatic

https://3646-fix-browser-font-sizing--65a6e2ed2314f7b8f98609d8.chromatic.com

---

## Description
Closes [#3646](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3646)

## QA Checklist
N/A

## Screenshots
N/A

## Acceptance criteria

## Definition of done


packages/css-library/src/stylesheets/base/va.scss
```diff
@@ -11,7 +11,7 @@ body {
 }
 
 html {
-  font-size: $em-base;
+  font-size: 100%; // Needs to be 100% to be able to grow and shrink with browser settings
```
jamigibbs, line 14: This [was changed to 16px](https://github.com/department-of-veterans-affairs/component-library/pull/1185/files) because of root font size updates related to USWDS v3 (it previously being 10px). Was this tested using Verdaccio to make sure everything still looks correct by default? 

Screenshot 2024-12-26 at 8 59 08 AM

jamigibbs (changes requested):

packages/css-library/src/stylesheets/base/va.scss
micahchiang: I agree here with Jami. If this hasn't been tested with Verdaccio, then it needs to be. The entire typography change initiative we completed last summer was centered around this.

jamigibbs: @Andrew565 I was thinking about this more and _think_ the `100%` font-size will still work because the [default font-size for most browsers](https://stackoverflow.com/questions/29511983/is-the-default-font-size-of-every-browser-16px-why) should already be `16px` without us needing to do anything special. That 100% value would then be based on that default. A base value of `1rem` might have the same effect too. A sanity check with Verdaccio or a release candidate would ease our minds though.

Andrew565: Working on testing this with Verdaccio.

Andrew565: Looks good with Verdaccio. Seen here is the "Very Large" font setting. The normal/medium font setting looks the same as it always does.
Screenshot 2025-01-02 at 1 02 38 PM

jamigibbs (approved):
Thanks for doing more testing on this! I think as long as a browser is defaulting to 16px (which most or all do?) this should work as expected.

My one last curiosity is if it should be `100%` or `1rem`. They both technically seem to do the same thing but I'm not sure if one would be preferred for some other cross-browser support situations that I'm not realizing.

The [USWDS documentation site](https://designsystem.digital.gov/) uses `1rem` for example.

Lastly, the variable `$em-base` now seems to be unused. Maybe there is a use for it somewhere but I just want to make note that it seems like cruft at this point.

PR #1446, opened 2025-01-07
va-button: add loading variation

powellkerry:
## Chromatic

https://3460-button-loading--65a6e2ed2314f7b8f98609d8.chromatic.com

## Description
Add loading variation for button.
The button should:
- act disabled (greyed out and disable all interactions)
- display an animated loading icon, the icon should not be animated for reduced motion users
- text should default to "Loading..." if the `text` prop is not set
- Screen readers should announce "loading" when the `loading` prop is true and "Loading complete" when it is set to false

Closes [3460](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3460)

## QA Checklist

## Screenshots
Screenshot 2025-01-07 at 9 01 21 AM

## Acceptance criteria

## Definition of done


packages/web-components/src/components/va-button/va-button.tsx
```diff
@@ -185,18 +205,27 @@ export class VaButton {
...
-          aria-label={label}
-          aria-describedby={ariaDescribedbyIds}
-          type={type}
-          part="button"
-        >
-          {back && !_continue && <va-icon icon="navigate_far_before" />}
-          {buttonText}
-          {_continue && !back && <va-icon icon="navigate_far_next" />}
-        </button>
+          {/* This span must always be present for changes to be announced for the loading prop. It will not show visually or be read without content*/}
```
powellkerry, line 208: I don't love the idea of this span being rendered for all buttons. I have tested it and it does not announce anything when there is not content, but it would be nice to figure out a way to only render it if the button will be used for loading. Hoping someone has an idea for how to do that.

danbrady (commented):
Design looks good. That animation is 🔥!

packages/web-components/src/components/va-button/va-button.tsx
```diff
@@ -50,6 +53,22 @@ export class VaButton {
    */
   @Prop({ reflect: true }) disabled?: boolean = false;
 
+  /**
+   * If `true`, the click event will not fire.
```
jamigibbs, line 57: Is this description accurate for the loading prop?

packages/web-components/src/components/va-button/va-button.scss
```diff
@@ -7,7 +7,37 @@
...
+
+  .loading-icon {
+    animation: spin 1.5s linear infinite;
+  }
+
+  .loading-icon.chromatic:after {
+    animation: none;
+  }
+
+  @media (prefers-reduced-motion) {
```
jamigibbs, line 25: This is a cool media feature. TIL!

packages/web-components/src/components/va-button/va-button.tsx
jamigibbs: I don't think it's a big deal to have a span always in the markup but was there some limitation with just doing conditional JSX? Something like:

```jsx
{this.loading && (
    <span class="loading-message" role="status">....</span>
)}
```

packages/web-components/src/components/va-button/va-button.tsx
```diff
@@ -50,6 +53,22 @@ export class VaButton {
...
+   */
+  @Prop({ reflect: true }) loading?: boolean = false;
+
+  @Watch('loading')
+  announceLoadingChange(newValue: boolean, oldValue: boolean) {
+    if (oldValue && !newValue) {
+      var me = this;
```
jamigibbs, line 64: Should this variable assignment be `let`?

packages/web-components/src/components/va-button/va-button.tsx
```diff
@@ -50,6 +53,22 @@ export class VaButton {
...
+      this.showCompletedMessage = true;
+      setTimeout(() => {
+        me.showCompletedMessage = false;
+      }, 3000);
```
jamigibbs, line 68: I'm not sure I understand the need for a timeout and why this specific time is used. Do you mind clarifying?

jamigibbs (approved):
Nothing blocking that I could see but I had just a few questions and comments. The animation sure is slick. 😎

packages/web-components/src/components/va-button/va-button.tsx
powellkerry: Yes, this is handled the same way as the disabled prop by `preventDefault()` and `preventPropogation()` on line 168.

packages/web-components/src/components/va-button/va-button.tsx
powellkerry: The `<span role="status">` will only be read by the screen reader if there are changes to the content of the span. This is the reasoning for the span always being rendered, even if the loading prop is false, because if we render it when the prop changes the reader will not count that as a change and "Loading" will never be read.

The same applies to this timeout. If there is not any time between when the completion message is set and when it is set to null, the reader won't detect the change and the message won't be read. Its an arbitrary number to guarantee the "Loading complete" is read.

jamigibbs: That makes sense. Thanks for the explanation!

packages/web-components/src/components/va-button/va-button.tsx
jamigibbs: oh ok yes that makes sense. I was expecting it to mention that it will display the loading indicator too.

powellkerry: oh, I should add that! Thanks!

packages/web-components/src/components/va-button/va-button.tsx
```diff
@@ -185,18 +205,27 @@ export class VaButton {
...
+          {/* This span must always be present for changes to be announced for the loading prop. It will not show visually or be read without content*/}
+          <span class="loading-message" role="status">
+            {this.loading ? 'Loading' : this.showCompletedMessage ? 'Loading complete' : null}
+          </span>
+          <button
+            class={buttonClass}
+            aria-disabled={ariaDisabled}
+            aria-busy={loading ? 'true' : undefined}
+            aria-label={label}
+            aria-live={loading ? 'polite' : 'off'}
```
rsmithadhoc, line 217: I don't believe we'll need this `aria-live` on the button itself, as the `span` will handle the live region. Leaving it in doesn't seem to cause any negative effects, but could lead to repetitive announcements. I tested this code without it and it announced as expected.

rsmithadhoc (approved):
This looked good in my testing, screen readers announced the loading state/complete as expected. Thanks!

Andrew565 (approved):

PR #1500, opened 2025-02-19
va-file-input: add percent upload support

it-harrison:
## Chromatic

https://prog-bar-file-input--65a6e2ed2314f7b8f98609d8.chromatic.com

---

## Description
Closes [3864](https://github.com/orgs/department-of-veterans-affairs/projects/768/views/17?pane=issue&itemId=99582576&issue=department-of-veterans-affairs%7Cvets-design-system-documentation%7C3864)

## QA Checklist

## Screenshots

https://github.com/user-attachments/assets/2c504c39-d661-44d4-a51c-30d03a8a7031

## Acceptance criteria

## Definition of done


packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -32,12 +33,6 @@ export class VaFileInput {
   private fileInputRef!: HTMLInputElement;
   private uploadStatus: 'idle' | 'success' = 'idle';
   private fileType?: string;
-  private defaultUploadMessage: HTMLElement = (
-    <span>
-      Drag a file here or{' '}
-      <span class="file-input-choose-text">choose from folder</span>
-    </span>
-  )
```
it-harrison, line 40: moved this variable into a function because when calling `forceUpdate` after reaching the end of upload Stencil was throwing errors relating to caching the JSX.

it-harrison:
@danbrady design question: I am using the `va-button-icon` for the "cancel" upload button...but `va-button-icon` doesn't work with the design because it appears to have extra margin on the left unless you hover. To match the design we would have to make an imposter component. 
Screenshot 2025-03-04 at 4 31 07 PM

danbrady:
@it-harrison Yeah, that's a bit of a quirk with the icon buttons. They have the same padding as our other buttons, but because they don't have a background in the default state they can look misaligned. However, it may be a bit hard to see but we compensated for that in the designs by reducing the left/right margin for the container of the buttons (or alternatively you could increase the left margin for the other elements):
image

BTW, the uploading progress looks great! When we did the designs, we weren't sure if the file weight would be known until the file was fully uploaded. So instead of showing an empty file weight, we used an italicized "Uploading..." label. I sort of like showing that label to describe what the progress bar is doing. (Also, I wonder if there is an accessibility reason to show it as well?) 

image
(Also showing the button container in this screenshot.)

it-harrison:
@danbrady after an update we have: 

https://github.com/user-attachments/assets/60d40a81-97c9-45d3-8955-a3cb6e54e0a7

danbrady:
LGTM! Nice work!

danbrady (approved):

packages/web-components/src/components/va-file-input/va-file-input.scss
```diff
@@ -194,6 +194,16 @@
   .thumbnail-error {
     color: $vads-color-error-dark;
   }
+
+  va-progress-bar {
+    font-size: var(--v3-font-base-size);
```
jamigibbs, line 199: I realize this is the only token we have right now that's for `16px` but it's such an awkward name. We created it temporarily when we were updating the base font sizes (before we started semantic naming). We should probably remove it from the CSS-Library now.

Are we sure this shouldn't be the `--vads-font-size-source-sans-normalized: 1.06rem;` token? This would make it `16.96px` though. 🤔 

https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/css-library/dist/tokens/css/variables.css#L151

There is also `--vads-font-size-root: 1rem;` (which should calculate to 16px).

danbrady: Ooh, good catch! I think all text should be normalized, so the value of this style should render as 16.96px. However, what text is this `font-size` property styling? I see it's on the `va-progress-bar` selector which is progress bar itself and has no visual text within it, no? Maybe this can be removed completely?

it-harrison: @danbrady @jamigibbs the `font-size` controls the height of the progress bar because it is in `em` units...and I think we want it to be 16px? So maybe use `--vads-font-size-root` ?

danbrady: Oh, ok! Yeah, let's keep it 16px then. Our designs aim to follow multiples of 8s for the most part. So that falls in line.

PR #1502, opened 2025-02-21
Adjust va-language-toggle font size and padding

ataker:
## Chromatic

https://lang-toggle-adjust-size--65a6e2ed2314f7b8f98609d8.chromatic.com

## Description
Follow up to [Language toggle accessibility updates #3564](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3564)

Revert font size to baseline. Increase padding to make touch target larger

## QA Checklist

## Screenshots
Screenshot 2025-02-21 at 10 27 47

## Acceptance criteria

## Definition of done

### DIFF
--- a/packages/web-components/src/components/va-language-toggle/va-language-toggle.scss
+++ b/packages/web-components/src/components/va-language-toggle/va-language-toggle.scss
@@ -11,7 +11,7 @@
     color: var(--vads-color-black);
     text-decoration: none;
     padding-top: 10px;
-    padding-bottom: 12px;
+    padding-bottom: 16px;
     &:focus {
       outline: 2px solid var(--vads-color-action-focus-on-light);
       outline-offset: 2px;
@@ -21,7 +21,7 @@
     &::part(anchor) {
       text-decoration: none;
       padding-top: 10px;
-      padding-bottom: 12px;
+      padding-bottom: 16px;
     }
     color: var(--vads-color-link);
     padding-bottom: 2px;
--- a/packages/web-components/src/components/va-language-toggle/va-language-toggle.scss
+++ b/packages/web-components/src/components/va-language-toggle/va-language-toggle.scss
@@ -1,6 +1,6 @@
 :host div {
   display: inline-block;
-  font-size: 1.2em;
+
   div.inner-div {
     border-right: 1px solid var(--vads-color-base-light);
     padding-right: 8px;
@@ -10,6 +10,8 @@
     font-weight: var(--font-weight-bold);
     color: var(--vads-color-black);
     text-decoration: none;
+    padding-top: 10px;
+    padding-bottom: 12px;
     &:focus {
       outline: 2px solid var(--vads-color-action-focus-on-light);
       outline-offset: 2px;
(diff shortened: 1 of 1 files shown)


jamigibbs (commented):
Do we want the hover state background to be that large still? It's probably possible to make the gray hover the same size as the link like it is normally but still have the touch/focus area larger?

Screenshot 2025-02-21 at 10 05 56 AM

Screenshot 2025-02-21 at 10 09 46 AM

Screenshot 2025-02-21 at 10 10 30 AM

mnahmias:
Why are these padding rules not using the vads spacing vars?

ataker:
> Do we want the hover state background to be that large still? It's probably possible to make the gray hover the same size as the link like it is normally but still have the touch/focus area larger?

Both the touch area and gray hover background are based on the padding, so I don't think we can have different sizes on those.

I did intentionally push these as large as (arguably) reasonable and I am ready to pull back.

LWWright7:
> Do we want the hover state background to be that large still? It's probably possible to make the gray hover the same size as the link like it is normally but still have the touch/focus area larger?
> 
> Screenshot 2025-02-21 at 10 05 56 AM
> 
> Screenshot 2025-02-21 at 10 09 46 AM
> 
> Screenshot 2025-02-21 at 10 10 30 AM

This is a good call-out Jami.  And honestly, very timely given what Barb presented on the focus state stuff this morning.  I think for accessibility sake that we should err on the side of caution here and use the larger focus and hover states here like on the top and bottom in your screenshots above.  

Maybe @rsmithadhoc could add his 2 cents on the accessibility perspective?

Also... @babsdenney & @danbrady may have a different opinion, so by all means, please chime in as well.

danbrady:
Yeah, I agree with @jamigibbs's approach if it's possible.

rsmithadhoc:
> > Do we want the hover state background to be that large still? It's probably possible to make the gray hover the same size as the link like it is normally but still have the touch/focus area larger?
> > Screenshot 2025-02-21 at 10 05 56 AM
> > Screenshot 2025-02-21 at 10 09 46 AM
> > Screenshot 2025-02-21 at 10 10 30 AM
> 
> This is a good call-out @jamigibbs . And honestly, very timely given what Barb presented on the focus state stuff this morning. I think for accessibility sake that we should err on the side of caution here and use the larger focus and hover states here like on the top and bottom in your screenshots above.
> 
> Maybe @rsmithadhoc could add his 2 cents on the accessibility perspective?
> 
> Also... @babsdenney & @danbrady may have a different opinion, so by all means, please chime in as well.

I believe the hover and focus states should match the size of the element. In this case, we are kind of treating a link like a button. With a link, it would just be the text getting the hover/focus state, like Jami is suggesting. But with a button, it would be the entire element, like it is now. I don't know that there is a specific rule regarding this, but typically that is just how `:hover` in CSS works, the clickable part of the element has those hover styles applied to it. If the click area is expanded, I would expect the hover state to encompass that whole area. So I'm on the side of the larger hover/focus, but I don't feel too strongly about it.

ataker:
Seems like we're okay with the larger hover area as a result of the larger tap target.

PR #1523, opened 2025-03-11
va-checkbox: swap out legend element 

it-harrison:
## Chromatic

https://3822-checkbox-update--65a6e2ed2314f7b8f98609d8.chromatic.com

---

## Description
Background:
1.  if the `description` prop is not `null`, `va-checkbox` renders a `legend` element with the prop text. 
2.  `va-checkbox` will only render content in the `slot="description"` if the `description` prop is `null`.

This PR swaps the `legend` referenced in point 1 above for a `div`, to avoid having multiple `legend`s in a `fieldset` when (if) multiple `va-checkbox` components with the `description` property set are used in `va-checkbox-group`. 

@Mottie asked why point 2 above is the case. I'm not sure why the component behaves as it does. Whatever is passed to `description` could also be passed to `slot="description"` Therefore, the `description` prop and does not need to exist. A quick search of `vets-website` showed the `description` prop is _not_ being used (but the slot is).

Pending team input, I'd say we should consider getting rid of the `description` prop.

Closes [3822](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3822)

## QA Checklist

## Acceptance criteria

## Definition of done


packages/web-components/src/components/va-checkbox/va-checkbox.tsx
```diff
@@ -219,7 +219,7 @@ export class VaCheckbox {
       'va-checkbox__container--tile--checked': tile && checked,
     });
     const descriptionClass = classnames({
-      'usa-legend': true,
+      'vads-u-margin-top--1p5': true,
```
it-harrison, line 222: I'm not sure if this is the correct margin.

powellkerry (approved):
LGTM

packages/web-components/src/components/va-checkbox/va-checkbox.tsx
jamigibbs: I think that's a good guess (I don't see anything in [Figma](https://www.figma.com/design/afurtw4iqQe6y4gXfNfkkk/VADS-Component-Library?node-id=137-17116&p=f&t=UjYhARqFoBAeg7Nb-0) to validate) but regardless of the margin size used, I'm wondering if we even want to use vads utility classes like this? Or should the style be added directly to the component's stylesheet? 

My instinct is that we should maintain component style encapsulation as much as possible. I know we've tried in the past to keep Formation styles outside of components for that reason and this feels like the same thing.

Screenshot 2025-03-12 at 10 08 46 AM

Mottie:
I just noticed this incorrect storybook title: "Title With Custom Content" should be "Tile With Custom Content", please fix it for me 🙏🏽

packages/web-components/src/components/va-checkbox/va-checkbox.tsx
it-harrison: @jamigibbs - good call!

jamigibbs (approved):
> Pending team input, I'd say we should consider getting rid of the description prop.

It does seem like a redundancy now. I think the initial build of the component intentionally wanted to enforce only strings in the description. But that has apparently evolved to allow for any markup (thus the slot).

Even though we aren't finding any implementations of the prop being used, it would still be a major API change and maybe we want to dig into the history a little more. Should we create a separate ticket to look at removing it?

PR #1526, opened 2025-03-12
Storybook: fixing Storybook web component types

Andrew565:
## Chromatic

https://9999-fixing-storybook-types--65a6e2ed2314f7b8f98609d8.chromatic.com

---
## Description
Closes [#3899](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/3899)

Adds a types definition file for Storybook so that we can get code completion and far fewer "red squigglies" TypeError errors while coding Storybook stories.

## Further Details and Notes

Hey, sorry in advance for all of the whitespace-only formatting changes! I wish I knew how to make my linter not do that so much!

~If you want to pull down these changes and test them locally, you will need to go into the `core` package and do a `yarn build` in order for the type changes I made in the components to reflect locally correctly. This is because the types.d.ts file in the storybook directory depends on the type definitions that come from core's dist folder.~

No longer requires building core to test, changed the types.d.ts "source of truth" to be the web-components/dist/components.d.ts instead, so just requires re-building web-components.

I've left comments on a majority of the things I tweaked, but feel free to reach out to me here or on Slack if something doesn't make sense!

## QA Checklist
- [X] Accessibility expert has signed off on code changes (if applicable. If not applicable provide reason why)
- [X] Designer has signed off on changes (if applicable. If not applicable provide reason why)

## Screenshots
N/A

## Acceptance criteria

## Definition of done
- [X] A link has been provided to the originating GitHub issue


packages/storybook/stories/va-back-to-top.stories.jsx
```diff
@@ -25,7 +25,7 @@ const Template = ({
       <div className="usa-grid usa-grid-full">
         <div className="usa-width-three-fourths">
           <h1>The top</h1>
-          {[...Array(displayAmount)].map(i => (
+          {[...Array(displayAmount)].map((_, i) => (
```
Andrew565, line 28: Fix for "keys must be unique error". `[...Array(displayAmount)]` generates an empty array where each 'value' of the array is "undefined", so this fix skips the value of the array and instead uses the array's index as the key.

packages/storybook/stories/va-checkbox-group-uswds.stories.jsx
```diff
@@ -316,6 +316,7 @@ const FormsPatternSingleTemplate = ({
         form-heading-level={1}
         form-heading="What's your relationship to the Veteran?"
         form-description="This is the additional form-description prop"
+        label={label}
```
Andrew565, line 319: `label` is a required prop for va-checkbox-group so added it here.

packages/storybook/stories/va-checkbox-uswds.stories.jsx
```diff
@@ -107,12 +109,12 @@ const IndeterminateTemplate = ({}) => {
   }, [checked]);
 
   const handleIndeterminate = () => {
-    const indeterminateCheckbox = document.querySelector(
-      '.indeterminate-checkbox',
+    const indeterminateCheckbox = /** @type {HTMLInputElement} */ (
```
Andrew565, line 112: Adding JSDOC type declaration to fix "element has no `checked` prop" error.

packages/storybook/stories/va-checkbox-uswds.stories.jsx
```diff
@@ -139,7 +141,9 @@ const IndeterminateTemplate = ({}) => {
   };
 
   const handleSelectAllToggle = event => {
-    const checkboxes = document.querySelectorAll('.example-checkbox');
+    const checkboxes = /** @type {HTMLInputElement[]} */ (
+      /** @type {unknown} */ (document.querySelectorAll('.example-checkbox'))
```
Andrew565, line 145: Adding JSDOC type declaration to make "no `checked` prop" error go away. Due to a quirk in typing logic, has to first be cast as `unknown` before it can be cast as `HTMLInputElement[]`, but trust me I've done this before, tested it and it works. :)

packages/storybook/stories/va-date.stories.jsx
```diff
@@ -61,7 +61,7 @@ const Template = ({
 const CustomValidationTemplate = ({ label, name, required, error, value }) => {
   const [dateVal, setDateVal] = useState(value);
   const [errorVal, setErrorVal] = useState(error);
-  const [invalidYear, setInvalidYear] = useState(null);
+  const [invalidYear, setInvalidYear] = useState(false);
```
Andrew565, line 64: `invalidYear` can be a boolean value or undefined, but not `null`.

packages/storybook/stories/va-file-input-multiple-uswds.stories.jsx
```diff
@@ -91,9 +91,13 @@ HeaderSize.args = {
 
 const additionalFormInputsContent = (
   <div>
-    <va-select label='What kind of file is this?' required>
-      <option key="1" value="1">Public Document</option>
-      <option key="2" value="2">Private Document</option>
+    <va-select label="What kind of file is this?" name="fileType" required>
```
Andrew565, line 94: `name` is a required prop for va-select so adding it here.

packages/storybook/stories/va-file-input-uswds.stories.jsx
```diff
@@ -170,7 +175,7 @@ const CustomValidationTemplate = ({
 
       const reader = new FileReader();
       reader.onloadend = () => {
-        const contents = reader.result;
+        const contents = /** @type {string} */ (reader.result);
```
Andrew565, line 178: Adding type declaration because `result` could be either a string or an ArrayBuffer and here we want it to always be treated like a string so we can utilize the `includes('X')` on line 179.

packages/storybook/stories/va-file-input-uswds.stories.jsx
```diff
@@ -338,17 +350,21 @@ const PercentUploadedTemplate = args => {
       setPercent(_prev => {
         if (_prev >= 100) {
           clearInterval(intervalId);
-          return null;
+          return 100;
```
Andrew565, line 353: Changing from `null` to `100` because the type checker expects this to always be a number. Felt 100 was safe since this is the percentage uploaded checker and we stop checking once it hits 100% or greater.

packages/storybook/stories/va-memorable-date-uswds.stories.jsx
```diff
@@ -70,7 +77,7 @@ const CustomValidationTemplate = ({ label, name, hint, required, error, value })
         hint={hint}
         required={required}
         error={errorVal}
-        invalidYear={!!errorVal || null}
+        invalidYear={!!errorVal || undefined}
```
Andrew565, line 80: `invalidYear` expects to be a boolean or undefined, not null.

packages/storybook/stories/va-notification.stories.jsx
```diff
@@ -81,7 +81,7 @@ const MultipleTemplate = ({
   return (
     <>
       <va-notification
-        visible="true"
+        visible={true}
```
Andrew565, line 84: Fix for expecting a boolean true rather than a string.

packages/storybook/stories/va-official-gov-banner.stories.jsx
```diff
@@ -19,34 +19,32 @@ const defaultArgs = {
...
-    document.querySelector('main').setAttribute('lang', lang);
+    document.querySelector('main')?.setAttribute('lang', lang);
   }, [lang]);
   return (
     <>
-      <button onClick={e => setLang('es')}>Español</button>
-      <button onClick={e => setLang('en')}>English</button>
-      <div style={{marginTop: '20px'}}>
-        <va-official-gov-banner 
+      <button onClick={_ => setLang('es')}>Español</button>
```
Andrew565, line 41: Fix for "e is defined but never used" warning.

packages/storybook/stories/va-process-list-uswds.stories.jsx
```diff
@@ -63,68 +63,96 @@ const Template = ({}) => {
...
+const HeaderSizeTemplate = defaultArgs => {
   return (
     <va-process-list>
-      <va-process-list-item header='Size h1' level='1'/>
-      <va-process-list-item header='Size h2' level='2'/>
-      <va-process-list-item header='Size h3' level='3'/>
-      <va-process-list-item header='Size h4' level='4'/>
-      <va-process-list-item header='Size h5' level='5'/>
-      <va-process-list-item header='Size h6' level='6'/>
+      <va-process-list-item header="Size h1" level={1} />
```
Andrew565, line 127: Fix for `level` expecting a number rather than a string.

packages/storybook/stories/va-text-input-uswds.stories.jsx
```diff
@@ -120,7 +120,12 @@ const Template = ({
       success={success}
       pattern={pattern}
       onBlur={e => console.log('blur event', e)}
-      onInput={e => console.log('input event value', e.target.value)}
+      onInput={e =>
+        console.log(
+          'input event value',
+          /** @type {HTMLInputElement} */ (e.target).value,
```
Andrew565, line 126: Adding type declaration for `element has no value prop` error.

packages/storybook/stories/wc-helpers.jsx
```diff
@@ -248,6 +248,7 @@ function CanvasLink() {
 /**
  * Return a component with Storybook docs blocks in a standard order.
  * Accepts a JSON object as a prop representing component information
+ * @param {{storyDefault: any, componentName?: string, data: any, children?: any}} arg
```
Andrew565, line 251: Adding type declaration because StoryDocs was throwing a type error about not having a `children` prop, so added it and made it optional and set to "any" as a cheap and easy way to remove this particular error.

packages/storybook/types.d.ts
```diff
@@ -0,0 +1,58 @@
+import { HTMLAttributes, Key } from 'react';
+import { Components as WCTypes } from '@department-of-veterans-affairs/component-library/dist/components/types';
+
+// Applies the component's type plus HTMLAttributes<HTMLElement> from React in order to satisfy the JSX "children" prop
+
+type IsElement = HTMLAttributes<HTMLElement>;
```
Andrew565, line 6: This `IsElement` type adds standard HTMLElement attributes like id and className and such to the Web Components.

packages/web-components/src/components.d.ts
```diff
@@ -551,7 +551,7 @@ export namespace Components {
         /**
           * Whether or not an analytics event will be fired.
          */
-        "enableAnalytics": boolean;
+        "enableAnalytics"?: boolean;
```
Andrew565, line 554: Majority of the Type Definition fixes seen in this file are because of changing "required" props into "optional" props. They all have defaults assigned to them in their respective component files, so making them optional doesn't change any base functionality.

micahchiang (commented):
In general I think these are good changes. Do you know how much additional configuration we would need to do to convert these story files to `.tsx` files so we can use inline typescript type declarations? The inline JSDoc doesn't feel great to me, but isn't necessarily a deal breaker.

Andrew565:
> In general I think these are good changes. Do you know how much additional configuration we would need to do to convert these story files to `.tsx` files so we can use inline typescript type declarations? The inline JSDoc doesn't feel great to me, but isn't necessarily a deal breaker.

I'm not positive but I think the majority of the 'work' would be just in renaming the files from `.jsx` to `.tsx`. I'll give it a try and see how it goes.

Andrew565:
@micahchiang I renamed the affected files from `.jsx` to `.tsx` and got it working (with a little extra configuration). The storybook files now work fine with either extension. Should I just finish converting the rest you think, or just leave them as-is until a time comes where a conversion is necessary?

packages/storybook/stories/va-checkbox-uswds.stories.tsx
```diff
@@ -139,7 +139,9 @@ const IndeterminateTemplate = ({}) => {
   };
 
   const handleSelectAllToggle = event => {
-    const checkboxes = document.querySelectorAll('.example-checkbox');
+    const checkboxes = document.querySelectorAll(
+      '.example-checkbox',
+    ) as unknown as HTMLInputElement[];
```
micahchiang, line 144: To avoid the double cast and `unknown`, can this be refactored to something like: 

```
const checkboxes = Array.from(document.querySelectorAll<HTMLInputElement>('.example-checkbox'))
```

packages/storybook/stories/va-file-input-multiple-uswds.stories.tsx
```diff
@@ -321,6 +343,7 @@ const CustomValidationTemplate = ({ label, name, accept, hint }) => {
       }
     });
 
+    // @ts-ignore
```
micahchiang, line 346: Could you add what rule this is ignoring just so there's a little more clarity?

it-harrison (approved):
looks good! very clean solution

packages/storybook/stories/va-file-input-uswds.stories.tsx
```diff
@@ -170,7 +174,7 @@ const CustomValidationTemplate = ({
 
       const reader = new FileReader();
       reader.onloadend = () => {
-        const contents = reader.result;
+        const contents = reader.result as String;
```
micahchiang, line 177: I think since we're always reading the result here on load end, it'll likely always be a string. So instead of casting it to a string we could probably do something like `const contents: String` instead. If we're not 100% confident the result will always be a string, then we could do something like `const contents: String = result.readAsText(file)`

micahchiang (approved):
Left some potential suggestions, but none of them are really blockers. Thanks for putting this together!

packages/storybook/stories/va-checkbox-uswds.stories.tsx
Andrew565: This worked, thanks!

packages/storybook/stories/va-file-input-uswds.stories.tsx
Andrew565: I wasn't able to make `contents: String` work here, nor was I able to make `.readAsText` work, so I'm just going to leave it with `as String` for now.

PR #1570, opened 2025-04-10
va-link-action: add primary-entry variation [Amends #4001]

humancompanion-usds:
## Chromatic

https://adds-va-link-action-primary-4001--65a6e2ed2314f7b8f98609d8.chromatic.com

---

## Description
Closes [4001](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/4001)

## QA Checklist

## Screenshots
Screenshot 2025-04-10 at 4 58 22 PM

### Hover state at 768px
Screenshot 2025-04-10 at 4 58 58 PM

### Focus state at small desktop
Screenshot 2025-04-10 at 4 59 31 PM

### Focus + hover at narrow viewport width
Screenshot 2025-04-10 at 5 00 00 PM

## Acceptance criteria

## Definition of done
- [x] A link has been provided to the originating GitHub issue


humancompanion-usds:
Irksome. The new test passed locally. I will dig into this tomorrow.

packages/storybook/stories/va-link-action.stories.tsx
```diff
@@ -26,6 +26,7 @@ const Template = ({
   'disable-analytics': disableAnalytics,
   href,
   text,
+  type, // Add the type parameter here
```
jamigibbs, line 29: There are a few left over prompt comments like this. Can those be cleaned up?

packages/web-components/src/components/va-link-action/va-link-action.css
```diff
@@ -85,4 +85,84 @@
...
+  padding: 15px 15px 15px 45px;
+  margin: 0 1.5rem 0 0;
+  clip-path: polygon(0 0, calc(100% - 15px) 0, 100% 50%, calc(100% - 15px) 100%, 0 100%);
+  transition: all 0.2s ease-in-out;
+  width: 95%;
+  line-height: 1.4;
+  min-height: 24px;
+  max-width: var(--vads-size-line-length-5);
+
+  @media (min-width: 768px) {
```
jamigibbs, line 109: Can we use a design system breakpoint custom property for this value? https://design.va.gov/foundation/breakpoints

packages/web-components/src/components/va-link-action/va-link-action.css
```diff
@@ -85,4 +85,84 @@
...
+
+    & va-icon.link-icon {
+      background-color: var(--vads-color-action-focus-on-light);
+    }
+    & .link-text {
+      color: var(--vads-color-action-focus-on-light)
+    }
+  }
+
+  /* Focus & Active state */
```
jamigibbs, line 140: There's some similarity with this action link variation and the recently added [va-critical-action component](https://adds-va-link-action-primary-4001--65a6e2ed2314f7b8f98609d8.chromatic.com/?path=/docs/components-va-critical-action--docs). The focus states seem very different though.

Screenshot 2025-04-11 at 9 05 33 AM

Screenshot 2025-04-11 at 9 07 10 AM

Should we consider syncing these up?

packages/web-components/src/components/va-link-action/va-link-action.css
```diff
@@ -85,4 +85,84 @@
...
+:host a.va-link--primary-entry {
+  position: relative;
+  display: inline-block;
+  background-color: var(--vads-color-primary);
+  color: var(--vads-color-white);
+  text-decoration: none;
+  padding: 15px 15px 15px 45px;
+  margin: 0 1.5rem 0 0;
+  clip-path: polygon(0 0, calc(100% - 15px) 0, 100% 50%, calc(100% - 15px) 100%, 0 100%);
+  transition: all 0.2s ease-in-out;
```
jamigibbs, line 103: a11y should confirm but I think this transition easing is not an ideal behavior when focused. It's challenging on the eyes to see the easing delay going from blue to yellow.

Adding `transition: none` to the focus styles might take care of it.

https://github.com/user-attachments/assets/e407ff01-d028-43a0-97cb-f2a18fc4ec0d

packages/web-components/src/components/va-link-action/va-link-action.stories.js
jamigibbs, line 1: Was this file meant to be added? It appears to be empty.

packages/web-components/src/components/va-link-action/va-link-action.css
humancompanion-usds: Yes. However, the AI was following other components where we are not doing that. I just grepped through. We have 21 components using pixels and only 4 using variables. Can we A. Fix this and B. add this to a lint or other checker/ I'm going to try adding things like this and other coding guidelines, with the team's blessing and collaboration of course, to a knowledge base that we can feed to the copilot PR checker. https://docs.github.com/en/copilot/customizing-copilot/adding-repository-custom-instructions-for-github-copilot

packages/web-components/src/components/va-link-action/va-link-action.css
humancompanion-usds: Yeah, this was definitely on my mind while creating this (my component was first to production though!) That lighter background for the critical action component is because it's already using the focus state yellow. The team explored changing it to another color but, we couldn't agree on something we liked. In short, we think this is maybe okay? We're going to find out when critical action gets tested.

packages/web-components/src/components/va-link-action/va-link-action.css
jamigibbs: > We have 21 components using pixels and only 4 using variables. Can we A. Fix this 

Yes! We've just recently started to flag this in code reviews and we should update older components.

> and B. add this to a lint or other checker

Yes! This should be less manual. I'll create a ticket so we can do both of those things.

packages/web-components/src/components/va-link-action/va-link-action.stories.js
humancompanion-usds: Removed!

packages/web-components/src/components/va-link-action/va-link-action.css
humancompanion-usds: Fair enough. Transition set to none.

jamigibbs: There is still a delay with the icon so the text and icon aren't in sync. That's also different from the other variation's too. Is it possible to remove the transition for the icon as well?


https://github.com/user-attachments/assets/357031a9-58dd-4473-9f34-9b2f46d96e6b

packages/web-components/src/components/va-link-action/va-link-action.scss
```diff
@@ -0,0 +1,171 @@
+@forward 'settings';
+@import '../../mixins/links.css';
+@import '../../mixins/accessibility.css';
+@import '~@department-of-veterans-affairs/css-library/dist/tokens/scss/variables.scss';
```
jamigibbs, line 4: We import [variables.css globally](https://github.com/department-of-veterans-affairs/component-library/blob/12e91c223ece239a8f0ce13a6a32203a997c9a23/packages/web-components/src/global/main.css#L1) and prefer to use custom properties (which are supported natively) instead of Sass where possible.

So this import can be removed and the media query can be changed from `$medium-screen` to `var(--medium-screen)`.

humancompanion-usds: Hmm. I tried that and it doesn't work. Thoughts? The only other media query that is using a variable is in va-table-inner and I can't confirm if that's working either.

humancompanion-usds: Are coding conventions like this one written down anywhere that I can reference?

jamigibbs: Ah, you're right the custom properties won't work in a media query so importing the Sass variables seems to be the right way forward here. Thanks for calling that out! 

As for conventions, I think we need to revisit our [contributing guidelines](https://design.va.gov/about/developers/contributing). Especially as more external teams are contributing to the system too.

jamigibbs: PR to update the documentation site with Sass breakpoint variables:

- https://github.com/department-of-veterans-affairs/vets-design-system-documentation/pull/4011

jamigibbs (approved):

PR #1586, opened 2025-04-30
va-additional-info: Fix TypeError on render

longmd:
## Chromatic

https://longmd-108748-fix-va-addtl-info-typeerror--65a6e2ed2314f7b8f98609d8.chromatic.com

---
- [x] Link to any related issues in the description so they can be cross-referenced.
- [x] Add the appropriate version patch label (`major`, `minor`, `patch`, or `ignore-for-release`).

## Description
After implementing Datadog logs on the multiple applications, there a consistent log emitting from the `va-additional-info` web component. The error message is `TypeError: null is not an object (evaluating 'e.scrollHeight')` This error is occurring on initial render where an attempt to calculate the max-height property before the shadow DOM is fully hydrated.

This PR defers execution of the calculate method to ensure the method runs after the browser has completed rendering and the shadow DOM is hydrated. Additionally, there is a check added to ensure the slot content is available to apply the max-height value.

## Related issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/108748

## QA Checklist

## Screenshots

Screenshot 2025-04-30 at 14 41 20

## Acceptance criteria

## Definition of done


packages/web-components/src/components/va-additional-info/va-additional-info.tsx
```diff
@@ -93,7 +95,9 @@ export class VaAdditionalInfo {
   }
 
   componentDidRender() {
-    this.updateInfoMaxHeight();
+    requestAnimationFrame
+      ? requestAnimationFrame(() => this.updateInfoMaxHeight())
+      : setTimeout(() => this.updateInfoMaxHeight(), 0);
```
longmd, line 100: Using `requestAnimationFrame` ensures that the method runs after the browser has completed rendering, which is ideal for DOM-related calculations like `scrollHeight`. The fallback to `setTimeout` ensures compatibility in older browsers where `requestAnimationFrame` is not available.

jamigibbs: Cool solution. TIL'd about `requestAnimationFrame`. 🤓 

In [the docs](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame), it gives this warning about using the first arg. Is that something we need to consider here?

> Warning: Be sure always to use the first argument (or some other method for getting the current time) to calculate how much the animation will progress in a frame — otherwise, the animation will run faster on high refresh-rate screens. For ways to do that, see the examples below.

It's also [widely supported](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame#browser_compatibility) and we [no longer supported ancient browsers like IE11](https://depo-platform-documentation.scrollhelp.site/developer-docs/browser-support-policy) so we probably don't need the setTimeout but it doesn't hurt anything to have it there if you prefer to keep it.

longmd: @jamigibbs , I went ahead and removed the `setTimeout`, since we have wide support with modern browsers.

For the warning about using the timestamp arg... that isn't anything we need to worry about here. The warning in the docs would apply if we were implementing animations that depend on time progression, like moving an element smoothly across the screen or fading in/out). Since we're just using this method to defer the update call until the DOM is fully rendered, we don't need to worry about timing.

jamigibbs (approved):

PR #1591, opened 2025-05-08
va-file-input: add aria-label to input for dictation

powellkerry:
## Chromatic

https://108718-va-file-input-dictation--65a6e2ed2314f7b8f98609d8.chromatic.com

## Description
It was reported that you could not use dictation to click on the file input and select a file. This PR adds an aria-label so that the field becomes actionable with dictation.

Closes [108718](https://github.com/department-of-veterans-affairs/va.gov-team/issues/108718)

## QA Checklist
- [ x] Designer has signed off on changes (if applicable. If not applicable provide reason why)

## Screenshots
No visual changes

## Acceptance criteria

## Definition of done
- [ x] A link has been provided to the originating GitHub issue

### DIFF
--- a/packages/web-components/src/components/va-file-input/va-file-input.tsx
+++ b/packages/web-components/src/components/va-file-input/va-file-input.tsx
@@ -524,7 +524,7 @@ export class VaFileInput {
           <input
             id="fileInputField"
             class="file-input"
-            aria-label={fileInputString}
+            aria-label={`Select a file to upload. Drag file here or ${fileInputString}`}
             style={{
               visibility: (this.uploadStatus === 'success' || uploadedFile) ? 'hidden' : 'unset',
             }}
--- a/packages/web-components/src/components/va-file-input/va-file-input.tsx
+++ b/packages/web-components/src/components/va-file-input/va-file-input.tsx
@@ -34,7 +34,7 @@ export class VaFileInput {
   private uploadStatus: 'idle' | 'success' = 'idle';
   private fileType?: string;
   private chooseFileString: string ='choose from folder';
-  private dragFileString: string = 'Drag file here or ';
+  private dragFileString: string = 'Drag a file here or ';
 
   @Element() el: HTMLElement;
   @State() file?: File;
--- a/packages/web-components/src/components/va-file-input/va-file-input.tsx
+++ b/packages/web-components/src/components/va-file-input/va-file-input.tsx
@@ -33,7 +33,8 @@ export class VaFileInput {
   private fileInputRef!: HTMLInputElement;
   private uploadStatus: 'idle' | 'success' = 'idle';
   private fileType?: string;
-  private fileInputString: string ='choose from folder';
+  private chooseFileString: string ='choose from folder';
+  private dragFileString: string = 'Drag file here or ';
 
   @Element() el: HTMLElement;
   @State() file?: File;
@@ -403,11 +404,11 @@ export class VaFileInput {
   disconnectedCallback() {
     this.el.removeEventListener('change', this.handleChange);
   }
-  private getDefaultUploadMessage(fileInputString: string) {
+  private getDefaultUploadMessage(chooseFileString: string) {
     return (
       <span>
-        Drag a file here or{' '}
-        <span class="file-input-choose-text">{fileInputString}</span>
+        {this.dragFileString}
+        <span class="file-input-choose-text">{chooseFileString}</span>
       </span>
     )
   }
@@ -426,7 +427,8 @@ export class VaFileInput {
       hint,
       file,
       uploadStatus,
-      fileInputString,
+      dragFileString,
+      chooseFileString,
       uploadMessage,
       headerSize,
       fileContents,
@@ -524,7 +526,7 @@ export class VaFileInput {
           <input
             id="fileInputField"
             class="file-input"
-            aria-label={`Select a file to upload. Drag file here or ${fileInputString}`}
+            aria-label={`${label}. ${dragFileString}${chooseFileString}`}
             style={{
               visibility: (this.uploadStatus === 'success' || uploadedFile) ? 'hidden' : 'unset',
             }}
@@ -553,7 +555,7 @@ export class VaFileInput {
               <div class={fileInputTargetClasses}>
                 <div class="file-input-box"></div>
                 <div class="file-input-instructions">
-                  {!!uploadMessage ? uploadMessage : this.getDefaultUploadMessage(fileInputString)}
+                  {!!uploadMessage ? uploadMessage : this.getDefaultUploadMessage(chooseFileString)}
                 </div>
               </div>
             </div>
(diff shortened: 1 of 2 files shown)


amyleadem (commented):
Functionality in Windows Voice Access is looking good! 

I performed the following tests on the [default file input](https://108718-va-file-input-dictation--65a6e2ed2314f7b8f98609d8.chromatic.com/iframe.html?args=&id=uswds-va-file-input-multiple--default&viewMode=story) and [default file input multiple](https://108718-va-file-input-dictation--65a6e2ed2314f7b8f98609d8.chromatic.com/iframe.html?args=&id=uswds-va-file-input-multiple--default&viewMode=story) components in Windows Voice Access:

- When the component is in its initialized state…
    - [x]  the voice command "Click choose from folder” opens the file dialog
- When the component is in a selected file state…
    - [x]  The voice command "Click change file" opens the file dialog
    - [x]  The voice command "Click delete" opens the delete modal
- When the component is the confirm delete state…
    - [x]  The voice command “Click yes remove this” closes the modal, removes the selected file, and returns the component to its initial state
    - [x]  The voice command “Click no keep this” closes the modal and returns to the previous selected file state
    - [x]  The voice command “Click close” closes the modal and returns to the previous selected file state

**issue (non-blocking):** In Windows Voice Access, the buttons for “Change file”, “Delete, “Yes, remove this”, and “No, keep this” have a pre-existing quirk where they are all functional via voice command, but register as having multiple instances on the page. This means that the user must select which instance of the button to trigger, even though each instance triggers the same expected functionality. Here's what happens: 
  1. After giving the command “Click [button name]”,  Voice Access displays the prompt “Click [button name]. Which one?”. 
  2. Then, Voice Access displays two numbers next to the button, "1" and "2". 
  3. Triggering either number with voice command successfully performs the expected button action, but it adds an extra step that could lead to confusion and frustration. 

Because the component is fully functional via voice command, I would say that this shouldn't need to be a blocker for the PR, but wanted to flag it for awareness and consideration for possible future improvement.

powellkerry:
@amyleadem I create a task to debug the issue with the edit and delete buttons: https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/4106

amyleadem (changes requested):
Wanted to cycle back to this because I realized that while this update might fix the issue for voice command users, it will likely negatively impact screen reader users. This is because it no longer includes the associated `<label>` content ("Select a file to upload") in the readout, and also does not include the full visible text. Here are some screenshots from the VoiceOver readout: 

**[Before:](https://design.va.gov/storybook/?path=/story/uswds-va-file-input--default)**
Includes associated label, but does not match visible text
✅ screen readers
❌  voice command (on Windows Voice Access)

image

**[After:](https://108718-va-file-input-dictation--65a6e2ed2314f7b8f98609d8.chromatic.com/?path=/story/uswds-va-file-input--default)**
Ignores the associated label, but includes incomplete visible text:
Includes associated label, but does not match visible text
❌ screen readers
✅ voice command (on Windows Voice Access)

image

Thinking that best practice would be to include both the label content and the visible text in the aria-label to generate something like this: `aria-label="Select a file to upload. Drag file here or choose from folder."`

Initial testing showed that the command "Click choose from folder" still worked when using this enhanced aria-label in both Windows and MacOS.

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -33,6 +33,7 @@ export class VaFileInput {
   private fileInputRef!: HTMLInputElement;
   private uploadStatus: 'idle' | 'success' = 'idle';
   private fileType?: string;
+  private fileInputString: string ='choose from folder';
```
jamigibbs, line 36: Will this string value ever change? Or would the variable be better defined as a constant?

jamigibbs (approved):
I just had a minor question about the string assignment. Thank you both for working on this!

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -523,6 +524,7 @@ export class VaFileInput {
           <input
             id="fileInputField"
             class="file-input"
+            aria-label={`Select a file to upload. Drag file here or ${fileInputString}`}
```
amyleadem, line 527: **suggestion:** Thinking we can make this more reusable and customizable by pulling in the label prop here. Additionally, we could make a new variable for the "Drag file here or" text so that we can use it in both `getDefaultUploadMessage` and here. Thinking the end result could be something like this: 
```suggestion
            aria-label={`${label}. ${dragFileText} ${chooseFileText}`}
```

amyleadem (changes requested):
@powellkerry thanks for the quick turnaround on this! Looks to be working as expected in both Windows and Mac. I added a suggestion below to make this more reusable. Let me know what you think!

packages/web-components/src/components/va-file-input/va-file-input.tsx
amyleadem: Also, just curious - should we build out an aria-label check in the tests for this component?

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -523,6 +526,7 @@ export class VaFileInput {
           <input
             id="fileInputField"
             class="file-input"
+            aria-label={`${label}. ${dragFileString}${chooseFileString}`}
```
amyleadem, line 529: **issue:** I noticed one small regression in the screen reader readout in the required variant. In VoiceOver, it no longer announces the word "Required".  Wondering if we could conditionally add the required text when `required` is true.

packages/web-components/src/components/va-file-input/va-file-input.tsx
```diff
@@ -402,12 +404,11 @@ export class VaFileInput {
...
     this.el.removeEventListener('change', this.handleChange);
   }
-
-  private getDefaultUploadMessage() {
+  private getDefaultUploadMessage(chooseFileString: string) {
     return (
       <span>
-        Drag a file here or{' '}
-        <span class="file-input-choose-text">choose from folder</span>
+        {this.dragFileString}
```
amyleadem, line 410: **question:** I noticed a small difference in how `chooseFileString` and `this.dragFileString` are being used here. Should we standardize so that both or neither are called with `this`?

amyleadem (changes requested):
Everything is working well in Windows and MacOS! 

When testing all the variants, I did notice one small regression in the required variant and had one quibble about code style. I've added details about both in the comment below. 

I _think_ that should be it. 
- [x] Confirm the component is accessible via Windows Voice Access and Mac Voice Control
- [x] Confirm code is reusable
- [x] Confirm tests are present
- [ ] Confirm no regressions in screen reader experience

amyleadem (approved):
LGTM! Thanks @powellkerry

1Copenut:
Thank you all for prioritizing this item and getting a fix out so quickly!

PR #1617, opened 2025-05-29
Storybook 8 migration

powellkerry:
## Chromatic

https://storybook8--65a6e2ed2314f7b8f98609d8.chromatic.com

## Description
Storybook 8 migration.

The major update is that version 8 does not support `*.stories.mdx` files anymore so those had to be renamed to `*.mdx`
There were also updates to make storybook compiler agnostic, so some changes needed to be made to our babel config.
Some add-ons were also moved.

## Definition of done
- [ ] Storybook works and renders stories as expected
- [ ] Mobile storybook integration works


jamigibbs (approved):
I think this looks great! The one difference I noticed is the mobile component section is automatically expanded in prod now where this one is not. It gets kind of hidden at the bottom of the sidebar. Are we able to control that so that it's expanded or maybe just moved up?

Screenshot 2025-05-30 at 8 25 43 AM

PR #2036, opened 2026-03-16
feat(web-components): add @guidanceHref to all top-level components

humancompanion-usds:
## Summary

- Backfills \`@guidanceHref\` docstag on all 36 top-level web components that were missing it
- Adds a local ESLint rule (\`require-guidance-href\`) that errors if a future component is merged without the tag — enforcement runs via \`yarn lint\` in \`packages/web-components\`
- Sub-components (\`va-accordion-item\`, \`va-radio-option\`, etc.) are excluded from both the backfill and the lint rule

## Background

The \`@guidanceHref\` tag was introduced in August 2022 and applied selectively to form components at the time. Layout, navigation, and general components were missed and never caught because there was no automated enforcement. This PR closes both gaps at once.

## Commits

1. **Backfill** — adds \`@guidanceHref\` to all 36 missing top-level components (one line per file)
2. **Enforcement** — adds \`eslint-rules/require-guidance-href.js\` custom ESLint rule + wires it into \`.eslintrc.json\` and the \`lint\` script

## Slugs needing URL verification

These components are new enough that their \`design.va.gov\` pages may not be published yet. The slugs follow the established naming convention but should be confirmed before merge:

| Component | Slug used |
|-----------|-----------|
| \`va-critical-action\` | \`critical-action\` |
| \`va-details\` | \`details\` |
| \`va-notification\` | \`notification\` |
| \`va-search-filter\` | \`search-filter\` |
| \`va-service-list-item\` | \`service-list-item\` |
| \`va-sort\` | \`sort\` |

## Test plan

- [x] Verify slug correctness for the 6 components listed above against design.va.gov
- [x] Confirm \`yarn lint\` passes with all existing \`@guidanceHref\` tags in place
- [ ] Confirm adding a new \`@Component\` without \`@guidanceHref\` triggers the lint error

### DIFF
--- a/packages/design-system-dashboard-cli/package.json
+++ b/packages/design-system-dashboard-cli/package.json
@@ -22,7 +22,7 @@
     "command-line-args": "5.2.1",
     "csv": "6.3.10",
     "fast-memoize": "2.5.2",
-    "glob": "13.0.2",
+    "glob": "13.0.3",
     "jsonexport": "3.2.0",
     "lodash": "4.17.23"
   },
--- a/packages/react-components/package.json
+++ b/packages/react-components/package.json
@@ -73,7 +73,7 @@
     "eslint-plugin-prettier": "3.4.1",
     "eslint-plugin-react": "7.37.5",
     "fs-extra": "9.1.0",
-    "glob": "13.0.2",
+    "glob": "13.0.3",
     "husky": "5.2.0",
     "is-docker": "2.1.1",
     "karma": "6.4.4",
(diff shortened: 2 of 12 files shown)


packages/web-components/eslint-rules/require-guidance-href.js
```diff
@@ -0,0 +1,83 @@
+'use strict';
+
+/**
+ * ESLint rule: require-guidance-href
```
jamigibbs: **suggestion**: I haven't been able to get this rule running yet but since this is the first custom rule we're adding to the component-library, I'd like to suggest we also include tests if possible. Here are [examples of tests](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/tree/master/packages/eslint-plugin/tests/lib/rules) for vets-website's custom rules.

jamigibbs (approved):
⚠️ This PR is approved but before merging, run `yarn dev` in the top level component-library directory so that an updated `packages/web-components/src/components.d.ts` file can be committed:

Image


And this looks much better now! Thank you. 🙌🏼

Image

